PR #771, opened 2014-03-23
Add MPICH support to MUMPS.

goxberry:
Add a conditional checking for the particular MPI implementation
installed (currently: open-mpi, mpich2), and then append flags
appropriate to that implementation.

### DIFF
--- a/mumps.rb
+++ b/mumps.rb
@@ -49,12 +49,21 @@ class Mumps < Formula
     make_args << "ORDERINGSF=#{orderingsf}"
 
     if build.with? :mpi
-      make_args += ["CC=#{ENV['MPICC']} -fPIC",
-                    "FC=#{ENV['MPIFC']} -fPIC",
-                    "FL=#{ENV['MPIFC']} -fPIC",
-                    "SCALAP=-L#{Formula["scalapack"].lib} -lscalapack",
-                    "INCPAR=-I#{Formula["open-mpi"].include}",
-                    "LIBPAR=$(SCALAP) -L#{Formula["open-mpi"].lib} -lmpi -lmpi_mpifh"]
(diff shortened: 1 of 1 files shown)


davydden:
@goxberry with regards to compile and link flags, aren't we safe if we compile with `mpic++`, `mpif90`, etc? I never used `MPICH` but this is my understanding of what those wrappers (also) do in case of `OpenMPI`.

goxberry:
@davydden No. MPICH and OpenMPI both provide compiler wrappers called `mpicc`, etc., and each wraps compilers with different flags, plus each provides different options for their respective wrappers.

davydden:
@goxberry sure, the flags are different. 

> Otherwise, that command should return the right compile flags for wrappers, which seems to be what is needed for the formula. The link flags for Open MPI should be taken from mpicc -showme:link.

I was referring to this part. IMHO one does not have to specify linking flags if things are compiled with wrappers. So no need to `mpicc -link_info`. 

Of course, that is a separate question from  `OpenMPI` vs `MPICH` consistency checks.

goxberry:
@davydden Ideally, yes, the compiler wrappers work for builds and querying flags is unnecessary. IIRC, the compiler wrappers didn't "just work" when writing the hypre formula; they weren't detected by `configure`. There are many possible solutions; that one happened to be the most expedient and least surprising. Another option would be to substitute `mpicc` in for `clang` in `CC`, etc., which I think is fine as long as switching MPI implementations forces a rebuild of all formulas depending on MPI, and the effects of those substitutions are limited to formula building software with MPI. Better still would be to ensure that MPI wrappers are properly detected by `configure` instead of manual flag hacking or redirecting `CC`,  etc.

PR #1069, opened 2014-07-20
root6: New formula

veprbl:
This is a proof-of-concept formula for root6. To make this compile into something usable I had to remove Homebrew's superenv clang/clang++ wrapper, because root6 would remember CC/CXX env from compile time and try to use it for cling runtime, which leads to fail. We probably should work with upstream to solve this problem gracefully.

The other problem is that I couldn't make root6 new interpreter fully work my systems (tried mac os non-Homebrew build, linux manual build, linux using binaries from cern AFS). I wouldn't say that it's completely broken, but it is not completely functional either. Please share your experiences with root6 in this ticket, that might help determine the scope of the problem.

I also took liberty to remove fake test and update zsh recipe.


iml:

mikemcquaid:
Killed this build after 4h43m.

veprbl:
I think this is ready for merge.

PR #1215, opened 2014-10-01
tophat: Fix URLs on older versions

jmchilton:
While actually potentially useful, this pull request is also a trial balloon to see if there is interest in accepting patches to older versions of the recipes (certainly understandable if it is felt that this is a waste of time so I will not be offended in the least if this is declined).

At any rate, the URL of TopHat has changed and older versions of the recipes fail. For every version since 1.4.1 - I checked out the commit representing the latest change to tophat at that version patched the URL and merge that change forward until HEAD. HEAD however was correct so it looks in the diff like there is no change.

This also exposed a bug in versions command - that I have documented [here](https://github.com/jmchilton/linuxbrew/compare/versions_fix?expand=1) - any interest in that being merged into linuxbrew? (I assume homebrew proper is uninterested since versions is deprecated).


tdsmith:
Thanks for the suggestion and for doing the legwork here. I think we will prefer not to use history to repair old formulas. If old versions of tophat are generally useful, a dedicated tap would be a great solution. Cheers!

PR #1235, opened 2014-10-06
Fixes with MOAB 4.7 release

rajeeja:
Changing website, link and sha.

### DIFF
--- a/moab.rb
+++ b/moab.rb
@@ -1,27 +1,30 @@
-require 'formula'
+require "formula"
 
 class Moab < Formula
-  homepage 'https://sigma.mcs.anl.gov/moab-library'
-  url 'ftp://ftp.mcs.anl.gov/pub/fathom/moab-4.7.0RC1.tar.gz'
-  sha1 'ad69a021b49d7a9870ff908945d8c9221f51b66b'
+  homepage "http://sigma.mcs.anl.gov"
+  url "http://ftp.mcs.anl.gov/pub/fathom/moab-nightly.tar.gz"
+  sha1 "78fddbda313e7f141014268a84c57ec23b6baf28"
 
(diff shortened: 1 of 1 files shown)


dpo:
@BrewTestBot test this please

dpo:
I think this needs to be rebased against `origin/master`.

moab.rb
```diff
@@ -1,27 +1,30 @@
...
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
 
-  depends_on :autoconf
-  depends_on :automake
-  depends_on :libtool
   depends_on 'netcdf'
   depends_on 'hdf5'
-  depends_on :fortran if build.with? 'check'
+  depends_on 'gcc'
```
iml: Don't depend on `gcc`, just `:fortran`. If it really doesn't build with clang, the `fails_with :clang` DSL should handle that. If the issue is libc++ incompatibility, pass `ENV.libstdcxx` to the install method.

moab.rb
```diff
@@ -1,27 +1,30 @@
-require 'formula'
+require "formula"
 
 class Moab < Formula
-  homepage 'https://trac.mcs.anl.gov/projects/ITAPS/wiki/MOAB'
-  url 'https://bitbucket.org/fathomteam/moab/get/4.6.2.tar.gz'
-  sha1 'cecd30aaf9bdbb58078e06b0097a9144cac63d7f'
+  homepage "http://sigma.mcs.anl.gov"
+  url "http://ftp.mcs.anl.gov/pub/fathom/moab-nightly.tar.gz"
```
iml: Need a stable release.

iml:
Since this was submitted on master, you might need to do a hard reset and recommit.

PR #1268, opened 2014-10-19
openalpr 1.2 (new formula)

twelve17:
I am submitting this formula to this tap, as it depends on opencv.

Note: this package depends on Tesseract 3.03, which is in RC status at
the moment (roadmap):

https://code.google.com/p/tesseract-ocr/

The formula will build with the currently released version of
tesseract, but will only run properly if the "--devel" version
of tesseract is installed, so I understand if this package
shouldn't be merged just yet.  If possible, I'd at least like to 
get it OK'd in the meantime so it is ready to go when tesseract 
3.03 is released.

This formula was originally reviewed by a developer on the
core homebrew project, here:

https://github.com/Homebrew/homebrew/pull/33351

### DIFF
--- a/openalpr.rb
+++ b/openalpr.rb
@@ -1,0 +1,59 @@
+require "formula"
+
+class Openalpr < Formula
+  homepage "https://www.github.com/openalpr/openalpr"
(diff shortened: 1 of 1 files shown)


dpo:
@BrewTestBot test this please

PR #1299, opened 2014-10-27
deal.II 8.2 (new formula)

davydden:
Added support for deal.II library with numerous dependencies
and options.

### DIFF
--- a/dealii.rb
+++ b/dealii.rb
@@ -8,20 +8,21 @@ class Dealii < Formula
   end
 
   option "with-testsuite", "Run full test suite (7000+ tests). Takes a lot of time."
-  depends_on "openblas"     => :optional
 
   depends_on "cmake"        => :build
   depends_on :mpi           => [:cc, :cxx, :f90, :recommended]
+  depends_on "openblas"     => :optional
+
+  depends_on "arpack"       => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : []) + ((build.with? "openblas") ? ["with-openblas"] : [])
   depends_on "boost"        => :recommended
   depends_on "hdf5"         => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : [])
-  depends_on "arpack"       => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : []) + ((build.with? "openblas") ? ["with-openblas"] : [])
-  depends_on "mumps"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
   depends_on "metis"        => :recommended
-  depends_on "opencascade"  => :recommended
-  depends_on "netcdf"       => [:recommended, "with-fortran", "with-cxx-compat"]
-  depends_on "p4est"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : []) if build.with? "mpi"
+  depends_on "mumps"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
   depends_on "muparser"     => :recommended
+  depends_on "netcdf"       => [:recommended, "with-fortran", "with-cxx-compat"]
+  depends_on "opencascade"  => :recommended
   depends_on "petsc"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
+  depends_on "p4est"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : []) if build.with? "mpi"
   depends_on "slepc"        => :recommended
   depends_on "trilinos"     => [:optional] + ((build.with? "openblas") ? ["with-openblas"] : [])
 
--- a/dealii.rb
+++ b/dealii.rb
@@ -13,18 +13,21 @@ class Dealii < Formula
   depends_on :mpi           => [:cc, :cxx, :f90, :recommended]
   depends_on "openblas"     => :optional
 
-  depends_on "arpack"       => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : []) + ((build.with? "openblas") ? ["with-openblas"] : [])
+  openblasdep = (build.with? "openblas") ? ["with-openblas"] : []
+  mpidep      = (build.with? "mpi")      ? ["with-mpi"]      : []
+
+  depends_on "arpack"       => [:recommended] + mpidep + openblasdep
   depends_on "boost"        => :recommended
-  depends_on "hdf5"         => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : [])
+  depends_on "hdf5"         => [:recommended] + mpidep
   depends_on "metis"        => :recommended
-  depends_on "mumps"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
+  depends_on "mumps"        => [:recommended] + openblasdep
   depends_on "muparser"     => :recommended
   depends_on "netcdf"       => [:recommended, "with-fortran", "with-cxx-compat"]
   depends_on "opencascade"  => :recommended
-  depends_on "petsc"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
-  depends_on "p4est"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : []) if build.with? "mpi"
+  depends_on "petsc"        => [:recommended] + openblasdep
+  depends_on "p4est"        => [:recommended] + openblasdep if build.with? "mpi"
   depends_on "slepc"        => :recommended
-  depends_on "trilinos"     => [:optional] + ((build.with? "openblas") ? ["with-openblas"] : [])
+  depends_on "trilinos"     => [:optional] + openblasdep
 
   def install
     args = %W[
--- a/dealii.rb
+++ b/dealii.rb
@@ -23,7 +23,7 @@ class Dealii < Formula
   depends_on "muparser"     => :recommended
   depends_on "petsc"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
   depends_on "slepc"        => :recommended
-  depends_on "trilinos"     => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
+  depends_on "trilinos"     => [:optional] + ((build.with? "openblas") ? ["with-openblas"] : [])
 
   def install
     args = %W[
--- a/dealii.rb
+++ b/dealii.rb
@@ -1,0 +1,80 @@
+class Dealii < Formula
+  homepage "http://www.dealii.org"
+  url "https://github.com/dealii/dealii/releases/download/v8.2.1/dealii-8.2.1.tar.gz"
+  sha256 "d75674e45fe63cd9fa294460fe45228904d51a68f744dbb99cd7b60720f3b2a0"
+
+  head do
+    url "https://github.com/dealii/dealii.git"
+  end
+
+  option "with-testsuite", "Run full test suite (7000+ tests). Takes a lot of time."
+  depends_on "openblas"     => :optional
+
+  depends_on "cmake"        => :build
+  depends_on :mpi           => [:cc, :cxx, :f90, :recommended]
+  depends_on "boost"        => :recommended
+  depends_on "hdf5"         => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : [])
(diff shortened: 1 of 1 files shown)


davydden:
whattodo:  there is a bug in cmake 3.1.0, dealii works with 3.0.2

dpo:
Maybe submit a formula for CMake 3.0.2 to homebrew/versions.

davydden:
submitted a PR for Cmake302: https://github.com/Homebrew/homebrew-versions/pull/617

davydden:
After Trilinos is sorted out #1779, I will clean this PR and make it ready.

dpo:
I pulled #1779.

davydden:
This should work as it is.

dealii.rb
```diff
@@ -0,0 +1,84 @@
...
+    version "8.3pre"
+  end
+
+  option "with-testsuite", "Run full test suite (7000+ tests). Takes a lot of time."
+
+  depends_on "cmake"        => :build
+  depends_on :mpi           => [:cc, :cxx, :f90, :recommended]
+  depends_on "boost"        => :recommended
+  depends_on "hdf5"         => [:recommended, (build.with? :mpi) ? "with-mpi" : ""]
+  depends_on "arpack"       => [:recommended, (build.with? :mpi) ? "with-mpi" : ""]
```
dpo: `build.with?` should now be called with strings instead of symols, i.e., `build.with? "mpi"`.

dealii.rb
```diff
@@ -0,0 +1,84 @@
...
+    args << "-DDEAL_II_WITH_METIS=OFF"  if build.without? "metis"
+    args << "-DDEAL_II_WITH_ARPACK=OFF" if build.without? "arpack"
+    args << "-DDEAL_II_WITH_P4EST=OFF"  if build.without? "p4est"
+    args << "-DDEAL_II_WITH_GSL=OFF"    if build.without? "gsl"
+    args << "-DDEAL_II_WITH_OPENCASCADE=OFF" if build.without? "opencascade"
+
+    if build.with? "petsc"
+      # make deal.II link against libpetsc rather than libpetsc.3.5.3
+      args << "-DPETSC_LIB=libpetsc"
+      args << "-DPETSC_DIR=#{Formula["petsc"].opt}"
```
dpo: Do you mean `opt_prefix` here?

dealii.rb
```diff
@@ -0,0 +1,84 @@
...
+      # run minimal test cases (8 tests)
+      system "make", "test"
+      log_name = "make-test.log"
+      system "make test 2>&1 | tee #{log_name}"
+      ohai `grep "tests passed" "#{log_name}"`.chomp
+      prefix.install "#{log_name}"
+      # run full test suite if really needed
+      if build.with? "testsuite"
+        system "make", "setup_tests"
+        system ("ctest -j" + Hardware::CPU.cores)
```
dpo: Maybe `system "ctest", "-j", Hardware::CPU.cores`?

davydden:
Thanks @dpo for reviewing the formula. I am AFK for several days, will look at your recent comments ASAP.

dpo:
@BrewTestBot test this please

dpo:
We now use the `sha256` instead of the `sha1`. Also, it's come up that we shouldn't pass symbols to `build.with?`, as in `build.with? :mpi` or `build.without? :fortran`. We should use `build.with? "mpi"` and `build.without? "fortran"` instead.

davydden:
one shall test it only after this PR is sorted out https://github.com/Homebrew/homebrew-science/pull/1924/files  as I use `with-openblas` option.

ashkan2200:
@dpo @davydden any chance this is going to be merged soon?

davydden:
we'll hit this https://github.com/Homebrew/homebrew/issues/38665 when trying to have different versions of `netcdf`: one from deal.ii which would need `--with-cxx-compat, --with-fortran` and another from `petsc` which has `--with-fortran` only...

any ideas?

davydden:
it seems something changed in homebrew, at least on `linuxbrew` i have `Error: undefined method `Mac' for OS:Module`

davydden:
and also `Error: undefined method `lib_prefix' for <Formulary::Formulae::Openblas:0x000000012d89c0>`

dealii.rb
```diff
@@ -0,0 +1,81 @@
...
+      -DCMAKE_INSTALL_PREFIX=#{prefix}
+      -DCMAKE_FIND_FRAMEWORK=LAST
+      -Wno-dev
+      -DDEAL_II_COMPONENT_COMPAT_FILES=OFF
+    ]
+    # constrain Cmake to look for libraries in homebrew's prefix
+    args << "-DCMAKE_PREFIX_PATH=#{HOMEBREW_PREFIX}"
+
+    if build.with? "openblas"
+      ext = OS.mac ? "dyld" : "so"
```
iml: `OS.mac?`

dealii.rb
```diff
@@ -0,0 +1,81 @@
...
+      args << "-DLAPACK_FOUND=true"
+      args << "-DLAPACK_INCLUDE_DIRS=#{Formula["openblas"].opt_include}"
+      args << "-DLAPACK_LIBRARIES=#{Formula["openblas"].lib_prefix}/libopenblas.#{ext}"
```
iml: `.opt_lib`

dealii.rb
```diff
@@ -0,0 +1,81 @@
+class Dealii < Formula
+  homepage "http://www.dealii.org"
+  url "https://github.com/dealii/dealii/releases/download/v8.2.1/dealii-8.2.1.tar.gz"
+  sha256 "d75674e45fe63cd9fa294460fe45228904d51a68f744dbb99cd7b60720f3b2a0"
+
+  head do
+    url "https://github.com/dealii/dealii.git", :branch => "master"
+    version "8.3pre"
```
iml: There are no versions for HEAD, and setting the master branch is unnecessary.

dealii.rb
davydden: supposedly it should be `ext = OS.mac? ? "dyld" : "so"` ?

dealii.rb
```diff
@@ -0,0 +1,80 @@
...
+    url "https://github.com/dealii/dealii.git"
+  end
+
+  option "with-testsuite", "Run full test suite (7000+ tests). Takes a lot of time."
+  depends_on "openblas"     => :optional
+
+  depends_on "cmake"        => :build
+  depends_on :mpi           => [:cc, :cxx, :f90, :recommended]
+  depends_on "boost"        => :recommended
+  depends_on "hdf5"         => [:recommended, (build.with? "mpi") ? "with-mpi" : ""]
```
dpo: Sorry, I'm the one who suggested this construct but it's invalid because `hdf5` doesn't have an option named `""`. It should be something like
````
depends_on "hdf5" => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : [])
````
Same for the others.

File: dealii.rb [hunk 8]
```diff
@@ -0,0 +1,80 @@
...
+
+  depends_on "cmake"        => :build
+  depends_on :mpi           => [:cc, :cxx, :f90, :recommended]
+  depends_on "boost"        => :recommended
+  depends_on "hdf5"         => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : [])
+  depends_on "arpack"       => [:recommended] + ((build.with? "mpi") ? ["with-mpi"] : []) + ((build.with? "openblas") ? ["with-openblas"] : [])
+  depends_on "mumps"        => [:recommended] + ((build.with? "openblas") ? ["with-openblas"] : [])
+  depends_on "metis"        => :recommended
+  depends_on "opencascade"  => :recommended
+  depends_on "netcdf"       => [:recommended, "with-fortran", "with-cxx-compat"]
```
dpo: Should there be a `depends_on :fortran` in this formula then?

davydden: My fault, i will just drop `with-fortran` from here. 
`dealii` does not need a fortran compiler. I had fortran here as `petsc` uses `netcdf` with fortran.
 
But as far as i recall, options should concatenate nicely, so if there is 
`netcdf --with-fortran` and we need now `netcdf --with-cxx-compat`, we should end up with `netcdf --with-fortran --with-cxx-compat`, right?

davydden: here's from a line from deal.ii cmake instrucitons (https://www.dealii.org/developer/index.html):
```
A working CXX compiler is needed. It is optional to provide a C or Fortran compiler.
```
but frankly I don't know why would the lib need them. 

@bangerth could you please comment why fortran compiler could be needed during the build?

dpo:
I'm getting
````
0  0x103ff3bc1  __assert_rtn + 144
1  0x104043431  ld::tool::InputFiles::findDylib(char const*, char const*) + 847
2  0x10403fc8b  mach_o::dylib::File<x86_64>::processIndirectLibraries(ld::dylib::File::DylibHandler*, bool) + 69
3  0x104043f2a  ld::tool::InputFiles::createIndirectDylibs() + 292
4  0x104044ea6  ld::tool::InputFiles::forEachInitialAtom(ld::File::AtomHandler&, ld::Internal&) + 526
5  0x10404ff50  ld::tool::Resolver::resolve() + 48
6  0x103ff4737  main + 689
A linker snapshot was created at:
	/tmp/libdeal_II.8.2.1.dylib-2015-03-22-155109.ld-snapshot
ld: Assertion failed: (counter() < 0xffff), function nextCounter, file /SourceCache/ld64/ld64-241.8/src/ld/ld.hpp, line 102.
````

**Comment by bangerth** [comment]
That's an internal error in the linker of the operating system -- I think there is little we can do about this from our side. What do you get when you google the error?

**Comment by davydden** [comment]
checked today, the current formula works for me both on `homebrew` (yosemite with all patches) and on `linuxbrew` (standalone OpenSuse).

**Comment by dpo** [comment]
I get the same on both Mavericks and Yosemite. The Yosemite machine is fresh.

**Comment by dpo** [comment]
@davydden Could you make `trilinos` `:optional` just to see if the bot builds the formula?

**Comment by davydden** [comment]
@dpo I made `trilinos` optional for a moment. I also restored `netcdf --with-fortran --with-cxx-compat`. That might not be related to your linking errors, but perhaps some dependency tracking in `homebrew` was not robust to concatenate `pets` and `trilinos` dependency `netcdf --with-fortran` and dealii `netcdf --with-cxx-compat` for the case when nothing is yet installed (fresh `homebrew`). Could you check which flags `netcdf` has at the end of dependencies installation?

File: dealii.rb [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+    args << "-DARPACK_DIR=#{Formula["arpack"].opt_prefix}" if build.with? "arpack"
+    args << "-DP4EST_DIR=#{Formula["p4est"].opt_prefix}" if build.with? "p4est"
+    args << "-DOPENCASCADE_DIR=#{Formula["opencascade"].opt_prefix}" if build.with? "opencascade"
+    args << "-DNETCDF_DIR=#{Formula["netcdf"].opt_prefix}" if build.with? "netcdf"
+    args << "-DPETSC_DIR=#{Formula["petsc"].opt_prefix}" if build.with? "petsc"
+    args << "-DSLEPC_DIR=#{Formula["slepc"].opt_prefix}" if build.with? "slepc"
+    args << "-DTRILINOS_DIR=#{Formula["trilinos"].opt_prefix}" if build.with? "trilinos"
+
+    mkdir "build" do
+      system "cmake", *args, "../"
```
dpo: This fails on Mountain Lion. Try `system "cmake", "..", *args`

dpo:
I'm using MPICH2. The linker error goes away if I use OpenMPI.

dpo:
Now should `trilinos` remain `:optional` or changed back to `:recommended`? We won't have bottles if we make it `:recommended` because the bot times out.

dpo:
I also recommend defining
````
openblasdep = (build.with? "openblas") ? ["with-openblas"] : []
mpidep = (build.with? "mpi") ? ["with-mpi"] : []
````
to streamline the dependencies. Perhaps you could also rearrange the dependencies (e.g., `cmake`, `:mpi` and `openblas` together, followed by the two definitions above, and the rest of the deps in alpha order)?!

davydden:
i would keep `trilinos` `:optional` then, at least until the bot can handle it. If somebody does not need trilinos, there will be  a bottle. Otherwise it will be compiling from source anyway with extra flag.

dpo:
Excellent. Many thanks, and thank you for your patience. This was a long-winded one.

PR #1314, opened 2014-10-31
Suite-Sparse: version 4.4.4.

dpo:
If compiling the MEX interfaces, your `mexopts.sh` should use gcc-4.9 and gfortran-4.9 to avoid a compilation error of SPQR with clang.

### DIFF
--- a/octave.rb
+++ b/octave.rb
@@ -5,7 +5,7 @@ class Octave < Formula
   mirror "https://ftp.gnu.org/gnu/octave/octave-3.8.2.tar.bz2"
   sha256 "83bbd701aab04e7e57d0d5b8373dd54719bebb64ce0a850e69bf3d7454f33bae"
   head "http://www.octave.org/hg/octave", :branch => "default", :using => :hg
-  revision 1
+  revision 2
 
   stable do
     # Allows clang 3.5 to compile with a recent libc++ release.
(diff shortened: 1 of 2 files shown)


dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

davydden:
if octave still does not work with current suite sparse, how about commenting out the dependency (making it effectively without-suite-sparse) until a formula for 4.0 is there, which hopefully compiles with the current suite sparse.

dpo:
Upgrading `octave` seems like a major pain no-one is willing to launch into. I'd rather have a separate `suite-sparse-421` formula. Do you have a need for 4.4.1?

davydden:
As there are some bug fixes, perhaps it is better to merge so that other software which use `suitesparse` (`petsc` and `trilinos`) are more robust. 
As a temporary measure for octave, i would add a note (`brew info`) saying that it is currently broken with the suite-sparse `4.4.1` and advise users to install with `--without-suitesparse`.
I understand that you don't want to break `octave` with this PR, but as a downside it prevents from using the software with lesser number of bugs in it. Just an opinion.

dpo:
Octave fails:
```
libtool: link: warning: library `/usr/local/Cellar/graphicsmagick/1.3.21/lib/libGraphicsMagick++.la' was moved.
../libtool: line 7486: cd: @@HOMEBREW_CELLAR@@/graphicsmagick/1.3.21/lib: No such file or directory
libtool: link: warning: cannot determine absolute directory name of `@@HOMEBREW_CELLAR@@/graphicsmagick/1.3.21/lib'
grep: @@HOMEBREW_CELLAR@@/graphicsmagick/1.3.21/lib/libGraphicsMagick.la: No such file or directory
/usr/local/opt/gnu-sed/bin/gsed: can't read @@HOMEBREW_CELLAR@@/graphicsmagick/1.3.21/lib/libGraphicsMagick.la: No such file or directory
libtool: link: `@@HOMEBREW_CELLAR@@/graphicsmagick/1.3.21/lib/libGraphicsMagick.la' is not a valid libtool archive
```

davydden:
it seems that your `graphicsmagick` is broken. have you tried reinstalling?

dpo:
It's on the bot, not my local machine. The same error occurred in another PR.

dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

PR #1324, opened 2014-11-04
ckon v0.7 (new formula)

tschaume:
See [project homepage](http://tschaume.github.io/ckon/) for details on the software. I know contributing one's own software is frowned upon but I thought this might be a nice addition for the high energy physics community using CERN's ROOT on a daily basis.


dpo:
Ping @tschaume

tschaume:
@dpo Sorry for the delay. I fixed everything you suggested. Especially the testing is meaningful now: An example source directory is cloned and `ckon dry` run over it. The output is cross-checked with the expected stdout. Nothing is actually compiled since that would require ROOT and YAML to be installed both of which aren't requirements to `ckon`.

ckon.rb
```diff
@@ -0,0 +1,84 @@
+class Ckon < Formula
+  homepage "http://tschaume.github.io/ckon/"
+  url "https://github.com/tschaume/ckon/archive/v0.7.1.tar.gz"
+  sha1 "15481a0232e7fe6584f91cd530d324e8d275f1c0"
```
dpo: Would you like to add a `head`?

tschaume:
Done. Good suggestion, thanks.

PR #1374, opened 2014-11-19
Add mlpack dependency on hdf5

macrael:
brew install mlpack failed because my machine did not have hdf5 installed. By running brew install hdf5 I was able to get brew install mlpack to work. I believe this is the correct way to add that dependency to the recipe but I am just guessing really.


dpo:
Closing in favor of #1819

moazzemi:
tnx.I had a random problem installing mlpack and installing hdf5 solved it.

PR #1467, opened 2014-12-15
ticcutils (latest) (new formula)

fbkarsdorp:


dpo:
Strangely, the test fails for me: https://gist.github.com/ed35177e95084899e66a

Also, could you please squash your commits together?

dpo:
This is not a new formula: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/ticcutils.rb

Please open a PR at Homebrew/homebrew if you believe this should be moved to Homebrew/Science.

PR #1468, opened 2014-12-15
timbl (latest) (new formula)

fbkarsdorp:
Tilburg Memory Based Learning


dpo:
Not a new formula: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/timbl.rb

PR #1469, opened 2014-12-15
timblserver (latest) (new formula)

fbkarsdorp:
server of TiMBL

PR #1471, opened 2014-12-15
mbt (latest) (new formula)

fbkarsdorp:
memory-based tagger used by frog


dpo:
Depends on #1469.

dpo:
The formula needs a version. So much has changed since this PR was opened that it would be better to open a new PR with the latest version (currently 3.2.16: https://github.com/LanguageMachines/mbt/releases). Thank you.

PR #1607, opened 2015-01-01
octave: Enable opt-out of Java dependency

rolk:
Mac OS X 10.10.1 Yosemite does not come with a Java Runtime Environment
bundled. Instead it has a sentinel program which alerts the user through
a message box that the runtime must be downloaded and installed.

The Autotools scripts detect this wrapper and allow to build the program
as if Java support was available. However, the programs are unusable
until a proper JRE is installed. This is particularily a problem when
installing other packages with Octave plugins, as the mkoctfile command
will now trip up a batch build.

This patch adds a --without-java option which is forwarded to the
configure script, as Octave is still usable without the Java support.
Thus, Octave can be installed on Yosemite without a JRE installed.

### DIFF
--- a/octave.rb
+++ b/octave.rb
@@ -65,6 +65,7 @@ class Octave < Formula
   option "without-fftw",           "Do not use FFTW (fft,ifft,fft2,etc.)"
   option "without-glpk",           "Do not use GLPK"
   option "without-hdf5",           "Do not use HDF5 (hdf5 data file support)"
+  option "without-java",           "Do not use Java Runtime Environment"
   option "without-qhull",          "Do not use the Qhull library (delaunay,voronoi,etc.)"
   option "without-qrupdate",       "Do not use the QRupdate package (qrdelete,qrinsert,qrshift,qrupdate)"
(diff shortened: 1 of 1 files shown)


octave.rb
```diff
@@ -65,6 +65,7 @@ class Octave < Formula
   option "without-fftw",           "Do not use FFTW (fft,ifft,fft2,etc.)"
   option "without-glpk",           "Do not use GLPK"
   option "without-hdf5",           "Do not use HDF5 (hdf5 data file support)"
+  option "without-java",           "Do not use Java Runtime Environment"
```
dpo: Thanks this is great. Would you mind adding the new dependency
````
depends_on :java if build.with? "java"
````
to the list of dependencies? The build will then abort if the JRE isn't installed.

rolk:
@dpo Sure. If you want the two commits squashed, just notify me before they are merged.

dpo:
The bot is happy. So yes, it would be great if you could squash your commits into one. Thanks!

rolk:
@dpo Squashed both commits into one final and force updated.

dpo:
Thanks!

PR #1615, opened 2015-01-03
Mumps: use FC as linker for shared libs.

dpo:


dpo:
@sjackman I'm getting a failed build on Travis-CI with the following:
````
==> Installing mumps dependency: cloog
Error: /home/travis/build/Homebrew/linuxbrew/opt/pkg-config not present or broken
````
Should we `brew install pkg-config` beforehand in the YAML script?

sjackman:
@dpo No, a workaround like that shouldn't be necessary. We should sort out why pkg-config isn't being correctly installed, when it is listed as a dependency of cloog.

PR #1617, opened 2015-01-03
Arpack: Omit vecLibFort on Linux.

dpo:


dpo:
@sjackman What's the recommended way to test for 64 bit architecture on Linux?

dpo:
@BrewTestBot test this please

arpack.rb
```diff
@@ -17,17 +17,19 @@ class Arpack < Formula
   depends_on :fortran
   depends_on :mpi => [:optional, :f77]
   depends_on "openblas" => :optional
-  depends_on "veclibfort" if build.without? "openblas"
+  depends_on "veclibfort" if build.without? "openblas" and OS.mac?
 
   def install
-    ENV.m64 if MacOS.prefer_64_bit?
+    ENV.m64 if OS.linux? or MacOS.prefer_64_bit?
```
sjackman: What's the purpose of this chunk, and is there an error message without it? It shouldn't be necessary on a native 64-bit Linux system, where 64-bit objects are the default.

dpo: What's the behavior of `MacOS.prefer_64_bit?` on Linux? Is `ENV.m64` always active there?

sjackman:
> What's the recommended way to test for 64 bit architecture on Linux?

Both `Hardware::CPU.is_64_bit?` and `MacOS.prefer_64_bit?` return the same value on Linux (but not on Mac). https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/os/mac.rb#L174

dpo:
Ok. It's quite odd that the `Mac` module would contain statements related to Linux.

sjackman:
Yep, there's lots of stuff in `MacOS` that has nothing to do with Mac specifically.

PR #1625, opened 2015-01-04
R: Adhere (more) to Ruby style guide.

dpo:
- Remove unnecessary "require formula"
- change single to double quotes.


dpo:
@BrewTestBot test this please

Issue #1624, opened 2015-01-04
pcl: Make error on buld target pcl_io


herx:
==> cmake -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/pcl/1.7.2 -DCMAKE_BUILD_TYPE=
==> make

[ 11%] Built target pcl_sample_consensus
cd /tmp/pcl-0N1AYE/pcl-pcl-1.7.2/macbuild/io && /usr/local/Cellar/cmake/3.1.0/bin/cmake -E cmake_symlink_library ../lib/libpcl_io.1.7.2.dylib ../lib/libpcl_io.1.7.dylib ../lib/libpcl_io.dylib
/usr/local/Cellar/cmake/3.1.0/bin/cmake -E cmake_progress_report /tmp/pcl-0N1AYE/pcl-pcl-1.7.2/macbuild/CMakeFiles  37 38 39 40
[ 11%] Built target pcl_io
make: *** [all] Error 2

READ THIS: http://git.io/brew-troubleshooting
If reporting this issue please do so at (not Homebrew/homebrew):
  https://github.com/homebrew/homebrew-science/issues

dpo:
There are several issues. First, `cmake` tries to detect if OpenMP is available. With clang, it isn't. But still the build fails because it's not finding `omp.h`. That should probably be reported upstream. Their [tutorial](http://www.pointclouds.org/documentation/tutorials/installing_homebrew.php#installing-homebrew) doesn't mention OpenMP.

The second issue is that if you build with `--cc=gcc-4.9`, it's not finding symbols of the form `Q_WS_X11`, which at QT symbols (again, `cmake` reports them as not found, but goes ahead and tries to compile and breaks down---that should probably be reported upstream too).

In discussion #1209, @fran6co suggests that building against QT5 would resolve the problem. Unfortunately, `cmake` then says
````
-- Found unsuitable Qt version "5.4.0" from /usr/local/opt/qt5/bin/qmake
````
So at this point, I think this should be discussed with the PCL folks.

fran6co:
The OpenMP thing is an old issue and it doesn't stop the build because of it.

I'm rebuilding it to check what is wrong, probably is related to the boost update to 1.57 + QT4. I'll see if I can make a patch for it for 1.7.2 and a PR for master.

fran6co:
[...]

PR #1629, opened 2015-01-05
[pcl] Fix boost 1.57

fran6co:
Fixes https://github.com/Homebrew/homebrew-science/issues/1624. Upstream patch https://github.com/PointCloudLibrary/pcl/pull/1062


fran6co:
It's erroring because it takes too long.

dpo:
@BrewTestBot test this please

dpo:
It doesn't seem to be building `vtk` `--with-qt` because the `sip` dependency isn't taken into account.

fran6co:
Isn't that an issue of vtk?

fran6co:
Weird, it has that option => https://github.com/Homebrew/homebrew-science/blob/master/vtk.rb#L47

fran6co:
It looks like it's using sip even without "--with-python"

fran6co:
What is blocking this PR? The errors in the build are not because of this formula.

dpo:
First we have to resolve the problem.

fran6co:
Found the problem, it's deep into homebrew core... Not sure what is the best course of action so I created an issue over there.

jacknagel:
The only way to work around this currently is to manually `brew install vtk --with-qt` first.

The problem is this code in the vtk formula:

```rb
  if build.with? "qt" or build.with? "qt5"
    if build.with? "python"
      depends_on "sip"
      depends_on "pyqt"
    end
  end
```

This conditional is in the class body, and therefore it is evaluated when the formula file is loaded. The `build` that is available in the class body has no way to know about the build options passed from another formula, e.g. it doesn't know about the `with-qt` being passed in by the pcl formula. It only knows about the contents of ARGV.

This problem has existed since the inception of the project.

Eventually using `build` in the class body will be deprecated, but we need new constructs that don't have this problem (e.g. https://gist.github.com/jacknagel/6731244) first.

fran6co:
Thanks @jacknagel, not sure what would be the best fix for the test system then @dpo.

jacknagel:
I should add that I've been working to make a fix possible for the better part of two years. It has required a lot of internal design changes and careful refactoring, and it is sometimes hard to keep up with other changes that are happening. We'll get there eventually, but I want to be careful not to make the same mistakes again.

fran6co:
Fixed @dpo, added sip and pyqt as a dependencies when building with qt. Couldn't think of a better work around.

fran6co:
Sorry to be a bother, but what is blocking this issue now?

dpo:
@BrewTestBot test this please

dpo:
Sorry, just lack of free time.

fran6co:
No prob, I thought that something else was blocking it. Thanks!

PR #1630, opened 2015-01-05
bioawk: Fix single quotes and add missing dependency

saketkc:


dpo:
Could you please squash your commits?

dpo:
Actually nevermind. Two commits are more logical here. Merged in 9dc509e.

PR #1631, opened 2015-01-05
TravisCI: Fix exceeding logs

saketkc:
Fixes: https://github.com/Homebrew/homebrew-science/pull/1629#issuecomment-68756002


dpo:
Ideally Travis should only build the `*.rb` files in `$TRAVIS_COMMIT`.

saketkc:
Working on it.

dpo:
Should we also systematically `apt-get install gfortran` and symlink gcc/g++ and gfortran into `$(brew --repo)/bin`? `gcc` takes forever to brew and fails to build every other time. In another project, I use [this setup](https://github.com/dpo/MUMPS.jl/blob/master/setup_travis_linux.sh) inspired by the Linuxbrew wiki, which speeds everything up dramatically. We can fix the `gcc` linux build in the mean time, but for now, it skews all the verdicts.

saketkc:
Sounds good! I will update `.travis.yml`. In my [other repo](https://github.com/saketkc/brewer/blob/master/.travis.yml), I simply rely on system installed `gcc`.

dpo:
That'd be fine too.

saketkc:
@dpo I am not sure, why the `test-bot` should consider anything not ending in `.rb` for testing. I can resort to passing changed formulae to test-bot(ending in `.rb`) rather than testing them by passing `$TRAVIS_COMMIT`, is there is no better solution.

dpo:
@saketkc I think it should only consider files ending in `*.rb`. Right now I have the impression that the bot declares failure because it can't build the formula `travis.sh`:
````
Error: No available formula for travis.sh 
````

saketkc:
I see. Should that be moved to a folder, so that it is not considered a formula inside the tap?
Like `travisCI/travis.sh`?

dpo:
Yes that might do it.

saketkc:
Looks like it is still failing, I will need to take a deeper look.

saketkc:
I think it should work now!
Check: https://travis-ci.org/saketkc/homebrew-science/builds/46276937

dpo:
Excellent! Thanks for the hard work. Funny that the brew bot would return with a failure message when there's nothing to build. Would you mind squashing?

dpo:
@saketkc The bot still tries to build `gcc` for almost every pull request. How about my suggestion to symlink the Linux `gcc` to `$(brew --repo)/bin`? We can do it with `apt-get install gcc-4.9` until we have bottles.

saketkc:
Will do so. I will have to rebase it with master before I send a new one.

sjackman:
> The bot still tries to build gcc for almost every pull request.

Though not every pull request. What's causing it to pull in GCC as a dependency?

dpo:
Anything that depends on `gcc`, on `:fortran`, or on anything that depends on those will trigger a `gcc` build.

saketkc:
I think it should be fixed now: https://travis-ci.org/saketkc/homebrew-science/builds/46517533#L456

PR #1637, opened 2015-01-06
bedtools: revert to 2.22.0

chapmanb:
There is an issue with the new naming checks in bedtools 2.22.1 that
breaks it when running intersections on GRCh37 files
(arq5x/bedtools2#166) and we're running into issues after pulling in
the new version (https://github.com/chapmanb/bcbio-nextgen/issues/716#issuecomment-68903703).
This patch reverts to the previous 2.22.0 until there is a new
release/workaround.

### DIFF
--- a/bedtools.rb
+++ b/bedtools.rb
@@ -4,16 +4,16 @@ class Bedtools < Formula
   homepage "https://github.com/arq5x/bedtools2"
   #doi "10.1093/bioinformatics/btq033"
   #tag "bioinformatics"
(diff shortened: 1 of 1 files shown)


saketkc:
Sorry about that @chapmanb. I was the one who updated it.

chapmanb:
No worries at all, thanks for keeping this up to date. We wouldn't have run into it and been able to provide feedback to Aaron and Neil without the update. I appreciate all your work.

PR #1639, opened 2015-01-06
trinity r20140717: Fix the build for Linuxbrew

sjackman:


sjackman:
@iml I fixed the build of Trinity for Linuxbrew. Can your read over this patch for me? It builds and tests okay on Linux. It builds on Mac (Yosemite) but I haven't tested because it requires Java 1.6 or 1.7 and doesn't work with 1.8.

```
Error, Trinity requires access to Java version 1.6 or 1.7.  Currently installed version is: java version "1.8.0_25"
Java(TM) SE Runtime Environment (build 1.8.0_25-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.25-b02, mixed mode)
Error: trinity: failed
```

iml:
Yea I'm using Yosemite bottles, which still fail `brew test bowtie`. I guess by this measure bowtie built from source on Yosemite is more broken than the bottle. Everything works fine with the Mavericks bottle. I think this is ready the merge, though.

sjackman:
Merged. I opened an upstream issue for Bowtie on Yosemite. https://github.com/BenLangmead/bowtie/issues/14

PR #1646, opened 2015-01-07
snpeff: update to 4.0e

heuermh:
Note also upstream fix has been merged (https://github.com/pcingola/SnpEff/commit/1d05064882190b501d9865a4d33920115dbfb497) so the workaround at lines 13-14 can be removed.


heuermh:
Thanks!

PR #1649, opened 2015-01-07
lumpy-sv 0.2.9

iml:


sjackman:
@iml Could you please add a doi and tag? Thanks!

PR #1654, opened 2015-01-08
wxmaxima: add gnuplot caveat

iml:
Remove redundant linkapp instructions.


MindTooth:
Reason for why you removed the linking of the wxMaxima.app?  Now it faults when run using `wxmaxima` from the terminal.  And after some trial and errors, I looked inside the folder to find the .app instead.

Pretty sure that one wants to exclusively run `wxMaxima.app`.

PR #1657, opened 2015-01-08
hdf5: change make check from default to optional

quantumsteve:
make check performs a lot of file IO and is especially slow on macs with hard drives. Therefore, the build-time tests should be optional. 

**Mac Pro, Mid 2012 (500 GB SSD)**
time brew install hdf5 --c++11 --with-fortran --with-fortran2003 --with-cxx
real	10m55.495s
user	11m51.631s
sys	3m37.435s

time brew install hdf5 --c++11 --with-fortran --with-fortran2003 --with-cxx --without-check
real	2m19.546s
user	7m25.132s
sys	1m34.924s

**MacBook Pro 15-inch, Early 2011 (500 GB HDD)**
time brew install hdf5 --c++11 --with-fortran --with-fortran2003 --with-cxx
real  39m39.680s
user 11m10.830s
sys   3m21.120s

time brew install hdf5 --c++11 --with-fortran --with-fortran2003 --with-cxx --without-check
real	2m28.229s
user	7m9.970s
sys	1m5.185s


quantumsteve:
The failure appears to be related to missing python libraries on the build server and not any changes in this pull request.

```
==> Installing shogun from Homebrew/homebrew-science
shogun: Unsatisfied dependency: numpy
Homebrew does not provide Python dependencies; install with:
  pip install numpy
shogun: Unsatisfied dependency: matplotlib
Homebrew does not provide Python dependencies; install with:
  pip install matplotlib
Error: Unsatisified requirements failed this build.
```

iml:
I'd be fine with this if we can add a test, like compiling a simple C/C++ program. See [armadillo](https://github.com/Homebrew/homebrew-science/blob/master/armadillo.rb#L29) for example.

dpo:
I agree. The tests are there for a reason and are important. It's always possible to build `--without-check`.

quantumsteve:
I added a test similar to armadillo where it compiles a simple c program that prints the version. 

In general I support comprehensive tests, but in this specific case the tests take a disproportionate amount of the install time.

PR #1659, opened 2015-01-08
yaggo 1.5.4: New formula

sjackman:
Yaggo is a tool to generate command line parsers for C++. Yaggo stands
for "Yet Another GenGetOpt" and is inspired by GNU Gengetopt.


dpo:
Out of curiosity, does this not belong in core?

sjackman:
@dpo Possibly yes, but the only software that uses it that I'm aware of is `jellyfish`.

PR #1664, opened 2015-01-08
sambamba 0.5.1

saketkc:


sjackman:
Just for the record.
```
==> Installing gcc dependency: cloog
Error: /home/travis/build/Homebrew/linuxbrew/opt/pkg-config not present or broken
```

saketkc:
Yes. Should be fixed by  #1631

sjackman:
Because of `brew install pkg-config`? Why doesn't it work without that? cloog correctly depends on `pkg-config`.

saketkc:
You have a point there. I was under the impression that it breaks because of a missing `pkg-config` since it is explicitly uninstalled in `.travis.yml`.

dpo:
I've noticed this too, and `brew install pkg-config` fixes it. I'm not sure why the dependency isn't taken into account properly.

sjackman:
It's possible it's a `hash` problem, and `hash -r` would resolve it, but that's a real shot in the dark. I doubt that's the case. We'd need to replicate it in an interactive shell session somehow to troubleshoot it. The workaround isn't so bad, but I don't like not understanding the root cause.

PR #1667, opened 2015-01-08
SimpleITK: update Java dependency.

dpo:


dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

simpleitk.rb
```diff
@@ -29,7 +28,7 @@ def install
     ]
     args << "-DBUILD_EXAMPLES=" + ((build.include? "examples") ? "ON" : "OFF")
     args << "-DWRAP_CSHARP=" + ((build.with? "csharp") ? "ON" : "OFF")
-    args << "-DWRAP_JAVA=" + ((build.with? "java") ? "ON" : "OFF")
+    args << "-DWRAP_JAVA=" + ((build.with? :java) ? "ON" : "OFF")
```
dpo: We've since learned that we should use `"java"` and not `:java` when calling `build.with?`.

PR #1669, opened 2015-01-08
travisCI: Move chmod to set permissions post tap-repair

saketkc:


saketkc:
Maybe, I should move it to #1631

dpo:
Yes, it would make sense in #1631.

sjackman:
Agreed.

PR #1672, opened 2015-01-09
Petsc 3.5.2 (update)

davydden:
added Superlu dependency (not to be confused with SuperLU_dist)


petsc.rb
```diff
@@ -57,9 +58,15 @@ def install
...
-      args << "--with-superlu_dist-lib=-L#{slu.opt_lib} -lsuperlu_dist"
+      slud = Formula["superlu_dist"]
+      args << "--with-superlu_dist-include=#{slud.opt_include}/superlu_dist"
+      args << "--with-superlu_dist-lib=-L#{slud.opt_lib} -lsuperlu_dist"
+    end
+
+    if build.with? "superlu"
+      slu = Formula["superlu"]
+      args << "--with-superlu-include=#{slu.opt_include}/superlu"
+      args << "--with-superlu-lib=-L#{slu.opt_lib} -lsuperlu_4.3"
```
dpo: Let's get rid of that version number in the SuperLU lib name. See #1674.

dpo:
Could you please squash? That will also trigger a new build.

PR #1674, opened 2015-01-09
SuperLU: Do not version lib name.

dpo:


dpo:
@BrewTestBot test this please

PR #1677, opened 2015-01-09
SuperLU: Fix audit warnings, add `test do` block.

dpo:


dpo:
@BrewTestBot test this please

PR #1679, opened 2015-01-09
travisCI: Fix exceeding logs using custom bash script

saketkc:
- This commit modifies .travis.yml to install the required dependencies
such as gcc, system wide, so that they do not have to be installed via
brew everytime.

- The travisCI/travis.sh bash script allows building on Travis without
  timing out and without exceeding the log limit.
  Credits to this stackoverflow answer: http://stackoverflow.com/a/26082445/756986

Signed-off-by: Saket Choudhary <saketkc@gmail.com>


.travis.yml
```diff
@@ -1,22 +1,41 @@
...
  - sudo rm -f /usr/local/clang-3.4/bin/llvm-config
+
+ # Install gcc-4.9 system wide[Default is gcc-4.6]
+ - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+ - sudo apt-get update -qq
+ - sudo apt-get install -y libyajl-dev libxml2-dev libxqilla-dev
+ - sudo apt-get install -y libstdc++-4.9-dev
+ - sudo apt-get install -y g++-4.9
+ - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.9 90
+ - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.9 90
```
dpo: Let's also do it with `gfortran-4.9`. Otherwise, any formula that `depends_on :fortran` will build `gcc` from source.

saketkc: Sure, thanks!

saketkc:
Do not merge, yet.
Still testing. 
https://travis-ci.org/saketkc/homebrew-science/builds/46515977

saketkc:
If https://travis-ci.org/saketkc/homebrew-science/builds/46517533 succeeds, it can be merged.

saketkc:
The build failed because I was doing `cat $BUILD_OUTPUT`.
Reverted to `tail -1000 $BUILD_OUTPUT`. I think this is good to merge.

saketkc:
Btw, `lampps.rb` is broken since that version is no longer available here: http://lammps.sandia.gov/tars/?C=M;O=D

dpo:
Many thanks!

dpo:
@saketkc When I look at #1681, it couldn't possibly have built on Linux. The downside of the bash script is that we don't see any detail any more. Now that the compilers are pre-installed, perhaps we should go for a finer workflow. Something like:
````
for file in $changed_files
do
    if [[ ${file: -3} == ".rb" ]]
    then
        brew install $(brew deps $file) >> $BUILD_OUTPUT 2>&1
        brew test-bot -v $file  # Show the output of this build!
    fi
done
````
What do you think?

saketkc:
@dpo I have tried to address this in #1687. To me it seems a good solution for now.

I would wait till https://travis-ci.org/saketkc/homebrew-science/builds/46698884 finishes.

PR #1683, opened 2015-01-11
snpEff: upgrade to 4.1

chapmanb:
This follows @heuermh upgrade to 4.0 to the latest 4.1 release from last
week. The 3.6 to 4.0 transition is pretty tricky as their are database
incompatibilities and the snpEff databases are not versioned as of 4.0.
4.1 introduces versioning so is an easier upgrade path from 3.6 and
should make things generally easier going forward.

### DIFF
--- a/snpeff.rb
+++ b/snpeff.rb
@@ -2,17 +2,9 @@ require 'formula'
 
 class Snpeff < Formula
   homepage 'http://snpeff.sourceforge.net/'
-  url 'https://downloads.sourceforge.net/project/snpeff/snpEff_v4_0_core.zip'
-  version '4.0e'
-  sha1 '969fa41cd774a33629f63938f665ba1f83f9adcc'
-
-  bottle do
-    root_url "https://downloads.sf.net/project/machomebrew/Bottles/science"
-    cellar :any
-    sha1 "d663e17fc44066a66ae8f05a99d2b7e4fd1fdda8" => :yosemite
-    sha1 "fdbf0db244297f35281850bb2381d7abbba0392e" => :mavericks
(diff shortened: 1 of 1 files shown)


heuermh:
Thank you, @chapmanb.

Sorry, perhaps I should have pinged you on the upgrade to 4.0 pull request; I didn't notice any problems when upgrading.  Was it that 4.0 tried to hit existing 3.6 databases and would break?  If so I wouldn't have run into that, since I'm always installing to a new machine/image.

snpeff.rb
```diff
@@ -2,17 +2,9 @@
...
-  version '4.0e'
-  sha1 '969fa41cd774a33629f63938f665ba1f83f9adcc'
-
-  bottle do
-    root_url "https://downloads.sf.net/project/machomebrew/Bottles/science"
-    cellar :any
-    sha1 "d663e17fc44066a66ae8f05a99d2b7e4fd1fdda8" => :yosemite
-    sha1 "fdbf0db244297f35281850bb2381d7abbba0392e" => :mavericks
-    sha1 "25acc2120b891e10c79530662578e24dc6ff20b8" => :mountain_lion
-  end
```
sjackman: In the future please don't remove the `bottle` section. The BrewTestBot will handle updating the bottles.

chapmanb:
Shaun -- thanks so much for merging and sorry about the mistake in removing the bottle section. I thought the bot magically merged it all back in.

Michael -- no worries at all, I appreciate the updates. I missed the issue too as it seems to only trigger in some evaluation cases. It was with older 3.6 databases and 4.0 code; the tricky part is that there isn't a clean way to know what version of snpEff a database came from pre-4.1. This should all be much easier with 4.1+ updates. Thanks again.

PR #1688, opened 2015-01-12
Petsc (update)

davydden:
Made superlu_dist and parmetis dependencies only when built with mpi


dpo:
Not sure I understand, `:mpi` isn't `:optional`.

davydden:
you are right! I was too quick to change it and thought that non-mpi version is also there.

PR #1701, opened 2015-01-14
simseq (new software)

erasche:
SimSeq simulates NGS data. Unfortunately it hasn't been updated in ~3 years, and I know the brew project generally discourages old/unmaintained software. However, it's the only NGS sequencing data simulator I'm aware of*, and I was learning to write brew formulae so I figured I might make a PR in case it's interesting to others.

I haven't added test cases, as it randomly generates the data (with no option for seeding).

If it isn't interesting/useful, please feel free to decline the PR!

\* I've since found [ART](http://www.niehs.nih.gov/research/resources/software/biostatistics/art/) which looks a lot more promising


saketkc:
https://github.com/lh3/wgsim

erasche:
Closing due to lack of interest/not worth time invested to resolve compilation issues (it depends on java7), and to try and test something which doesn't produce reproducible results.

PR #1704, opened 2015-01-15
Suite-sparse (4.4.2)

davydden:
Update to 4.4.2. According to authors this fixes a rarely
occuring but serious bug that exists in CHOLMOD 3.0.0 through 3.0.3.,
and in SuiteSparse 4.3.0. through 4.4.1. The bug can cause a segfault.


dpo:
I submitted #1314 before 4.4.2 came out. We should upgrade because of the bug fix. The problem is that it breaks `octave`. Let me see if the patch is obvious.

davydden:
I noticed you removed `depends_on "homebrew/versions/gcc43" => [:build, "enable-fortran"] if build.with? "matlab"`. Perhaps this should be removed here as well. I do not have Matlab, so I can't test.

davydden:
p.s. there is `octave` `3.8.2` already (August 2014). If something is still broken in `3.8.1`, perhaps it is fixed in `3.8.2`?

dpo:
Good point. I'll have a look. I'm going to close this if you don't mind so I don't get confused. It seemed easier to update #1314 to version 4.4.2.  Thanks for the update!

davydden:
np

PR #1707, opened 2015-01-16
update snpEff sha1 value

mdeboth:
snpEff_v4_1_core.zip was modified 2015-1-14 and has a different sha1 value. It was breaking bcbio-nextgen downstream for me. 
https://github.com/chapmanb/bcbio-nextgen/issues/724

PR #1708, opened 2015-01-16
snpEff: update to 4.1a

chapmanb:
The snpEff 4.1 release download changed to 4.1a, breaking checksums
(chapmanb/bcbio-nextgen#724). This updates to the latest version. I'll
also raise an issue upstream about maintaining a consistent download
target.


chapmanb:
Sorry I missed #1707 before submitting. Matt is too quick for me. The only advantage of this one is that it includes the 'a' subrelease in the version, which is the underlying issue behind the updated download but either pull will fix the underlying issue. Thanks again.

Issue #1716, opened 2015-01-17
travisCI does not build for pull requests


saketkc:
I guess after merging https://github.com/Homebrew/homebrew-science/pull/1679, the travis build is green  for all requests. Probably because of https://github.com/Homebrew/homebrew-science/blob/master/travisCI/travis.sh#L29

```
export changed_files=`git diff-tree --no-commit-id --name-only -r $TRAVIS_COMMIT`
```
I am not really sure, why this does not work out. It could be related to https://github.com/travis-ci/travis-ci/issues/1719 and https://github.com/travis-ci/travis-ci/issues/2666 .

dpo:
Yes I've noticed. One possibility would be to use `set -evx` at the top of `travis.sh` so that all commands and what they evaluate to are printed on the console.

saketkc:
I am working on a solution. Will update.

saketkc:
Do you suggest changing to the `brew test-bot` method? It is I think okay to assume the travis files would not be changed that often.

The reason I say this is because `test-bot` handles what I am trying to use, it fetches the difference between the origin and upstream branch.

dpo:
Were you able to test it? I wish the test-bot build could be verbose. How do we get the log in case of failure?

saketkc:
Closed by #1717

saketkc:
This was due to https://github.com/blog/1941-organization-approved-applications

Mike has re-enabled it.

PR #1717, opened 2015-01-17
travisCI: Revert to test-bot testing strategy

saketkc:
This commit reverts to the test-bot based testing strategy,
drifting away from strategy used in #1679 as it did not
build PRs. The only downside being it will also test
changes in files which are not formulae(such as .travis.yml)
but it is safe to assume, these will not change as frequently.

See #1716


dpo:
Were you able to test it? I wish the test-bot build could be verbose. How do we get the log in case of failure?

saketkc:
This currently dumps the last 3Mb of the log. Would that be sufficient?

dpo:
If the `test-bot` script does a verbose build, yes, that's probably enough. The Travis limit is 4Mb I believe.

saketkc:
I have updated it. It's a bit more intelligent now. However, the script just compares the difference between current commit and it's parent, so if you want to build each of your commit, you need to push one commit at a time.

See https://github.com/saketkc/homebrew-science/pull/7

dpo:
We'll improve this as we go. Thanks for your hard work!

saketkc:
For some reason, the PRs do build. But github doesn' show them. I do not know why.
https://travis-ci.org/Homebrew/homebrew-science/pull_requests

PR #1721, opened 2015-01-18
arpack 3.2.0

iml:


dpo:
Was this closed because of the failed test?

iml:
They moved the test binaries into a `make check` which failed (locally) with `bug_1323_single`. It wasn't obvious why that failed and not `bug_1323_double`.

PR #1740, opened 2015-01-22
mlpack 1.0.12

saketkc:


dpo:
Are you a `mlpack` user? Do you have an opinion on #1374?

saketkc:
I am pretty new. Thanks for pointing that out. It builds fine on my system, because probably I do have `hdf5` installed.

Let's wait for the build to finish. I can update the formula accordingly. And since github no longer shows us travis status, it is here: https://travis-ci.org/Homebrew/homebrew-science/builds/47860631

saketkc:
The failure on jenkins is due to flloating point precision. I will report that upstream.

saketkc:
And yes, that `hdf5` issue is persistent too

dpo:
Looks like it should be added to `armadillo`.

saketkc:
I will send a new one, for `armadillo` first.

saketkc:
Looks like it is going to require fixing couple of packages, first of them being `arpack-ng`:
https://travis-ci.org/saketkc/homebrew-science/builds/47992713#L349

dpo:
Ah, I think we have to `apt-get install libblas-dev liblapack-dev` on the Linux bot, no?

PR #1746, opened 2015-01-23
harvest-tools 1.1 (new formula)

iml:
Part of the Harvest suite and build dependency of gingr and parsnp.


sjackman:
Odd Mountain Lion failure:
```
gcc -Isrc -I/usr/local/opt/protobuf/include -mmacosx-version-min=10.7 -stdlib=libc++ -o harvesttools src/harvest/AnnotationList.o src/harvest/harvest.o src/harvest/HarvestIO.o src/harvest/LcbList.o src/harvest/parse.o src/harvest/PhylogenyTree.o src/harvest/PhylogenyTreeNode.o src/harvest/ReferenceList.o src/harvest/TrackList.o src/harvest/VariantList.o src/harvest/pb/harvest.pb.o libharvest.a /usr/local/opt/protobuf/lib/libprotobuf.a -lstdc++ -lz -lm -lpthread
  "google::protobuf::MessageFactory::InternalRegisterGeneratedFile(char const*, void (*)(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&))", referenced from:
      protobuf_AddDesc_harvest_2fpb_2fharvest_2eproto() in harvest.pb.o
```

iml:
Yea not sure, maybe libstdcxx vs libcxx conflict with protobuf?

sjackman:
Yeah, I think something like that. It looks like a standard library conflict. I have no Mountain Lion with which to test.

PR #1750, opened 2015-01-23
SuperLU_dist: Explict MPI compilers, tests.

dpo:
Also remove unnecessary "require formula".

This makes it possible to build SuperLU_dist against MPICH.


dpo:
@saketkc Travis is again trying to build `gcc`. Any idea why? Also it looks like Linuxbrew will pick up an existing install of MPI, so we should consider `apt-get install`ing either OpenMPI or MPICH to speed things up.

saketkc:
I will take a look.

dpo:
Thanks!

saketkc:
This is due to https://github.com/Homebrew/homebrew-science/blob/master/travisCI/travis.sh#L50

dpo:
What's the solution? `| grep -v gcc`?

dpo:
Or we should ask the Travis guys for a much larger timeout.

saketkc:
Give me a day. I think we can make it more intelligent.

saketkc:
@dpo: https://travis-ci.org/saketkc/homebrew-science/builds/48125625
Does this look good?

See Lines 328-333

saketkc:
Sorry, not there yet.

PR #1758, opened 2015-01-24
Scalapack (update)

davydden:
output the summary of tests to the user

### DIFF
--- a/scalapack.rb
+++ b/scalapack.rb
@@ -40,7 +40,11 @@ class Scalapack < Formula
     mkdir "build" do
       system 'cmake', '..', *args
       system 'make all'
-      system 'make test' if build.with? 'check'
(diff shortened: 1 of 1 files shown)


dpo:
Is this really useful? The build fails as soon as a test fails, and the detailed results may be found in `~/Library/Logs/Homebrew/scalapack`.

davydden:
I am not sure that the build will fail if some tests fail in general. Did you experience that? 
If they fail to compile - this is one story, however if the output of the test fails against some blessed output - the build will happily go on, IMHO.

dpo:
At least for `scalapack` I'm sure the build fails if a single test fails. Sometimes building `--with-shared-libs` is enough to make it fail.

davydden:
ok. did not know that.

PR #1759, opened 2015-01-24
ParMETIS: Build against brewed METIS lib.

dpo:
- Remove unnecessary require formula.


dpo:
@saketkc I think we need to add `.linuxbrew/lib` to `LD_LIBRARY_PATH` don't we?

PR #1763, opened 2015-01-25
travisCI: Silent apt-get installs, tap hb/python, set paths

saketkc:


dpo:
Looks great! Should we also `apt-get install` one of OpenMPI and MPICH? The `:mpi` dependency seems to pick it up and that could greatly speed things up.

saketkc:
Done.

PR #1767, opened 2015-01-26
Multi-threaded built of nfft (optional)

cmarquardt:
The nfft library offers to build a multi-threaded version along with the serial one, but the former is implemented using OpenMP - so it can't be build with clang at present.

I've changed the nfft formula to allow both the serial and multi-threaded build with `--with-openmp --cc=gcc-x.y`. I also modified the test to link against the multithreaded version if available.

The default behaviour remains unchanged; it builds the serial library with clang. If `--with-openmp`is requested, but the compiler is still clang, a warning is given (`opoo ...`), and the serial version is build instead.


nfft.rb
```diff
@@ -14,24 +16,80 @@ class Nfft < Formula
 
   def install
     args = %W[--disable-debug --disable-dependency-tracking --prefix=#{prefix}]
+    if build.with? "openmp"
+      if ENV.compiler == :clang
+        opoo "Clang does not support OpenMP. Compile with gcc (--cc=gcc-x.y) if this is not acceptable."  
+      else
+        args << "--enable-openmp"
+      end
+    end
```
dpo: Why not simply enable openMp if building with gcc?

cmarquardt: Mostly because I don't know what the preferred policy Homebrew has with respect to OpenMP; the short paragraph on OpenMP in the [CONTRIBUTIONS](https://github.com/Homebrew/homebrew-science/blob/master/CONTRIBUTING.md) file doesn't say that much. For nfft itself, the multi-threaded build doesn't happen by default; one has to explicitly request it with an option to configure.  

But I'm happy to automatically do the threaded built as soon as the chosen compiler is not clang.

Hmm - are there still versions of gcc available in Homebrew which do not support OpenMP?

Thanks for your advice,

   Christian.

cmarquardt: Implemented dpo's suggestion...

cmarquardt:
My apologies if the question is stupid...

The build in Travis CI fails; as far as I understand from the log, the following happens:

 - the build happens with gcc on linuxbrew
 - a multithreaded bottle is build (apparently successfully)
 - the multi-threaded test fails because the compiler can't find fftw3.h, which is part of the (required) fftw3 library
 - fftw3 is (re-?) installed after the test fails

Shouldn't the installation of the required fftw3 occur before the build? Is that a hickup in the CI, or do I misinterpret the logging?

Many thanks,

   Christian.

saketkc:
Yes, the travis setup is buggy right now. We are working on it.

nfft.rb
```diff
@@ -14,24 +12,78 @@ class Nfft < Formula
...
+          fftw_cleanup_threads();
+          return 0;
+        }
+      EOS
+      system ENV.cc, "test.c", "-I#{include}", "-L#{lib}", "-lnfft3_threads", "-L#{HOMEBREW_PREFIX}/lib", "-lfftw3_threads", "-o", "test"
+      system "./test"
+    end
   end
 
+  def caveats
```
dpo: This only applies when building with `clang`.

nfft.rb
```diff
@@ -14,24 +12,78 @@ class Nfft < Formula
...
+    A multi-threaded version of the NFFT library can be build with
+    Homebrew's GNU C compiler, using
+
+      brew install --cc=gcc-x.y nfft
+
+    which will create both serial and multi-threaded NFFT libraries.
+
+    Linking against the serial libraries:
+
+       $CC ... -L`brew --prefix`/lib -lnfft -lfftw3
```
dpo: Please use `-L#{opt_lib} -lnfft -lfftw3`

dpo: Also `$CC` should probably be `#{ENV["CC"]}`

cmarquardt: Thanks!

$CC was meant symbolic; and CC isn't necessarily defined... I've replaced it with simply gcc.

I can can squash the commits, btw...

iml:
Keep in mind brew doesn't remember anything in `ENV` outside the install method, so if you build with `--cc=gcc-4.9` it will run the clang test anyway (unless you pass the compiler to the `test` command, or have it set with `HOMEBREW_CC`).

cmarquardt:
I've reduced the testing to just a serial build which should always work. Also squashed the commits into a single one.

The build still fails as it can't find the fftw3 header files, although the fftw3 formula is required. Maybe still an issue in the Travis-Ci setup? Like the links from the homebrew's include directory into fftw's cellar being missing?

saketkc:
I can check the travis setup. Give me a day.

cmarquardt:
Sorry for pushing - any news on the CI?

saketkc:
No. Just ignore it for the moment.

PR #1769, opened 2015-01-27
Cytoscape 3.2.0 (new formula)

kozo2:
Cytoscape is an open source software platform for visualizing complex networks and integrating these with any type of attribute data. A lot of Apps are available for various kinds of problem domains, including bioinformatics, social network analysis, and semantic web.


kozo2:
I will re-send this pull request from kozo2:cytoscape branch.

PR #1771, opened 2015-01-28
root: recommended dependency gsl, enable-mathmore

tschaume:
To avoid annoying build errors when using ROOT's MathMore library, I'd suggest to simply make `gsl` a recommended dependency. Any concerns?


root.rb
```diff
@@ -55,6 +56,10 @@ def install
       --mandir=#{man}
     ]
 
+    if build.with? "gsl"
+      args << "--enable-mathmore"
+    end
+
```
dpo: Please make this a  one liner: `args << "--enable-mathmore" if build.with? "gsl"`.

PR #1774, opened 2015-01-28
Opium 3.8

davydden:
Opium is a software to generate pseudopotentials for ab initio calculations in condensed matter physics, computational chemistry and material science.

### DIFF
--- a/opium.rb
+++ b/opium.rb
@@ -1,0 +1,77 @@
+class Opium < Formula
+  homepage "http://opium.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/opium/opium/opium-v3.8/opium-v3.8-src.tgz"
+  sha256 "edee6606519330aecaee436ee8cfb0a33788b5677861d59e38aba936e87d5ad3"
+
+  depends_on :fortran
+  depends_on "veclibfort"
+  depends_on "numdiff"
+
+  patch :DATA
+
+  def install
+    system "./configure", "LDFLAGS=-lveclibfort"
+    cd "src"
+    system "make", "all-subdirs"
+    system "make", "opium"
+    cd ".."
+    bin.install "opium"
+    (share/"opium-testdir").install Dir["testdir/*"]
+  end
+
+  def caveats; <<-EOS
+    Example input and output files are located in #{share}/opium-testdir
+    For a discussion on failing test fe.cpi and pt.cpi see
+    https://sourceforge.net/p/opium/mailman/message/33287478/
+    EOS
+  end
+
+  # TODO: increase execution time, otherwise tests fail at the very end or greping
+  test do
+    cd "#{share}/opium-testdir"
+    system "./run"
+    system "./comp", "cpi"
+    ohai `grep "differs from" "OUT"`.chomp
+    system "./clean"
+  end
+end
+
+__END__
+diff --git a/src/Makefile b/src/Makefile
+index 09b3356..5b2afe6 100644
+--- a/src/Makefile
++++ b/src/Makefile
+@@ -2,7 +2,7 @@ SHELL=/bin/sh
+ 
+ include ../makevars
+ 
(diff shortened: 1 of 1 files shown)


dpo:
Any idea why the test fails on Yosemite?

davydden:
does it run completely for you?

For me I have differences in `fe.cpi`, and `pt.cpi`. I wrote to the mailing list to if it is indeed some issue or `blessed` files are not up-to-date

davydden:
Note that also removed compiling `lapack` and substituted it with `veclibfort`, but this shouldn't be an issue anyway.

dpo:
It builds fine for me under 10.9 and 10.10, and the tests run, but I also get
````
==> +++  File "fe.cpi" differs from file "results/fe.cpi"
+++  File "pt.cpi" differs from file "results/pt.cpi"
````
on both platforms.

**Comment by davydden** [comment]
well, i asked developers, so far no answer: https://sourceforge.net/p/opium/mailman/message/33287478/

**Comment by davydden** [comment]
just contacted the project coordinator E.J. Walter, hope there will be a resolution to this issue.

**Comment by davydden** [comment]
Here is the comment from the project coordinator E.J.Walter:
>The differences you see for fe.cpi and pt.cpi are most likely due to the use of gfortran on the mac.  My tests use f2c + fort on the mac since the gfortran compiler (on both mac and linux) is still a little unstable and causes some (usually small) numerical differences to creep up.  If you run configure and gfortran is found as the only fortran compiler a warning stating this is issued.
>
>I am a little surprised by the size of the of the error for fe.cpi.  
>
>I would be very happy to see the opium package to be part of homebrew-science.  If I get some time, I will look into the numerical issues surrounding gfortran on mac and linux.

so i guess i will modify Caveats accordingly and it should be good to go.

**Comment by dpo** [comment]
Not sure what he means by "fort" or by "gfortran ... is still a little unstable".

**Comment by davydden** [comment]
i suppose what he means is that `gfortran`, being a free compiler, is not as reliable to produce the correct machine code as, say, `ifort` (Intel Fortran Compiler), which you have to buy. It is also related to which implementation of BLAS/Lapack is used.

File: opium.rb [hunk 1]
```diff
@@ -0,0 +1,77 @@
+class Opium < Formula
+  homepage "http://opium.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/opium/opium/opium-v3.8/opium-v3.8-src.tgz"
+  sha1 "9649e79e2393fe35e7575ab31dffd2b1f0437e71"
```
dpo: Would you mind updating this to a `sha256`?

opium.rb
```diff
@@ -0,0 +1,77 @@
...
+  patch :DATA
+
+  def install
+    system "./configure", "LDFLAGS=-lveclibfort"
+    cd "src"
+    system "make", "all-subdirs"
+    system "make", "opium"
+    cd ".."
+    bin.install "opium"
+    (share/"opium-testdir").install Dir["testdir/*"]
```
dpo: Does `opium` not have a `make install`?

davydden: funny enough - not.

dpo:
@BrewTestBot test this please

davydden:
I am working on a Formula for `ape` pseudo-potential project http://www.tddft.org/programs/APE/ (it's also on MacPorts https://trac.macports.org/browser/trunk/dports/science/ape/Portfile). It does pretty much the same thing (at least to my understanding), but looks like a more active project. The current version does not compile against `libxc` 2.2.0. I notified developers and the promised to have a look at it. `homebrew-science` could have both and remove either if it is not used much.

dpo:
Should this still be pulled, or is it superseded by `ape`?

davydden:
I think this PR can be closed without pulling. I will keep the branch, just in case...

dpo:
Ok thanks. Feel free to reopen if need be.

PR #1775, opened 2015-01-29
update statismo's formula for v0.10.2

arnaudgelas:


dpo:
Seems like the `:python` dependency isn't picked up:
````
make[2]: *** No rule to make target `/usr/local/Cellar/python/2.7.8_2/Frameworks/Python.framework/Versions/2.7/lib/libpython2.7.dylib', needed by `bin/ITKHeaderTest2'.  Stop.
````

arnaudgelas:
indeed python is needed for building tests (which are not installed), should I turn it off (i.e. BUILD_TESTING) or should I add dependency to python?
I'd rather go for for the first option...

dpo:
What surprises me is that `:python` is a dependency of both `vtk` and `insighttoolkit`. It might be because they're tagged as `:build` dependencies only.

arnaudgelas:
if I am not mistaken ''':python''' is needed for python wrapping for both vtk and itk. For ITK, since BUILD_TESTING is turned off, there is no need of python... Here I could follow the same pattern and turn off BUILD_TESTING too

dpo:
That seems to make sense.

arnaudgelas:
ok, I'll squash these 2 commits together

arnaudgelas:
see https://github.com/Homebrew/homebrew-science/pull/1778

PR #1779, opened 2015-01-29
Trilinos: comprehensive revamping.

### LINKED ISSUE #887: Trilinos failed to build on 10.9.2
Hi,  I could not build trilinos.   Please see the following gist link for more information.  Thank you in advance for your help.

https://gist.github.com/anonymous/d198b172b02be96338da

dpo:
Many thanks to @davydden for his contributions.

Closes #887.


dpo:
Well I'm not sure what this is about:
````
Error: /usr/local/opt/pkg-config not present or broken
````

File: trilinos.rb [hunk 1]
```diff
@@ -1,52 +1,228 @@
...
+  # are commented out with #- and failure reasons are documented.
+
+  option "with-cholmod", "Build with Cholmod (Experimental TPL) from suite-sparse"
+  #-option "with-csparse", "Build with CSparse (Experimental TPL) from suite-sparse" # Undefined symbols for architecture x86_64: "Amesos_CSparse::Amesos_CSparse(Epetra_LinearProblem const&)"
+
+  depends_on :mpi           => [:cc, :cxx, :recommended]
+  depends_on :fortran       => :recommended
+  depends_on :x11           => :recommended
+
+  depends_on :python        => ["numpy", :recommended]
```
iml: If this was meant to check for the NumPy module it should be `depends_on "numpy" => :python`

dpo: That's right. Thanks!

iml:
Ah you will probably run into problems with `:python` and module deps since the bot will use Homebrew Python, which doesn't include any of those modules. So the 'Homebrew' way is to avoid `:python` and only link against system Python/NumPy by default. Or depend on homebrew/python/numpy.

dpo:
I suppose I could `depend_on "homebrew/python/numpy" as you suggest. That's what `opencv.rb` does. Will the bot tap `/homebrew/python`?

iml:
Yes the bot will work with that cross-tap dependency, and bottle against it. I felt is was appropriate with OpenCV because compiles its bindings in a module at build-time and users were unlikely to want Apple's old Python/NumPy for development. If this can use any old NumPy, however, system Python should suffice.

dpo:
Trilinos can probably use an older NumPy, but I'd really like to produce bottles including the Python module for this. On a laptop that doesn't have a SSD, this can take ~2 hours to build, not counting the dependencies.

dpo:
`brew deps doxygen --with-dot` knows that `libpng` is a dependency (via `graphviz`), but somehow `brew install --only-dependencies` doesn't include `libpng`:
````
Error Message

failed: brew install --only-dependencies --verbose --build-bottle trilinos

Stacktrace

        ==> Installing trilinos from Homebrew/homebrew-science
==> Installing dependencies for trilinos: gmp, mpfr, libmpc, isl, cloog, gcc, openssl, libevent, open-mpi, readline, sqlite, gdbm, python, homebrew/python/numpy, pcre, swig, xz, cmake, pkg-config, colpack, adol-c, boost, cppunit, doxygen, hwloc, metis, mumps, szip, hdf5, netcdf, parmetis, veclibfort, scalapack, scotch, tbb, suite-sparse, superlu, superlu_dist, hypre, glpk, glm
````

**Comment by jacknagel** [comment]
```
==> Installing trilinos dependency: doxygen
Error: /usr/local/opt/libpng not present or broken
Please reinstall libpng. Sorry :(
```
An error like this is almost always because somewhere in the dependency tree, a dependency is being declared conditionally. In this case, it's [here](https://github.com/Homebrew/homebrew/blob/c3796cd121c2cb8e9a04a620ba0ebd2c6806e838/Library/Formula/doxygen.rb#L19).

I've fixed it in Homebrew/homebrew@786928034027ab1c21c88ef62d884fab954c8aa0. Update this to use `with-graphviz` instead of `with-dot` and it should work.

jacknagel:
By the way, I noticed that symbols (rather than strings) are being passed to `build.with?` in this PR. This only works by accident of implementation, and I'm not sure it should be relied on at this point.

dpo:
Many thanks @jacknagel! Is there a consensus about constructs of the form
````
option "with-feature"
depends_on "package" if build.with? "feature"
````
It always seems redundant to me, and in this case actually caused trouble. Should those systematically be simplified to
````
depends_on "package" => :optional
````

To be sure I understand your second point, should we always use `build.with? "fortran"` and `build.with? "mpi"` instead of `:fortran` and `:mpi`? I typically use symbols because there are no formulas called `fortran` or `mpi` (and `:mpi` can represent either `open-mpi` or `mpich`). It seems like using symbols would be consistent with `depends_on :mpi`, but I'm sure I don't understand the fine details.

davydden:
does one need to get rid of `require "formula"`, `audit --strict` shows that it is now unnecessary.

jacknagel:
>Is there a consensus about constructs of the form [...] It always seems redundant to me, and in this case actually caused trouble.

In cases where it is reasonable for the option name to be `with-<dependency name>`, then we should always be using `=> :optional` and `=> :recommended`. There are some cases where it doesn't quite make sense, though, so there are is some remaining usage of `depends_on ... if`. Eventually we will have a way to declare optional dependencies that don't match the option name exactly, and then we can get rid of the rest.

>To be sure I understand your second point, should we always use build.with? "fortran" and build.with? "mpi" instead of :fortran and :mpi?

We never intentionally supported passing symbols to `build.with?`. We always pass strings for these in core, so I think it would be good to match that here for consistency across repos.

dpo:
@davydden We removed `require "formula"` a while back. Are you looking at an old version?

davydden:
@dpo sorry, i did not notice that it is indeed removed.

dpo:
@BrewTestBot test this please

dpo:
@mikemcquaid Is there a 60 minutes timeout on the bot? This formula can take almost double that amount of time, which is why it'd be great to bottle it.
````
==> brew bottle --rb trilinos --root-url=https://downloads.sf.net/project/machomebrew/Bottles/scienceBuild timed out (after 60 minutes). Marking the build as aborted.
````
Is there a way to request a longer timeout?

mikemcquaid:
There is. I'm afraid there's not really a current workaround as the bot is a shared resource and it stops being a shared resource when there's builds taking too long :sob:

PR #1783, opened 2015-01-30
Updating SYMPHONY to latest release, removing fails_with for clang, and ...

tkralphs:
...removing running of fulltest

OK, let's see if this works.


dpo:
Spic and span! Thank you.

PR #1789, opened 2015-02-03
salmon 0.2.7: New formula (does not compile)

sjackman:
Salmon is a tool for transcript quantification from RNA-seq data.


sjackman:
See https://github.com/kingsfordgroup/sailfish/issues/69

sjackman:
Possible workaround is to set `-DFETCH_BOOST=TRUE`. See https://github.com/kingsfordgroup/sailfish/blob/master/README.md#installation

PR #1807, opened 2015-02-09
armadillo: Allow build with hdf5

saketkc:
As per https://github.com/Homebrew/homebrew-science/pull/1740#issuecomment-70956142


saketkc:
Have been trying to get this and new `mlpack` build on travis.
https://travis-ci.org/saketkc/brewer/builds/50284325

@dpo Any clues why boost fails to install? (https://travis-ci.org/saketkc/brewer/builds/50284325#L1394)

dpo:
@saketkc Something seems to be wrong with gcc/g++:
````
g++-4.9: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
````
See https://travis-ci.org/saketkc/brewer/builds/50284325#L1319. Perhaps we should install gcc-4.8 instead of 4.9 on Linux?! I'm frankly not quite sure.

**Comment by saketkc** [comment]
Oh yes, missed that! I have been using `gcc-4.9` for quite sometime. In fact I brewed one this afternoon on a new Ubuntu machine.

But yes, worth a shot again. I think I have screwed the `travis` setup more by those PRs.

File: armadillo.rb [hunk 1]
```diff
@@ -12,13 +12,24 @@ class Armadillo < Formula
     sha1 "64ed43d130bfab2333c39ceb64eb2adea0506e2a" => :mountain_lion
   end
 
+  option "with-hdf5", "Enable the ability to save and load matrices stored in the HDF5 format"
+
   depends_on "cmake" => :build
   depends_on "arpack"
+  depends_on "openblas" => :build
```
iml: Is this for the Linux build?

saketkc: Yes. Should I add  a `OS.linux?` clause?

dpo:
Travis builds are a real burden to debug. Many thanks for your hard work here!

armadillo.rb
iml: Yeah, otherwise we're adding a whole different BLAS on Mac. Should perhaps be a runtime dependency as well, unless it's statically linked on Linux?

saketkc:
Fixed, travis still failing.

saketkc:
@dpo Possible culprit: RAM size[1]. I have all reasons to believe so, because I have seen it happening sporadically.

The trick to avoid reinstalling `gcc` is to create empty directory `Cellar/gcc`. Fails with `gcc-4.6` too as well.

https://travis-ci.org/saketkc/brewer/builds/50325703
[1] http://stackoverflow.com/questions/13103396/c11-g-4-7-internal-compiler-error

saketkc:
On a second though `4.5GB` of a RAM isn't less, so I am not really sure.

saketkc:
Can this be merged? I have another PR in queue that depends on this.

iml:
Ya one sec.

PR #1808, opened 2015-02-10
butterflow 0.1.5 (new formula)

dthpham:
A tool that lets you make slow motion and smooth motion videos from the command line.


dthpham:
Travis build is failing because `gcc-4.9.2` fails to build.

saketkc:
Yes, it's a known issue. I am unable to build your formula on linux because `opencv` seems to be broken at the moment.

dpo:
Should this be in science or in core?

dthpham:
@dpo I submitted this to core before but it was rejected because of a dependency to homebrew-science: https://github.com/Homebrew/homebrew/pull/34205.

dpo:
@dthpham Ok I see thanks.

PR #1813, opened 2015-02-11
opencv: Fix for linuxbrew

saketkc:
Testing if this goes through on Jenkins.


saketkc:
Not sure why python binding fails on devel.

PR #1821, opened 2015-02-12
htslib, samtools, bcftools: update to 1.2

chapmanb:
Hi all;
This pulls the latest releases in the samtools/bcftools/htslib releases. Thanks much.


dpo:
One formula per pull request, please.

iml:
It's OK for these to be bumped together since they are each coupled to htslib.

PR #1839, opened 2015-02-16
bless 0.24: New formula

sjackman:
BLESS: Bloom filter-based error correction solution for
high-throughput sequencing reads


sjackman:
I believe this error is due to google-sparsehash being compiled with clang and bless being compiled with GCC.

```
                 from count_solid_kmers.cpp:1:
/usr/local/include/sparsehash/sparse_hash_map:98:27: error: 'hash' in namespace 'std' does not name a template type
           class HashFcn = SPARSEHASH_HASH<Key>,   // defined in sparseconfig.h
                           ^
```

PR #1845, opened 2015-02-17
xrmc 6.5.0 (new formula)

tschoonj:
This is a formula for XRMC, a tool for the prediction of X-ray imaging and spectroscopy experiments using Monte Carlo simulations.

The build appears to work fine, but I get the following strange message at the begin:

```
```

which is surprising because there is no Fortran compiler required by the Formula: perhaps this is because of the OpenMP requirement?

At the end I get the following message:

```
Warning: xrmc dependency gcc was built with a different C++ standard
library (libstdc++ from clang). This may cause problems at runtime.
```

It seems harmless though: the test runs fine...

Please consider this package for inclusion in homebrew-science.

Thanks in advance!


iml:
I would expect the Fortran message with the `xmi-msim` option but not otherwise. The C++ warning is from GCC and the openmp requirement and can be ignored.

tschoonj:
The Fortran message is quite strange indeed but it appears to be harmless.

Anything else I can do to get the pull request accepted?

tschoonj:
@iml Could you take another look at this please?

Thanks

tschoonj:
@dpo Could you take a look at this please?

Thanks

dpo:
@BrewTestBot test this please

dpo:
Would you mind changing the sha1 to a sha256? You can obtain it with `shasum -a 256`. Thanks!

tschoonj:
done!

dpo:
Ok thanks! It builds fine for me and the tests pass. Please squash your commits.

tschoonj:
squashed and waiting to build

dpo:
Any idea why the build fails on Mountain Lion?
````
detector.cpp: In member function 'int detectorarray::ForcedAcquisition(basesource**, photon*, randmt_t**)':
detector.cpp:187: error: expected '#pragma omp' clause before 'collapse'
detector.cpp: In member function 'int detectorarray::UnforcedAcquisition(basesource**, photon*)':
detector.cpp:279: error: expected '#pragma omp' clause before 'collapse'
````

tschoonj:
yes: it's using llvm-gcc that supports only an old version of the OpenMP standard, which does not support the collapse statement. I thought that the :openmp statement enforced switching to gcc49?

dpo:
`llvm-gcc` probably gets picked up because it supports OpenMP. Perhaps you should add a `fails-with` block to avoid it being picked up. I'm not quite sure why this only happened on Mountain Lion.

tschoonj:
Mountain Lion is the last version that has an Xcode that ships with `llvm-gcc`. I will just list `gcc` as a dependency, which should fix the problem. I will do this tomorrow.

dpo:
The `fails_with :clang` block isn't necessary I think because `needs :openmp` explicitly excludes `:clang`. It just wasn't excluding `:llvm` on Mountain Lion.

tschoonj:
@dpo I have removed the `fails_with :clang` block and squashed all commits into one: could you please pull in this formula? 

Thanks

(PS: also the gnudatalanguage formula should be fine by now)

dpo:
It seems the bottles aren't available. Sorry this is taking a while and I keep getting distracted.

dpo:
@BrewTestBot test this please

dpo:
@mikemcquaid I'm having trouble fetching bottles from Bintray:
````
$ brew fetch --force-bottle xrmc.rb
==> Fetching xrmc from homebrew/homebrew-science
==> Downloading https://homebrew.bintray.com/bottles-science/xrmc-6.5.0.mavericks.bottle.tar.gz

curl: (22) The requested URL returned error: 401 Unauthorized
````
Any idea what might be wrong?

mikemcquaid:
Have you used `brew pull` to publish them?

dpo:
I have and that worked fine:
````
$ brew pull --bottle https://github.com/Homebrew/homebrew-science/pull/1845
==> Applying patch
Applying: xrmc 6.5.0 (new formula)
==> Patch closes issue #1845
Switched to a new branch 'pull-bottle-1845'
==> Applying patch
Applying: xrmc 6.5.0 (new formula)
Applying: xrmc: add 6.5.0 bottle.
First, rewinding head to replay your work on top of it...
Applying: xrmc: add 6.5.0 bottle.
Switched to branch 'master'
Your branch is ahead of 'origin/master' by 1 commit.
  (use "git push" to publish your local commits)
Updating d219db3..4d24609
Fast-forward
 xrmc.rb | 7 +++++++
 1 file changed, 7 insertions(+)
Deleted branch pull-bottle-1845 (was 4d24609).
==> Patch changed:
 xrmc.rb | 47 +++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 47 insertions(+)
````
My `BINTRAY_USER` and `BINTRAY_KEY` also look correct.

mikemcquaid:
It should say something about publishing files on Bintray.

mikemcquaid:
Your login details do not seem to be correct and/or you need to run `brew update` because the files are not published here: https://bintray.com/homebrew/bottles-science/xrmc/6.5.0/view

dpo:
Oh well when I followed the link you just gave and clicked `Publish`, the bottles became available. Does that have to be done manually?

mikemcquaid:
It's done by `brew pull`. Please `brew update` and make sure that you have exported those variables correctly: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/pull.rb#L160-L180

tschoonj:
Thanks!

dpo:
I `brew update` several times a day. The bottles were uploaded, just not published. It might have been a temporary glitch because I was just able to fetch another bottle (#1918).

PR #1847, opened 2015-02-18
salmon 0.3.0: Needs C++11

sjackman:
Fix error: no member named 'all_of' in namespace 'std'
See https://github.com/kingsfordgroup/sailfish/issues/74


sjackman:
See https://stackoverflow.com/questions/24764478/clang-doesnt-recognize-stdall-of-in-algorithm

sjackman:
Hmm. Still doesn't build on Mountain Lion, even with `needs :cxx11`.

sjackman:
Salmon does require C++11.

sjackman:
Jellyfish fails to build on Mountain Lion for the same reason.

iml:
Might need to set `ENV.libstdcxx if MacOS.version <= :mountain_lion`?

sjackman:
Thanks for the suggestion, Ian. I'm trying it out at https://github.com/Homebrew/homebrew-science/pull/1858

sjackman:
@iml Still fails on Mountain Lion, even with `ENV.libstdcxx if MacOS.version <= :mountain_lion`.
```
  CXX    lib/misc.lo
lib/misc.cc:86:11: error: no member named 'all_of' in namespace 'std'
  if(std::all_of(arg.begin(), arg.end(), isblunt))
     ~~~~~^
1 error generated.
```

PR #1854, opened 2015-02-19
MUMPS: Add mumps_simple ressource.

dpo:


dpo:
@BrewTestBot test this please

PR #1858, opened 2015-02-19
salmon 0.3.0: Fix for Mountain Lion

sjackman:
Fix error: no member named 'all_of' in namespace 'std'


sjackman:
@BrewTestBot test this please

sjackman:
Still fails, even with `ENV.libstdcxx if MacOS.version <= :mountain_lion`.
```
  CXX    lib/misc.lo
lib/misc.cc:86:11: error: no member named 'all_of' in namespace 'std'
  if(std::all_of(arg.begin(), arg.end(), isblunt))
     ~~~~~^
1 error generated.
```

PR #1876, opened 2015-02-24
NEST NEural Simulation Tool

physicalist:
NEST is a simulator for spiking neural network models that focuses on the dynamics, size and structure of neural systems rather than on the exact morphology of individual neurons.

See http://nest-initiative.org/

### DIFF
--- a/nest.rb
+++ b/nest.rb
@@ -1,0 +1,62 @@
+require 'formula'
+
+# NEST is the NEural Simulation Tool
+
+class Nest < Formula
+  homepage 'http://www.nest-initiative.org'
+  url 'http://www.nest-initiative.org/download/gplreleases/nest-2.2.2.tar.gz'
+  sha1 '275a1d658fdbb5f9f4ffcadd878c341eac495169'
+
+  option "without-python", "Do not build python bindings (saves some time but not recommended; pyNN wouldn't work, among other things)"
+  option "without-gsl", "Don't use Gnu scientific library (GSL) for extended numerical possibilites (NOT recommended, WILL break some neuron models)"
+
+  option "enable-pthread", "Use pthreads"
+  option "enable-openmpi", "Use OpenMPI"
+
+  fails_with :clang do
+    build 500
+    cause <<-EOS.undent
+      Build will fail due to missing implementation of OpenMP standard in Clang.
+      As a workaround, use --without-openmp as build parameter, or rather, use
+      GCC as preferred compiler.
+      EOS
+  end
+
+  depends_on :python => :recommended
+  depends_on "gsl" => :recommended
+  depends_on "open-mpi" if build.include? 'enable-openmpi'
+
+  depends_on "homebrew/python/scipy" if build.include? "python"
+
+  def install
+    args = [ "--disable-debug",
+             "--disable-dependency-tracking",
+             "--prefix=#{prefix}" ]
+
+    # Clang still doesn't have OpenMP support!
+    args << '--without-openmp' if ENV.compiler == :clang
+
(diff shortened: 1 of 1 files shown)


nest.rb
```diff
@@ -0,0 +1,86 @@
+require 'formula'
+
+# NEST is the NEural Simulation Tool
+
```
iml: The above lines are unnecessary. Also please use all double-quotes with Ruby code.

nest.rb
```diff
@@ -0,0 +1,86 @@
...
+class Nest < Formula
+  homepage 'http://www.nest-initiative.org'
+  url 'http://www.nest-initiative.org/download/gplreleases/nest-2.4.2.tar.gz'
+  sha1 '86da83c9747c898e63d00a2f99159700a68d8b36'
+
+  option "without-python", "Do not build python bindings (saves some time but not recommended; pyNN wouldn't work, among other things)"
+  option "without-gsl", "Don't use Gnu scientific library (GSL) for extended numerical possibilites (NOT recommended, WILL break some neuron models)"
+
+  option "enable-pthread", "Use pthreads"
+  option "enable-openmpi", "Use OpenMPI"
```
iml: Options should be `with-pthread` or `without-pthread` such that the DSL is `build.with?` or `build.without?` later on. `build.include?` is deprecated.

nest.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  test do
+    # simple check whether NEST was compiled & linked
+    system "#{bin}/nest", "--version"
+
+    # use pyNEST's built-in test suite
+    python do
+      system python, "-c", "import nest; nest.test()"
+    end
```
iml:
```ruby
Language::Python.each_python(build) do |python, _version|
  system python, "-c", "import nest; nest.test()"
end

File: nest.rb (hunk 1 above)
Comment (physicalist) [inline nest.rb]: When I created the formula, `require` was still required :)

**Comment by dpo** [comment]
@physicalist Will this be updated?

File: nest.rb [hunk 4]
```diff
@@ -0,0 +1,86 @@
...
+
+# NEST is the NEural Simulation Tool
+
+class Nest < Formula
+  homepage 'http://www.nest-initiative.org'
+  url 'http://www.nest-initiative.org/download/gplreleases/nest-2.4.2.tar.gz'
+  sha1 '86da83c9747c898e63d00a2f99159700a68d8b36'
+
+  option "without-python", "Do not build python bindings (saves some time but not recommended; pyNN wouldn't work, among other things)"
+  option "without-gsl", "Don't use Gnu scientific library (GSL) for extended numerical possibilites (NOT recommended, WILL break some neuron models)"
```
dpo: These options are redundant with the `depends_on` lines below.

nest.rb
dpo: Please remove the `openmpi` option.

nest.rb
```diff
@@ -0,0 +1,86 @@
...
+      > error: field of type 'std::vector<double>::const_iterator'
+      > (aka '__wrap_iter<const_pointer>') has private constructor
+      > next_(0),
+      > ^
+     EOS
+  end
+
+  depends_on :python => :recommended
+  depends_on "gsl" => :recommended
+  depends_on "open-mpi" if build.include? 'enable-openmpi'
```
dpo: Shoud be `depends_on :mpi`.

nest.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  depends_on "homebrew/python/scipy" if build.include? "python"
```
dpo: `build.with? "python"`

PR #1884, opened 2015-02-26
sundials: fix #1814

### LINKED ISSUE #1814: Sundials: segmentation fault
While trying to install petsc on Yosemite 10.10.2, brew installs sundials as a dependency and adds the `--with-mpi` flag. If my understanding is correct, when we add a flag, brew automatically recompiles the dependency with the correct flags and do not use bottles. The compilation of sundials ends up with an error.

Using the `--debug` flag for `install`, I noticed that the sundial installation ends up with a segfault when executing `./cvodes/serial/cvsHessian_ASA_FSA`.

This can be worked around by adding the `without-check` flag to the dependency in petsc, but does not solves this segmentation fault issue.
According to their [website](https://computation.llnl.gov/casc/sundials/download/download.php), the version of sundial provided in homebrew is the latest (the release date of 2.5.0 is March 2012).

What could be the solution to this problem ?
- removing the dependency on sundials from petsc ?
- adding the `without-check` flag (might not be safe as the segfault could come from the library itself)

aancel:


sundials.rb
```diff
@@ -16,6 +16,11 @@ class Sundials < Formula
 
   option "without-check", "Skip build-time checks and examples (not recommended)"
 
+  # Fix for issue #1814: Segmentation fault on Yosemite
+  if MacOS.version == :yosemite && (build.with?("mpi") || build.with?("openblas"))
+    env :std
+  end
+
```
dpo: Thanks! Could you please make this a one-liner
````
env :std if ...
````

Comment (aancel) [inline sundials.rb]: Ok, I'll change this.
The travis build seems to be crashing at the same step as my compilation was.
Should I also add a test for linux systems, and go with the classic environment in this case ? (The test seems to be "OS.linux?" from other formulae)

Comment (dpo) [inline sundials.rb]: Yes we can try that. It is strange though that the build appears to have succeeded at https://travis-ci.org/Homebrew/homebrew-science/builds/52276580#L180 but failed later. Yes, you can use `OS.mac?` and `OS.linux?`.

**Comment by dpo** [comment]
You're right. I'm not sure why the Linux build fails. To figure it out, someone needs to build at a Linux terminal and be able to look at the logs in detail. For now, we're only concerned with OSX build though.

Would you mind squashing your commits together?

**Comment by dpo** [comment]
I'm a bit worried because I'm having to use `--env=std` on a bunch of formulas that otherwise segfault. I wonder if this is due to a recent change in superenv.

**Comment by aancel** [comment]
Indeed, it is annoying. Would it be possible to try and indetify which environment variable causes the segfault ?
FYI, I tried to build sundials with ubuntu 12.04 (the version used by travis) and 14.04 using docker, but I was not able to reproduce the issue encountered on travis ...

**Comment by saketkc** [comment]
Travis setup is a bit buggy now. It should be safe to ignore that.

**Comment by davydden** [comment]
as I mentioned here https://github.com/Homebrew/homebrew-science/issues/1814 with mid-January homebrew sundials was installed without an error for me. On the Linux machine it worked with Ubuntu 12.04 as well (`brew install sundials –with-mpi`)... strange...

**Comment by davydden** [comment]
p.s. I am also on Yosemite

**Comment by dpo** [comment]
After a `brew update`, are you able to build `sundials` from source on Yosemite?

**Comment by davydden** [comment]
just reinstalled `sundials` on Ubuntu, no issues... will try on os-x next...

**Comment by davydden** [comment]
A friend of mine checked that `brew install sundials --with-mpi` worked on Mavericks without an issue after `brew udpate`. I will try it on Yosemite later as I run some calculations nonstop currently...

**Comment by dpo** [comment]
It's known to build on Mavericks.

**Comment by dpo** [comment]
@BrewTestBot test this please

**Comment by davydden** [comment]
`brew reinstal sundials` worked just fine for me on Yosemite. 
```
Reinstalling sundials with --with-mpi
```

davydden:
and I did `brew update` just before, to be sure

dpo:
I'm frankly not sure what to do about this. @jacknagel, any idea why some of us need `--env=std` on Yosemite? Using Superenv leads to a segmentation fault.

jacknagel:
I will take a look if I find some time, maybe over the weekend.

dpo:
Closing this PR as it was determined in #1814 that the issues arises from using MPICH.

PR #1886, opened 2015-02-27
butterflow 0.1.7

dthpham:
Version bump. Also added `:python` as a dependency.


tdsmith:
Why add `:python`?

dthpham:
From my limited testing, it works without it. 

But right now, the bottled `butterflow` package segfaults when it's run. I found that the problem was because `butterflow` was linking to the system `python` and `opencv` was linking to a brewed one. The simplest fix is adding `:python` to force both to depend on a brewed `python`.

tdsmith:
Okay, thanks. I'll unbottle this as well, I think; bottling things that depend on `:python` is bad because introducing a dependency on Homebrew python isn't good.

FWIW, Python extension modules don't need to have explicit links to framework binaries on OS X and probably shouldn't. Instead of building with `-lpython` or `-framework Python`, extension modules should use `-undefined dynamic_lookup` so that python symbols are learned at import time. distutils seems to do this by default. `otool -L _module.so` shouldn't mention `Python`. This allows modules to be imported from foreign Pythons without segfaulting the interpreter. Butterflow's linkage was correct in 0.1.5 but it has explicit links to a Python framework in 0.1.7.

tdsmith:
#2005 will remove opencv's explicit link to Python.

dthpham:
Thanks for the tip. I updated butterflow so that it doesn't have explicit links anymore and removed `:python` from the formula.

tdsmith:
Just resolved the opencv linkage issue; will take a look at this again tomorrow. Thanks for your patience and sorry for the delay!

tdsmith:
It builds fine but it doesn't seem to work; can you tell me how to invoke this against one of your samples? I see errors like

```
av: could not open input
  File "/usr/local/Cellar/butterflow/0.1.8/libexec/bin/butterflow", line 9, in <module>
    load_entry_point('butterflow==0.1.8', 'console_scripts', 'butterflow')()
  File "/usr/local/Cellar/butterflow/0.1.8/libexec/lib/python2.7/site-packages/butterflow/cli.py", line 148, in main
    project = Project.new(src_path)
  File "/usr/local/Cellar/butterflow/0.1.8/libexec/lib/python2.7/site-packages/butterflow/project.py", line 38, in new
    return cls(video_path)
  File "/usr/local/Cellar/butterflow/0.1.8/libexec/lib/python2.7/site-packages/butterflow/project.py", line 33, in __init__
    self.vid_info = LibAvVideoInfo(video_path)
  File "/usr/local/Cellar/butterflow/0.1.8/libexec/lib/python2.7/site-packages/butterflow/media/info.py", line 36, in __init__
    raise RuntimeError('could not get video info: {}'.format(video_path))
RuntimeError: could not get video info: vid-5s-vs.mp4
```

Does ffmpeg need more flags?

tdsmith:
Nevermind, I think I did something dumb.

PR #1905, opened 2015-03-05
gnudatalanguage 0.9.5 (new formula)

tschoonj:
Most optional dependencies have been added. The remaining ones first
need to be added to Homebrew.

A remaining issue here is the OpenMP support: I added the option but the
build fails at the linking stage due to some error I cannot fix. This
build is quite complicated because I have to force g++ to link against libc++
instead of the default libstdc++. I am hoping that someone else can fix
this:

```
  "___cxa_throw_bad_array_new_length", referenced from:
      GDLInterpreter::l_deref(ProgNode*)      in GDLInterpreter.cpp.o
      lib::stregex_fun(EnvT*)      in basic_fun.cpp.o
      lib::transpose(EnvT*)     in basic_fun.cpp.o
      lib::sort_fun(EnvT*)     in basic_fun.cpp.o
      lib::median(EnvT*)     in basic_fun.cpp.o
      PreAllocPListT<BaseGDL*, 64ull>::push_back(BaseGDL*) in basic_fun.cpp.o
      lib::routine_names_value(EnvT*)      in basic_fun_jmg.cpp.o
      ...
```


gnudatalanguage.rb
```diff
@@ -0,0 +1,190 @@
...
+  sha256 "cc9635e836b5ea456cad93f8a07d589aed8649668fbd14c4aad22091991137e2"
+
+  depends_on "cmake" => :build
+  depends_on "pkg-config" => :build
+  depends_on "plplot"
+  depends_on "gsl"
+  depends_on "readline"
+  depends_on "graphicsmagick"
+  depends_on "netcdf"
+  depends_on "hdf4" => :optional
```
dpo: Should be `homebrew/versions/hdf4`

dpo:
Have you tried `ENV.libcxx`?

tschoonj:
I tried ENV.libcxx but it didn't make any difference whatsoever. OpenMP forces me to use g++, but g++ really does not like to be coupled with clang's libc++, unless you start overriding the default include search path as well as the libraries to link against.

Basically everything compiles fine, but things go wrong at the linking stage when it complains about the undefined reference to `___cxa_throw_bad_array_new_length`, which is a standard exception, that is present in libstdc++, but not in libc++. I guess the compiler screwed up here somehow because I am pretty confident that libc++ also offers this exception.

tschoonj:
Could you restart this build please: something went wrong while downloading the source tarball during the Mountain Lion build

Thanks

tschoonj:
@dpo If you agree, can you also pull in this formula?

Thanks!

dpo:
Could you please remove all the commented out code?

tschoonj:
@dpo Done!

tschoonj:
Thanks! @dpo @iml

PR #1906, opened 2015-03-05
R: enable memory profiling.

### LINKED ISSUE #1890: R installation does not include the cairo graphics devices
After installing R ([r.rb](https://github.com/Homebrew/homebrew-science/blob/master/r.rb)), I found that I was missing files necessary for the cairo-based graphics (like the `svg` function).

It seems that they need to be installed separately with `make cairodevices` as described [here](http://cran.r-project.org/doc/manuals/r-release/R-admin.html#Building-the-cairo-devices-files). Currently there is no option to do this with the Homebrew formula. I don't know any Ruby or how to use make, but I feel like this should be possible; I'm sure it was just overlooked.

dpo:
Fixes #1890.


dpo:
@BrewTestBot test this please

dpo:
Bot down?

PR #1909, opened 2015-03-06
diamond 0.7.1: New formula

sjackman:
DIAMOND is a new high-throughput program for aligning a file
of short reads against a protein reference database such as
NR, at 20,000 times the speed of BLASTX, with high
sensitivity.


sjackman:
Requires Boost and OpenMP.
```
checking for boost/program_options.hpp... yes
checking for the Boost program_options library... no
configure: error: cannot find the flags to link with Boost program_options
```

PR #1911, opened 2015-03-06
update qualimap to last version with important fix

lpantano:
Hi,

it would be great to have the last version, that have fixed important bugs in the last month.

cheers


iml:
The regex trips on that URL, so the version will need to be set to something like `version "06-03-15"`.

lpantano:
anything I can improve to get this merge?

iml:
Thanks for your contribution!

PR #1470, opened 2014-12-15
libfolia (latest) (new formula)

fbkarsdorp:
XML library used by frog and ucto

PR #1913, opened 2015-03-08
Migrate ticcutils, timbl, wopr from core.

dpo:
Addresses #1470.
Core pull request: https://github.com/Homebrew/homebrew/pull/37450.


bfontaine:
Note that the latest Wopr version is 1.36.1.

PR #1916, opened 2015-03-09
abinit: makes scalapack, netcdf and libXC useable

mtorrent:
I have modified this existing formula in order to add the following new capabilities:
1- Use of scalapack for parallel runs (recommended); I have simply added "+scalapack" to the "--with-linalg-flavor" option and have linked scalapack formula.
2- Use of libXC exchange-correlation library ; current libXC 2.2 is not allowed by ABINIT 7.10.4 build system (will be supported by v8) but a simple change in configure script make it useable. I made it with a single "inreplace" statement.
3- Use of netCDF (highly recommended for ABINIT); this is not possible for ABINIT v7.10.4 because of a bug occurring when netCDF library is linked and ESTF_IO library is not (and this is precisely the case for homebrew has no ETSF IO formula is available). This bug can be corrected and impacts 3 files. I have added a (temporary) patch for that. This bug has been committed upstream and will be corrected in the next version of the code.
4- Use of ABINIT specific parallel compilation: "make mj4". Improves compilation speed. 

Marc, ABINIT developer


abinit.rb
```diff
@@ -70,8 +89,11 @@ def install
       args << "--with-fft-libs=-L#{Formula["fftw"].opt_lib} -lfftw3 -lfftw3f -lfftw3_mpi -lfftw3f_mpi"
     end
 
+    args << "--with-dft-flavor=#{dft_flavor}"
+    args << "--with-trio-flavor=#{trio_flavor}"
+
     system "./configure", *args
-    system "make"
+    system "make", "mj4"
```
dpo: Homebrew runs `make` with concurrent tasks by default. Does this do something else?

mtorrent: "make mj4" (an alias for "make multi multi nproc=4") is supposed to compile Abinit in 2 stages : 1-sequential make of plugins, 2-parallel make of abinit main program. This is mandatory to compile safely.
But, you're probably right: as all plugins are transmitted as formulas, compilation should begin at stage 2 and this make target is probably useless... I try to suppress it.

dpo:
Sorry for the delay. There are quite a few audit warnings (see `brew audit abinit --strict`). Could you also please set the `sha256`, which now replaces the `sha1`?

mtorrent:
Done!

dpo:
@BrewTestBot test this please

PR #1917, opened 2015-03-09
Petsc (openblas/Linuxbrew)

davydden:
Minor modifications to Petsc to make it build on Linuxbrew.

### DIFF
--- a/petsc.rb
+++ b/petsc.rb
@@ -24,7 +24,6 @@ class Petsc < Formula
   depends_on :x11 => :optional
   depends_on "cmake" => :build
 
-  depends_on "openssl"
   depends_on "superlu"      => :recommended
   depends_on "superlu_dist" => :recommended
   depends_on "metis"        => :recommended
@@ -38,6 +37,7 @@ class Petsc < Formula
   depends_on "suite-sparse" => :recommended
   depends_on "netcdf"       => ["with-fortran", :recommended]
   depends_on "fftw"         => ["with-mpi", "with-fortran", :recommended]
+  depends_on "openblas"     => :optional
 
   #TODO: add ML, YAML dependencies when the formulae are available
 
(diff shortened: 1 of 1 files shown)


petsc.rb
```diff
@@ -86,6 +87,14 @@ def install
     args << "--with-mumps-dir=#{oprefix("mumps")}" if build.with? "mumps"
     args << "--with-x=0" if build.without? "x11"
 
+    # if build with openblas, need to provide lapack as well.
+    # Link against system lapack library if available?
```
dpo: Yes, here we should link against `libblas` and `liblapack` if not building with `openblas`. See, e.g., the `arpack` formula.

davydden:
I have some problems when using `petsc` when linking `deal.II` tests on Linux:
```
/home/davydden/.linuxbrew/lib/libpetsc.so: undefined reference to `SSL_CTX_free@OPENSSL_1.0.0'
```
so pls don't rush merging it ;-) there could be more changes...

petsc.rb
davydden: I removed that comment as i think `./configure` shall pick up system libraries otherwise. I have the whole toolchain installed with `openblas`, so i would propose to keep it like that until someone tests it with native libraries.

davydden:
the issue above is not there now.

petsc.rb
```diff
@@ -86,6 +89,13 @@ def install
     args << "--with-mumps-dir=#{oprefix("mumps")}" if build.with? "mumps"
     args << "--with-x=0" if build.without? "x11"
 
+    # if build with openblas, need to provide lapack as well.
+    if build.with? "openblas"
+      exten = (OS.mac?) ? "dylib" : "so"
+      args << ("--with-blas-lib=#{Formula["openblas"].opt_lib}/libopenblas."+exten)
```
dpo: I suggest `"--with-blas-lib=#{Formula["openblas"].opt_lib}/libopenblas.#{exten}"`

davydden:
do we want to link against `veclibfort` here as well?

davydden:
p.s. PETSc is written in C, so we ain't use `veclibfort` here.

dpo:
You're right.

PR #1918, opened 2015-03-09
Slepc (update for Linuxbrew)

davydden:
Minor update to Slepc to work correctly on Linuxbrew.

### DIFF
--- a/slepc.rb
+++ b/slepc.rb
@@ -14,6 +14,7 @@ class Slepc < Formula
   deprecated_option "complex" => "with-complex"
 
   option "with-complex", "Use complex version by default. Otherwise, real-valued version will be symlinked"
+  option "without-check", "Skip run-time tests (not recommended)"
 
   depends_on "petsc" => :build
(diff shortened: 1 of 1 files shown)


dpo:
Now you're hitting #1884.

davydden:
I think the issue you mentioned is related, but separate entity.

slepc.rb
```diff
@@ -43,7 +43,8 @@ def install
     ENV["PETSC_DIR"] = "#{Formula["petsc"].opt_prefix}/#{petsc_arch_complex}"
     system "./configure", "--prefix=#{prefix}/#{petsc_arch_complex}", *args
     system "make"
-    system "make", "test"
+    # TODO investigate why complex tests fail to run on Linuxbrew
+    system "make", "test" if OS.mac?
```
dpo: I would prefer a `--without-check` option. Linux failures should be reported upstream.

dpo:
@BrewTestBot test this please

PR #1919, opened 2015-03-09
Superlu (update for Linuxbrew)

davydden:
minor update to Superlu to make it usable in Linuxbrew.


dpo:
Is it feasible to coerce SuperLU into building a dynamic lib?

davydden:
It seems MacPorts build a shared library out of it: https://trac.macports.org/browser/trunk/dports/math/superlu/Portfile

davydden:
I would propose to keep it one step at a time, `-fPIC` helps with building dynamic `petsc` libraries on Linux.

PR #1921, opened 2015-03-09
Hypre (update for Linuxbrew)

davydden:
Added -fPIC flags, which are crucial for Linuxbrew.


hypre.rb
```diff
@@ -71,6 +71,13 @@ def install
       config_args << "--without-fei" if build.without? "fei"
       config_args << "--without-mli" if build.without? "mli"
 
+      # on Linux many Homebrew ports willl fail to build
```
dpo: What do you mean by "Homebrew ports"?

PR #1922, opened 2015-03-09
P4est (update with openblas)

davydden:
Added openblas option to p4est.

### DIFF
--- a/p4est.rb
+++ b/p4est.rb
@@ -19,6 +19,7 @@ class P4est < Formula
 
   depends_on :mpi => [:cc, :cxx, :f77, :f90]
   depends_on :fortran
+  depends_on "openblas" => :optional
 
   def install
     ENV["CC"]       = ENV["MPICC"]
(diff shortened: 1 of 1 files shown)


p4est.rb
```diff
@@ -28,9 +29,16 @@ def install
     ENV["CFLAGS"]   = "-O2"
     ENV["CPPFLAGS"] = "-DSC_LOG_PRIORITY=SC_LP_ESSENTIAL"
 
+    if build.with? "openblas"
+      blas = (OS.mac?) ? "BLAS_LIBS=#{Formula['openblas'].opt_lib}/libopenblas.dylib" : "BLAS_LIBS=#{Formula['openblas'].opt_lib}/libopenblas.so"
+    else
+      blas = (OS.mac?) ? "BLAS_LIBS=#{Formula['veclibfort'].opt_lib}/libveclibfort.dylib" : "" #"BLAS_LIBS=/usr/lib/libblas.so"
```
dpo: You need to add a dependency on `veclibfort`. Also the library name is `libvecLibFort`.

dpo: Please also remove the comment.

p4est.rb
```diff
@@ -19,6 +19,8 @@ class P4est < Formula
 
   depends_on :mpi => [:cc, :cxx, :f77, :f90]
   depends_on :fortran
+  depends_on "openblas" => :optional
+  depends_on "veclibfort" => :optional if OS.mac?
```
dpo: Here I also suggest `depends_on "veclibfort" if build.without?("openblas") && OS.mac?`

dpo: On Linux, don't you need to set `blas` to `-lblas -llapack`?

dpo:
@BrewTestBot test this please

PR #1923, opened 2015-03-09
Parmetis (update for Linuxbrew)

davydden:
Added extra LDFLAGS and CFLAGS.
-lm (libm) was needed to compile on Linuxbrew and -fPIC helps with
linking shared libraries on Linuxbrew.
Other CFLAGS are taken from the Petsc build script.

### DIFF
--- a/parmetis.rb
+++ b/parmetis.rb
@@ -22,7 +22,8 @@ class Parmetis < Formula
   patch :DATA
 
   def install
-    ENV["LDFLAGS"] = "-L#{Formula["metis"].lib} -lmetis"
+    ENV["LDFLAGS"] = "-L#{Formula["metis"].lib} -lmetis -lm"
+
     system "make", "config", "prefix=#{prefix}", "shared=1"
     system "make", "install"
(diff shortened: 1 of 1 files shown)


parmetis.rb
```diff
@@ -22,7 +22,9 @@ class Parmetis < Formula
   patch :DATA
 
   def install
-    ENV["LDFLAGS"] = "-L#{Formula["metis"].lib} -lmetis"
+    ENV["LDFLAGS"] = "-L#{Formula["metis"].lib} -lmetis -lm"
+    ENV["CFLAGS"]  = "-fPIC -Wall -Wwrite-strings -Wno-strict-aliasing -Wno-unknown-pragmas"
```
dpo: Isn't `-fPIC` activated by `shared=1`? The rest is just warning flags.

davydden: i think you are right as static library is not built at all. `petsc` builds fine on `linuxbrew` without `-fPIC` so I remove modification to flags altogether.

PR #1924, opened 2015-03-09
Trilinos (numerous updates)

davydden:
Minor updates to Trilinos:

1) Made python optional as otherwise other dependent formulas can't be
built on Linuxbrew. 2) Added support for Openblas. 3) Specify
MPI base dir to avoid picking up system's MPI on Linuxbrew
4) Manually specify MUMPS library names to make it link against MUMPS 5.0

### DIFF
--- a/trilinos.rb
+++ b/trilinos.rb
@@ -1,7 +1,7 @@
 class Trilinos < Formula
   homepage "http://trilinos.sandia.gov"
   url "http://trilinos.org/oldsite/download/files/trilinos-11.12.1-Source.tar.bz2"
-  sha1 "f24f401e2182003eb648d47a8e50a6322fdb79ec"
+  sha256 "a41539414529c65905260b3befe3aee4f1dd1015ff719f0f6b5a10902d576fda"
   head "https://software.sandia.gov/trilinos/repositories/publicTrilinos", :using => :git
 
   option "with-teko",  "Enable the Teko secondary-stable package"
@@ -25,6 +25,8 @@ class Trilinos < Formula
 
   depends_on "cmake"        => :build
   depends_on "pkg-config"   => :build
+
+  depends_on "openblas" => :optional
 
   depends_on "adol-c"       => :recommended
   depends_on "boost"        => :recommended
@@ -88,6 +90,22 @@ class Trilinos < Formula
                -DTPL_ENABLE_Matio=OFF
                -DSacado_ENABLE_TESTS=OFF]
 
+    # constrain Cmake to look for libraries in homebrew's prefix
+    args << "-DCMAKE_PREFIX_PATH=#{HOMEBREW_PREFIX}"
+
+    # on Linux Trilinos might pick up wrong MPI.
+    # Can't specify "open-mpi" location as other (mpich)
+    # implementations may be used.
+    args << "-DMPI_BASE_DIR:PATH=#{HOMEBREW_PREFIX}" if build.with? "mpi"
+
+    # BLAS / LAPACK support
+    if build.with? "openblas"
+      args << "-DBLAS_LIBRARY_NAMES=openblas"
+      args << "-DBLAS_LIBRARY_DIRS=#{Formula['openblas'].opt_lib}"
+      args << "-DLAPACK_LIBRARY_NAMES=openblas"
+      args << "-DLAPACK_LIBRARY_DIRS=#{Formula['openblas'].opt_lib}"
+    end
+
     args << "-DEpetraExt_ENABLE_TESTS=OFF" if build.with? "hypre"
 
     args << "-DTrilinos_ASSERT_MISSING_PACKAGES=OFF" if build.head?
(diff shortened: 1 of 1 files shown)


trilinos.rb
```diff
@@ -88,6 +94,22 @@ def install
                -DTPL_ENABLE_Matio=OFF
                -DSacado_ENABLE_TESTS=OFF]
 
+     # constrain Cmake to look for libraries in homebrew's prefix
+     args << "-DCMAKE_PREFIX_PATH=#{HOMEBREW_PREFIX}"
+ 
+    # on Linux Trilinos might pick up wrong MPI
+    if build.with? "mpi"
+      args << "-DMPI_BASE_DIR:PATH=#{Formula['open-mpi'].opt_prefix}"
+    end
```
dpo: This introduces an explicit dependency on `open-mpi` and will prevent building with `mpich`.

davydden: I agree. Is there any way we can query for opt_prefix of current `:mpi`? In worst case, I could point to Homebrew location.

dpo: `HOMEBREW_PREFIX` will do on OSX, but I'm not sure what will happen on Linux.

davydden: i'll give it a try once I resolve some other issues...

davydden:
`veclibfort` instead of default `-framework Accelerate`?

dpo:
Wow! The build actually completed under 60 minutes on Yosemite!

trilinos.rb
```diff
@@ -19,13 +19,19 @@ class Trilinos < Formula
   depends_on :fortran       => :recommended
   depends_on :x11           => :recommended
 
-  depends_on :python        => :recommended
+  # Make python optional as numpy does not build on Linuxbrew
```
dpo: Did you report the build issues over at homebrew/python? It would be great to fix this.

davydden: just did it https://github.com/Homebrew/homebrew-python/issues/209

tdsmith: I don't use Linuxbrew and can't support it; someone will have to tell me how to fix this.

dpo: Yes we'll try to fix it. This is just to keep track of the issue. Thanks.

dpo: @davydden If you're not planning to look into the issue, just restore the `:python` dependency so it builds by default on OSX and use `--without-python` on Linux.

davydden: what happens if another formula has `trilinos` as `recommended`? It seems that the `re-build` will be invoked if it is build without python?

trilinos.rb
```diff
@@ -128,8 +146,9 @@ def install
     args << onoff("-DTPL_ENABLE_HWLOC:BOOL=",       (build.with? "hwloc"))
     args << onoff("-DTPL_ENABLE_HYPRE:BOOL=",       (build.with? "hypre"))
 
-    # METIS conflicts with ParMETIS in Trilinos config, see TPLsList.cmake in the source folder
-    if (build.with? "metis") && (build.without? "parmetis")
+    # Although METIS seems to conflicts with ParMETIS in Trilinos config (see TPLsList.cmake in the source folder),
+    # we still need to provide METIS_INCLUDE_DIRS so that metis.h is picked up on Linuxbrew.
+    if (build.with? "metis") # && (build.without? "parmetis")
```
dpo: Let's also remove the commented bit in the line above and we'll restart a build. Thanks!

trilinos.rb
```diff
@@ -137,7 +156,14 @@ def install
       args << "-DTPL_ENABLE_METIS:BOOL=OFF"
     end
 
-    args << onoff("-DTPL_ENABLE_MUMPS:BOOL=",       (build.with? "mumps"))
+    # A hack for mumps 5.0
+    # TODO: use extra LIBRARY_NAMES with 5.0 only?
+    if (build.with? "mumps")
+      args << "-DTPL_ENABLE_MUMPS:BOOL=ON"
+      args << "-DMUMPS_LIBRARY_DIRS=#{Formula["mumps"].opt_prefix}"
+      args << "-DMUMPS_LIBRARY_NAMES=dmumps;pord;mumps_simple;mumps_common"
```
dpo: You probably don't need `mumps_simple` here.

dpo:
@BrewTestBot test this please

trilinos.rb
dpo: Yes that will trigger a rebuild (or ask the user to rebuild).

davydden: How about adding "if OS.mac?" to the dependency so that on Linuxbrew formulae which use trilinos do not try to build it with python and numpy consequently?

dpo: That would work but it would be avoiding the real issue. We really should fix numpy.

trilinos.rb
davydden: you are right, removed it.

dpo:
There's a segmentation fault in there.

davydden:
Must be related to the new `mumps`, unless it was there before... How did you hit it? Have you notified trilinos folks?

trilinos.rb
tdsmith: Fixed!

davydden: thanks a lot @tdsmith

tdsmith:
The numpy dependency shouldn't cause test-bot issues now that Homebrew/homebrew@1d2e3ae110542862f88b2fd281af326686bfc1c4 has landed.

dpo:
Many thanks @tdsmith! This should be a great improvement!

dpo:
@BrewTestBot test this please

PR #1925, opened 2015-03-09
Suite sparse (update for Linuxbrew)

davydden:
Provide extra CFLAGS.

### DIFF
--- a/suite-sparse.rb
+++ b/suite-sparse.rb
@@ -33,12 +33,23 @@ class SuiteSparse < Formula
        "SuiteSparse_config/SuiteSparse_config.mk"
 
     make_args = ["INSTALL_LIB=#{lib}", "INSTALL_INCLUDE=#{include}"]
-    make_args << "BLAS=" + ((build.with? "openblas") ? "-L#{Formula["openblas"].opt_lib} -lopenblas" : "-framework Accelerate")
+    if build.with? "openblas"
+      make_args << "BLAS=-L#{Formula["openblas"].opt_lib} -lopenblas"
+    elsif OS.mac?
+      make_args << "BLAS=-framework Accelerate"
+    else
+      make_args << "BLAS=-lblas -llapack"
+    end
+
     make_args << "LAPACK=$(BLAS)"
     make_args += ["SPQR_CONFIG=-DHAVE_TBB",
                   "TBB=-L#{Formula["tbb"].opt_lib} -ltbb"] if build.with? "tbb"
     make_args += ["METIS_PATH=",
                   "METIS=-L#{Formula["metis4"].opt_lib} -lmetis"] if build.with? "metis4"
+
+    # Add some flags for linux
+    # -DNTIMER is needed to avoid undefined reference to SuiteSparse_time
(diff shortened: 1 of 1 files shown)


dpo:
@davydden Please hang on. The Linux builds on Travis are still not quite resolved and all these PRs are going to put considerable stress on the OSX test bot. We have to resolve the first few formulas before sending them all to BrewBot and Travis.

dpo:
@davydden PS : many thanks for all your updates!

davydden:
No problems, there is no hurry!

suite-sparse.rb
```diff
@@ -40,6 +40,12 @@ def install
     make_args += ["METIS_PATH=",
                   "METIS=-L#{Formula["metis4"].opt_lib} -lmetis"] if build.with? "metis4"
 
+    # Add some flags for linux
+    # -DNTIMER is needed to avoid undefined reference to SuiteSparse_time
+    if not(OS.mac?)
+      make_args << "CF=-fPIC -O3 -fno-common -fexceptions -DNTIMER $(CFLAGS)"
+    end
+
```
dpo: Please make this a one liner.

suite-sparse.rb
```diff
@@ -17,6 +17,7 @@ class SuiteSparse < Formula
 
   depends_on "tbb" => :recommended
   depends_on "openblas" => :optional
+  depends_on "veclibfort" => :optional if OS.mac?
```
dpo: `depends_on "veclibfort" if build.without?("openblas") && OS.mac?` (see, e.g., `arpack`).

davydden: it is not a mandatory dependence, so i guess at least something like `depends_on "veclibfort" => :optional if build.without?("openblas") && OS.mac?`

dpo: I think we should use `veclibfort` instead of `Accelerate` as it fixes some bugs.

iml:
Please remember to squash commits when they're ready to be merged and prefix the message with the formula name. This applies to the other PRs as well.

dpo:
Let's hope TravisCI comes back to life soon.

suite-sparse.rb
davydden: I removed `veclibfort` as this is not  a fortran-only code, and `veclibfort` (https://github.com/mcg1969/vecLibFort) is aimed to fix issues when compiling `fortran` code, as far as I understand

dpo:
@BrewTestBot test this please

dpo:
AMD is Fortran and it will be compiled with `gfortran`.

davydden:
yes, but there is C code as well (most of it).

dpo:
The entire library doesn't have to be Fortran to make `veclibfort` relevant. Fortunately, `amd.f` doesn't appear to make BLAS calls directly, so I think we're ok.

davydden:
i was not sure if `veclibfort` can be provided to link `C`/`C++` files.

dpo:
`veclibfort` is a thin interface to `Accelerate` that shadows just a few BLAS functions.

PR #1927, opened 2015-03-09
igv 2.3.42

sjackman:


iml:
Looks like 2.3.43 just came out.

sjackman:
@iml My preference is to commit as many releases as possible, so that they're in the record, and bottles of historical versions are available. It could help reproducibility down the road and supporting installing a specific version of some software.

iml:
Ah, sorry I'll keep that in mind.

PR #1935, opened 2015-03-11
openblas: bottle against core2 arch

### LINKED ISSUE #1930: openblas bottles for sandybridge, segfaults on penryn
Sorry if I missed an earlier discussion about this. When I `brew install openblas` I get a bottle containing /usr/local/Cellar/openblas/0.2.13/lib/libopenblas_sandybridgep-r0.2.13.dylib.

Unfortunately, on my vintage laptop, this means `brew install numpy --with-openblas` gives me a numpy that segfaults on its tests.

My config:
```
darkside:homebrew-pypi-poet tim$ brew config
HOMEBREW_VERSION: 0.9.5
ORIGIN: git@github.com:homebrew/homebrew.git
HEAD: 14fc4411c8ebea423c5cbc999af967e8e6f4d2fe
Last commit: 43 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_CELLAR: /usr/local/Cellar
CPU: dual-core 64-bit penryn
OS X: 10.10.2-x86_64
Xcode: 6.1.1
CLT: 6.1.1.0.1.1416017670
Clang: 6.0 build 600
X11: 2.7.7 => /opt/X11
System Ruby: 2.0.0-p481
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /Users/tim/.rvm/rubies/ruby-2.1.2/bin/ruby
Java: 1.8.0_20
```

Should openblas have bottles? Can Homebrew be more aggressive about telling openblas to bottle for the core2 arch? Alternatively but probably less favorably, a `pour_bottle?` method could decide whether or not to pour the bottle depending on the user's arch.

tdsmith:
Fixes #1930.


openblas.rb
```diff
@@ -17,9 +17,22 @@ class Openblas < Formula
...
+    ENV["TARGET"] = "CORE2" if build.bottle?
+
     # Must call in two steps
     system "make", "FC=#{ENV['FC']}", "libs", "netlib", "shared"
     system "make", "FC=#{ENV['FC']}", "tests"
     system "make", "PREFIX=#{prefix}", "install"
   end
+
+  def caveats
+    if Tab.for_formula(self).built_as_bottle
```
tdsmith: @jacknagel, should I avoid this?

jacknagel: I think the name `built_as_bottle` is kind of a implementation detail. Can we use `build.bottle?` here or are you looking to handle the `brew info` case? We can add another method to Tab that isn't coupled to the json key name if we need to.

tdsmith: I was thinking it would be good to show this for `info`, yeah. What would you call the method? `built_as_bottle?`?

jacknagel: I'll have to think about it. I've tried to keep the Tab and BuildOptions interfaces as similar as possible... for example, [`build` inside tests is actually the tab object](https://github.com/Homebrew/homebrew/blob/eec62ee3a74dec29eefb1cb7d263f23f8cf329e9/Library/Homebrew/formula.rb#L635).

PR #1939, opened 2015-03-12
lmod 5.9: New formula

sjackman:
Lmod is a Lua based module system that easily handles the MODULEPATH
Hierarchical problem. Environment Modules provide a convenient way to
dynamically change the users' environment through modulefiles. This includes
easily adding or removing directories to the PATH environment variable.
Modulefiles for Library packages provide environment variables that specify
where the library and header files can be found.


sjackman:
```
Error: Failure while executing: /usr/bin/curl -f#LA Homebrew\ 0.9.5\ (Ruby\ 2.0.0-481;\ Mac\ OS\ X\ 10.10.2) --silent --fail -o /dev/null -I https://github.com/BrewTestBot/homebrew-science/compare/homebrew:master...pr-1939
```

sjackman:
@drpowell @tseemann Let's move the discussion to a public forum, here, if that suits you.
@drpowell Would you post your private e-mail to me here?

drpowell:
My group is setting up a server for bioinformatics and as you are no doubt aware installing scientific software can be a real pain.  We are using ansible to provision and configure the machine with the goal to make it easy re-do on other machines (https://github.com/MonashBioinformaticsPlatform/bio-ansible).  We are also using a module system, Lmod, to allow multiple versions of the same software, and to keep the user PATH manageable.

My thought was that linuxbrew might be able to be used here since it initially installs software into the Cellar, and then links the binaries into a common directory.  I was hoping it would be possible to modify linuxbrew so that instead of linking, we create an Lmod module file specifying the location of the binaries in the Cellar.  Do you have any idea how difficult it would be to use linuxbrew in that way?

My current thought is to use a special user to run linuxbrew and install software, with the cellar in a sensible location.  Then have a separate script to parse the contents of "brew info --json=v1 <package>" to create a module file for each package pointing to the cellar location.  Thoughts?

sjackman:
An older such module system is called [modules](http://modules.sourceforge.net/), which is also packaged for Homebrew.

I think a module system is a perfect for Homebrew. Here's the twenty-line shell function named [require](https://github.com/sjackman/dotfiles/blob/master/.functions#L23-L41) that I wrote and use. e.g. `require samtools` or `require samtools/1.2` to set the `PATH` appropriately. It does not handle setting include and library locations, which is handled by Homebrew during the build process.

I'm completely unfamiliar with `lmod`, but I think writing module files based on the Homebrew installation of a formula seems pretty reasonable.

sjackman:
When I configure `lmod` with `--prefix=/usr/local/Cellar/lmod/5.9` (which is standard for Homebrew), lmod appends `lmod/5.9` to that path and installs most of its files in `/usr/local/Cellar/lmod/5.9/lmod/5.9` (except for `LICENSE`, `README` and a symlink from `5.9` to `lmod`). Does that directory structure make sense, or should I configure `lmod` with `--prefix=/usr/local/Cellar`? I need insight from a user of `lmod`?

sjackman:
@drpowell Can you test the current `lmod` formula and see whether it works for you?
`brew install https://raw.githubusercontent.com/sjackman/homebrew-science/lmod/lmod.rb`

drpowell:
Yes, I've used [modules](http://modules.sourceforge.net/) before.  Lmod is even backward compatible with the same module definition files

@sjackman That formula works.  It does seem odd to be in `lmod/5.9/lmod/5.9` but probably fine.  To use it, you need to `source /usr/local/Cellar/lmod/5.9/lmod/5.9/init/$shell`, and set your `MODULEPATH` to somewhere that contains the modules definitions.

sjackman:
Would it make more sense to install it in `/usr/local/Cellar/lmod/5.9/init/…` instead?

drpowell:
Yes, that probably makes more sense.

I have written a basic script to write the Lmod module definition.  It only adds the cellar bin directory to the PATH at the moment - no handling of library paths nor dependencies.  
https://github.com/MonashBioinformaticsPlatform/bio-ansible/blob/master/scripts/brew2lmod.rb

And I've incorporated it into an ansible task to install a package with linuxbrew, and make it available as an Lmod pakage:
https://github.com/MonashBioinformaticsPlatform/bio-ansible/blob/master/tasks/brew_package.yml

sjackman:
```
❯❯❯ brew install --build-bottle lmod
❯❯❯ brew bottle -v lmod
==> Determining lmod bottle revision...
==> Bottling lmod-5.9.yosemite.bottle.tar.gz...
tar czf /usr/local/Library/Taps/homebrew/homebrew-science/lmod-5.9.yosemite.bottle.tar.gz lmod/5.9
Warning: String '/usr/local/opt' still exists in these files:
/usr/local/Cellar/lmod/5.9/lmod/5.9/libexec/Configuration.lua
 --> match 'tbl.path_lua   = { k = "Path to Lua"                   , v = "/usr/local/opt/lua/bin",        }' at offset 0xf8c
/usr/local/Cellar/lmod/5.9/lmod/5.9/libexec/MasterControl.lua
 --> match 'local luaprog = "/usr/local/opt/lua/bin/lua"' at offset 0xfdf
/usr/local/Cellar/lmod/5.9/lmod/5.9/libexec/MT.lua
 --> match 'local path   = "/usr/local/opt/lua/bin:" .. os.getenv("PATH")' at offset 0xc8ee
/usr/local/Cellar/lmod/5.9/lmod/5.9/libexec/myGlobals.lua
 --> match 'LUAC_PATH = "/usr/local/opt/lua/bin/luac"' at offset 0x1155
Warning: String '/usr/local/Cellar' still exists in these files:
/usr/local/Cellar/lmod/5.9/lmod/5.9/init/env_modules_python.py
 --> match 'commands = os.popen('/usr/local/Cellar/lmod/5.9/lmod/lmod/libexec/lmod python %s %s'\' at offset 0x44
/usr/local/Cellar/lmod/5.9/lmod/5.9/init/lmod_bash_completions
 --> match '/usr/local/Cellar/lmod/5.9/lmod/lmod/libexec/lmod bash -t avail  2>&1 > /dev/null | sed ' /:$/d; s#/*$##g;'' at offset 0x13
 --> match '/usr/local/Cellar/lmod/5.9/lmod/lmod/libexec/lmod bash -t spider  2>&1 > /dev/null' at offset 0x96
 --> match '/usr/local/Cellar/lmod/5.9/lmod/lmod/libexec/lmod bash -t list  2>&1 > /dev/null | sed ' /^ *$/d; /:$/d; s#/*$##g;'' at offset 0x108
 --> match '/usr/local/Cellar/lmod/5.9/lmod/lmod/libexec/lmod bash -t list  2>&1 > /dev/null | sed ' /^ *$/d; /:$/d; s#/*$##g; s|^|-|g;'' at offset 0x1a3
/usr/local/Cellar/lmod/5.9/lmod/5.9/init/perl
 --> match 'eval `/usr/local/Cellar/lmod/5.9/lmod/lmod/libexec/lmod perl @_`;' at offset 0x1c
/usr/local/Cellar/lmod/5.9/lmod/5.9/libexec/Configuration.lua
 --> match 'tbl.prefix     = { k = "Lmod prefix"                   , v = "/usr/local/Cellar/lmod/5.9",             }' at offset 0xec6
 --> match 'tbl.mpath_root = { k = "MODULEPATH_ROOT"               , v = "/usr/local/Cellar/lmod/5.9/modulefiles",    }' at offset 0x14ea
/usr/local/Cellar/lmod/5.9/lmod/5.9/modulefiles/Core/lmod/5.9.lua
 --> match 'prepend_path('PATH','/usr/local/Cellar/lmod/5.9/lmod/lmod/libexec')' at offset 0x49
/usr/local/Cellar/lmod/5.9/lmod/5.9/modulefiles/Core/settarg/5.9.lua
 --> match 'local base        = "/usr/local/Cellar/lmod/5.9/lmod/lmod/settarg"' at offset 0x0

./lmod-5.9.yosemite.bottle.tar.gz
  bottle do
    sha256 "83f677cd50814d2084defb2721cf0026ab1a20849d298b241c5564118f2ce05b" => :yosemite
  end
```

sjackman:
Although I'd prefer to install to move everything to `Cellar/lmod/5.9` rather than `Cellar/lmod/5.9/lmod/5.9`, lmod installs files that references the symlink `lmod/lmod -> 5.9`, which may confuse Homebrew. It's surmountable, I think, but would require a bit of hacking that I don't have time for at the moment. I'd be happy to accept a pull request. I'm gong to merge this PR as is without reorganizing the directory structure.

sjackman:
Scratch that. I came up with a simple solution.

```ruby
    # Move prefix/lmod/VERSION/lmod/VERSION/* to prefix/lmod/VERSION/*
    mv Dir[prefix/"lmod/#{version}/*"], prefix
    rmdir prefix/"lmod/#{version}"
    (prefix/"lmod").install_symlink ".." => version
```

sjackman:
@BrewTestBot test this please

sjackman:
```
❯❯❯ brew pull https://github.com/Homebrew/homebrew-science/pull/1939 --bottle
==> Applying patch
Applying: lmod 5.9: New formula
==> Patch closes issue #1939
Error: Failure while executing: /usr/bin/curl -f#LA Homebrew\ 0.9.5\ (Ruby\ 2.0.0-481;\ Mac\ OS\ X\ 10.10.2) --silent --fail -o /dev/null -I https://github.com/BrewTestBot/homebrew-science/compare/homebrew:master...pr-1939
```

PR #1953, opened 2015-03-14
prokka 1.11

### LINKED ISSUE #1947: Prokka 1.10 - missing dependencies
@sjackman  Shaun I am hoping you can help me update the prokka.rb recipe

1. It need the Perl module "Time::Piece" which is not installed on many OS by default (I thought it was a core module but it seems not)

2. Can it be made to install 'infernal' dependency *before* the 'hmmer' dependency to avoid the #1931 problem?

3. Can we upgrade it to prokka 1.11 (don't use 1.12 yet)

```
  url "http://www.vicbioinformatics.com/prokka-1.11.tar.gz"
  sha1 "4e6424366d0203717d44e33700bd88d84da94ccb"
```

I'm in the process of setting up a new compute system and am using LinuxBrew as the basis, so once I get the basics set up I will start doing proper pull requests for all this stuff rather than burdening you all!

Torst

saketkc:
Fixes #1947
Made infernal a hard dependency


sjackman:
Thanks for your contribution, Saket.

PR #1954, opened 2015-03-14
r: add caveat about libs_user

tdsmith:
R won't install packages to a user lib directory unless it exists; if it exists, it will use it by default, instead of writing packages to the Cellar. Packages installed to the user lib directory survive minor R upgrades; packages installed to the Cellar do not.

The default pathname is [different on Linux](https://stat.ethz.ch/R-manual/R-devel/library/base/html/libPaths.html) so this won't work there.

We should probably do some symlink magic in the Cellar like we do for pythons in core so that this Just Works for the global package paths, but I'm afraid of breaking Linux.


r.rb
```diff
@@ -154,6 +162,11 @@ def caveats
         (where <version> can be found by running `java -version` or `locate jni.h`)
       EOS
     end
+    s
+  end
+
+  def xy
+    version.to_s.slice(/(\d.\d)/) || "3.1"
```
tdsmith: The `|| "3.1"` here provides a cheat to use when `version` is `HEAD`; not sure if this whole expression is better or worse than just `"3.1"`.

iml:
:+1:

PR #1956, opened 2015-03-14
mlst 1.0 (new formula)

tseemann:
Scan contig files against PubMLST typing schemes.


tseemann:
Improved test & squashed!

tseemann:
@sjackman Should be good to go now.

sjackman:
Done.

PR #1958, opened 2015-03-14
oases 0.2.08 (new formula)

tseemann:
I've had to use 'head' because it uses git submodules which aren't included in .tar.gz downloads.


saketkc:
It looks like the `Makefile` supports passing `VELVET_DIR` as an option.
You can declare `vevlet` as a dependency and then pass it's installation like `Formula["velvet"].lib`

tseemann:
It directly compiles, then links in the ```*.o``` object files from the velvet source code, it does not use a library.

Will that still work?

saketkc:
Looking at the Makefile, it looks like it would require number of changes. You can declare `velvet` as a resource. [Example](https://github.com/Homebrew/homebrew-science/blob/master/sratoolkit.rb#L19)

And for avoiding the `head` warning, you could use a github commit archive download, something like:
https://github.com/dzerbino/oases/archive/645af43cdf.tar.gz

tseemann:
Thanks @saketkc for the "resource" tip. I'll have another attempt soon.

saketkc:
Just a minor correction, if you name the resource `velvet`, you can pass the path like:
`make VELVET_DIR=#{velvet}`

sjackman:
You can open an issue upstream asking the Daniel to tag a release on GitHub.

tseemann:
@BrewTestBot test this please

tseemann:
@BrewTestBot test this please

sjackman:
Nice one, Torsten. That wasn't an easy formula.

PR #1961, opened 2015-03-15
insighttoolkit 4.7.1

iMichka:
Updated to ITK 4.7.1

I have great news. The Python wrapping works again (after a lot of work done fixing it during the last year). For the moment one can only build it using the HEAD version. The next stable version (ITK 4.8.0) will bring this feature back.

### DIFF
--- a/insighttoolkit.rb
+++ b/insighttoolkit.rb
@@ -1,11 +1,8 @@
-require 'formula'
-
 class Insighttoolkit < Formula
-  homepage 'http://www.itk.org'
-  url 'https://downloads.sourceforge.net/project/itk/itk/4.6/InsightToolkit-4.6.1.tar.gz'
-  sha1 '95b879b17a8e2c4727b363037a3af4377095422d'
-  head 'git://itk.org/ITK.git'
-  revision 1
+  homepage "http://www.itk.org"
+  url "https://downloads.sourceforge.net/project/itk/itk/4.7/InsightToolkit-4.7.1.tar.gz"
+  sha1 "70815c884b82899c51c7563296ba4b9ac0bf5e26"
+  head "git://itk.org/ITK.git"
 
   bottle do
     root_url "https://downloads.sf.net/project/machomebrew/Bottles/science"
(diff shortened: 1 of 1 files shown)


insighttoolkit.rb
```diff
@@ -2,8 +2,8 @@
 
 class Insighttoolkit < Formula
   homepage 'http://www.itk.org'
-  url 'https://downloads.sourceforge.net/project/itk/itk/4.6/InsightToolkit-4.6.1.tar.gz'
-  sha1 '95b879b17a8e2c4727b363037a3af4377095422d'
+  url 'https://downloads.sourceforge.net/project/itk/itk/4.7/InsightToolkit-4.7.1.tar.gz'
+  sha1 '70815c884b82899c51c7563296ba4b9ac0bf5e26'
   head 'git://itk.org/ITK.git'
   revision 1
```
iml: Remove the revision when upgrading to new versions.

insighttoolkit.rb
```diff
@@ -33,11 +33,10 @@ class Insighttoolkit < Formula
   option 'remove-legacy', 'Disable legacy APIs'
   option 'with-review', 'Enable modules under review'
 
-  if build.with? 'python'
+  if build.with? 'python' and !build.head?
```
iml: `if build.with?("python") && build.stable?` See `brew audit --strict`

iMichka:
Brew audit should be happier now :)

iMichka:
The yosemite build passed, the two others were aborted (don't know why). Did you only trigger one build ?

Homebrew Science Pull Requests » mountain_lion completed with result ABORTED
Homebrew Science Pull Requests » yosemite completed with result SUCCESS
Homebrew Science Pull Requests » mavericks completed with result ABORTED

iml:
The bot currently times out after an hour, so VTK couldn't be bottled and as this depends on VTK it also times out. Since Yosemite gets in just under the wire, it still might be possible to bottle things again with some tweaking. Otherwise this will need to be pulled sans bottle.

iMichka:
Maybe we can get this in without bottles for the moment.
Does the time out include the build of all dependencies ? Maybe it should only take into account the build time of the updated package (though that may be difficult to implement). Or would it possible to "whitelist" ITK, as this is a package known to need some time to build. And those big packages are then also especially interesting to have as bottles to speed up installation :)

iml:
The timeout is per-machine for each PR, including dependencies. I'm not sure we can make exceptions at this point but @mikemcquaid would be able to answer that. The bots have been pretty busy recently. I'll go ahead pull this with just the Yosemite bottle, though.

mikemcquaid:
We can't really make exceptions for now because the bot machines are having issues. Once they get resolved the speed will increase.

PR #1966, opened 2015-03-16
opencv 2.4.11

kashif:


iml:
I will look into this later but I'm a bit concerned that the bottle here appears to be less than half the size of the existing one.

kashif:
true... one possible reason could be that it is not building for some architecture e.g. `x86_64`... I am also checking...

iml:

tdsmith:
It would be nice to land this together with #2005.

tdsmith:
Accepted as 98e9515. Thanks!

PR #1967, opened 2015-03-16
Artemis 16.0.11 (new formula)

tseemann:
Artemis is a free genome browser and annotation tool that allows visualisation of sequence features, next generation data and the results of analyses within the context of the sequence, and also its six-frame translation.
Artemis is written in Java, and is available for UNIX, Macintosh and Windows systems. It can read EMBL and GENBANK database entries or sequence in FASTA, indexed FASTA or raw format. Other sequence features can be in EMBL, GENBANK or GFF format.


artemis.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    bin.write_jar_script java/jar, "dnaplotter", opts
+    inreplace bin/"dnaplotter", "-jar", "-cp"
+    inreplace bin/"dnaplotter", '"$@"', 'uk.ac.sanger.artemis.circular.DNADraw "$@"'
+  end
+
+  test do
+    system "grep", "Artemis", "#{bin}/art"
+    system "grep", "Act", "#{bin}/act"
+    system "grep", "DNA", "#{bin}/dnaplotter"
```
sjackman: This test only check that the specified string exists in the shell script, which will always pass. Did you mean to run the shell script? If so, see [IGV](https://github.com/Homebrew/homebrew-science/blob/master/igv.rb#L18) for an example.

sjackman:
Please squash the two commits to a single commit.

sjackman:
For future PR, you can amend the commit using `git commit --amend` followed by `git push --force`.

artemis.rb
tseemann: No, that's a whoopsie. Will change to IGV example. Thanks!

tseemann:
Not sure if I messed up the commit squash ... my git-foo is lacking. I tried to do "git reset HEAD~2" ... "git push --force".

saketkc:
It looks okay. However the test block requires invocation of `swing` based GUI and is hence failing. Is there an option to run it without invoking of GUI?

tseemann:
No, they are purely GUI. Now I remember why I had the 'wierd' tests @sjackman asked me about!

@saketkc What is best strategy for GUI-only apps in test clause?

saketkc:
I am not really sure about that. But, I don't think there is a way to test such apps then, unless they come with some bundled standalone script that can be run headlessly.

sjackman:
If there's no way to test them without invoking the GUI, I'd recommend from the `test do` block and add a comment that the apps are GUI only.

tseemann:
@sjackman Ok, I have removed the test block and replaced with a comment about being GUI only. Should be good to go now.

tseemann:
Do I need to put a dummy test block in?

```
Error: 1 problems in 1 formulae
==> audit problems
artemis:
```

sjackman:
Perhaps just to silence the audit warning. I've committed it anyways.

PR #1970, opened 2015-03-17
rstudio server 0.98 (new formula)

randy3k:
I made a PR in the core homebrew https://github.com/Homebrew/homebrew/pull/37668, but it was advised that
`rstudio-server` should go to homebrew/science.


rstudio-server.rb
```diff
@@ -0,0 +1,109 @@
+class RstudioServer < Formula
+  homepage "http://www.rstudio.com"
+  url "https://github.com/rstudio/rstudio/archive/v0.98.1103.tar.gz"
+  sha256 "084049aae03cbaaa74a848f491d57cffab5ea67372ece0e256d54e04bd5fc6da"
+
+  depends_on "ant" => :build
+  depends_on "cmake" => :build
+  depends_on "homebrew/versions/boost150" => :build
+  depends_on "homebrew/science/r"
```
dpo: There's no need to specify `/homebrew/science` here.

randy3k:
I have updated the `depends_on` command.

iml:
Don't know why there are issues with Mountain Lion now, and not earlier. Pulling without those bottles rather than spam the bot further.

randy3k:
May be due to the version of `ant` or `java`?

I see the failing line is 
```
BUILD FAILED
/private/tmp/rstudio-server20150323-70318-5mhh49/rstudio-0.98.1103/src/gwt/build.xml:108: Class not found: javac1.8
```
and `javac1.8` error is a result of `ant` < v1.9 [(link)](http://stackoverflow.com/questions/20702626/javac1-8-class-not-found).

iml:
Ah, OK it looks like `ant` is keg-only for `MacOS.version < :mavericks` and it's picking up system ant there.

PR #1973, opened 2015-03-17
Muscle 3.8.31 (enable RHEL Linux support, does not affect Mac OS X)

tseemann:
The Makefile for Muscle attempts to create a statically linked binary (libstdc++) only for Linux.

This will fail in the general case on modern Linux. There is no need for it to be static. The Mac version simply links dynamically. Best to be consistent.


muscle.rb
```diff
@@ -22,6 +22,13 @@ def install
               "#include <mach/task_info.h>",
               "#include <mach/vm_statistics.h>\n#include <mach/task_info.h>"
 
+    # This patch makes 3.8.31 build on RHEL 7.x
+    # It ONLY affects Linux (in an "if Linux" clause in the 'mk' script)
+    # It is unnecessary to create a static binary
```
sjackman: Please report in a comment the error message that you're fixing.

tseemann: @sjackman ok, added.

sjackman: Hmm. I don't see it. Did you push it?

tseemann: Sorry I thoguht you meant a comment in the code, here is the Linux error message:

```
strip: 'muscle': No such file
ls: cannot access muscle: No such file or directory
sum: muscle: No such file or directory
make: *** [muscle] Error 1
```

tseemann:
@sjackman Any chance this one can be accepted now?   So many things depend on it.

sjackman:
Done. Thanks for your contribution, Torsten.

PR #1978, opened 2015-03-17
SCAMP 2.0.4: New formula

mwcraig:
This pull request adds scamp, used to calibrate both positions and magnitudes (brightnesses) of objects in astronomical images.

There are a number of patches; all but one modify the code to use the Accelerate framework for linear algebra instead of ATLAS. One modification fixes a bug in the released version of scamp.

All of the modifications have been reported via email to the upstream maintainer. 

In the interests of complete disclosure, as far as I can tell `make check` doesn't actually do anything in scamp.

Please let me know if there are changes that need to be made before merging.

### DIFF
--- a/scamp.rb
+++ b/scamp.rb
@@ -1,0 +1,48 @@
+class Scamp < Formula
+  homepage "http://www.astromatic.net/software/scamp"
+  url "http://www.astromatic.net/download/scamp/scamp-2.0.4.tar.gz"
+  sha256 "cbcd57f5042feefa081dc0c5ff07f7f50114a7ef41e79c060ed163caae119d41"
+
+  depends_on "fftw"
+  depends_on "cdsclient"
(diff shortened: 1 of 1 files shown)


mwcraig:
The changes I made should fix the build errors. Apologies that the `test do` block doesn't actually do a very useful test.

dpo:
Were the patched submitted upstream?

mwcraig:
Yes; I expect the patch for the segfault will be included in a future release (though time between releases is very long for this software) but the patches to use the Accelerate framework instead of ATLAS may never be integrated.

iml:
Can you squash these to a single commit, that will also trigger a new build.

mwcraig:
@iml --squashed, but didn't seem to trigger the new build...

iml:
I think the bots have been disabled for today.

iml:

PR #1984, opened 2015-03-19
Updated to kraken-0.10.5-beta.tgz

nitesh1989:


saketkc:
`install.sh` will require a patch.
From `install.sh`:
```
# Perl cmd used to canonicalize dirname - "readlink -f" doesn't work
# on OS X.
export KRAKEN_DIR=$(perl -MCwd=abs_path -le 'print abs_path(shift)' "$1")
```

saketkc:
```
$ brew audit kraken.rb --strict

==> audit problems
kraken:
 * Stable: SHA1 checksums are deprecated, please use SHA256
 * `require "formula"` is now unnecessary
```

PR #1989, opened 2015-03-20
octave 3.8.2

jawshooah:
https://savannah.gnu.org/bugs/?41337 was fixed in 3.8.2, so the corresponding patch is no longer needed.


jawshooah:
@BrewTestBot Test this please

tdsmith:
dynare needs to be rebuilt against new octave; can you add a commit to this PR that bumps dynare's revision?

jawshooah:
@tdsmith Will that work?

iml:
I'll pull this in with #2022 tomorrow or this weekend.

PR #1990, opened 2015-03-20
kraken 0.10-5-beta

saketkc:
Should supersede @nitesh1989's #1984 if it succeeds.


nitesh1989:
This works on my machine. Thanks @saketkc !

saketkc:
Did you test it on a Mac? I don't have access to one. 
@BrewTestBot seems to be slow today

nitesh1989:
Yes, I did. It works fine. It installs like its supposed to.

PR #1995, opened 2015-03-22
graph-tool: support building for both python version at the same time

tnilanon:
- support building for both python version at the same time

### DIFF
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -15,6 +15,7 @@ class GraphTool < Formula
   option "with-gtk+3", "Build with gtk+3 support for interactive plotting"
 
   cxx11 = MacOS.version < :mavericks ? ["c++11"] : []
+  with_pythons = build.with?("python3") ? ["with-python3"] : []
 
   depends_on "pkg-config" => :build
   depends_on "boost" => cxx11
@@ -24,30 +25,32 @@ class GraphTool < Formula
   depends_on "gtk+3" => :optional
   depends_on :python => :recommended
   depends_on :python3 => :optional
+  depends_on "boost-python" => cxx11 + with_pythons
 
   if build.with? "gtk+3"
     depends_on "gnome-icon-theme"
     depends_on "librsvg" => "with-gtk+3"
+    depends_on "pygobject3" => with_pythons
   end
 
-  if build.with? "python3"
-    depends_on "boost-python" => cxx11 + ["with-python3"]
-    depends_on "py3cairo" if build.with? "cairo"
-    depends_on "pygobject3" => "with-python3" if build.with? "gtk+3"
-    depends_on "matplotlib" => :python3
-    depends_on "numpy" => :python3
-    depends_on "scipy" => :python3
-  elsif build.with? "python"
-    depends_on "boost-python" => cxx11
+  if build.with? "python"
     depends_on "py2cairo" if build.with? "cairo"
-    depends_on "pygobject3" if build.with? "gtk+3"
     depends_on "matplotlib" => :python
     depends_on "numpy" => :python
     depends_on "scipy" => :python
   end
 
+  if build.with? "python3"
+    depends_on "py3cairo" if build.with? "cairo"
+    depends_on "matplotlib" => :python3
+    depends_on "numpy" => :python3
+    depends_on "scipy" => :python3
+  end
+
   def install
     ENV.cxx11
+
+    system "./autogen.sh" if build.head?
 
     config_args = %W(
       --disable-debug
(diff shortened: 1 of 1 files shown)


graph-tool.rb
```diff
@@ -56,22 +61,32 @@ def install
...
+      begin
+        if version.to_s.start_with?("3")
+          inreplace "configure", "libboost_python", "libboost_python3"
+        elsif version.to_s.start_with?("2")
+          inreplace "configure", "libboost_python3", "libboost_python"
+        else
+          opoo "Python version not recognized"
+        end
+      rescue Utils::InreplaceError
+      end
```
tdsmith: I'd reduce this begin..end block to just:

```
if python == "python3"
  inreplace ...
end
```

I promise that `each_python` always runs in the order `python`, `python3` and that `python` will never take any other value. If `python4` is ever released, we can deal with that later. ;)

graph-tool.rb
```diff
@@ -56,22 +61,32 @@ def install
...
+
+      mkdir "build-#{python}-#{version}" do
+        system "../configure", "PYTHON_EXTRA_LDFLAGS= ", *config_args, *config_args_x
```
tdsmith: may as well wrap `PYTHON_EXTRA_LDFLAGS` into `config_args`

graph-tool.rb
```diff
@@ -56,22 +61,32 @@ def install
...
 
-    system "./autogen.sh" if build.head?
-    inreplace "configure", "libboost_python", "libboost_python3" if build.with? "python3"
-    system "./configure", "PYTHON_EXTRA_LDFLAGS= ", *config_args
-    system "make", "install"
+    Language::Python.each_python(build) do |python, version|
+      config_args_x = ["PYTHON=#{python}"]
+      config_args_x << "CFLAGS=#{`#{python}-config --cflags`.chomp}"
+      config_args_x << "LDFLAGS=#{`#{python}-config --ldflags`.chomp}"
+      config_args_x << "LIBS=#{`#{python}-config --libs`.chomp}"
```
tdsmith: What happens without these extra FLAGS, out of curiosity? I'm surprised graph-tool can't figure this out on its own.

graph-tool.rb
```diff
@@ -15,6 +15,7 @@ class GraphTool < Formula
   option "with-gtk+3", "Build with gtk+3 support for interactive plotting"
 
   cxx11 = MacOS.version < :mavericks ? ["c++11"] : []
+  tags = []
```
tdsmith: Can you call this `python_tags` or `with_pythons`? To make what this is more obvious I'd probably prefer to write something here like the `cxx11` expression above instead of pushing to the list in the `if` below.

graph-tool.rb
tnilanon: Thanks! How do I go about doing this? `git revert`?

graph-tool.rb
tnilanon: I tried `config_args << "PYTHON_EXTRA_LDFLAGS= "` but the space gets dropped by `configure`. Any idea?

graph-tool.rb
tnilanon: `configure` works fine without `CFLAGS` and `LIBS`. This is me just avoiding corner (and weird) cases.

graph-tool.rb
tnilanon: That is a great idea!

graph-tool.rb
tdsmith: Just make an additional commit; you can squash all your commits with `git rebase -i` and force-push to your branch on github if you like.

graph-tool.rb
tdsmith: Can you put it first in `config_args`? They should be logically the same.

Also, just remembered: `system *list1, *list2` doesn't work on Ruby 1.8 (on e.g. Mountain Lion, IIRC); can you use `*(config_args + config_args_x)`?

graph-tool.rb
tdsmith: Is it correct to get rid of all of these and set `PYTHON_EXTRA_LDFLAGS` to the output of `python-config --ldflags`?

graph-tool.rb
tnilanon: Nevermind. I don't know what I did differently last time. It works now.

tnilanon:
Should I force the user to install `numpy`, `scipy`, and `matplotlib` via `Homebrew` like they do in `scipy` and `matplotlib` formula? i.e. `depends_on "numpy" => "with-python3"` instead of `depends_on "numpy" => :python3`

graph-tool.rb
tnilanon: :done

graph-tool.rb
tnilanon: :done

graph-tool.rb
tnilanon: :done

graph-tool.rb
tnilanon: :done

tdsmith:
No, since installing those with `pip` is a good option. Something similar to a `NumpyRequirement` that's satisfied by being able to import numpy and falls back to installing the homebrew/python/numpy formula otherwise would be fine to me but I don't know if other maintainers agree.

tnilanon:
Please don't review this yet

tnilanon:
Built & tested

iml:
I'm fine with having the user install their own modules here. I assume `:numpy` hasn't been merged yet. Please squash these to a single commit (`git rebase -i @~12` or so and push with `--force`).

tnilanon:
squashed down to one commit

tdsmith:
:+1: :dancer: Thanks for your work here!

tnilanon:
np! thx for your help too. :snowflake:

PR #1996, opened 2015-03-22
readseq 2.1.19 (new formula)

tseemann:
ReadSeq is a program and library for conversion of biosequence data from one format to another, useful in various bioinformatics programs and services. It is written in Java, though an earlier version in C remains available.


readseq.rb
```diff
@@ -0,0 +1,22 @@
+
```
sjackman: Please remove this leading blank line.

PR #1999, opened 2015-03-23
bowtie2 2.2.5

saketkc:


iml:
I think this was my doing but should only `bowtie`, `bowtie-build`, and `bowtie-inspect` be staged into bin? If so you can install everything to libexec then symlink those three to bin. Otherwise forget what I said.

PR #2003, opened 2015-03-24
freec 7.2 (new formula)

tseemann:
Control-FREEC is a tool for detection of copy-number changes and allelic imbalances (including LOH) using deep-sequencing data. Control-FREEC automatically computes, normalizes, segments copy number and beta allele frequency (BAF) profiles, then calls copy number alterations and LOH. The control (matched normal) sample is optional.


tseemann:
Looks like "values.h" should be "limits.h"
http://lists.apple.com/archives/scitech/2003/Nov/msg00189.html

tdsmith:
Ha, it's item 20 in their FAQ: http://bioinfo.curie.fr/projects/freec/FAQ.html

I was looking at their website hoping to find some trivial, small data to test this with but their test data set is 150 MB which is probably on the large side.

tseemann:
I've managed to mess my git branch up and can't figure out how to recover. Will submit a fresh pull request. Sorry.

PR #2005, opened 2015-03-24
opencv: don't explicitly link python

tdsmith:
This makes opencv safe to bottle without a :python dependency on OS X by preventing segfaults arising from opencv being built with one python and imported from another.

I assume passing the correct library is still important on Linux. Would be grateful for review from @iml et alia.

References Itseez/opencv#3866 and Itseez/opencv#3865.


iml:
Looks good, thanks for submitting this upstream. 

Here's my peculiar cmake output:
```
--   Python:
--     Interpreter:                 /usr/local/bin/python2 (ver 2.7.9)
--     Libraries:                   /usr/lib/libpython2.7.dylib (ver 2.7.6)
--     numpy:                       /usr/local/lib/python2.7/site-packages/numpy/core/include (ver 1.9.2)
--     packages path:               lib/python2.7/site-packages
```
However, importing the module works from system python after unlinking the above brewed NumPy. I wonder if this gets us closer to not defaulting to brewed-numpy, e.g. having the bottles build with their old system NumPy but allowing the user to pick up something newer at runtime.

tdsmith:
I'm not sure. numpy needs to be importable at run time in the user's environment for the installation to be complete and useful. Our options might look like:
 * depends_on "numpy" => :python -- lame, can't be bottled
 * depends_on NumpyRequirement => :recommended which falls back to homebrew/python/numpy -- introduces a hard dependency on the numpy formula for anyone installing from bottles but avoids it for source-builders
 * build against system numpy and print a caveat that you might need to pip install numpy to actually use the bindings in your environment -- makes life harder for things that depend on opencv's python bindings
 * depend on homebrew/python/numpy
 * build and vendor numpy as a resource: avoids cross-tap dependency, dunno if it's okay to bottle

a NumpyRequirement here seems fine to me but I think it will be hard to do better.

iml:
NumpyRequirement seems like the most flexible and provides the best path from our current configuration, which seems I think is quite functional and descriptive but isn't very friendly toward all the other ways of one might prefer to have NumPy installed.

tdsmith:
No go on NumpyRequirement -- check the bot failure: http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/1379/

We'd have to teach fetch how to fetch from untapped taps first.

iml:
Ah ok, probably worth revisiting that later along with https://github.com/Homebrew/homebrew/pull/36810.

tdsmith:
Not sure why the bot is still rejecting numpy! This worked in #1966: http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/1361/version=yosemite/

dpo:
We have the same problem in #1924.

tdsmith:
@BrewTestBot test this please

tdsmith:
Figured out the magic; `depends_on "homebrew/python/numpy" => :optional` will succeed but a depends_on without a tag will fail because of funny conditional ordering in DependencyCollector#parse_string_spec. Opened #38424.

tdsmith:
The upstream PRs were accepted so we can drop the patches on the next OpenCV releases.

PR #1818, opened 2015-02-11
shogun 4.0.0

saketkc:

PR #2010, opened 2015-03-25
shogun 4.0.0

iml:
WIP. Fails to build Python bindings with latest SWIG.


saketkc:
Thanks @iml. This should close #1818?

iml:
Ah, sure. We'll need to get some help from upstream I think.

iml:
https://github.com/shogun-toolbox/shogun/issues/2619#issuecomment-63199857 seems to indicate SWIG 2 is needed :(

dpo:
Why the dupe with #1818? For Swig 2.08, see https://gist.github.com/dpo/d6cdc6b8f043fc50b2d4

iml:
I wanted to test it on the bots. This should be resolved by https://github.com/shogun-toolbox/shogun/commit/fef8937d215db7f26e15cdc8b21201ee52e69e1f but I had to set `-D__ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES=0` instead.

iml:

shogun.rb
```diff
@@ -1,83 +1,80 @@
...
+  url "http://shogun-toolbox.org/archives/shogun/releases/4.0/sources/shogun-4.0.0.tar.bz2"
+  sha256 "ab39d3cc27fc1ddf6128f6e3fc60217e73a843b51c20f5d3ceb69e6565a43695"
+
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+  depends_on "xz"
+  depends_on "lzo"
+  depends_on "snappy"
+  depends_on "hdf5"
+  depends_on "protobuf" => MacOS.version < :mavericks ? "c++11" : ""
```
dpo: This causes an error because `protobuf` has no option named `""`. You could try
````
depends_on "protobuf" => MacOS.version < :mavericks ? ["c++11"] : []
````

iml: Thanks

PR #2015, opened 2015-03-26
links-scaffolder 1.1 (new formula)

warrenlr:
changed links.rb for links-scaffolder.rb


sjackman:
Thanks for the LINKS formula, Rene.

PR #2021, opened 2015-03-27
etsf_io 1.0.4 (new formula)

mtorrent:
A library of F90 routines to read/write the ETSF file format, called ETSF_IO.
This library is used by the ABINIT code, already available in homebrew.


dpo:
@BrewTestBot test this please

PR #2022, opened 2015-03-27
octave: don't depend on :tex

### LINKED ISSUE #2008: Octave Docstring Installation Error
I've been trying to setup Octave (a package in the homebrew/science tap). I ran:

    brew install octave

and it works fine, but running:

    octave --silent

produces the error:

    warning: docstring file '/usr/local/Cellar/octave/3.8.1_1/share/octave/3.8.1/etc/built-in-docstrings' not found

I looked online it appears some people have fixed this by installing Octave with the `--without-docs` argument, so I uninstalled and reinstalled Octave with this argument:

    brew uninstall --force octave
    brew install octave --without-docs

Still, Octave complains about not having docstrings and I'm not sure how to fix this. I've also reported it [here](http://stackoverflow.com/q/29238053/1944335) and was instructed to create an issue in the repository so people could look into it.

Any help you could provide would be greatly appreciated! Thanks!

iml:
Enable built-in docs by default. 

Fixes #2008.


iml:

iml:
Passes all tests and the help/doc commands seem to work fine. I've cleared out MacTex locally and tested the bottles. Dynare should be handled in another PR so as not to go over the time limit.

PR #2026, opened 2015-03-27
Moved eigen and openexr to be optional

tzickel:
these packages are not needed for the general public, someone keeps adding them because he is using them....


iml:
These are being kept on for ffmpeg interoperability, so we don't have to specify additional options for dependencies that automatically rebuild opencv `with-ffmpeg`. The inconvenience of installing them and not needing them is less than needing them and not knowing why.

PR #2029, opened 2015-03-28
Diamond 0.7.7 (updated formula + compile from source)

tseemann:
Author now has tagged releases on github.
I am compiling from source (uses BOOST).
Author has changed to a fixed Makefile which includes some Darwin logic for the -lrt issue on OSX.
Unable to test on OSX this weekend.


sjackman:
Thanks,  Torsten.

PR #2040, opened 2015-04-03
qualimap: last version with cleaner log and minor bug fixed

lpantano:
There is a better version that avoid to full the log file with the same message. Would be good to have that to update bcbio-nextgen as well. 

The only thing is that bottles need to be updated, but I guess only some people can do it?

thanks

### DIFF
--- a/qualimap.rb
+++ b/qualimap.rb
@@ -2,9 +2,9 @@ require 'formula'
 
 class Qualimap < Formula
   homepage 'http://qualimap.bioinfo.cipf.es/'
-  url 'https://bitbucket.org/kokonech/qualimap/downloads/qualimap-build-06-03-15.tar.gz'
-  sha1 'fc0a227d2739a1e92b95cfabee074dc54d4c7ce6'
-  version '06-03-15'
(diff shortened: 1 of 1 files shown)


qualimap.rb
```diff
@@ -12,7 +12,6 @@ class Qualimap < Formula
     sha256 "1916bbaaf3d8e81cb365fa38d55ca45d80e500d917f73b5da8b207286bdf7900" => :mavericks
```
lpantano: all these hash are wrong, but don't know who has to update https://homebrew.bintray.com/bottles-science

thanks

iml: Those are updated by @BrewTestBot

iml:
Just a warning for this style of versioning. Putting the day in the beginning can cause an upgrade to not be triggered if the new release is an earlier day in a later month. Homebrew doesn't yet have any rules for versioning but the more popular convention is YYYYMMDD, which increments like an integer.

lpantano:
thanks. Will talk to the author about this, hopefully future version will be better.

PR #2042, opened 2015-04-04
SMALT 0.7.6 (new formula)

tseemann:
SMALT efficiently aligns DNA sequencing reads with a reference genome. Reads from a wide range of sequencing platforms, for example Illumina, Roche-454, Ion Torrent, PacBio or ABI-Sanger, can be processed including paired reads.


tseemann:
@sjackman bagged and tagged!

smalt.rb
```diff
@@ -0,0 +1,18 @@
+class Smalt < Formula
+  homepage "https://www.sanger.ac.uk/resources/software/smalt/"
+  # tag "bioinformatics"
+  # doi ""
```
sjackman: Delete this line if there's no publication or preprint.

sjackman:
Thanks, Torsten.

PR #2044, opened 2015-04-04
Suite-Sparse: define matlab inside if block.

dpo:
For a strange reason, the Matlab build could be triggered even if `--with-matlab` is not specified.


dpo:
@BrewTestBot test this please

PR #2047, opened 2015-04-05
InterViews 18 (new formula)

tbekolay:
InterViews is a C++ graphical interface builder for X-windows.
It has a long history, but at this point is mostly irrelevant
except for some legacy science software.

The version included in this formula is used by the NEURON
neural simulator, and has several modifications from the original
InterViews source.

----

This is my first attempt at writing a Homebrew formula, so I would greatly appreciate any criticism, and am happy to make changes as suggested. I'm very comfortable with git / Github, but my Ruby knowledge is rather minimal as I'm a Python programmer. In particular, here are some things that I'm unsure about:

1. Should this formula be more intimately linked to NEURON? E.g., this could be called `neuron-iv` instead of `inter-views`. I'm not sure of any other programs that use `inter-views`.
2. The build system of InterViews puts binaries in architecture specific directories; for me, this is `x86_64/bin`, but I believe `ppc` and `i386` are possibilities as well. For now, I'm just installing symlinks to the files in `x86_64`, but this should likely be handled better. I tried to grep for other formulas that handle this gracefully, as installing to architecture specific directories must be a common thing to solve, but I couldn't find much. Are there examples of other formulas that handle this?
3. InterViews contains a GUI demo that is useful for making sure that the build was successful. However, this (and all other binaries) are graphical programs using X11; they need user intervention to close once they've been opened. I used this demo in the `test do` block, but I suspect that this won't work for any automated testing. Is there a suitable way to test this GUI library? Or should I test that certain files exist and are executable?

Thanks!


inter-views.rb
```diff
@@ -0,0 +1,21 @@
+class InterViews < Formula
+  homepage "http://www.neuron.yale.edu/neuron/"
+  url "http://www.neuron.yale.edu/ftp/neuron/versions/v7.3/iv-18.tar.gz"
+  sha256 "a875692a20211e0856e9e283ab9ef5da022b4d49853aa7f2f734104f399e7af1"
+
+  depends_on :x11
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
```
dpo: I would suggest installing to `#{libexec}` instead of `#{prefix}` in this case.

inter-views.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    system "make"
+    system "make", "install"
+    bin.install_symlink "../x86_64/bin/iclass"
+    bin.install_symlink "../x86_64/bin/idemo"
+    bin.install_symlink "../x86_64/bin/idraw"
```
dpo: Then you can symlink `#{libexec}/x86_64/bin/...`

dpo:
Thanks for your contribution! Here are some comments:

1. It's best for this formula to remain `inter-views` so others can depend on it (who knows, it could happen).
2. It seems you can do `./configure --prefix=#{libexec} --exec-prefix=#{libexec}` and that should take care of the `x86_86` (or whatever) part of the path.
3. Indeed the demo is what's causing the build bot to fail currently. It's probably ok to just check that `idemo` is in `#{bin}` and is executable in this case.

tbekolay:
Thanks so much for your feedback @dpo! I've made the changes you suggested, rebased to master, and force pushed this branch. Happy to make any other improvements you can see also!

One quick question, how did you find the `--exec-prefix` flag? I'm no too familiar with Make; is there some way to interrogate Makefiles for their flags, or did you read through the Makefile?

dpo:
If you do `brew install -i inter-views` you will drop to a shell where the package has been unpacked. You can look and play around like in a normal shell. I found `--exec-prefix` by looking at `./configure --help`.

PR #2048, opened 2015-04-05
NEURON neural simulator

tbekolay:
NEURON is a simulator for biologically realistic neurons.
It includes a GUI that can be used to build 3D models of neuron
morphology and run experiments by manipulating parameters,
a scripting interface called Hoc for building models,
and Python bindings to build models with Python rather than Hoc.

----

This PR depends on the InterViews PR that I just submitted (#2047). Again, this is my first shot at a Homebrew formula, so criticism would be appreciated. In particular:

1. The patches made are mostly Homebrew specific. They've also been previously reported in the NEURON forums (unfortunately, they don't have a bug tracker), and workarounds were given rather than solutions.
2. Again, puts binaries in architecture specific directories. Not sure how best to move these to `bin`.

Thanks!


tseemann:
@tbekolay I'm also recently new to brew packaging.  You seem to have 2 commits in this pull request. The maintainers will want separate pull requests for each formula I think.

tbekolay:
Hi @tseemann! Yes, the first commit is from another PR I submitted (#2047) which this PR depends on. I wasn't sure the best way to separate them; perhaps wait until the other one is merged?

tseemann:
I went through the same problem. You have to do the pull requests each in their own branch. 
```
git checkout master
git checkout -b neuron-7.3
# edit neuron.rb
git commit neuron.rb
git push
```
Then come to github and submit the pull request.
Then repeat for the other formula on a different branch.

It is sort of explained here:
https://github.com/Homebrew/homebrew/tree/master/share/doc/homebrew#contributing
But not 100% obvious when you are just learning git like me.

tbekolay:
Sorry, I should have said that I deliberately had both of them in this branch. This branch depends on that other branch, which is to say that the NEURON formula depends on the InterViews formula, which is the subject of #2047. So, if I based this branch off of master, it wouldn't build or pass tests. Once the other PR is merged, I can rebase this branch to master, but for the time being I wanted to submit this anyhow to get feedback on this formula. If this is premature, then I can close this and wait for #2047 to resolve (its only real purpose is to support this formula though, so I'm not sure how best to manage this given Homebrew's policy of one formula per PR).

neuron.rb
```diff
@@ -0,0 +1,112 @@
...
+  # 2. The build attempts to do python setup.py install and passes a
+  #    --home argument, which conflicts with the --prefix argument
+  #    that Homebrew sets up through an RC file.
+  #    The patch removes the python setup.py install invocation;
+  #    we instead do this manually after make install.
+  patch :DATA
+
+  def install
+    iv_path = Formula["inter-views"].opt_prefix
+    args = ["--with-iv=#{iv_path}"]
```
dpo: If you don't reuse `iv_path` below, let's make this a one liner:
````
args = ["--with-iv=#{Formula["inter-views"].opt_prefix}"]
````

File: neuron.rb [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+
+    if build.with? "mpi"
+      args << "--with-paranrn"
+    end
```
dpo: `args << "--with-paranrn" if build.with? "mpi"`

neuron.rb
```diff
@@ -0,0 +1,112 @@
+class Neuron < Formula
+  homepage "http://www.neuron.yale.edu/neuron/"
+  url "http://www.neuron.yale.edu/ftp/neuron/versions/v7.3/nrn-7.3.tar.gz"
+  sha256 "71cff5962966c5cd5d685d90569598a17b4b579d342126b31e2d431128cc8832"
+
+  depends_on "inter-views"
+  depends_on :python => :recommended
+  depends_on :mpi => :optional
```
dpo: If you only need support for certain languages, you can specify them here, e.g., `[:cc, :optional]`

neuron.rb
```diff
@@ -0,0 +1,112 @@
...
+    iv_path = Formula["inter-views"].opt_prefix
+    args = ["--with-iv=#{iv_path}"]
+
+    if build.with? "mpi"
+      args << "--with-paranrn"
+    end
+
+    if build.with? "python"
+      args << "--with-nrnpython"
+    end
```
dpo: `args << "--with-nrnpython" if build.with? "python"`

neuron.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          *args
+
+    system "make"
+    system "make", "install"
```
dpo: Is there some form of `make test` that wouldn't involve the GUI?

neuron.rb
```diff
@@ -0,0 +1,112 @@
...
+    rm_rf "#{prefix}/../mos2nrn.app"
+    rm_rf "#{prefix}/../neurondemo.app"
+    rm_rf "#{prefix}/../nrngui.app"
+
+    bin.install_symlink "../x86_64/bin/mknrndll.sh"
+    bin.install_symlink "../x86_64/bin/modlunit"
+    bin.install_symlink "../x86_64/bin/mos2nrn"
+    bin.install_symlink "../x86_64/bin/neurondemo"
+    bin.install_symlink "../x86_64/bin/nrngui"
+    bin.install_symlink "../x86_64/bin/nrniv"
```
dpo: There might be a similar `--exec-prefix=...` game to play here.

neuron.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  test do
+    system "#{bin}/nrniv", "--version"
+    if build.with? "python"
+      system "python", "-c", "import neuron; neuron.test()"
+    end
```
dpo: `system "python", "-c", "import neuron; neuron.test()"` if build.with? "python"` (we like one liners)

neuron.rb
dpo: The documentation for `--with-mpi` says that it will build Sundials. It would be good to see if `neuron` can be built against the Homebrew `sundials` instead.

neuron.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    system "make"
+    system "make", "install"
+
+    cd "src/nrnpython" do
+      system "python", *Language::Python.setup_install_args(prefix)
+    end
+
+    # Since we manually install Python, we don't want lib to be linked
+    rm_rf "#{lib}"
```
dpo: Could this be handy?
````
  --enable-pysetup=installoption
                          Execute 'python setup.py install installoption' as
                          the last installation step. --disable-pysetup or an
                          installoption of 'no' means do NOT execute 'python
                          setup.py...' The default installoption is
                          '--home=<prefix>'
````

File: neuron.rb (hunk 7 above)
Comment (tbekolay) [inline neuron.rb]: Thanks! I do too; just wasn't sure of the correct Ruby syntax coming from a Python background :smile:

File: neuron.rb [hunk 9]
```diff
@@ -0,0 +1,112 @@
+class Neuron < Formula
+  homepage "http://www.neuron.yale.edu/neuron/"
+  url "http://www.neuron.yale.edu/ftp/neuron/versions/v7.3/nrn-7.3.tar.gz"
+  sha256 "71cff5962966c5cd5d685d90569598a17b4b579d342126b31e2d431128cc8832"
+
+  depends_on "inter-views"
+  depends_on :python => :recommended
```
tdsmith: If this uses distutils to build the python bits, a :python dependency probably isn't necessary since a build with system python should import fine from any other python 2.7. You can verify this is safe by running `otool -L` against any .so files installed to site-packages; this is safe unless there is a link to libpython or a Python framework executable in the otool output.

Unless there's a reason not to build the Python bits, you can make this
`depends_on :python if MacOS.version <= :snow_leopard`

neuron.rb
dpo: No problem.

neuron.rb
tbekolay: Does `python setup.py install` using system Python work without `sudo`?

tdsmith: Yep, because we tell it to install to our prefix instead of to /Library.

tbekolay:
Thanks a lot for all the feedback! I've taken it into account and rebased to master. I've also added a caveat to document something that the NEURON install instructions notes about XQuartz.

Unfortunately, using `--exec-prefix` (for both this and `inter-views`) breaks NEURON because it relies on the `*.la` files that Homebrew removes (for good reason, I know). I was able to patch this in the makefile to get it to build, but unfortunately NEURON has a mechanism to build new neuron models on the fly, and that mechanisms also depends on `*.la` files, so a quick patch isn't going to be sufficient here. I believe inter-views and NEURON need `pkg-config` files added so that the `*.la` files aren't required anymore... or perhaps just those scripts need to be fixed. In either case, this stuff is unfortunately not in the area of my expertise, and I don't have much time to dig deeper at the moment, so I'm just updating this PR for now to let interested parties know what its current state is.

I'm going to ask on the NEURON forums for some help on this for the time being; if anyone sees this and wants to help out, feel free to contact me as well!

tdsmith:
You can add `skip_clean :la` to the Formula class body to retain the .la files, if that's the correct thing to do.

tbekolay:
Thanks @tdsmith! I set that, and now it works great! I think this should be good to go.

Oh, one other note, I changed the `depends_on` line to `depends_on "homebrew/science/inter-views"`. Is this the correct thing to do for dependencies in taps like this one?

iml:
No, just `depends_on "inter-views"` for formulae in the same tap.

tdsmith:
You don't need to qualify references to formulas in the same tap; "inter-views" is fine.

tbekolay:
Thanks, updated!

dpo:
There are a few audit issues left:
````
==> audit problems
neuron:
 * Non-libraries were installed to "/usr/local/Cellar/neuron/7.3/lib"
Installing non-libraries to "lib" is discouraged.
The offending files are:
  /usr/local/Cellar/neuron/7.3/lib/nrnconf.h
 * Non-executables were installed to "/usr/local/Cellar/neuron/7.3/bin"
The offending files are:
  /usr/local/Cellar/neuron/7.3/bin/macnrn.term
  /usr/local/Cellar/neuron/7.3/bin/nrniv_makefile
  /usr/local/Cellar/neuron/7.3/bin/nrnmech_makefile
  /usr/local/Cellar/neuron/7.3/bin/nrnoc_makefile
````
Should the `.h` file go to `include`?

tbekolay:
I'm not sure if `nrnconf.h` is ever used, but moving it to include seems prudent.

Removing `macnrn.term` is no problem, but NEURON uses those makefile as part of its ability to compile new NEURON models at run time (same reason it needs the `.la` files. Is there a way to mark these files as necessary? In the meantime I'll search for a way to put these somewhere more reasonable...

dpo:
You might be able to run `configure` with `--bindir="#{libexec}/bin"`, and then create symbolic links in `bin`. See, e.g., `asl.rb` for an example.

tbekolay:
That seems to work well! Updated and force pushed. Not sure why the Jenkins build failed though...

dpo:
Thank you for your contribution!

tbekolay:
Thanks for all your feedback! Looking forward to contributing more in the future :smile_cat:

PR #2056, opened 2015-04-07
VAGUE 1.0.5 (new formula)

tseemann:
VAGUE is a vague acronym for "Velvet Assembler Graphical Front End", which means it is a GUI for the Velvet de novo assembler. The command line version of Velvet can be complicated for beginners to use, but VAGUE makes it clear and simple.


tseemann:
@BrewTestBot test this please

vague.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  url "http://www.vicbioinformatics.com/vague-1.0.5.tar.gz"
+  sha256 "7144bef4a5c8ccb3d0e6b27201a506e1cfa25812fb9bbe54842dc910100c5907"
+
+  depends_on "velvet"
+
+  def install
+    opts = "-Xmx3750m"
+    jar = "vague.jar"
+    java = share/"java"
```
iml: JARs are usually installed to `libexec`

PR #2057, opened 2015-04-07
VarScan 2.3.7 (new formula)

tseemann:
VarScan is a platform-independent mutation caller for targeted, exome, and whole-genome resequencing data generated on Illumina, SOLiD, Life/PGM, Roche/454, and similar instruments. The newest version, VarScan 2, is written in Java, so it runs on most operating systems.


tseemann:
@BrewTestBot test this please

PR #2060, opened 2015-04-07
PoSSuM-search 2.0 (new formula)

tseemann:
Significant speedup of database searches with HMMs by search space reduction with PSSM family models. (No source code to be found)


possum-search.rb
```diff
@@ -0,0 +1,28 @@
...
+  version "2.0"
+
+  if OS.mac?
+    url "http://bibiserv.cebitec.uni-bielefeld.de/resources/download/possumsearch/PoSSuM-2_0-darwin9.7.0.amd64.tar.gz"
+    sha256 "9087d690bfc816e16c83ef956e5f8b9f1da2ee8d3414275a9922ef9641bb267a"
+  elsif OS.linux?
+    url "http://bibiserv.cebitec.uni-bielefeld.de/resources/download/possumsearch/PoSSuM-2_0-linux-gnu.x86_64.tar.gz"
+    sha256 "1cc656a1288277c9515572f7c3a5b032cc0dea569045d7b1579afae6ebbf3440"
+  else
+    ohoe "Unsupported operating system"
```
dpo: It's `onoe`, not `ohoe`. But this part isn't necessary. Homebrew only runs on OSX and Linux.

tseemann: @dpo 
I was just practicing defensive programming :) This raises a bigger issue in that Linux exists on many architectures, and if there are binary components and not compiling from src, it won't work anyway.

Should it just be ```if OS.mac XXXX else YYYY``` then ?

tseemann: Ok, i've tweaked the formula.

possum-search.rb
```diff
@@ -0,0 +1,26 @@
...
+    url "http://bibiserv.cebitec.uni-bielefeld.de/resources/download/possumsearch/PoSSuM-2_0-darwin9.7.0.amd64.tar.gz"
+    sha256 "9087d690bfc816e16c83ef956e5f8b9f1da2ee8d3414275a9922ef9641bb267a"
+  else
+    url "http://bibiserv.cebitec.uni-bielefeld.de/resources/download/possumsearch/PoSSuM-2_0-linux-gnu.x86_64.tar.gz"
+    sha256 "1cc656a1288277c9515572f7c3a5b032cc0dea569045d7b1579afae6ebbf3440"
+  end
+
+  def install
+    bin.install Dir["bin/possum*"]
+    share.install "share/PoSSuM"
```
dpo: This should probably be `(share/"possum_search").install ...`

tseemann: ok, fixed

tdsmith:
Binary-only and non-free:
> The PoSSuM2 software package, including the program PoSSuMsearch2, is available free of charge for non-commercial research institutions.

My personal inclination would be to decline to accept this package but I'll defer to the other homebrew-scientists.

tseemann:

tseemann:
```
Error Message

failed: brew doctor
Stacktrace

        Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: Homebrew's sbin was not found in your PATH but you have installed
formulae that put executables in /usr/local/sbin.
Consider setting the PATH for example like so
    echo 'export PATH="/usr/local/sbin:$PATH"' >> ~/.bash_profile

Warning: Some installed formula are missing dependencies.
You should `brew install` the missing dependencies:

    brew install epstool erlang fftw fontconfig freetype gd ghostscript gl2ps gmp gnuplot graphicsmagick arpack glpk hdf5 octave qhull qrupdate suite-sparse421 veclibfort imagemagick jpeg libpng libtiff libtool little-cms2 lua pcre plotutils pstoedit pyqt qscintilla2 qt sip szip tbb unixodbc wxmac xz

Run `brew missing` for more details.
```

tseemann:
This binary package has been moved to `tseemann/bioinformatics-linux`.

ilovezfs:
These binaries (I noticed you closed a couple other PRs for the same reason) you're sequestering in your own tap should probably be published via Homebrew Cask if you can be bothered to make that happen. CC @vitorgalvao

vitorgalvao:
Yes, these seem like acceptable candidates to Homebrew-Cask. If you want to submit them as casks, we’ll be happy to take a look.

ilovezfs:
Thanks @vitorgalvao!

tseemann:
@vitorgalvao @ilovezfs   I don't think Casks are supported on Linuxbrew.

vitorgalvao:
@tseemann They are not (since they pertain mostly to macOS GUI apps, there’s no point), so if you need that, running your own tap seems to indeed be best.

ilovezfs:
@tdsmith if these binaries don't have a home in cask, then maybe we need to reconsider the policy in science of not accepting them. Given that homebrew/binary is being deprecated too "because cask" that puts this stuff in Catch-22 situation that isn't exactly being addressed if the solution is to ship things off to TorstenTown.

vitorgalvao:
Please correct me if I’m wrong, but Linuxbrew is not an official part of Homebrew. HB is for macOS first and other systems second.

HBC has no work done to function on Linux (nor do we intend to), but if Linuxbrew were to unofficially do the work and support, I don’t think anyone would object (i.e. we’re not actively against Linix, or anything like that).

I’ve repetedly said taps are my favourite part of HB. Not being tied to the decisions of the main repo is a wonderful feature.

With that in mind, I wonder if changing a policy based on a single small group of formulas not working on an not offiially supported OS is the way to go. I’d argue not, especially since the problem is effectively solved in a supported way with taps.

Again, if I’m wrong in any of my statements, please do correct me. Also, to be clear, this is just an opinion: I’m not in charge of that particular policy.

ilovezfs:
>Please correct me if I’m wrong, but Linuxbrew is not an official part of Homebrew

The science tap tries to maintain full compatibility on both OSs to the extent possible so binaries-that-would-go-in-the-science-tap-if-they-weren't-binaries are in an awkward spot right now.

PR #2061, opened 2015-04-07
RazerS 3.4.0 (new formula)

tseemann:
RazerS is a read mapping program with adjustable sensitivity based on counting q-grams. In this work, we propose the successor RazerS 3, which now supports shared-memory parallelism, an additional seed-based filter with adjustable sensitivity, a much faster, banded version of the Myers’ bit-vector algorithm for verification, memory-saving measures and support for the SAM output format. This leads to a much improved performance for mapping reads, in particular, long reads with many errors. We extensively compare RazerS 3 with other popular read mappers and show that its results are often superior to them in terms of sensitivity while exhibiting practical and often competitive run times. In addition, RazerS 3 works without a pre-computed index.


tseemann:
@BrewTestBot test this please

razers.rb
```diff
@@ -0,0 +1,29 @@
...
+  version "3.4.0"
+
+  if OS.mac?
+    url "http://packages.seqan.de/razers3/razers3-3.4.0-Mac-x86_64.tar.bz2"
+    sha256 "487a9a4b9fc006ee32f7c50fa64d78fc848a8bb36ab70ab794b8c207fde310d4"
+  elsif OS.linux?
+    url "http://packages.seqan.de/razers3/razers3-3.4.0-Linux-x86_64.tar.bz2"
+    sha256 "e7c10341a92ec77e82556969a410e97f6d5349d605157f33b0f6599afe25a871"
+  else
+    ohoe "Unsupported operating system"
```
dpo: This part isn't necessary.

tseemann: @dpo You mean the ```else ohoe "Unsupported..."``` ?

iml:
Can this not be built from source?

razers.rb
dpo: Yes.

tseemann:
It possibly could be at a future point in time, but it depends on someone getting the SeqAn library compiled and then getting it all working with BOOST too I think.

iml:
I'll try updating seqan.

tseemann:
@iml Good luck! https://github.com/Homebrew/homebrew-science/pull/2121

iml:
Yea would it be appropriate to have an all-inclusive seqan that builds RazerS and everything else? I haven't tested beetl or kat against seqan2 so I expect a formula with the old version will be needed for compatibility.

tseemann:

tseemann:
I am closing this, as it is binary based. Install via `tseemann/bioinformatics-linux/razers` if you want it.

PR #2074, opened 2015-04-08
fwdpp 0.2.9

vancleve:
New "make check" also enabled and deleted old "test do" section.


vancleve:
I looked at the failures and I can't quite get why they happened. Three of them are due to a missing "szip". But fwdpp doesn't require szip; it only requires zlib.

The remaining error on mountain lion appears to be configure complaining about missing the dependency libsequence, but libsequence is already listed as a dependency.

dpo:
It's good to have a `test do` section as it runs a test in a real situation, after the package was installed. Why was it removed?

vancleve:
I thought the 'make check' ran tests, but it doesn't, so I readded the unit tests in a `test do` section

vancleve:
It looks like the tests are failing to execute, but there is no information why. They execute fine on my Mac OS 10.10.3 systems.

Any thoughts?

dpo:
It's picking up the file named "`.`" and is trying to execute it.

There are also a few audit issues (check with `brew audit ./fwdpp.rb --strict`) and your PR has picked up some extraneous commits.

vancleve:
I've cleaned up the audit issues on a new clone of the main repo. Should I do another pull request with this new repo and delete the current pull request?

As for the execution issues, the formula specifically looks in the "unit" folder for regular files that are executable and thus should avoid trying to execute any directories including `"."`. Looking at the Stacktrace, it appears the build is beginning to run the first unit test, "crossoverTest", and fails there.

dpo:
A new PR would be best, but let's wait until the build passes.

When I built it locally, I got this:
````
$ brew test -v ./fwdpp.rb
Testing fwdpp
==> .
sh: line 0: .: filename argument required
.: usage: . filename [arguments]
Error: fwdpp: failed
Failed executing: .
/usr/local/Library/Homebrew/formula.rb:739:in `block in system'
/usr/local/Library/Homebrew/formula.rb:696:in `open'
/usr/local/Library/Homebrew/formula.rb:696:in `system'
/Users/dpo/dev/homebrew-science/fwdpp.rb:28:in `block (2 levels) in <class:Fwdpp>'
/Users/dpo/dev/homebrew-science/fwdpp.rb:28:in `each'
/Users/dpo/dev/homebrew-science/fwdpp.rb:28:in `block in <class:Fwdpp>'
/usr/local/Library/Homebrew/formula.rb:638:in `block in run_test'
/usr/local/Library/Homebrew/formula.rb:634:in `run_test'
/usr/local/Library/Homebrew/cmd/test.rb:38:in `block (2 levels) in test'
/usr/local/Library/Homebrew/cmd/test.rb:37:in `block in test'
/usr/local/Library/Homebrew/cmd/test.rb:15:in `each'
/usr/local/Library/Homebrew/cmd/test.rb:15:in `test'
/usr/local/Library/brew.rb:135:in `<main>'
````
You're right that it's not the same failure as on the bot.

**Comment by dpo** [comment]
When I run the test manually, I get
````
$ /usr/local/Cellar/fwdpp/0.2.9/share/unit/crossoverTest
Running 2 test cases...

*** No errors detected
````
so somehow `crossoverTest` returns a nonzero exit code.

File: fwdpp.rb [hunk 1]
```diff
@@ -13,13 +17,18 @@ class Fwdpp < Formula
...
+    share.install 'examples' # install examples
+    share.install 'unit'     # install unit tests
   end
 
   test do
-    # run one of the example programs (https://github.com/molpopgen/fwdpp#diploid)
-    system "#{share}/examples/diploid 1000 10 10 1000 50 1 272"
+    # run unit tests compiled with 'make check'
+    if build.with? "check"
+      Dir.chdir("#{share}/unit")
```
dpo: This works:
````
Dir["#{share}/unit/*"].each { |f| system f if ((File.file? f) && (File.executable? f)) }
````

iml:
Remember to checkout a feature branch if you decide to open a new PR, rather than the master branch.

dpo:
````
configure: error: Sequence/SimDataIO.hpp not found. libsequence (www.molpopgen.org) >= 1.7.8 required.
````
Is there anything special about `libsequence` on 10.8?

vancleve:
Not that I know, though I'd have to ask the developer to be sure.

vancleve:
It's weird though that 10.8 is failing simply to find the header files from `libsequence`. This initially happened when I submitted the pull request too.

iml:
It's possible libsequence is broken on mountain lion.

vancleve:
Right, I suppose so. It's weird since the libsequence must have passed checks. In any case, is there a way to specify that the formula can only be built for > 10.8?

PR #2077, opened 2015-04-09
r: cairographics needs pkg-config

### LINKED ISSUE #1890: R installation does not include the cairo graphics devices

iml:
Fixes #1890.


iml:
Bot keeps making `revision 4` bottles..

iml:
The cairo capabilities seem to be enabled on what it bottled, fortunately. I'll try this on a new branch, perhaps @BrewTestBot has an older branch of the same name.

tdsmith:
The bottle revision is because 14d4ba2 appears to be a 3.1.3 R formula with bottle revision 3 (even though those bottles point to the old 3.1.2 bottles; they were updated to the 3.1.3 bottles and the revision was wiped in the next commit).

tdsmith:
(but the bottle-revision logic is "look through history for the highest bottle revision that has ever appeared in a formula for this version of the package and add 1," so it detects this).

iml:
I see, that is unfortunate. So really the bottle block should be cleared along with with any version upgrades.

jacknagel:
I think that we could possibly fix this, at least for the most common cases. If we detect the version has increased between two successive revisions, then we can throw out the bottle revision.

iml:
@jacknagel Could we just throw out (or not consider for max) the last non-zero revision for a given pkg_version? A new pkg_version seems to imply resetting the bottle revision.
```ruby
{#<PkgVersion:0x007f8d2c09b620
  @revision=0,
  @version=#<Version::FromURL:0x007f8d2c09b800 @version="3.1.3">>=>
  [0, 0, 3],
 #<PkgVersion:0x007f8d2ab295d0
  @revision=1,
  @version=#<Version::FromURL:0x007f8d2ab29670 @version="3.1.2">>=>
  [3, 3, 3, 2, 2, 2, 1, 1, 1, 0, 0, 0, 0],
 #<PkgVersion:0x007f8d297a9550
  @revision=0,
  @version=#<Version::FromURL:0x007f8d297941a0 @version="3.1.2">>=>
  [0, 0, 0]
}

PR #2079, opened 2015-04-09
cdhit: enable openmp support when compiling with gcc

ntkme:


cd-hit.rb
```diff
@@ -7,7 +7,10 @@ class CdHit < Formula
   sha1 '744be987a963e368ad46efa59227ea313c35ef5d'
 
   def install
-    system "make"
+    args = []
+    args << "openmp=yes" if ENV.compiler != :clang
```
dpo: Please make this a one liner: `args = (ENV.compiler == :clang) ? [] : ["openmp=yes"]`

PR #2085, opened 2015-04-10
igv 2.3.48

ntkme:
This PR updates IGV and igvtools to 2.3.48.


ntkme:
I'm having the same error as the bot has when running test using ssh and X window forwarding. Running test command locally works fine.

iml:
That's expected for IGV (and most GUI apps).

PR #2087, opened 2015-04-10
Add pkg-config file for openni2

hgaiser:


openni2.rb
```diff
@@ -39,6 +39,21 @@ def install
...
+includedir=${prefix}/include/ni2
+
+Name: OpenNI2
+Description: A general purpose driver for all OpenNI cameras.
+Version: 2.2.0.33
+Cflags: -I${includedir}
+Libs: -L${libdir} -lOpenNI2 -L${libdir}/OpenNI2/Drivers -lOniFile -lPS1080
+"
+    File.open('openni2.pc', 'w') {|f| f.write(pkg_config_file)}
+    (lib + 'pkgconfig').install 'openni2.pc'
```
iml: Use a here document:
```ruby
(lib/"pkgconfig/openni2.pc").write <<-EOS.undent
  prefix=#{prefix}
  exec_prefix=${prefix}
  libdir=${exec_prefix}/lib/ni2
  includedir=${prefix}/include/ni2

  Name: OpenNI2
  Description: A general purpose driver for all OpenNI cameras.
  Version: 2.2.0.33
  Cflags: -I${includedir}
  Libs: -L${libdir} -lOpenNI2 -L${libdir}/OpenNI2/Drivers -lOniFile -lPS1080
EOS

iml:
This should really be submitted upstream, though, there's no reason for Homebrew to maintain to maintain pkg-config files. That should be distributed by the software.

PR #2092, opened 2015-04-11
vcflib 20140412 (new formula)

tseemann:
Still lobbying Erik to tag a release: https://github.com/ekg/vcflib/issues/76

vcflib provides methods to manipulate and interpret sequence variation as it can be described by VCF. It is both an API for parsing and operating on records of genomic variation as it can be described by the VCF format, and a collection of command-line utilities for executing complex manipulations on VCF files.


tseemann:
I have bad feeling you won't accept HEAD-only formulae?

tseemann:
I've learnt I can tag a commit/revision in Github and give it my own version. Hooray!

vcflib.rb
```diff
@@ -0,0 +1,35 @@
...
+    %w[bed2region vcf_strip_extra_headers vcfbiallelic vcfcomplex
+       vcfindelproximity vcfindels vcfmultiallelic vcfnobiallelicsnps vcfnoindels
+       vcfnosnps vcfqualfilter vcfremovenonATGC vcfsnps vcfvarstats].each do |file|
+      inreplace "bin/#{file}", "/usr/bin/perl", "/usr/bin/env perl"
+    end
+
+    %w[vcf2bed.py vcf2sqlite.py vcfclearid vcfclearinfo
+       vcffirstheader vcfnulldotslashdot].each do |file|
+      inreplace "bin/#{file}", "/usr/bin/python", "/usr/bin/env python"
+    end
```
iml: Since you already submitted a patch for this could you use that instead? It's something upstream should fix anyway and makes it easier to track progress.
```
# Fix /usr/bin hashbang: https://github.com/ekg/vcflib/pull/84
patch do
  url "https://patch-diff.githubusercontent.com/raw/ekg/vcflib/pull/84.diff"
  sha256 ""
end

**Comment by tseemann** [comment]
@lml @sjackman I've implemented the "patch" method suggested above. hopefully good to go!

File: vcflib.rb [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+  homepage "https://github.com/ekg/vcflib"
+  # tag "bioinformatics"
+  url "https://github.com/ekg/vcflib.git",
+    :revision => "b1e9b31d2d95f957f86cdfd1e7c9ec25b4950ee8"
+  version "20140412"
+
+  # Fix /usr/bin hashbang: https://github.com/ekg/vcflib/pull/84
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/ekg/vcflib/pull/84.diff"
+    sha256 ""
```
iml: Fill this in, it should give a warning and the values with `brew install` or `brew fetch`

vcflib.rb
```diff
@@ -0,0 +1,30 @@
+
```
sjackman: Please delete this blank line.

vcflib.rb
sjackman: Please add the SHA.

vcflib.rb
```diff
@@ -0,0 +1,30 @@
...
+  # Fix /usr/bin hashbang: https://github.com/ekg/vcflib/pull/84
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/ekg/vcflib/pull/84.diff"
+    sha256 ""
+  end
+
+  def install
+    if OS.mac?
+      inreplace "tabixpp/Makefile", "SUBDIRS=.", "SUBDIRS=.\nLOBJS=tabix.o"
+      inreplace "smithwaterman/Makefile", "LDFLAGS=-Wl,-s", "LDFLAGS=-Wl"
```
sjackman: This line doesn't look right. It should probably be `"LDFLAGS=-Wl,-s", "LDFLAGS="`.
Using the `ld -s` option is a rather obscure way of stripping the output file. Running `strip` or even better `install -s` is much nicer and more portable. Would you like to open a PR upstream?

vcflib.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  # Fix /usr/bin hashbang: https://github.com/ekg/vcflib/pull/84
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/ekg/vcflib/pull/84.diff"
+    sha256 ""
+  end
+
+  def install
+    if OS.mac?
+      inreplace "tabixpp/Makefile", "SUBDIRS=.", "SUBDIRS=.\nLOBJS=tabix.o"
```
sjackman: What error message does this patch fix? Please include the error message in the formula. i.e.
`# Fix error: undefined reference to symbol 'frobnicate'`

vcflib.rb
sjackman: Is this patch even necessary? It only results in a warning: `ld: warning: option -s is obsolete and being ignored`, or is the goal to avoid stripping?

vcflib.rb
tseemann: Done.

vcflib.rb
tseemann: Done.

vcflib.rb
tseemann: Done.

vcflib.rb
tseemann: Wierd error that only that solves

ld: internal error: atom not found in symbolIndex(__ZNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEE6assignINS_16reverse_iteratorINS_11__wrap_iterIPcEEEEEENS_9enable_ifIXsr21__is_forward_iteratorIT_EE5valueERS5_E4typeESD_SD_) for architecture x86_64

tseemann:
I'm not sure what all this means:

http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/1726/

tdsmith:
Mashed the rebuild button; we'll give it another spin.

vcflib.rb
tdsmith: Agree that Shaun's fix looks more correct (i.e. dropping `-Wl`, which indicates that the following option is to be given to the linker)

tseemann:
@tdsmith I think that fixed it!

vcflib.rb
```diff
@@ -0,0 +1,31 @@
...
+    url "https://patch-diff.githubusercontent.com/raw/ekg/vcflib/pull/84.diff"
+    sha256 "e4ad6d11b864c47834be5eed67edeb6d6bb52ed83bbb0f93683def3fa1735458"
+  end
+
+  def install
+    if OS.mac?
+      # FIX => missing makefile var: https://github.com/ekg/tabixpp/issues/5
+      inreplace "tabixpp/Makefile", "SUBDIRS=.", "SUBDIRS=.\nLOBJS=tabix.o"
+      # FIX => ld: internal error: atom not found in symbolIndex(<snip>) for architecture x86_64
+      inreplace "smithwaterman/Makefile", "LDFLAGS=-Wl,-s", "LDFLAGS=-Wl"
```
sjackman: This line isn't right. The `-Wl` gets misinterpreted as a warning flag instead of a linker flag if it's not given its mandatory argument. Try instead `inreplace "smithwaterman/Makefile", "LDFLAGS=-Wl,-s", "LDFLAGS="`

tseemann: Thanks for the tip. Trying now.

vcflib.rb
```diff
@@ -0,0 +1,33 @@
+class Vcflib < Formula
+  desc "Command-line tools for manipulating VCF files"
+  homepage "https://github.com/ekg/vcflib"
+  head "https://github.com/ekg/vcflib.git"
+  # tag "bioinformatics"
+  url "https://github.com/ekg/vcflib.git",
+    :revision => "b1e9b31d2d95f957f86cdfd1e7c9ec25b4950ee8"
```
sjackman: I've asked the upstream maintainers to tag a stable release.
https://github.com/ekg/vcflib/issues/95

sjackman:
Thanks again for your contribution, Torsten.

PR #2093, opened 2015-04-12
Update openbr.rb

shashanksingh:
Change version from 0.4.1 to 0.5 with appropriate download url.

### DIFF
--- a/openbr.rb
+++ b/openbr.rb
@@ -2,9 +2,8 @@ require "formula"
 
 class Openbr < Formula
   homepage "http://www.openbiometrics.org/"
-  url "https://github.com/biometrics/openbr/releases/download/v0.4.1/openbr0.4.1_withModels.tar.gz"
-  version "0.4.1"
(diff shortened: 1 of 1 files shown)


openbr.rb
```diff
@@ -2,9 +2,8 @@
 
 class Openbr < Formula
   homepage "http://www.openbiometrics.org/"
-  url "https://github.com/biometrics/openbr/releases/download/v0.4.1/openbr0.4.1_withModels.tar.gz"
-  version "0.4.1"
-  sha1 "1cf3b9b6fad1c717d8b0256b3a22d138510572e9"
+  url "https://github.com/biometrics/openbr/releases/download/v0.5.0/OpenBR-0.5.0-osx.zip"
+  version "0.5.0"
```
iml: Please update the checksum. Also, we need the source archive not precompiled binaries.

shashanksingh: Sure , thanks will do.

dpo: Please also use the `sha256` instead of `sha1`.

dpo:
@shashanksingh

PR #2095, opened 2015-04-13
Update armadillo to 5.000.0

zeeMonkeez:


iml:
The test fails because the format of `arma_version` changed to 5.0.0 for some reason.

dpo:
Thanks. Could you please squash your commits?

iml:
Sorry for not being clear, I think you can leave off the version and ignore the test. The failure is really a false-positive because the arma_versions are defined as integers not strings. Thanks!

zeeMonkeez:
Seems like we would not want the test to fail? I did ask Conrad Sand, author of armadillo, about the inconsistent version strings, but haven't received a response yet.
Guess this problem will go away once three-figure minor versions are reached again ...

zeeMonkeez:
ugh, I don't understand the failure message.

saketkc:
@BrewTestBot test this please

dpo:
SEems like a bot issue. I'm having the same in #2113.

saketkc:
It doesn't seem to be picking up `test this please` also, right?

dpo:
@BrewTestBot test this please

PR #2098, opened 2015-04-13
libsequence: build shared libs, make check

iml:


iml:
`cstdint` needs to be prefixed with `td` in order to build with `stdlibcxx`. Enforcing `libcxx` is not an appropriate cudgel to get around this because linking errors will crop up (see #2074).

iml:

PR #2102, opened 2015-04-14
graph-tool: bump version to 2.2.40

count0:

PR #2108, opened 2015-04-16
graph-tool: Bump version to 2.2.40

count0:


dpo:
Should #2102 be closed?

count0:
Yes, I just closed it.

graph-tool.rb
```diff
@@ -1,8 +1,7 @@
 class GraphTool < Formula
   homepage "http://graph-tool.skewed.de/"
-  url "http://downloads.skewed.de/graph-tool/graph-tool-2.2.35.tar.bz2"
-  sha1 "f75a31dec45843beff18eb6b5ce8eda5a0645277"
-  revision 1
+  url "http://downloads.skewed.de/graph-tool/graph-tool-2.2.40.tar.bz2"
+  sha1 "6713ab4e55c6b84620479253d99345a010c1d206"
```
dpo: Could you please use the `sha256`?

dpo:
I'm getting a test error on Mavericks:
````
$ brew test ./graph-tool.rb
Testing graph-tool
==> python test.py
  File "test.py", line 1, in <module>
    import graph_tool.all as gt
  File "/usr/local/lib/python2.7/site-packages/graph_tool/all.py", line 35, in <module>
    from graph_tool.draw import *
AttributeError: 'module' object has no attribute 'draw_hierarchy'
Error: graph-tool: failed
````

**Comment by count0** [comment]
This is an upstream bug when pycairo and Gtk are not available (they should be optional, but are highly recommended). This is already fixed in git, and will be in the next version.

**Comment by tdsmith** [comment]
If that commit applies cleanly on top of the release we can carry it as a patch. If you want cairo and pygtk to be opt-out instead of opt-in, making them :recommended deps instead of :optional will achieve that.

**Comment by count0** [comment]
I've just done that. The gtk+ visualization part of graph-tool is very important, and I think most users will be expecting it anyway.
I could also include a patch, but I'd rather leave that for the next release. If you insist on the patch, you would have to tell me how to include it in the formula (I'm totally new to homebrew).

**Comment by tdsmith** [comment]
We'd add a block that looks like

```ruby
stable do
  # fix import if gtk+3 or cairo isn't present
  # https://github.com/Homebrew/homebrew-science/pull/2108#issuecomment-94138693
  patch do
    url "https://github.com/count0/graph-tool/commit/c8f99e7c.diff"
    sha256 "ecebc6b311ea438506e32a4ea0f74964e661118a562b7bc71563dfb9c3cf4407"
  end
end
```

to the Formula class body, near the top of the file.

Thanks for sending us changes; I appreciate it.

count0:
I hope that does it.

dpo:
I'm constantly getting
````
$ brew test -v ./graph-tool.rb
Testing graph-tool
==> python test.py
Dynamic session lookup supported but failed: launchd did not provide a socket path, verify that org.freedesktop.dbus-session.plist is loaded!
````
Sorry I'm not sure who outputs this error message.

dpo:
Ah, sorry. I had missed the caveats from `dbus`. It's all good now.

PR #2111, opened 2015-04-16
nextflow 0.13.2: Depends on Java 1.7+

sjackman:
Change the Java dependency from 1.7 to 1.7+


sjackman:
I'm off on vacation for a week. Have fun in my absence. :sunglasses:

dpo:
Enjoy the vacation!

iml:
Have fun!

PR #2116, opened 2015-04-17
r: avoid deprecated bundled libs

tdsmith:
per changelog, R's internal zlib, bzip2, xz and PCRE are deprecated in
favor of external libraries.


iml:
Is being more explicit with configure worthwhile?
```
  --with-libpng           use libpng library (if available) [yes]
  --with-jpeglib          use jpegfplib library (if available) [yes]
  --with-libtiff          use libtiff library (if available) [yes]
  --with-system-zlib      use system zlib library (if available) [yes]
  --with-system-bzlib     use system bzlib library (if available) [yes]
  --with-system-pcre      use system PCRE library (if available) [yes]
  --with-system-tre       use system tre library (if available) [no]
  --with-system-xz        use system xz (lzma) library (if available) [yes]

tdsmith:
They all default on so I don't think it matters; I ran the build locally to see what configure did and it picked up the right things.

iml:
:shipit:

PR #2126, opened 2015-04-19
ABINIT 7.10.4

mtorrent:
Push Abinit to new version 7.10.4:
- The patch correcting netCDF bugs isn't needed anymore
- Now that a formula is available for ETSF_IO, it can be used for abinit


abinit.rb
```diff
@@ -60,7 +55,13 @@ def install
...
     end
 
-    if build.with? "netcdf"
+    if build.with? "etsf_io"
+      trio_flavor = "netcdf+etsf_io"
+      args << "--with-etsf-io-incs=-I#{Formula["etsf_io"].opt_include}"
+      args << "--with-etsf-io-libs=-L#{Formula["etsf_io"].opt_lib} -letsf_io_low_level -letsf_io_utils -letsf_io"
+      args << "--with-netcdf-incs=-I#{Formula["netcdf"].opt_include}"
+      args << "--with-netcdf-libs=-L#{Formula["netcdf"].opt_lib} -lnetcdff -lnetcdf"
+    elsif build.with? "netcdf"
```
iml: This won't work if netcdf is disabled by default with `:optional`, as above. Perhaps netcdf should be recommended and etsf_io optional? Or create some option(s) for finer control.

abinit.rb
```diff
@@ -60,7 +55,13 @@ def install
       args << "--with-linalg-libs=-L#{Formula["veclibfort"].opt_lib} -lveclibfort"
     end
 
-    if build.with? "netcdf"
+    if build.with? "etsf_io"
+      trio_flavor = "netcdf+etsf_io"
+      args << "--with-etsf-io-incs=-I#{Formula["etsf_io"].opt_include}"
+      args << "--with-etsf-io-libs=-L#{Formula["etsf_io"].opt_lib} -letsf_io_low_level -letsf_io_utils -letsf_io"
+      args << "--with-netcdf-incs=-I#{Formula["netcdf"].opt_include}"
+      args << "--with-netcdf-libs=-L#{Formula["netcdf"].opt_lib} -lnetcdff -lnetcdf"
```
iml: This won't work if netcdf is disabled by default with `:optional`, as above. Perhaps netcdf should be recommended and etsf_io optional? Or create some option(s) for finer control.

mtorrent:
netcdf is a mandatory dependence of etsf_io; I think this should work.
3 possibilities:
- with-etsf-io : netcdf is already installed
- without-etsf-io, with-netcdf : install netcdf if necessary
- without-etsf-io, without-netcdf: nothing done

iml:
I would make `etsf_io` and `netcdf` both `:recommended` and put the args for each under ifs. If you want to robustify the etsf_io conditional you can add something like 
```
if build.with? "etsf_io"
  fail "Building with etsf_io support requires netcdf" if build.without? "netcdf"

mtorrent:
OK, done!

iml:
Pulled 30947713013. Thanks!

PR #2133, opened 2015-04-20
FastML 3.1 (new formula)

tseemann:
The FastML server is a bioinformatics tool for the reconstruction of ancestral sequences based on the phylogenetic relations between homologous sequences. The server runs several algorithms that reconstruct the ancestral sequences with emphasis on an accurate reconstruction of both indels and characters. For character reconstruction the previously described FastML algorithms [1, 2] are used to efficiently infer the most likely ancestral sequences for each internal node of the tree. Both joint and the marginal reconstructions are provided. For indels reconstruction the sequences are first coded according to the indel events detected within the multiple sequence alignment (MSA) [3] and then a state-of-the-art likelihood model is used to reconstruct ancestral indels states [4, 5]. The server results are the most probable sequences, together with posterior probabilities for each character and indel at each sequence position for each internal node of the tree. FastML is generic and is applicable for any type of molecular sequences (nucleotide, protein, or codon sequences).


tseemann:
All tests passed, looks the @BrewTestBot bug again.

tseemann:
@BrewTestBot test this please

tseemann:
I'll try and find time to see why this fails on OSX but ok on Linux...

```
ranlib libgainLoss.a
g++ -L../../libs/phylogeny  gainLoss.o libgainLoss.a ../../libs/phylogeny/libEvolTree.a   -o gainLoss
cp gainLoss ../../bin/gainLoss
make: *** [programs.install] Error 2
```

dpo:
This is the error:
````
In file included from /usr/include/c++/4.2.1/bits/os_defines.h:61:
/usr/include/unistd.h:538:6: error: conflicting types for 'getopt'
int      getopt(int, char * const [], const char *) __DARWIN_ALIAS(getopt);
         ^
../../libs/phylogeny/getopt.h:147:12: note: previous declaration is here
extern int getopt ();
           ^
`````
It's only occurring on Mountain Lion.

**Comment by tseemann** [comment]
@dpo - what is the correct way to gracefully reject <= :mountain_lion ?

**Comment by iml** [comment]
`depends_on :macos => :mavericks`, although that might affect Linuxbrew @sjackman. Personally I think  `depends_on :macos => :x` should be ignored on Linuxbrew since it's a useful pattern on Mac.

**Comment by sjackman** [comment]
Is it not ignored on Linux? If not, pull request would be happily accepted.

File: fastml.rb [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+
+  url "http://fastml.tau.ac.il/source/FastML.v3.1.tgz"
+  sha256 "16c8631a4186f434f81f5b7e8c1147660e79b025b3c93a6db9889c30477bd5f4"
+ 
+  depends_on :macos => :mavericks   # won't build on [Mountain] Lion
+
+  def install
+    mkdir "bin"
+    system "make", "install"
+    bin.install Dir["bin/*"]
```
tdsmith: Why is this necessary?

fastml.rb
```diff
@@ -0,0 +1,23 @@
...
+  # doi "10.1093/bioinformatics/18.8.1116"
+
+  url "http://fastml.tau.ac.il/source/FastML.v3.1.tgz"
+  sha256 "16c8631a4186f434f81f5b7e8c1147660e79b025b3c93a6db9889c30477bd5f4"
+ 
+  depends_on :macos => :mavericks   # won't build on [Mountain] Lion
+
+  def install
+    mkdir "bin"
+    system "make", "install"
```
tdsmith: How does it know where to install?

tdsmith:
Oh, `make install` just installs into its own build directory? That's kinda weird and upstream might fix it one day so if we can do something that looks more idiomatic I think that'd be good.

fastml.rb
tseemann: A hard-coded ```binDir=../../bin``` two levels down.

tseemann: @tdsmith You haven't been around a lot of bioinformatics software have you? ;-)  

This sort of Makefile behaviour is typical. Because most users don't have root access.

tdsmith: Eh, well spotted; I spend way more time with pipettes than sequences.

PR #2143, opened 2015-04-23
fasttree: refactor, simplify fails_with

iml:
LLVM has been added to the :openmp collection.


tseemann:
@lml Hooray! thanks :)

fasttree.rb
```diff
@@ -14,39 +14,30 @@ class Fasttree < Formula
...
 
-  # See discussion (https://github.com/Homebrew/homebrew-science/pull/2134#issuecomment-94607619)
-  fails_with :llvm
-
-  fails_with :clang do
-    build 425
-    cause "segmentation fault when running Fasttree"
-    # See also discussion to use -DNO_SSE (https://github.com/Homebrew/homebrew-science/pull/96)
-  end
+  needs :openmp
```
dpo: @tseemann @sjackman Is this a bug? Shouldn't it have `if build.with? "openmp"`, or does `fasttree` only build with gcc (in which case a `fails_with` would be more appropriate)?

PR #2145, opened 2015-04-24
MLPack: version 1.0.12.

dpo:


saketkc:
Thanks!
Sorry, I am busy with my screening exams.

dpo:
@iml Is this what you were recommending?

iml:
Yeah, but just `ENV.cxx11` since it's non-optional, and `needs` only affects compiler selection.

dpo:
Thanks @iml! We finally have bottles for Mountain Lion! There's still a compilation error in the test on ML that I'm trying to resolve.

dpo:
Thanks @iml and @saketkc !

PR #2147, opened 2015-04-24
Flux Simulator 1.2.1 (new formula)

roryk:
Flux Simulator models RNA-seq experiments in silico.
    It can generate reads with various experimental biases
    profiles and simulate gene expression.


flux-simulator.rb
```diff
@@ -0,0 +1,17 @@
+require 'formula'
```
dpo: This construct is now obsolete.

flux-simulator.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class FluxSimulator < Formula
+  homepage "http://sammeth.net/confluence/display/SIM/Home"
+  version "1.2.1"
```
dpo: `version` is redundant with the version from the `url`.

flux-simulator.rb
```diff
@@ -0,0 +1,17 @@
...
+  url "http://artifactory.sammeth.net/artifactory/barna/barna/barna.simulator/1.2.1/flux-simulator-1.2.1.tgz"
+  sha1 "49a14b34e8700bc225175c6ebc87a343a3c41355"
```
dpo: We now use the `sha256` instead. You can find it using `shasum -a 256`.

roryk:
Thanks @dpo I made those changes.

dpo:
````
C: 13: 12: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

1 file inspected, 1 offense detected
Error: 3 problems in 1 formulae

==> audit problems
flux-simulator:
 * Stable: version 1.2.1 is redundant with version scanned from URL
 * Use `system 'flux-simulator", "-h'` instead of `system 'flux-simulator -h'` 
 * JARs were installed to "/usr/local/Cellar/flux-simulator/1.2.1/lib"
Installing JARs to "lib" can cause conflicts between packages.
For Java software, it is typically better for the formula to
install to "libexec" and then symlink or wrap binaries into "bin".
See "activemq", "jruby", etc. for examples.
````

**Comment by roryk** [comment]
Thanks, swapped to double quotes.

**Comment by dpo** [comment]
What of the `jar` files?

**Comment by roryk** [comment]
Fixed.

**Comment by dpo** [comment]
@BrewTestBot test this please

File: flux-simulator.rb [hunk 4]
```diff
@@ -0,0 +1,16 @@
+class FluxSimulator < Formula
+  homepage "http://sammeth.net/confluence/display/SIM/Home"
+  version "1.2.1"
+  url "http://artifactory.sammeth.net/artifactory/barna/barna/barna.simulator/#{version}/flux-simulator-#{version}.tgz"
+  sha256 "9e4da0b4e1c21bddd9f8d0cca67bd6a55de43c08db40833ac2fbb531f3a8cff1"
+
+  def install
+    inreplace "bin/flux-simulator", 'libdir="$dir/../lib"', 'libdir="$dir/../libexec"'
+    libexec.install Dir["lib/*.jar"]
+    bin.install Dir["bin/flux-simulator"]
```
dpo: This should just be `"bin/flux-simulator"`.

dpo:
The bot gives an audit warning about the `version` being redundant, but I think what you did makes sense.

roryk:
Thank you, sorry for all the back and forth and not conforming to the style guides. I made the change.

dpo:
Thanks for the update. Actually, the test fails:
````
==> brew test --verbose Homebrew/science/flux-simulator			 FAILED
Testing flux-simulator
==> flux-simulator -h
Flux-Simulator v1.2.1 (Flux Library: 1.22)

[ERROR] 
[ERROR] No parameter file specified !
[ERROR] 

Error: flux-simulator: failed
Failed executing: flux-simulator -h
````
I'm not familiar with this software. Could you update the test?

roryk:
Sorry again, it looks like it returns a non-zero code for --help and --version. -o which prints an example parameter file to start from returns 0 though. If that works it means everything should be functional.

dpo:
Thanks!

roryk:
Thanks @dpo for your patience.

PR #2149, opened 2015-04-24
FastUniq 1.1 (new formula)

tseemann:
We present FastUniq as a fast de novo tool for removal of duplicates in paired short reads. FastUniq identifies duplicates by comparing sequences between read pairs and does not require complete genome sequences as prerequisites. FastUniq is capable of simultaneously handling reads with different lengths and results in highly efficient running time, which increases linearly at an average speed of 87 million reads per 10 minutes.


fastuniq.rb
```diff
@@ -0,0 +1,20 @@
+class Fastuniq < Formula
+  homepage "http://sourceforge.net/projects/fastuniq/"
+  # tag "bioinformatics"
+  # doi "10.1371/journal.pone.0052249"
+
+  url "https://downloads.sourceforge.net/project/fastuniq/FastUniq-1.1.tar.gz"
+  sha256 "9ebf251566d097226393fb5aa9db30a827e60c7a4bd9f6e06022b4af4cee0eae"
+
+  def install
+    inreplace "source/fastq_uniq.c", "FSATQ", "FASTQ"  # fix typo
```
dpo: I'd suggest adding this as a small patch, reporting it upstream, and including a link to the report here as a comment.

tseemann: I'm regretting fixing their typo now.... they last touched their project  3 years ago!
```Last Update: 2012-05-13```

dpo: What are the consequences of this typo?

tseemann: It's just printed in the --help synposis. No logic change.

dpo: It's a legit change. Thanks!

PR #2151, opened 2015-04-24
Megahit 0.2.1 (new version)

tseemann:
New --prefix option and --cpu-only is default.


megahit.rb
```diff
@@ -3,9 +3,9 @@ class Megahit < Formula
   # doi "arXiv:1409.7208"
   # tag "bioinformatics"
 
-  url "https://github.com/voutcn/megahit/archive/v0.2.0-a.tar.gz"
-  version "0.2.0-a"
-  sha256 "5b30a163027b248bf2d96c766c3bcf1f4fe50466e368a3844b3c4e0a2491d8c4"
+  url "https://github.com/voutcn/megahit/archive/v0.2.1.tar.gz"
+  version "0.2.1"
```
dpo: `version` is redundant and is causing an audit warning.

tseemann: Thank you @dpo. It was needed for the old "-a" version but not anymore! I've fixed it. I appreciate the time you took to check the build failure for me.

PR #2153, opened 2015-04-24
tbl2asn 24.3

ntkme:
This PR updates it to the latest version and fixes checksum mismatch during the installation.


tseemann:
Thanks for this up @ntkme !
I wonder what they changed in this minor version.

ntkme:
Not sure, but you can see that the link are still same as before, which causes checksum mismatch.

tseemann:
@ntkme Yes, NCBI are very annoying with this software. They just release new versions without notice and with the same filename :-1:

PR #2156, opened 2015-04-24
PEAT 1.0b1 (new formula) - need help with boost problem

tseemann:
PEAT, a fast and highly accurate adapter-trimming algorithm, designed specifically for paired-end sequencing. PEAT requires no a priori adapter sequence and significantly outperforms other widely used adapter trimmers in both simulated and real life paired-end sequencing libraries. Moreover, in order to process single-end sequencing data, we also provide single-end trimmer in PEAT.


tseemann:
I can build it manually on the command line but not via Brew. I am confused!

```
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
thread.cpp:(.text+0x1bd): undefined reference to `pthread_key_create'
thread.cpp:(.text+0x233): undefined reference to `pthread_create'
/bio/linuxbrew/lib/libboost_thread-mt.a(thread.o): In function `boost::thread::start_thread_noexcept(boost::thread_attributes const&)':
thread.cpp:(.text+0x2b2): undefined reference to `pthread_create'
/bio/linuxbrew/lib/libboost_thread-mt.a(thread.o): In function `boost::thread::detach()':
hyphy.rb:    The help has been installed to #{HOMEBREW_PREFIX}/share/hyphy.
thread.cpp:(.text+0x1bd): undefined reference to `pthread_key_create'
thread.cpp:(.text+0x16d): undefined reference to `pthread_key_create'
thread.cpp:(.text+0x19aa): undefined reference to `pthread_join'
/bio/linuxbrew/lib/libboost_thread-mt.a(thread.o): In function `boost::detail::get_current_thread_data()':
thread.cpp:(.text+0x146): undefined reference to `pthread_getspecific'
/bio/linuxbrew/lib/libboost_thread-mt.a(thread.o): In function `boost::detail::set_current_thread_data(boost::detail::thread_data_base*)':
```

dpo:
It looks like it built on Mavericks and Yosemite; there are just some minor audit warnings to fix.

The failure is on Mountain Lion. Does this need to build in c++11 mode? Try adding `ENV.cxx11` at the top of the `install` method.

iml:
Might need to also build boost with c++11 on mountain lion, assuming that's the issue and not something else.

peat.rb
```diff
@@ -0,0 +1,28 @@
...
+  
+  depends_on "cmake" => :build
+  depends_on "boost"
+  
+  def install
+    cmake_args = std_cmake_args
+    cmake_args << "-DBOOST_ROOT=#{Formula["boost"].opt_prefix}"
+    system "cmake", ".", *cmake_args
+    system "make"
+    bin.install "bin/PEAT"
```
tdsmith: is there no `make install`?

peat.rb
```diff
@@ -0,0 +1,28 @@
...
+  # tag "bioinformatics"
+  # doi "10.1186/1471-2105-16-S1-S2"
+
+  url "https://github.com/jhhung/PEAT/archive/v1.0-beta.1.tar.gz"
+  sha256 "dc7216aeccc8f57885a84e3f762ff7281542c5f062a9457061e5da295331d867"
+  version "1.0b1"
+  
+  depends_on "cmake" => :build
+  depends_on "boost"
+
```
tdsmith: add `needs :cxx11` here

peat.rb
```diff
@@ -0,0 +1,28 @@
...
+  # doi "10.1186/1471-2105-16-S1-S2"
+
+  url "https://github.com/jhhung/PEAT/archive/v1.0-beta.1.tar.gz"
+  sha256 "dc7216aeccc8f57885a84e3f762ff7281542c5f062a9457061e5da295331d867"
+  version "1.0b1"
+  
+  depends_on "cmake" => :build
+  depends_on "boost"
+  
+  def install
```
tdsmith: add `ENV.cxx11` here

peat.rb
```diff
@@ -0,0 +1,28 @@
...
+  homepage "https://github.com/jhhung/PEAT/"
+  # tag "bioinformatics"
+  # doi "10.1186/1471-2105-16-S1-S2"
+
+  url "https://github.com/jhhung/PEAT/archive/v1.0-beta.1.tar.gz"
+  sha256 "dc7216aeccc8f57885a84e3f762ff7281542c5f062a9457061e5da295331d867"
+  version "1.0b1"
+  
+  depends_on "cmake" => :build
+  depends_on "boost"
```
tdsmith: this can be:
`depends_on "boost" => (MacOS.version < :mavericks ? "cxx11" : [])`

peat.rb
```diff
@@ -0,0 +1,28 @@
...
+    system "make"
+    bin.install "bin/PEAT"
+    doc.install "README.md", "LICENSE"
+    (share / "peat").install Dir["test_file/*"]
+  end
+
+  test do
+    assert_match "trimming", shell_output("PEAT 2>&1", 1)
+    assert_match "single", shell_output("PEAT single 2>&1", 1)
+    assert_match "paired", shell_output("PEAT paired 2>&1", 1)
```
tdsmith: Can we add a test that tests the functionality of the application? Downloading a small data set as a resource is fair.

tdsmith: Testing against the test files installed to `share` would be :100:

peat.rb
tseemann: Will this affect if OS.linux?

peat.rb
tseemann: Nope. or "clean" !

peat.rb
tseemann: I knew someone would notice those! ;-)

peat.rb
tdsmith:
```
tim@ramnode:~$ brew irb
==> Interactive Homebrew Shell
Example commands available with: brew irb --examples
irb(main):001:0> MacOS.version < :mavericks
=> true
irb(main):002:0> OS.linux?
=> true
```

so this will always require boost to be rebuilt in c++11 mode on Linux; I'm not sure if that's the right thing to do or not (but this is probably safest?).

tseemann: It wants to build boost every time it builds PEAT now (on Linux). I'm getting confused. sorry!

tdsmith: If you change the conditional to
`OS.mac? && MacOS.version < :mavericks`
you won't have to rebuild boost. If it still works for you that way, then I'm happy.

tseemann:
@tdsmith I don't think I'm going to be able to build this in Brew properly, it's exceeding my skills and patience.

They distribute Mac and Linux binaries for a reason it seems, and I tried to do it from source, but i've failed. I will go back to putting it my own repo as a binary formula unless someone wants to take this on.

If they get it built, I am happy to write a proper test that uses the sample input files.

xu-cheng:
@tseemann If you remove `needs :openmp`, it should build fine. 

The link problem you encountered is because the program is built by gcc(system clang doesn't support openmp and you are asking for it) while boost is built by clang.

tseemann:
I messed up my git branch for this and have started again. I will incorporate the ideas in this thread.

PR #2165, opened 2015-04-27
libdivsufsort 2.0.1 (new formula)

tseemann:
This library provides a simple and an efficient C API to construct a suffix array and a Burrows-Wheeler transformed string from a given string over a constant-size alphabet. The algorithm runs in O(n log n) worst-case time using only 5n+O(1) bytes of memory space, where n is the length of the string.

[dependency for Andi software]


libdivsufsort.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    mkdir "libdivsufsort-build" do
+      args = std_cmake_args
+      system "cmake", "..", *args
+      system "make", "install"
+    end
+  end
+
+  test do
+    # no test due to being library only
```
dpo: Would it be difficult to write a short program that uses the library and run it as a test?

sjackman: Is there any Homebrew infrastructure to automate this, for example simply to test that a particular symbol exists in the library, as done by autoconf? We could use `nm` to check that the symbol exists.

sjackman: I'm a fine of making the test extra simple. Comprehensive tests are the author's responsibility, not Homebrew.

dpo: Yes and no. I find linking a short program against a library is a great way to make sure the install is solid. The library may be there and tests might pass during build, but that doesn't prove it's functional when installed.

sjackman: Yeah, true. I feel that build a test application is going to be 5-10 lines of boiler plate code that I'd rather not see replicated in every formula. I'd like to see the equivalent of autoconf's `	AC_CHECK_LIB([pthread], [pthread_create])`, such as
```ruby
test do
    check_library lib/"libpthread.so", "pthread_create"
end
```

dpo: Agree that that would come in handy. Slightly more thorough test, like in [armadillo.rb](https://github.com/Homebrew/homebrew-science/blob/master/armadillo.rb#L35) or [mlpack.rb](https://github.com/Homebrew/homebrew-science/blob/master/mlpack.rb#L58) are pretty easy to implement though, and they'll let you know immediately if something's glaring. Sometimes, such examples ship with the library!

tseemann: I added this formula to allow me to add a formula for Andi which uses it.

In the short term Andi will test if it is working or not.

libdivsufsort.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  url "https://github.com/y-256/libdivsufsort/archive/2.0.1.tar.gz"
+  sha256 "9164cb6044dcb6e430555721e3318d5a8f38871c2da9fd9256665746a69351e0"
+
+  depends_on "cmake" => :build
+
+  def install
+    mkdir "libdivsufsort-build" do
+      args = std_cmake_args
+      system "cmake", "..", *args
```
sjackman: This should be simply `system "cmake", "..", * std_cmake_args`

libdivsufsort.rb
sjackman: Although a test would be nice, I'm okay with merging without a test.

tdsmith: Remove the `test` block entirely if you don't intend to add one, though; it will fail a strict audit but that's a correct failure.

libdivsufsort.rb
tseemann: Agreed. I think I was testing fiddling with some compile options but removed them ultimately. I've simplified it now.

libdivsufsort.rb
tdsmith: Building and running the example from the README would be great: https://github.com/y-256/libdivsufsort

Installing the examples in the tarball to `share` and building one of those would be great, too.

tseemann:
Hooray, I made it!  Once this is in, I will test it on Andi software.

sjackman:
I added a test.

```ruby
  test do
    (testpath/"test.c").write <<-EOS
      #include <divsufsort.h>
      #include <stdio.h>
      int main() {
        const unsigned char s[] = "panamabananas";
        int sa[sizeof s];
        int i;
        divsufsort(s, sa, sizeof s);
        for (i = 0; i < sizeof s; ++i)
          printf(" %d", sa[i]);
        return 0;
      }
    EOS
    system ENV.cc, "test.c", "-I#{include}", "-L#{lib}", "-ldivsufsort", "-o", "test"
    assert_equal shell_output("./test"), " 13 5 3 1 7 9 11 6 4 2 8 10 0 12"
  end
```

sjackman:
See here: https://stepic.org/lesson/Suffix-Arrays-310/step/1

PR #2171, opened 2015-04-28
placnet 1.01 (new formula)

tseemann:
Placnet is a new tools for plasmid analysis in NGS projects. Placnet has been written in Perl. It's been optimized to work with Illumina sequences but it also works with 454, Iontorrent or any of the actual sequence technologies.

The input of placnet is a set of contigs and one or more SAM files with the mapping of the reads against the contigs. Placnet obtains a set of files, easily opened on Cytoscape software or other network tools.


placnet.rb
```diff
@@ -0,0 +1,37 @@
...
+  sha256 "c25704c5ba7086121cc20309bc93f7ed123ca68a036c8678d7197ae6690dec08"
+  version "1.01"
+
+  resource "manual" do
+    url "https://downloads.sourceforge.net/project/placnet/Manual_v1.01.pdf"
+    sha256 "9fdedc8539546c4662ed7d17069cea4773b4bfe88edc29cc97a8cdcfae32b87c"
+  end
+
+  def install
+    inreplace "placnet.pl", "/usr/bin/perl", "/usr/bin/env perl"
```
tdsmith: Why not prefer system perl?

placnet.rb
```diff
@@ -0,0 +1,37 @@
+class Placnet < Formula
+  homepage "https://sourceforge.net/projects/placnet"
+  # tag "bioinformatics"
+  # doi "10.1371/journal.pgen.1004766"
+
+  depends_on "Getopt::Long" => :perl
```
tdsmith: Can you download this as a `resource`, stage it to libexec, and modify PERLPATH at runtime with `write_env_script` to point to where it's installed so that the formula Just Works? This is what we usually do in core for applications that require Python modules; grepping should identify some examples of doing this with perl as well.

placnet.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def caveats
+    caveats = <<-EOS.undent
+      This tool relies on GeneMark to run which requires an acdemic licence
+      from http://exon.gatech.edu/GeneMark/license_download.cgi
+    EOS
+  end
+
+  test do
+    assert_match "Plasmid", shell_output("placnet 2>&1", 1)
```
tdsmith: Is there a test we can do that exercises the functionality of the application?

placnet.rb
```diff
@@ -0,0 +1,37 @@
...
+    inreplace "placnet.pl", "/usr/bin/perl", "/usr/bin/env perl"
+    bin.install "placnet.pl" => "placnet"
+    resource("manual").stage do
+      doc.install "Manual_v1.01.pdf"
+    end
+  end
+
+  def caveats
+    caveats = <<-EOS.undent
+      This tool relies on GeneMark to run which requires an acdemic licence
```
tdsmith: typo: "academic"

placnet.rb
```diff
@@ -0,0 +1,37 @@
...
+      from http://exon.gatech.edu/GeneMark/license_download.cgi
+    EOS
```
tdsmith: I'm not sure what this means; does this mean that users will need to separately download and install GeneMark before this formula is useful?

placnet.rb
```diff
@@ -0,0 +1,37 @@
+class Placnet < Formula
+  homepage "https://sourceforge.net/projects/placnet"
+  # tag "bioinformatics"
+  # doi "10.1371/journal.pgen.1004766"
+
+  depends_on "Getopt::Long" => :perl
+  depends_on "blast"
+  depends_on "velvet" => :optional
+
+  url "https://downloads.sourceforge.net/project/placnet/placnet.pl"
```
tdsmith: Can we request a versioned filename so that this doesn't break on updates?

placnet.rb
tseemann: What do you mean?  
the depends Getopt::Long line just tests that module is available no matter where?
(i don't have brew perl and that's what it does)

placnet.rb
tseemann: It just checks the module is installed. It is standard with more recent Perl.

placnet.rb
tseemann: No. Unable to run without the caveats.

placnet.rb
tdsmith: Rephrased: why is this inreplace useful?

placnet.rb
tseemann: Correct. There are other examples of this in homebrew. The s/w is "free" but needs a click-through email licence.

placnet.rb
tdsmith: If it should always exist, there's no need to test for it; if it might not exist, we should always download and install it.

placnet.rb
tseemann: If by "we" you mean me, but I don't have the time to chase them up now. I'm already chasing 6 projects on versions.  Happy to put this one on the backburner, not urgent for me.

placnet.rb
tdsmith: Okay, that's fine! Can this language be simplified to:

> placnet requires GeneMark, which can be obtained from http://exon.gatech.edu/GeneMark/license_download.cgi

placnet.rb
tdsmith: :+1:

placnet.rb
tseemann: That's not what every other recipe does. The default behavior of ```depends_on 'xxx::yyy' => :perl``` is to print a message ```Brew doesn't do perl, use cpan -i xxx:yyy to install it```. Same for python where it says to use pip.

placnet.rb
tdsmith: Yes, "you," sorry. :)

placnet.rb
tseemann: Because *every* perl/python script should have that. There is **no** guarantee where Perl will be installed. This will ensure the one the sydadmin intended to be run (ie. first in PATH) is used.

tseemann:
```
==> placnet 2>&1
Error: placnet: failed
```

The authors have Spanish accent characters in their name. I've had this problem before. Solution?

placnet.rb
tdsmith: Using :language dependencies isn't allowed in Homebrew/homebrew; they're still used here but I think it's better to avoid them.

Using CPAN or shelling out to pip to install into a user's regular package library is bad because that means a) network traffic we can't cache and b) messing with the user's current development environment; installing them to our own private paths is more stable and less intrusive.

If this module exists on Mountain Lion but maybe doesn't exist on $linux_distro then this is probably fine.

placnet.rb
tdsmith: /usr/bin/perl is guaranteed to exist on OS X.

tseemann: I am one of those annoying Brew users who primarily target LinuxBrew, as most bioinformatics apps run on Linux servers. I am "forced" to get it working in OS X, which I am happy to do.

What if the user installs the Brew version of the latest Perl?

Shouldn't Brew be using that one?

tdsmith:
Does either
* placing `# encoding: utf-8` at the very top of the formula, or
* setting `Encoding.default_external = Encoding::UTF_8` at the top of `test`
resolve the accent problem?

placnet.rb
tdsmith: Usually "no," because what's the difference? Always using the system interpreter is probably more deterministic.

A system without a /usr/bin/perl would be pretty weird; if the fear is that it's potentially ancient because maybe we're running on legacy RHEL or something then this is fine.

tseemann: The difference affects the search PATH for modules.

tdsmith: Sure, but in Homebrew/homebrew it shouldn't matter (goes the logic),
because all of the language module dependencies should be vendored by the
formula into a private path that's injected into the runtime environment by
a wrapper script.

mscook:
@tdsmith - can we please encourage the use of /usr/bin/env perl and *NOT* /usr/bin/perl for the reason that software is often meant to be run on different platforms (and the latter can cause heaps and heaps of problems).

mscook:
@tseemann - will the sourceforge distribution of placnet actually do anything? It was of my belief that the required databases and other tools were missing.

tdsmith:
> can we please encourage the use of /usr/bin/env perl and NOT /usr/bin/perl

I'm sympathetic to the idea that this may be useful for Linuxbrew users who are not system administrators. As always, it would be great to send changes like this which aren't Homebrew-specific to the upstream developers.

tseemann:
@tdsmith 
Is there something Brew assumes that requires it to be in that location? 
If so, I will change it.
If not, I would have thought having cross-platform compatibilty would be helpful.

tdsmith:
It's fine. It would be great to have this changed upstream.

tseemann:
This software is problematic and I do not wish to add it to Brew any longer.

PR #2177, opened 2015-04-30
neuron: add linux support

ntkme:
This RP enables linuxbrew to build neuron.


neuron.rb
```diff
@@ -28,6 +28,12 @@ class Neuron < Formula
   patch :DATA
 
   def install
+    if OS.mac?
+      inreplace "configure", "$IV_LIBDIR/libIVhines.la", "$IV_LIBDIR/libIVhines.dylib"
+    else
+      inreplace "configure", "$IV_LIBDIR/libIVhines.la", "$IV_LIBDIR/libIVhines.so"
+    end
+
```
dpo: Other formulas use `dylib = OS.mac? "dylib" : "so"`, which is a bit more concise.

neuron.rb
```diff
@@ -55,6 +58,8 @@ def install
     end
 
     ln_sf Dir["#{libexec}/lib/*.dylib"], lib
+    ln_sf Dir["#{libexec}/lib/*.so.*"], lib
+    ln_sf Dir["#{libexec}/lib/*.so"], lib
```
dpo: Can these be just `#{libexec}/lib/*.#{dylib}`?

ntkme: On linux we also need to link "*.so.X" and "*.so.X.Y.Z".
I think it won't hurt by explicitly listing them all.

PR #2178, opened 2015-04-30
OpenCV: update devel to RC1

nicolacastaman:


tdsmith:
Did you test this locally? I think that the patch in the `devel` block should no longer apply since the change should have been accepted upstream; you'll probably have to remove it, or else I'll have to investigate why it didn't stick.

tdsmith:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support.

Just for future reference, our favorite commit message format is like `packagename 1.2.3 (devel)` for devel version bumps, squashed to a single commit.

PR #2180, opened 2015-05-01
lmfit 5.1 (new formula)

T045T:
Added a formula for lmfit ( http://apps.jcns.fz-juelich.de/doku/sc/lmfit ).

No test block because I wasn't sure about compiling a test program in a test - both how to best do it and whether it should be done at all.

### DIFF
--- a/lmfit.rb
+++ b/lmfit.rb
@@ -11,6 +11,9 @@ class Lmfit < Formula
     (share/"lmfit/demos").install Dir["demo/*.c"]
   end
 
+  def caveats
+    "Demo sources have been placed in " + (share/"lmfit/demos")
+  end
 
   test do
     # Compile the demo programs from share and check that they exit nonzero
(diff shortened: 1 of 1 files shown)


lmfit.rb
```diff
@@ -0,0 +1,103 @@
...
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS
+    #include "lmcurve.h"
+    #include "lmmin.h"
```
tdsmith: Can you indent the test code another 2 spaces?

tdsmith:
The test looks good! If you want you can install the demo from the tarball with e.g. `(share/"lmfit").install Dir["demo/*"]` or whatever and then build it from there in the test block.

dpo:
+1 for the demo.

T045T:
I changed the formula to install the demo sources - since they're supposed to demonstrate how the library works, I thought this made more sense than copying the binaries.

I also added a caveat to inform the user about those demo files. Is that okay?

dpo:
Great. Thank you for your contribution!

PR #2192, opened 2015-05-02
Opencl

glehmann:
add clfft and clblas opencl libraries


clblas.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  patch do
+    # install clBLAS cmake configuration and version files
+    url "https://github.com/clMathLibraries/clBLAS/commit/554d61ec834c72b6ce727b9232ecbf9864cf9d8c.diff"
+    sha256 "49bd44ef1c2cb6e0e63ab5286d2aa185224a529c04d3d9efff45533c984d6bf8"
+  end
+
+  def install
+    args = std_cmake_args
+    args << "-DBUILD_TEST=OFF"
```
dpo: What kind of test does this build? Could it be an option to build it?

clfft.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  patch do
+    # don't force the usage of libc++
+    url "https://github.com/clMathLibraries/clFFT/pull/73.diff"
+    sha256 "ad7d8b858027e8562b5478793583299f2dfc00621919c62ee199e1885af09a99"
+  end
+
+  def install
+    cd "src"
+    system "cmake", ".", "-DBUILD_TEST:BOOL=OFF", *std_cmake_args
```
dpo: What kind of test does this build? Could it be an option to build it?

glehmann:
They are extra tests that require extra dependencies. The client in these libraries is also able to run some tests - this is what is done in the test section :-)

dpo:
Thank you.

PR #2193, opened 2015-05-02
z3 4.4.0

dtrebbien:


z3.rb
```diff
@@ -1,17 +1,9 @@
...
   head "https://github.com/Z3Prover/z3.git"
 
-  bottle do
-    root_url "https://homebrew.bintray.com/bottles-science"
-    cellar :any
-    sha256 "2ce2e28b1ed47fd2df764fdba8ed61fa6896f6cb4f443afed7dcee6fae69b080" => :yosemite
-    sha256 "cb5f83f5a2b2d263d51ae8df5c0cbf7fdee4d72d58c26a17ea57f19bf2a96ccb" => :mavericks
-    sha256 "9e58713737a98619cc05c0f89fb63ba61ed66c4a98beeea650d6a83c4bdb8beb" => :mountain_lion
-  end
-
```
dpo: Please don't remove the `bottle` section. It will be updated by the bot.

dtrebbien: @dpo: Do I just remove the sha256 lines in the bottle section?

dpo: No, there's no need to touch the bottle section at all.

dpo: I'm not sure if I was very clear. Could you please restore the bottle section as if you had never removed it?

dtrebbien:
Hi @dpo.  I have pushed a new commit which just changes the url and sha256 lines.

dpo:
Thanks!

PR #2194, opened 2015-05-02
aribas 1.64 (new formula)

dtrebbien:


aribas.rb
```diff
@@ -0,0 +1,104 @@
...
+          k5 = 327843840;
+      var
+          A, n: integer;
+          S: integer;
+      begin
+          A := zz*k1;
+          S := A * k2;
+          n := 1;
+          while A > 0 do
+                  A := A * ((6*n-5)*(6*n-3)*(6*n-1));
```
dpo: Seems like the indentation needs to be fixed here.

aribas.rb
```diff
@@ -0,0 +1,104 @@
...
+          x := isqrt(zz*zz*k3)*k6;
+          x := (zz * x) div Saux(zz);
+          return (x div 2**16);
+      end;
+
+      pi_chud(50).
+
+      exit.
+    EOS
+    system "#{bin}/aribas pi.ari | grep -q 3_14159_26535_89793_23846_26433_83279_50288_41971_69399_37510"
```
dpo: Great test!

aribas.rb
dtrebbien: Yes, you're right.  New commit uploaded.

aribas.rb
dtrebbien: Thanks :smile:

PR #2205, opened 2015-05-05
update vips.rb to 8.0

jcupitt:
I removed the checksums on the old bottled versions, hope that's correct.


iml:
The bottle block is managed by @BrewTestBot so that should be left unchanged. I'll add a note mentioning that in the contributing guide since it's a common misunderstanding. Also, for version upgrades the preferred commit message is just `forumla_name version` i.e. `vips 8.0.2`. Thanks!

guoxiao:
see #2228

jcupitt:
I'll close this, please merge #2228 instead. Thanks!

PR #2207, opened 2015-05-05
GMT 5.1.2

claudiodsf:


gmt.rb
```diff
@@ -2,15 +2,15 @@
...
-  mirror "ftp://gd.tuwien.ac.at/pub/gmt/legacy/gmt-5.1.1-src.tar.bz2"
-  sha1 "ff64936dfdec8a57a89d29f505f27e435169d33f"
+  url "ftp://ftp.soest.hawaii.edu/gmt/gmt-5.1.2-src.tar.bz2"
+  mirror "ftp://ftp.star.nesdis.noaa.gov/pub/sod/lsa/gmt/gmt-5.1.2-src.tar.bz2"
+  mirror "ftp://ftp.iris.washington.edu/pub/gmt/gmt-5.1.2-src.tar.bz2"
+  mirror "ftp://ftp.iag.usp.br/pub/gmt/gmt-5.1.2-src.tar.bz2"
+  mirror "ftp://ftp.scc.u-tokai.ac.jp/pub/gmt/gmt-5.1.2-src.tar.bz2"
+  mirror "ftp://gmt.mirror.ac.za/pub/gmt/gmt-5.1.2-src.tar.bz2"
+  mirror "ftp://ftp.geologi.uio.no/pub/gmt/gmt-5.1.2-src.tar.bz2"
+  mirror "ftp://gd.tuwien.ac.at/pub/gmt/gmt-5.1.2-src.tar.bz2"
```
dpo: 1. Is there really a *need* for so many mirrors?
2. If yes, a `version` variable would serve the formula well.

claudiodsf: Probably not, but I'm unsure about which one to choose.
@joa-quim, any thought?

joa-quim:
DCW is a relatively small package that is not even often updated. Hawaii is the home place so it must remain but being an island in the path off hurricanes it sometimes suffers power outages so we need a mirror or two. I would suggest the

ftp://ftp.star.nesdis.noaa.gov/pub/sod/lsa/gmt/gmt-5.1.2-src.tar.bz2
ftp://gd.tuwien.ac.at/pub/gmt/gmt-5.1.2-src.tar.bz2

or even only the nooa one if the decision is to keep one mirror only.

dpo:
One or two mirrors at most would be sufficient.

dpo:
@claudiodsf

claudiodsf:
Sorry for this late reaction :wink:

dpo:
Thanks!

PR #2219, opened 2015-05-07
alpscore 0.4.5

aeantipov:
The ALPSCore project, based on the ALPS (Algorithms and Libraries for Physics Simulations) project, provides generic algorithms and utilities for physics problems. It strives to increase software reuse in the physics community.


alpscore.rb
```diff
@@ -0,0 +1,122 @@
+require 'formula'
```
dpo: This line is now obsolete and can be removed.

alpscore.rb
```diff
@@ -0,0 +1,122 @@
...
+
+class Alpscore < Formula
+  homepage "http://alpscore.org"
+  url "alpscore"
```
dpo: Here we need a `url` to the stable release. Probably the one that's on line 9.

alpscore.rb
```diff
@@ -0,0 +1,113 @@
...
+  sha256 "b043f5043f6fdca5efd8e1fc2ba0d893da0fd04bff8adaa213c797b44d68e72e"
+
+  # fetch current version fro git (fix with first release)
```
dpo: This comment and the one on line 9 below are unnecessary.

alpscore.rb
```diff
@@ -0,0 +1,113 @@
...
+  url "https://github.com/ALPSCore/ALPSCore/archive/v0.4.5.tar.gz"
+
+  # head version checked out from git
+  head "https://github.com/ALPSCore/ALPSCore.git"
+
+  # options
+  option "with-static", "Build static instead of shared libraries"
+  option "without-mpi", "Disable building with MPI support"
```
dpo: This is redundant with the dependency on `:mpi`.

alpscore.rb
```diff
@@ -0,0 +1,113 @@
...
+  option "with-doc",    "Build documentation"
+  option "with-check",  "Build and run shipped tests"
+
+  # Dependencies
+  # cmake
+  depends_on "cmake"   => :build
+  # boost - check mpi and c++11
+  boost_options = []
+  boost_options << "with-mpi" if build.with? "mpi"
+  boost_options << "without-single" if build.with? "mpi"
```
dpo: The two lines above could be grouped: `boost_options += ["with-mpi", "without-single"] if build.with? "mpi"`

alpscore.rb
```diff
@@ -0,0 +1,113 @@
...
+  boost_options << "c++11" if build.cxx11?
+  depends_on "boost" => boost_options
+  # mpi, hdf5
+  if build.cxx11?
+      depends_on "open-mpi" => ["c++11", :recommended] if build.with? "mpi" # boost has troubles with mpich
```
dpo: It seems like this will be handled by `boost`, won't it?

alpscore.rb
```diff
@@ -0,0 +1,113 @@
...
+      depends_on "hdf5" => ["c++11"]
+  else
+      depends_on :mpi      => [:cc, :cxx, :recommended] if build.with?"mpi"
```
dpo: Space missing before `"mpi"`.

dpo:
Thanks for your contribution! Please run `brew audit --strict ./alpscore.rb` for style and syntax issues.

alpscore.rb
aeantipov: boost has 

```
if build.cxx11?
depends_on "icu4c" => [:optional, "c++11"]
depends_on "open-mpi" => "c++11" if build.with? "mpi"
```
Have no idea how to do a proper workaround

alpscore.rb
aeantipov: Not sure if I get it. mpi is an optional dependency

aeantipov:
Dominique, thanks a lot! 
I worked through the updates and ran strict audit.

alpscore.rb
dpo: Sorry, the mpi dependency is a bit messy here (like in `boost`). I think it should be sufficient to keep your `option "without-mpi"` and to pass `"with-mpi"` to `boost` if the user builds `with-mpi`. So lines 26 and 29 should be taken care of by `boost`.

alpscore.rb
```diff
@@ -0,0 +1,108 @@
...
+        alps::accumulators::accumulator_set set;
+        set << alps::accumulators::FullBinningAccumulator<double>("a");
+        set["a"] << 2.9 << 3.1;
+      }
+    EOS
+    args_compile = %W[]
+    args_compile << "test.cpp" << "-lalps-accumulators" << "-lalps-hdf5" << "-lalps-utilities"
+    args_compile << "-lboost_filesystem-mt" << "-lboost_system-mt" << "-lboost_program_options-mt"
+    if build.with? "mpi"
+      args_compile << "-lmpi" << "-lmpi_cxx" << "-lboost_mpi-mt"
```
dpo: This is a bit risky because those libs aren't the same for OpenMPI and MPICH. It would be safer to compile with `ENV["MPICXX"]` instead of `ENV.cxx`. `MPICXX` will automatically add the appropriate libraries.

alpscore.rb
```diff
@@ -0,0 +1,108 @@
...
+      int main()
+      {
+        alps::accumulators::accumulator_set set;
+        set << alps::accumulators::FullBinningAccumulator<double>("a");
+        set["a"] << 2.9 << 3.1;
+      }
+    EOS
+    args_compile = %W[]
+    args_compile << "test.cpp" << "-lalps-accumulators" << "-lalps-hdf5" << "-lalps-utilities"
+    args_compile << "-lboost_filesystem-mt" << "-lboost_system-mt" << "-lboost_program_options-mt"
```
dpo: Let's populate the `args_compile` list when it's declared instead of adding to it.

alpscore.rb
aeantipov: Trying this, but ENV["MPICXX"] doesn't seem to work

aeantipov:
It seems I will need help with the mountain lion error:
```       
 ==> Installing alpscore from homebrew/homebrew-science
Error: undefined method `[]' for :required:Symbol
```
It probably comes from older ruby on 10.8 and passes on 10.9, 10.10

dpo:
It's enough to say `depends_on :mpi => [:cc, :cxx] if build.with? "mpi"`.

alpscore.rb
```diff
@@ -0,0 +1,114 @@
...
+  boost_options = []
+  boost_options += ["with-mpi", "without-single"] if build.with? "mpi"
+  boost_options << "c++11" if build.cxx11?
+  depends_on "boost" => boost_options
+  # hdf5
+  if build.cxx11?
+    depends_on "hdf5" => ["c++11"]
+  else
+    depends_on "hdf5"
+  end
```
dpo: Let's make this a one liner: `depends_on "hdf5" => ((build.cxx11?) ? ["c++11"] : [])`

alpscore.rb
```diff
@@ -0,0 +1,114 @@
...
+      args << "-DBuildStatic=OFF"
+      args << "-DBuildShared=ON"
+    end
+
+    # handle mpi
+    if build.with? "mpi"
+      args << "-DENABLE_MPI=ON"
+    else
+      args << "-DENABLE_MPI=OFF"
+    end
```
dpo: `args << ("-DENABLE_MPI=" + ((build.with? "mpi") ? "ON" : "OFF"))`

alpscore.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    # documentation
+    if build.with? "doc"
+      args << "-DDocumentation=ON"
+    else
+      args << "-DDocumentation=OFF"
+    end
```
dpo: Same here.

alpscore.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    # do the actual install
+    mkdir "tmp"
+    chdir "tmp"
+    # the source is at parent dir
+    args << ".."
+    system "cmake", *args
+    if build.with? "check"
+      system "make"
+      system "make", "test"
```
dpo: Let's move `system "make"` outside the `if` and say `system "make", "test" if build.with? "check"`

aeantipov:
Done.

alpscore.rb
```diff
@@ -0,0 +1,95 @@
+class Alpscore < Formula
+  homepage "http://alpscore.org"
+  sha256 "b043f5043f6fdca5efd8e1fc2ba0d893da0fd04bff8adaa213c797b44d68e72e"
+
+  url "https://github.com/ALPSCore/ALPSCore/archive/v0.4.5.tar.gz"
+
+  head "https://github.com/ALPSCore/ALPSCore.git"
+
+  # options
```
dpo: Thanks for your patience. Please bear with me for just a few more small changes. It seems most of the comments are unnecessary and can be removed.

alpscore.rb
```diff
@@ -0,0 +1,95 @@
...
+    args << ("-DENABLE_MPI=" + ((build.with? "mpi") ? "ON" : "OFF"))
+
+    # documentation
+    args << ("-DDocumentation=" + ((build.with? "doc") ? "ON" : "OFF"))
+
+    # check
+    args << ("-DTesting=" + ((build.with? "check") ? "ON" : "OFF"))
+
+    # do the actual install
+    mkdir "tmp"
```
dpo: This shoud be a `mkdir "tmp" do ... end` block

aeantipov:
No problems, thanks for help!

jacknagel:
These two lines:

```rb
option "without-mpi", "Disable building with MPI support"
depends_on :mpi => [:cc, :cxx] if build.with? "mpi"
```

are equivalent to:

```rb
depends_on :mpi => [:cc, :cxx, :recommended]
```

dpo:
They are now that we've gotten rid of the implicit dependency on `open-mpi` (implied by `boost`).

aeantipov:
fixed

dpo:
Thanks for your patience!

PR #2225, opened 2015-05-08
quest 2.4.0 (new formula)

saketkc:
No longer maintained. 
QuEST is a statistical software for analysis of ChIP-Seq data


saketkc:
@iml Makes sense to put this on homebrew-science? It is no longer maintained, so probably will go to my personal tap otherwise.

iml:
I'm fine with including it here.

PR #2229, opened 2015-05-09
field3d: bump revision

tdsmith:
Link against current ilmbase.


tdsmith:
blocking #2229.

tdsmith:
Electing to bump this in #2216 instead because otherwise we'll have to revision oiio to keep it from breaking, when we really just want to upgrade it.

PR #2232, opened 2015-05-09
analysis 0.8.7

### LINKED ISSUE #2231: analysis.rb out of date
This package is currently not installable via brew:

1. It links to a source package at molpopgen.org.  That site no longer provides releases.  Everything is on git.
2. The specific version is too old to be compatible with the version of libsequence installed using brew.

tdsmith:
Closes #2231.

### DIFF
--- a/libsequence.rb
+++ b/libsequence.rb
@@ -14,11 +14,13 @@ class Libsequence < Formula
     sha256 "d3734a60eece746f0e72bc58275350ed196c55dd1fe12000a74738a14547b3a7" => :mavericks
   end
 
-  depends_on :macos => :mavericks
   depends_on "boost" => :build
   depends_on "gsl"
 
+  needs :cxx11
+
   def install
+    ENV.cxx11
     system "./configure", "--prefix=#{prefix}",
                           "--docdir=#{doc}",
                           "--mandir=#{man}",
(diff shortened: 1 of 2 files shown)


libsequence.rb
```diff
@@ -14,11 +14,13 @@ class Libsequence < Formula
...
   end
 
-  depends_on :macos => :mavericks
   depends_on "boost" => :build
   depends_on "gsl"
 
+  needs :cxx11
+
   def install
+    ENV.cxx11
```
tdsmith: @iml, in #2098 you said that this is wrong, but I'm not sure I understand why, since the libsequence README says it requires a c++11 compiler. Am I missing something?

tdsmith:
Patch submitted as https://github.com/molpopgen/analysis/pull/3.

libsequence.rb
iml: I was referring to `ENV.libcxx if MacOS.version < :mavericks` as being wrong. `ENV.cxx11` and boost cxx11 looks correct. I must have thought a lot of patching was needed to get the older compiler to support cstdint, but that's a standard C++11 header.

tdsmith: :+1: thanks for clarifying!

PR #2238, opened 2015-05-11
New formula to install bio++ library

kgori:
Bio++ is a set of C++ libraries for Bioinformatics, including sequence analysis, phylogenetics, molecular evolution and population genetics.


biopp.rb
```diff
@@ -0,0 +1,68 @@
...
+    sha256 "c7ec73a5af84808362f301479c548b6a01c47a66065b28a053ff8043409e861a"
+  end
+
+  depends_on "cmake" => :build
+
+  def install
+    %w[bppcore bppseq bppphyl bpppopgen bppseqomics bppraa].each do |r|
+      resource(r).stage do
+        Dir.mkdir "build"
+        Dir.chdir("build") do
```
dunn: These two lines can be replaced with `mkdir "build" do`

steeplejack:
This formula should hopefully be OK to merge now.

kgori:
Can this be merged?

biopp.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    end
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make", "bppphylomics-shared", "bppphylomics-static"
+      system "make", "install"
+    end
+  end
+  test do
+    File.exist? "#{include}/Bpp.all"
```
tdsmith: Can the test build and link a really simple program against the library? Just a few lines from e.g. http://biopp.univ-montp2.fr/wiki/index.php/Alphabet would be great!

kgori: Thanks for the suggestion. It works on my machine, but we'll see what jenkins has to say about it.

tdsmith:
Thanks! I squashed your commits and changed the test to use a heredoc instead of string concatenation and then to invoke the compiled binary; the result is in e1fc14b95a. Homebrew thrives because of pull requests like this!

kgori:
Great, thanks!

PR #2249, opened 2015-05-13
gnudatalanguage OpenMP support

tschoonj:
made possible using clang-omp!


gnudatalanguage.rb
```diff
@@ -27,6 +27,13 @@ class Gnudatalanguage < Formula
   depends_on :x11
   depends_on :python => :optional
 
+  option "with-openmp", "Build with OpenMP support"
+
+  if build.with?("openmp")
+    depends_on "clang-omp" => :build
+    depends_on "libiomp"
+  end
```
xu-cheng: Don't add `clang-omp`. Instead you should just add `needs :openmp if build.with? "openmp"`.

gnudatalanguage.rb
```diff
@@ -39,6 +46,13 @@ def install
     args << "-DHDF=OFF" if build.without?("hdf4")
     args << "-DPYTHON=OFF" if build.without?("python")
     args << "-DWXWIDGETS=OFF" << "-DPSLIB=OFF"
+    if build.with?("openmp")
+      args << "-DOPENMP=ON"
+      args << "-DCMAKE_CXX_COMPILER=clang-omp++"
+      args << "-DCMAKE_C_COMPILER=clang-omp"
```
xu-cheng: Don't set compiler.

tschoonj:
needs :openmp means that gcc will be used right? In that case compilation will fail badly due to libc++ vs libstdc++ issues.

xu-cheng:
> needs :openmp means that gcc will be used right? In that case compilation will fail badly due to libc++ vs libstdc++ issues.

I'm working on this in Homebrew/homebrew#39676 

You shouldn't set compiler as it should be gcc for linuxbrew or tigerbrew.

tschoonj:
Cool! didn't realize you were working on this!

Ok, will close this one for now then and wait until you merge in your work. If I change it to :openmp at this point, compilation is doomed to fail

PR #2261, opened 2015-05-14
datamash 1.0.6 (new formula)

sjackman:
GNU datamash is a command-line program which performs basic
numeric, textual and statistical operations on input textual
data files.


mikemcquaid:
@sjackman FYI I'd wait for it to be remove from core before merging into a tap. No worries this time though, thanks!

sjackman:
Ah, good to know. Thanks, Mike.

PR #2263, opened 2015-05-14
grabix: bump version to 0.1.5

brentp:


grabix.rb
```diff
@@ -1,8 +1,8 @@
 class Grabix < Formula
   homepage "https://github.com/arq5x/grabix"
   # tag "bioinformatics"
-  url "https://github.com/arq5x/grabix/archive/1.4.tar.gz"
-  sha1 "687c6874741127c012e84cad4d85a5116afc1034"
+  url "https://github.com/arq5x/grabix/archive/0.1.5.tar.gz"
+  sha1 "5542ecb027b615e10b7f462161b553953d5394af"
   head "https://github.com/arq5x/grabix.git"
```
dpo: We now use the `sha256`. You can obtain it with `shasum -a 256`. Thanks!

brentp: it's the shasum of the tar ball? I had it as the git sha.

dpo: Yes, it's the shasum of the tarball: `shasum -a 256 1.5.tar.gz`.

PR #2272, opened 2015-05-17
wfdb-10.5.24 (new formula)

avivrosenberg:
The WFDB software package (http://physionet.org/physiotools/).

WFDB is a software library for viewing, analyzing, and creating
recordings of physiologic signals. It provides uniform access to
digitized signals and annotations in a wide variety of formats,
including those in PhysioBank (http://physionet.org/physiobank/).

### DIFF
--- a/wfdb.rb
+++ b/wfdb.rb
@@ -1,0 +1,63 @@
+class Wfdb < Formula
+  desc "a software library for working with physiologic signals"
+  homepage "http://physionet.org/physiotools/"
+  url "https://github.com/bemoody/wfdb/archive/10.5.24.tar.gz"
+  sha256 "be3be34cd1c2c0eaaae56a9987de13f49a3c53bf1539ce7db58f885dc6e34b7b"
+
+  head do
+    url "https://github.com/bemoody/wfdb.git"
+  end
+
+  depends_on :arch => :intel
+
+  def install
+    ENV.deparallelize
+
+    if build.head?
+      # We need to set the package version manually, otherwise the configure script will prompt user for it...
+      # We'll take the version from the NEWS file:
+      news_version = %q(`head -1 ../NEWS | awk '{printf "wfdb-%s", $1}'`)
+      inreplace "configure", /^PACKAGE=`.*`$/, "PACKAGE=#{news_version}"
+    end
+
+    # Configure paths
+    system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
+
+    # Force compilation, prevent "install up to date"
+    system "sh", "-c", "echo '.PHONY: install' >> Makefile"
+
+    # Compile and install
+    system "make", "install"
+
+    # Install some example c-code that comes with the package
+    (share/"wfdb").install "examples"
+
+    # For some reason the configure script doesn't install the man pages properly
+    # even though '--mandir' is used. Manual fix:
+    share.install(prefix/"man") if File.exist?(prefix/"man")
+  end
+
+  def caveats; <<-EOS.undent
+    WFDB Example programs have been installed to:
+      #{share}/wfdb/examples
+    EOS
+  end
+
+  ###
+  # Compile and run a small test program from the examples
+  test do
+    # Use wfdb-config to get the location of headers and libs
+    cflags = `#{bin/"wfdb-config"} --cflags`.chomp
+    libs = `#{bin/"wfdb-config"} --libs`.chomp
+
+    # Compare them to Homebrew paths
+    assert_equal("-I#{include}", cflags, "wrong cflags from wfdb-config")
+    assert_equal("-L#{lib} -lwfdb", libs, "wrong libs from wfdb-config")
+
+    # Compile the test program
+    system ENV.cc, cflags, "-o", "wfdbversion", share/"wfdb/examples/wfdbversion.c", *libs.split
+
+    # Run the program
+    system "./wfdbversion"
+  end
+end


wfdb.rb
```diff
@@ -0,0 +1,70 @@
...
+      inreplace "conf/darwin-slib.def", / -arch ppc/, ""
+    end
+
+    # Configure paths
+    system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
+
+    # Needed, otherwise build fails on stable version
+    cd "lib"
+    system "make", "setup"
+    cd ".."
```
dpo: This should be
````
cd "lib" do
  system "make", "setup"
end
````

File: wfdb.rb [hunk 2]
```diff
@@ -0,0 +1,70 @@
...
+
+    # Compile and install
+    system "make", "install"
+
+    # Install some example c-code that comes with the package
+    (share/"wfdb").install "examples"
+
+    # For some reason the configure script doesn't install the man pages properly
+    # even though '--mandir' is used. Manual fix:
+    mv "#{prefix}/man", "#{man}" if File.exist? "#{prefix}/man"
```
dpo: Could it be because `#{prefix}/man` doesn't exist? Have you tried `man.mkdir`?

wfdb.rb
avivrosenberg: Thanks, now fixed.

wfdb.rb
avivrosenberg: Thank, I tried creating it before the `make install` part, but it still doesn't work - the man pages get installed to `$(brew --prefix)/Cellar/wfdb/<version>/man/` anyway...

Maybe this could be fixed by patching the configure script or one of the Makefiles, however that seems more complicated... And maybe the maintainer will fix this issue in the future.

avivrosenberg:
About the audit errors,
 - splitting the `system` commands so that it has several arguments produces different results, for some reason. It just doesn't work if I change it to what the audit suggests. This is true for both the `echo` and the compilation in the test block. I'm not sure what the problem is, but I'd love to know...
 - The non-executables in the `bin` directory are scripts that the `wfdb` package installed together with the binaries. I guess it expects them to be there.

avivrosenberg:
@dpo, any chance to merge this?
Please look at what I wrote about the audit errors. I'm not sure how (if at all) to fix, any suggestions are welcome.

Thanks.

wfdb.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  depends_on :arch => :intel
+
+  def install
+    ENV.deparallelize
+
+    if build.head?
+      # We need to set the package version manually, otherwise the configure script will prompt user for it...
+      # We'll take the version from the NEWS file:
```
tdsmith: If this is e.g. `10.5.23`, just use `version` (which is parsed from the URL) instead of parsing it from NEWS.

wfdb.rb
tdsmith: Do `share.install prefix/"man"` unconditionally instead. If it breaks in a future version, it will be obvious why, which is better than carrying dead code.

wfdb.rb
tdsmith: Just noticed `if build.head?` here, nevermind.

wfdb.rb
```diff
@@ -0,0 +1,70 @@
...
+    # Configure paths
+    system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
+
+    # Needed, otherwise build fails on stable version
+    cd "lib" do
+      system "make", "setup"
+    end
+
+    # This forces compilation, prevents "install up to date"
+    system "echo '.PHONY: install' >> Makefile"
```
tdsmith: `system "sh", "-c", "echo '.PHONY: install' >> Makefile"` will shut up the audit; I think that's a reasonable thing to do here.

wfdb.rb
```diff
@@ -0,0 +1,70 @@
...
+  test do
+    # Use wfdb-config to get the location of headers and libs
+    cflags = `#{bin/"wfdb-config"} --cflags`.chomp
+    libs = `#{bin/"wfdb-config"} --libs`.chomp
+
+    # Compile the program
+    system "#{ENV.cc} #{cflags} -o wfdbversion #{share}/wfdb/examples/wfdbversion.c #{libs}"
+
+    # Print output, should be a string with the WFDB version
+    puts `./wfdbversion`
```
tdsmith: Please run this with `system` so the error level is checked.

wfdb.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  ###
+  # Compile and run a small test program from the examples
+  test do
+    # Use wfdb-config to get the location of headers and libs
+    cflags = `#{bin/"wfdb-config"} --cflags`.chomp
+    libs = `#{bin/"wfdb-config"} --libs`.chomp
+
+    # Compile the program
+    system "#{ENV.cc} #{cflags} -o wfdbversion #{share}/wfdb/examples/wfdbversion.c #{libs}"
```
tdsmith: Ah, quoting this fails because #{libs} needs to be evaluated by the shell. Just use:

`system ENV.cc, "-I#{include}", "-o", "wfdbversion", share/"wfdb/examples/wfdbversion.c", "-L#{lib}", "-lwfdb"`

avivrosenberg:
@tdsmith thanks for your feedback!

I think I've fixed everything, except for the conditional man dir copying... If I don't condition it, the install fails with the HEAD version.

In the meantime the maintainer of wfdb released version 10.5.24, so this is now the stable version for the formula. This also allowed me to remove some of the hacks that were required to get the previous stable version to compile...

avivrosenberg:
Added a `desc` as suggested by the audit in the build server (my version of homebrew was outdated so it didn't complain about this).

tdsmith:
Looks like a spurious build failure on 10.8, kicking it off again...

tdsmith:
Opened #2420, which is this commit with some style edits and without installing the environment scripts, which aren't used by any of the binaries.

wfdb.rb
```diff
@@ -0,0 +1,63 @@
...
+      # We'll take the version from the NEWS file:
+      news_version = %q(`head -1 ../NEWS | awk '{printf "wfdb-%s", $1}'`)
+      inreplace "configure", /^PACKAGE=`.*`$/, "PACKAGE=#{news_version}"
+    end
+
+    # Configure paths
+    system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
+
+    # Force compilation, prevent "install up to date"
+    system "sh", "-c", "echo '.PHONY: install' >> Makefile"
```
tdsmith: Can you report this as a bug to the upstream maintainers?

tdsmith:
Thanks for the PR! Homebrew thrives because of contributions like yours.

PR #2275, opened 2015-05-18
OTPS 2 (new formula)

zimonkaizoku:
This is a new formula for the OSU Tidal Prediction Software
(OTPS2). The software can be used for tidal prediction and
is a recommended dependency for MB-System (mbsystem.rb).


otps.rb
```diff
@@ -0,0 +1,22 @@
+class Otps < Formula
+  homepage "http://volkov.oce.orst.edu/tides/tpxo8_atlas.html"
+  url "ftp://anonymous:anonymous%40homebrew.com@ftp.oce.orst.edu/dist/tides/OTPS2.tar.Z"
+  sha256 "c3e15679515e6e358d173eb9b46276c51ceb5bcd80ecb50822030e1d2900c361"
+
+  depends_on "gcc"
```
dpo: If this doesn't build with `clang`, then there should be a`fails_with` block instead of a dependency on `gcc`. Please grep other formulas for `fails_with`.

otps.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def install
+    system "make", "extract_HC"
+    system "make", "predict_tide"
+    system "make", "extract_local_model"
+
+    prefix.install Dir["*"]
```
dpo: Is there no `make install`?

otps.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    puts
+    puts "  Don\'t forget to install a tide model."
+    puts "  E.g. run 'brew home otps' to donwload the TPXO8-atlas model."
+    puts "  Each tide model needs to be linked into:"
+    puts "    /usr/local/Cellar/otps/2/DATA/Model_<model_name>"
+    puts
```
dpo: The messages should go in a `caveats` method.

zimonkaizoku:
Any news here? The mbsystem formula is not working anymore and I would like to add OTPS as a recommended dependency for the new mbsystem formula update. If it can not be merged I can link the dependency to an external tap, but if possible I would like to keep things in homebrew-science.
Thanks

dpo:
@BrewTestBot test this please

dpo:
```
==> audit problems
homebrew/science/otps:
 * 'DATA' was found, but no '__END__'
 * `option` (line 8) should be put before `depends_on` (line 6)
 * Formula should have a desc (Description).
```

PR #2284, opened 2015-05-20
Updated Alembic formula

fish2000:
* removed `require "formula"` line
* Added `:recommended` flags to hdf5 and openexr requirements,
  [per the upstream README](https://github.com/alembic/alembic/blob/e6c90d4faa32c4550adeaaf3f556dad4b73a92bb/README.txt#L13-L26)
* Updated `url` and `head` to point at current project repository
  (specifically, `url` points to a zipped source release instead of an SCM tag)
* Updated stable version to 1.5.8 and added sha256 hash
* Links to libc++
* [amend] Trimmed trailing whitespace

### DIFF
--- a/alembic.rb
+++ b/alembic.rb
@@ -1,18 +1,17 @@
-require "formula"
 
 class Alembic < Formula
   homepage "http://alembic.io"
-  head "https://code.google.com/p/alembic/", :using => :hg
-  url "https://code.google.com/p/alembic/", :using => :hg,
-    :tag => "1_05_04"
-  version "1.5.4"
+  url "https://github.com/alembic/alembic/archive/1.5.8.zip"
+  sha256 "1f347a0e310d096370fe2afd3a83674e9a00776f1ab74ec601fddbf8657ed568"
+  version "1.5.8"
+  head "https://github.com/alembic/alembic.git", :using => :git
 
   needs :cxx11
-
-  depends_on "cmake" => :build
-  depends_on "boost"
-  depends_on "hdf5"
+  depends_on "cmake"    => :build
+  depends_on "hdf5"     => :recommended
+  depends_on "openexr"  => :recommended
   depends_on "ilmbase"
+  depends_on "boost"
 
   def install
     cmake_args = std_cmake_args + %W[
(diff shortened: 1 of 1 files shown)


alembic.rb
```diff
@@ -1,19 +1,18 @@
...
-  url "https://code.google.com/p/alembic/", :using => :hg,
-    :tag => "1_05_04"
-  version "1.5.4"
-
+  url "https://github.com/alembic/alembic/archive/1.5.8.zip"
+  sha256 "1f347a0e310d096370fe2afd3a83674e9a00776f1ab74ec601fddbf8657ed568"
+  version "1.5.8"
+  head "https://github.com/alembic/alembic.git", :using => :git
+  
   needs :cxx11
```
dpo: Please remove trailing whitespace (see also line 27).

alembic.rb
```diff
@@ -21,12 +20,12 @@ def install
       -DUSE_ARNOLD=OFF
       -DUSE_MAYA=OFF
       -DUSE_PYALEMBIC=OFF
-      -DCMAKE_CXX_FLAGS='-std=c++11'
+      -DCMAKE_CXX_FLAGS='-std=c++11\ -stdlib=libc++'
     ]
```
dpo: Isn't this ensured by `ENV.cxx11`?

fish2000: I thought it would be, too – but it appeared not to be. Alembic’s CMake infrastructure is very heavy-duty and I couldn’t find an obvious spot wherein those flags were being clobbered from within – and the original formula author had included the explicit `-DCMAKE_CXX_FLAGS` argument to deal with it, presumably.

That’s why that flag is there – I myself would be personally thrilled to either have it somewhere less ad-hoc-ish, or nix it altogether, if someone in the know w/r/t the Alembic build system could point me in the right direction.

alembic.rb
fish2000: No problem – will update and squash when I get back to the terminal (~15min), thanks for pointing that out.

dpo:
@Brewtetest this please

dpo:
@BrewTestBot test this please

dpo:
```
==> Downloading https://github.com/alembic/alembic/archive/1.5.8.zip
==> Downloading from https://codeload.github.com/alembic/alembic/zip/1.5.8
Warning: Formula reports different SHA256: 1f347a0e310d096370fe2afd3a83674e9a00776f1ab74ec601fddbf8657ed568
Downloaded to: /Library/Caches/Homebrew/alembic-1.5.8.zip
SHA1: 0c5f5ca0f490c467c367288237e861fb36cae7d3
SHA256: e09ebe3d038c8f0867a4d83dcc6c29d577382567965db7f8ffe7c4205b028ca2
```

PR #2295, opened 2015-05-21
Bump revisions for new gcc

### LINKED ISSUE #2286: R 3.2.0: libgfortran.3.dylib: image not found
After upgrading `gcc` to 5.1.0 I got the following error from R:

```
❯❯❯ R
dyld: Library not loaded: /usr/local/lib/gcc/4.9/libgfortran.3.dylib
  Referenced from: /usr/local/Cellar/r/3.2.0/R.framework/Versions/3.2/Resources/lib/libR.dylib
  Reason: image not found
[1]    85177 trace trap  R
```

Installing `gcc49` should work around this issue, but `gcc49` is built without fortran by default. `brew install gcc49 --with-fortran` should work around this issue—I'm testing this now—but requires building `gcc49` from source.

@mikemcquaid What's the recommended way of rebuilding the R bottle with `gcc` 5.1.0? I'm guessing that we're going to need to do this for a lot of formulae.

iml:
Fixes #2286.

This will take a while.


iml:
Need to split this up due to the 2hr limit.

PR #2301, opened 2015-05-22
Bump revisions for new gcc (part 3)

iml:


dpo:
`slepc` depends on `petsc`. Will it build against the `petsc` you're submitting in this PR? I was thinking the first batch almost has to be one formula at a time, but I may be pessimistic.

iml:
It should, unless something has changed, and depending on the order. We'll see how it goes.

PR #2304, opened 2015-05-22
blast 2.2.30: Remove include and lib

sjackman:
Libraries and headers conflict with ncbi-c++-toolkit
libproj.a conflicts with proj


iml:
I feel like this renders the dll and static options obsolete if we're just keeping the binary. Might as well always build a static binary.

sjackman:
Good point. Removing `lib` breaks `--without-static`. Hmmm. I guess those shared libraries should be moved to `libexec` and the binaries linked against that if we wan't to keep the `--without-static` functionality. Can fixing up the library locations stored in the executables be done after building using `install_name_tool` and does Homebrew have any tools to make that job easier?

sjackman:
The build failure on Mountain Lion appears to be a `brew doctor` issue unrelated to `blast`.

http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/2012/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/doctor/
```
Warning: Some installed formula are missing dependencies.
You should `brew install` the missing dependencies:

    brew install czmq fontforge glibmm libsigc++ py2cairo pygobject zeromq
```

sjackman:
```
❯❯❯ du -sh with-static without-static
636M	with-static
113M	without-static
```

sjackman:
Here's an example of using `install_name_tool` in [tmv-cpp](https://github.com/Homebrew/homebrew-science/blob/master/tmv-cpp.rb#L29).

sjackman:
[`Keg::fix_install_names`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/keg_fix_install_names.rb#L5) does something like what we'd want to do to BLAST. Is this worth the effort, or should we just remove `--without-static`?

sjackman:
Never mind my yammering. The simple solution of `configure --libdir=#{libexec}` works just fine.

PR #2311, opened 2015-05-25
Gtk+ goes quartz only

tschoonj:
This branch continues my work in Homebrew/homebrew#39868 regarding removing X11 support in Gtk+ and its dependencies cairo and pango in favor of quartz.

Unless I am wrong (please correct me if so), there are only 2 packages in homebrew-science with a mandatory or recommended dependency on gtk+:

- [x] osgearth
- [x] pspp

This branch should not be merged in until:

1. Homebrew/homebrew#39868 is merged in
2. the bottles in this branch have been rebuilt against those that were created for Homebrew/homebrew#39868

### DIFF
--- a/osgearth.rb
+++ b/osgearth.rb
@@ -1,5 +1,6 @@
 class Osgearth < Formula
   homepage "http://osgearth.org"
+  desc "a geospatial SDK and terrain engine for OpenSceneGraph applications."
   url "https://github.com/gwaldron/osgearth/archive/osgearth-2.6.tar.gz"
   sha256 "965c93837520ab9538038843ff83ee1903548f8be191ee211e40abb0e6c1bb4a"
 
@@ -18,6 +19,7 @@ class Osgearth < Formula
   depends_on "minizip" => :recommended
   depends_on "v8" => :optional
   depends_on "tinyxml" => :optional
+  depends_on MinimumMacOSRequirement => :mavericks
 
   resource "sphinx" do
     url "https://pypi.python.org/packages/source/S/Sphinx/Sphinx-1.2.1.tar.gz"
--- a/osgearth.rb
+++ b/osgearth.rb
@@ -24,12 +24,6 @@ class Osgearth < Formula
     sha256 "182e5c81c3250e1752e744b6a35af4ef680bb6251276b49ef7d17f1d25e9ce70"
   end
 
-  # all merged upstream, remove on next version
-  # find a v8 lib: https://github.com/gwaldron/osgearth/pull/434
-  # find JavaScriptCore lib: https://github.com/gwaldron/osgearth/pull/435
-  # find libnoise lib: https://github.com/gwaldron/osgearth/pull/436
-  patch :DATA
-
   def install
     if (build.with? "docs-examples") && (!which("sphinx-build"))
       # temporarily vendor a local sphinx install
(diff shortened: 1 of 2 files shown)


tschoonj:
Could anyone comment on these build errors I am getting? I guess the following are due to some homebrew bug?

````
        ==> Determining osgearth bottle revision...
Error: undefined method `>' for nil:NilClass
Please report this bug:
/usr/local/Library/Homebrew/cmd/bottle.rb:134:in `bottle_formula'
/usr/local/Library/Homebrew/cmd/bottle.rb:292:in `block in bottle'
/usr/local/Library/Homebrew/cmd/bottle.rb:291:in `each'
/usr/local/Library/Homebrew/cmd/bottle.rb:291:in `bottle'
/usr/local/Library/brew.rb:140:in `<main>'
````

The ML build issues for osgearth are more interesting though: looks like some linking error. Perhaps the libraries were not provided in the right order? This is the (patched) latest version of osgearth by the way, which was not yet updated on Homebrew.

**Comment by tschoonj** [comment]
@mikemcquaid 

Could you have a look at this please? Thanks!

**Comment by mikemcquaid** [comment]
@tschoonj Have retriggered the job as this was a bug yesterday.

**Comment by tschoonj** [comment]
Thanks but the Mountain Lion build error remains as I expected :-(

Can you think of any reason why the ML linker would do this differently compared to Mavericks and Yosemite?

**Comment by rleigh-dundee** [comment]
@tschoonj ML uses clang++ but with GCC libstdc++ for some sort of transitional reasons, and I've encountered numerous issues with it which you don't get with pure GCC/libstdc++ or clang++/libc++.  It's fundamentally broken in many ways, and I ended up dropping support for it in my projects due to it being such a mess.  You might well be tripping up on one of these defects.

**Comment by tschoonj** [comment]
Although I have seen my share of libc++ vs libstdc++ madness this one looks a bit different to me.
The undefined references appear to be related to functions in `libosg.dylib` and other libraries which are definitely put in the linker command.

File: osgearth.rb [hunk 1]
```diff
@@ -23,23 +21,21 @@ class Osgearth < Formula
...
-    sha1 "448cdb89d96c85993e01fe793ce7786494cbcda7"
+    sha256 "182e5c81c3250e1752e744b6a35af4ef680bb6251276b49ef7d17f1d25e9ce70"
   end
 
-  # all merged upstream, remove on next version
-  # find a v8 lib: https://github.com/gwaldron/osgearth/pull/434
-  # find JavaScriptCore lib: https://github.com/gwaldron/osgearth/pull/435
-  # find libnoise lib: https://github.com/gwaldron/osgearth/pull/436
+  # fixes support for latest osg release 3.3.3
+  # based on https://github.com/gwaldron/osgearth/commit/48264aebc7b01f46de05c5a63b8466aafbda5098
```
mikemcquaid: So is this already upstream and just waiting for a new release?

mikemcquaid:
@tschoonj Worst case can always restrict this to 10.9 and above.

osgearth.rb
tschoonj: That's it basically indeed.

I got the information from here: https://github.com/gwaldron/osgearth/issues/588
I had to modify the diff a little bit to get it working

tschoonj:
How do I implement such a version restriction? Haven't seen one in a Formula so far...

rleigh-dundee:
@tschoonj try

```
depends_on MinimumMacOSRequirement => :mavericks
```

Needed this for the first time today by complete coincidence and I didn't see any current usage either, but it does appear to work!  https://github.com/Homebrew/homebrew-science/pull/2326/files#diff-0cdaeb4fa14b9af05dd33275e4359ff1R24

tschoonj:
@rleigh-dundee Thanks for the trick!

Build works fine now, as expected. Will mark the formula as done for now.

I am quite busy the next two weeks but afterwards I will continue with 'quartzing' homebrew-x11...

mikemcquaid:
@BrewTestBot test this please

mikemcquaid:
@BrewTestBot test this please

tschoonj:
@mikemcquaid Could you also close the one in homebrew-games? No rebuild is necessary there since all formulas could be migrated to gtk3. Thanks!

mikemcquaid:
@BrewTestBot test this please

PR #2313, opened 2015-05-26
clustal-omega 1.2.1

gecgooden:


sjackman:
Thanks for your contribution, George.

PR #2323, opened 2015-05-28
xraylib formula fixes

tschoonj:
python bindings code was wrong
audit --strict compliance


tschoonj:
Thanks for rebuilding the package, it can be pulled in now

PR #2326, opened 2015-05-28
bioformats-cpp 5.1.2 (new formula)

rleigh-dundee:
The Bio-Formats C++ libraries and tools for reading and writing microscopy imaging file formats. It currently supports reading and writing of Baseline TIFF and OME-TIFF formats, with support for metadata using the OME-XML data model.


tseemann:

rleigh-dundee:
Is this OK to merge, or are there any changes you want making?

bioformats-cpp.rb
```diff
@@ -0,0 +1,95 @@
...
+  option "with-qt5", "Build with Qt5 (used for OpenGL image rendering)"
+  option "without-docs", "Build API reference and manual pages"
+
+  depends_on "boost"
+  depends_on "cmake" => :build
+  depends_on "libpng"
+  depends_on "libtiff"
+  depends_on "xerces-c"
+  depends_on "graphicsmagick" => :optional if build.with? "check"
+  depends_on "qt5" => :optional if build.with? "qt5"
```
tdsmith: The `if` at the end is redundant and should be removed.

bioformats-cpp.rb
```diff
@@ -0,0 +1,95 @@
...
+  def install
+    ENV.prepend_create_path "PYTHONPATH", buildpath/"python/lib/python2.7/site-packages"
+    ENV.prepend_path "PATH", buildpath/"python/bin"
+    ENV.prepend_create_path "CMAKE_PROGRAM_PATH", buildpath/"python/bin"
+    resources.each do |r|
+      r.stage do
+        system "python", *Language::Python.setup_install_args(buildpath/"python")
+      end
+    end
+    ENV.prepend_path "PATH", "#{HOMEBREW_PREFIX}/opt/qt5/bin" if build.with? "qt5"
```
tdsmith: superenv should manage this automatically; did you find that this was necessary on e.g. Linux?

bioformats-cpp.rb
```diff
@@ -0,0 +1,95 @@
...
+  option "without-check", "Skip build time tests (not recommended)"
+  option "with-qt5", "Build with Qt5 (used for OpenGL image rendering)"
+  option "without-docs", "Build API reference and manual pages"
+
+  depends_on "boost"
+  depends_on "cmake" => :build
+  depends_on "libpng"
+  depends_on "libtiff"
+  depends_on "xerces-c"
+  depends_on "graphicsmagick" => :optional if build.with? "check"
```
tdsmith: Even if graphicsmagick is only needed for --with-check, let's just include it unconditionally since class-level conditionals are problematic and with-check is the default.

bioformats-cpp.rb
```diff
@@ -0,0 +1,95 @@
...
+  depends_on "qt5" => :optional if build.with? "qt5"
+  depends_on "glm" => :build if build.with? "qt5"
+  depends_on "doxygen" => :build if build.with? "docs"
+  depends_on "graphviz" => :build if build.with? "docs"
+
+  # Needs clang/libc++ toolchain; mountain lion is too broken
+  depends_on MinimumMacOSRequirement => :mavericks
```
tdsmith: Does adding `ENV.cxx11` to the head of `install` allow this to build reasonably on 10.8? If this is a decision bioformats made, I'm fine with it.

bioformats-cpp.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", buildpath/"python/lib/python2.7/site-packages"
+    ENV.prepend_path "PATH", buildpath/"python/bin"
+    ENV.prepend_create_path "CMAKE_PROGRAM_PATH", buildpath/"python/bin"
+    resources.each do |r|
+      r.stage do
+        system "python", *Language::Python.setup_install_args(buildpath/"python")
+      end
+    end
```
tdsmith: Should this block be conditional on `build.with? "docs"`?

bioformats-cpp.rb
```diff
@@ -0,0 +1,95 @@
...
+        system "ctest", "-V"
+      end
+
+      system "make", "install"
+    end
+  end
+
+  test do
+    system "bf-test", "--usage"
+    system "bf-test", "info", "--usage"
```
tdsmith: If you could load an image file, that would be a good test. We have some small fixtures; you can grep through formulae for "fixture" to see how to use them.

tdsmith:
Looking good otherwise; sorry for the delay!

bioformats-cpp.rb
rleigh-dundee: Not really.  I think the `:cxx11` is redundant, so I'll remove it.  The problem with 10.8 is the hybrid clang++/libstdc++: there are some fundamental parts of the standard library which are broken with 10.8; I was manually working around this and having our CI system catch any regressions, but in the end it just got too painful--we hit a problem without a reasonable workaround, so at that point we dropped 10.8 due to the ongoing maintenance burden.

bioformats-cpp.rb
rleigh-dundee: Unfortunately, while we can do such tests as part of the unit tests during the build, we don't have any suitable image files available post-install.  I'll look at installing our samples for the next point release.

bioformats-cpp.rb
rleigh-dundee: I found it necessary on MacOSX outside the context of a homebrew formula; removing here since it is indeed unnecessary.

rleigh-dundee:
@tdsmith I've pushed an updated formula which should address all your comments.

rleigh-dundee:

rleigh-dundee:
Hmm, the jenkins jobs failed, but due to a git error.  Not sure what I can do to correct this (if anything).  I tried pushing an updated commit again to see if it was a one-off failure, but it failed in the same way again.

bioformats-cpp.rb
tdsmith: Okay! (FWIW: `ENV.cxx11` adds `-std=c++11 -stdlib=libc++` to compiler invocations.)

tdsmith:
Kicked off a new build.

bioformats-cpp.rb
rleigh-dundee: At the moment, we support both C++98 and C++11, so there's no strict requirement for C++11 but we can use a subset of C++11 features if available, otherwise falling back to their Boost equivalents.  It's probably a bit more flexible for end users to use C++98 at the moment, rather than forcing the newer standard; but I fully expect we'll move to C++11 at some future point (basically when the minimum version of all the compilers we support allows use of C++11 features).

tdsmith:
Looks good; fixed the audit warning myself. Thanks for your contribution and your patience!

PR #2328, opened 2015-05-29
bam-readcount 0.7.4

saketkc:


saketkc:
I am not sure why the test fails on `mavericks` and `yosemite`.
It passes on `linux` too.

PR #2332, opened 2015-05-30
abyss 1.9.0

sjackman:
Add `needs :openmp`


tseemann:
```
configure: error: ABySS should be compiled with Google sparsehash to
	reduce memory usage. It may be downloaded here:
	http://code.google.com/p/google-sparsehash

	If you do not wish to use sparsehash, specify --without-sparsehash.
```

sjackman:
Oddly it succeeds on Mountain Lion but fails on Mavericks and Yosemite.

sjackman:
http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/2070/version=yosemite/console
```
checking google/sparse_hash_map usability... no
checking google/sparse_hash_map presence... yes
configure: WARNING: google/sparse_hash_map: present but cannot be compiled
configure: WARNING: google/sparse_hash_map:     check for missing prerequisite headers?
configure: WARNING: google/sparse_hash_map: see the Autoconf documentation
configure: WARNING: google/sparse_hash_map:     section "Present But Cannot Be Compiled"
configure: WARNING: google/sparse_hash_map: proceeding with the compiler's result
configure: WARNING:     ## ----------------------------------- ##
configure: WARNING:     ## Report this to abyss-users@bcgsc.ca ##
configure: WARNING:     ## ----------------------------------- ##
checking for google/sparse_hash_map... no
```

sjackman:
Ah, it's because `google-sparsehash` is compiled with clang, but `abyss` is being compiled with `gcc` because I added `needs :openmp`. The solution is probably to add `google-sparsehash` as a resource to `abyss`.

sjackman:
Work around is
```sh
brew install google-sparsehash abyss --cc=gcc-5
```

sjackman:
For now I'm going to make `--without-openmp` the default behaviour, which was the case with the previous version of this formula, and add the option `--with-openmp`.

sjackman:
@mikemcquaid The BrewTestBot is getting a `401 Unauthorized` error when uploading to BinTray. Any idea why?

http://bot.brew.sh/job/Homebrew%20Bottles/23929/console
```
curl -I --fail --output /dev/null https://homebrew.bintray.com/bottles-science/abyss-1.9.0.mavericks.bottle.tar.gz
curl: (22) The requested URL returned error: 401 Unauthorized
```

sjackman:
I've tried this a handful of times and get the same 401 error. http://bot.brew.sh/job/Homebrew%20Bottles/23935/console

sjackman:
Ah, it looks like a previous build that succeeded only on Mountain Lion was uploaded to [Bintray](https://bintray.com/homebrew/bottles-science/abyss/1.9.0/view#files), and it appears was also published, for some reason. I suspect that's the source of the BrewTestBot error. I'll delete the bottle and try the build again.

xu-cheng:
> I've tried this a handful of times and get the same 401 error.

Sorry, my fault to fail to silent this output. It should have 401 error which indicates there isn't an already published bottle.

> and it appears was also published, for some reason.

We used to auto publish bottles. But due to security reason, this has been changed. Now, bottles won't be auto published and overwriting published bottles are also forbidden.

sjackman:
**To do**: `google-sparsehash` needs to bundled into `abyss` as a resource to be able to add `needs :openmp` and compile with `gcc`.

PR #2333, opened 2015-05-30
meme: Enable parallel, add doi,tag,desc

saketkc:


saketkc:
@iml Any clue, what's going wrong here?

iml:
I don't know about the specific error but Open MPI and OpenMP are two different things. The former is supported through the `:mpi` dependency, with `open-mpi` being the default formula.

saketkc:
@mikemcquaid There seems to be an issue with the `bot` specific to `homebrew-science` PRs.

mikemcquaid:
Just trigger a rebuild if you get that error.

saketkc:
Thanks!
Looks like it's been happening more often today.

mikemcquaid:
We have no control over it, unfortunately. If you can work out how to fix it in the upstream Jenkins Git plugin that'd be good. I suspect it may be related to force-pushing.

saketkc:
Looks like a known issue, with a possible fix?
https://github.com/janinko/ghprb/issues/148#issuecomment-110622022

mikemcquaid:
@saketkc Yeh, I'm not sure if that fix is user-configurable. If you know what settings I should change I'm all ears.

saketkc:
Not really. Not even a Jenkins apprentice.

meme.rb
```diff
@@ -16,10 +20,15 @@ class Meme < Formula
     with other packages.
   EOF
 
+  option "without-openmpi", "Disable parallel support"
+  depends_on "open-mpi" => [:cc] if build.with? "openmpi"
```
iml: The `:cc` doesn't apply here. Remove the option and use `depends_on :mpi => [:recommended, :cc]`, which should set the appropriate environment and make mpidir below unnecesssary. There are other MPI implementations, so the `:mpi` dependency is supposed to handle chosing between them.

saketkc:
@iml Can you trigger a build?
Giving up on force push.

tdsmith:
Triggered a rebuild.

PR #2334, opened 2015-05-30
Simulate-PCR 1.2 (new formula)

tseemann:
Assessing primer specificity and predicting both desired and off-target amplification products is an essential step for robust PCR assay design. This script predicts potential polymerase chain reaction (PCR) amplicons in a large sequence database such as NCBI nt from either singleplex or a large multiplexed set of primers, allowing degenerate primer and probe bases, with target mismatch tolerance and amplicon length range to be set by the user. The PCR amplicon simulation code also annotates amplicons with gene information automatically downloaded from NCBI, and optionally it can predict whether there are also TaqMan/Luminex probe matches within predicted amplicons. It is an open source command line Perl script called simulate_PCR.pl that calls the BLAST (Altschul, et al., 1990) programs makeblastdb, blastn, and blastdbcmd, and the NCBI efetch utility (http://www.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi ).


sjackman:
> that calls the … the NCBI efetch utility

Should `edirect` be an optional or recommended dependency?

tseemann:
Added optional dependency for 'edirect'.

simulate-pcr.rb
```diff
@@ -0,0 +1,22 @@
...
+  depends_on "blast"
+  depends_on "Bio::Perl" => :perl
+  depends_on "edirect" => :optional
+
+  def install
+    bin.install "simulate_PCR"
+  end
+
+  test do
+    assert_match "amplicon", shell_output("simulate_PCR 2>&1", 255)
```
sjackman: This should be `#{bin}/simulate_PCR`. I've fixed it up.

sjackman:
Thanks, Torsten.

PR #2335, opened 2015-05-30
Fix Python support in lp_solve.py

jaredwhite128:
Moved --no-user-cfg argument to setup.py's global scope, since it was formerly being passed to the "install" subcommand.


tdsmith:
Thanks!

PR #2336, opened 2015-05-31
openblas: bottle dynamic arch

tdsmith:
openblas supports optimizing for several architectures at compile time
and dynamically selecting the right arch at runtime, which should give
better performance than optimizing for core2.


tdsmith:
FWIW: openblas.dylib's install_name points to the architecture-specific library, like `/usr/local/opt/openblas/lib/libopenblasp-r0.2.14.dylib`. It could break things if people reinstall this bottled openblas after installing something --with-openblas, but that'll also happen at minor version changes.

Issue #2338, opened 2015-05-31
sratoolkit error: /lib/libbz2.a(blocksort.o): could not read symbols: Bad value


tseemann:
I tried with gcc 5.1 and 4.8 but same problem. 

Here's the gist-logs: https://gist.github.com/93f9d4a53cfa4ae4209f

```
gcc -shared -o /bio/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/linux/gcc/x86_64/rel/lib/libncbi-vdb.so.2.5.0 -Wl,-soname,libncbi-vdb.so.
2 -L/bio/linuxbrew/lib -Wl,-rpath,/bio/linuxbrew/lib -DNDEBUG  -m64  -DNDEBUG  -m64   -L/bio/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/
linux/gcc/x86_64/rel/lib -L/bio/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/linux/gcc/x86_64/rel/ilib -Wl,-Bstatic -Wl,-whole-archive -la
lign-reader -lascp -laxf -lsraxf -lwgsxf -lvxf -lsrareader -lsradb -lvdb-blast -lvdb -lkdb -lalign-access -lncbi-bam -lvfs -lkxml -lkxfs
 -lkns -lkfg -lkrypto -lksrch -lkfs -lklib -lkproc -lkq -lkfc -lbz2 -lz -lngs-jni -lngs-py -lngs -Wl,-no-whole-archive -Wl,-Bdynamic -ld
l -lpthread -lxml2 -lm
/usr/bin/ld: /bio/linuxbrew/lib/libbz2.a(blocksort.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a sha
red object; recompile with -fPIC
/bio/linuxbrew/lib/libbz2.a(blocksort.o): could not read symbols: Bad value
collect2: error: ld returned 1 exit status
make[3]: *** [/bio/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/linux/gcc/x86_64/rel/lib/libncbi-vdb.so] Error 1
make[3]: Leaving directory `/bio/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/linux/gcc/x86_64/rel/obj/libs/ncbi-vdb'
make[2]: *** [std] Error 2
[...]

PR #2341, opened 2015-06-01
hmmer2: sha256, desc, +pthreads, compile bug fix with opt/lib/

tseemann:


hmmer2.rb
```diff
@@ -20,12 +21,15 @@ def install
     # Fix "make: Nothing to be done for `install'."
     rm "INSTALL"
 
-    system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
+    # Fix "evd_test.c:(.text.startup+0x79c): undefined reference to `FreeHistogram'"
+    ENV["LDFLAGS"]=""
```
tdsmith: Do you know what flag is causing trouble? I'm worried this is masking some deeper problem.

tseemann: I think you are right. And I think it might be Linux specific. See #2338 too. I did a recent fresh brew install on weekend from scratch and these are cropping up.

PR #2343, opened 2015-06-01
sratoolkit: fix compile fail on Linux - Issue #2338

tseemann:


tseemann:

sratoolkit.rb
```diff
@@ -32,6 +33,11 @@ class Sratoolkit < Formula
 
   def install
     ENV.deparallelize
+
+    # Fix: libbz2.a(blocksort.o): relocation R_X86_64_32 against `.rodata.str1.1'
+    # https://github.com/Homebrew/homebrew-science/issues/2338
+    ENV["LDFLAGS"]="" if OS.linux?
```
sjackman: I'm not a fan of a whole-sale clearing of `LDFLAGS`. The one offending flag should be removed.

sjackman:
@tseemann Can you report the output of
```sh
ls `brew --prefix bzip2`/lib
```

It looks as though the issue is that it's compiling PIC (position-independent code for a shared object) objects against the static version of `-lbzip2`, which is a no-no. It should preferentially link against `libbz2.so` rather than `libbz2.a`.

tseemann:
Here you go @sjackman 
```
% ls -l `brew --prefix bzip2`/lib
-r--r--r--. 1 tseemann domain^users 358330 May 31 10:11 libbz2.a
lrwxrwxrwx. 1 tseemann domain^users     15 May 31 10:11 libbz2.so -> libbz2.so.1.0.6
lrwxrwxrwx. 1 tseemann domain^users     15 May 31 10:11 libbz2.so.1 -> libbz2.so.1.0.6
lrwxrwxrwx. 1 tseemann domain^users     15 May 31 10:11 libbz2.so.1.0 -> libbz2.so.1.0.6
-r--r--r--. 1 tseemann domain^users 241535 May 31 10:11 libbz2.so.1.0.6
```

sjackman:
It was just a sanity check. Looks fine to me. sratoolkit is being dumb.

sjackman:
It fails for me for a different reason.
https://gist.github.com/sjackman/9756930a4b386b7a3f70#file-06-make-L1094
```
gcc -o /gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/linux/gcc/x86_64/rel/ilib/sra-dflt-schema -L/gsc/btl/linuxbrew/lib -Wl,--dynamic-linker=/gsc/btl/linuxbrew/opt/glibc/lib/ld-linux-x86-64.so.2 -Wl,-rpath,/gsc/btl/linuxbrew/lib -DNDEBUG -m64 -DNDEBUG -m64 sra-dflt-schema.o -L/gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/linux/gcc/x86_64/rel/lib -L/gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.0/ncbi-vdb/linux/gcc/x86_64/rel/ilib -Wl,-Bstatic -lkapp-norsrc -lvdb -laxf -lsraxf -lwgsxf -lvxf -lalign-reader -lkdb -lvfs -lkns -lksrch -lkrypto -lkfg -lkfs -lksproc -lklib -ljudy -Wl,-Bdynamic -ldl -Wl,-Bstatic -lz -lbz2 -Wl,-Bdynamic -lpthread -lm
sra-dflt-schema.o: In function `KMain':
sra-dflt-schema.c:(.text+0x49f): undefined reference to `KDirectoryNativeDir_v1'
sra-dflt-schema.c:(.text+0x6bb): undefined reference to `KDirectoryRelease_v1'
```

tseemann:
Be good if they could fix it at the NCBI Hackathon!
http://ncbiinsights.ncbi.nlm.nih.gov/2015/05/22/ncbis-first-hackathon-advanced-bioinformatic-analysis-of-next-gen-sequencing-data/

tseemann:
@sjackman The new sratoolkit 2.5.2 formula seems to build on Linux for me now. You too?

sjackman:
I get this strange error message:
```
gcc -shared -o /gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.2/ncbi-vdb/linux/gcc/x86_64/rel/lib/libkdf5.so.2.5.2 -Wl,-soname,libkdf5.so.2  -DNDEBUG  -m64  -DNDEBUG  -m64  hdf5dir.pic.o hdf5file.pic.o hdf5arrayfile.pic.o -L/gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.2/ncbi-vdb/linux/gcc/x86_64/rel/lib -L/gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.2/ncbi-vdb/linux/gcc/x86_64/rel/ilib -lhdf5 -Wl,-Bstatic -Wl,-whole-archive -lz -lbz2 -Wl,-no-whole-archive -Wl,-Bdynamic -ldl -lpthread -lm
hdf5file.pic.o: In function `strtod':
hdf5file.c:(.text+0x5a0): multiple definition of `strtod'
hdf5dir.pic.o:hdf5dir.c:(.text+0x7e0): first defined here
hdf5file.pic.o: In function `strtol':
hdf5file.c:(.text+0x5b0): multiple definition of `strtol'
hdf5dir.pic.o:hdf5dir.c:(.text+0x7f0): first defined here
```
https://gist.github.com/sjackman/6e716c5a41cfa8a3da44

kwrodarmer:
It's recommended that you use the libraries designated by the sratoolkit rather than trying to override them. Over the years we have had to deal with many users with disparate systems and have arrived at linking in zlib and bz2 statically, since we cannot depend upon them in all cases to be available externally.

As to how we build the sources, they have to be PIC because we use the static libraries to build one or more dynamic libraries. This is another reason that we build our own zlib/bz2. If you are trying to override our library selections, that may be a no-no.

sjackman:
Ah, yes. That's the problem. Thanks for the explanation. You've statically linked `glibc` into your copies of `libz` and `libbz2`, so you're resulting executable will have multiple copies of `glibc` in it. There's some definite downsides to that approach. For one, having multiple copies of `glibc` will increase the footprint of your executable code, and you'll get worse instruction cache (icache) performance. Would you consider providing an option to use the system versions of `libz` and `libbz2`. You can keep the current behaviour as default, if you like.

```
-Wl,-Bstatic
-Wl,-whole-archive
-lz
-lbz2
-Wl,-no-whole-archive
-Wl,-Bdynamic
```

Could you elaborate on what issue you've had linking against the system `libz` and `libbz2`? Working with Linuxbrew, I've seen all sorts of crazy library issues, but I've never had a problem with these two particular libraries.

kwrodarmer:
Our mix has evolved over time, and yes part of the problem has been to deal with Windows peculiarities, especially since we used to build Windows code from Linux make...

We are very well aware of the benefits of shared libraries, and had originally designed the VDB system as a collection of several. It's true that use of shared libraries will decrease the footprint, but in this case by relatively small amounts by today's standards (libz.a is 145K while libbz2.a is 88K). The icache impact might be an issue if VDB were calling into other (shared) libraries that used a different copy of the code than we do, but since we do not use other libraries in any time critical code, the only zlib instructions in cache will be our own. There is a slight increase in physical RAM pressure, but this would be noticed mainly if several different toolkit tools were operating at once.

There are downsides to shared libraries, too. For example there can be subtle changes in ABI. There should not be, but it happens. There's also the issue that when we test against a particular version of a 3rd party library, there's a risk that an updated version can invalidate our tests and something will break. Shouldn't happen, but again it's a risk. To avoid the latter risk I can link against a particular version, but then that creates the potential for multiple versions of the shared library being mapped at once.

The problems we had were learned over years of finding people trying to build from source on systems that had inadequate or outdated development systems. Often, they were missing support for a library entirely, or else they had the shared libraries and no static libraries or headers. We just had an issue related to a CentOS system with a missing or inadequate `<limits.h>`. On Windows the problem was worse for 32-bit builds due to multiple calling conventions (__especially__ for libz). We needed to ensure that libz/libbz2 were present, because not all users have an adequate development environment.

We are open to adapting the build and welcome your suggestions for changes. We only ask that you keep in mind that we are focusing more on developing new technologies that are pretty cool, so if there are workarounds you can employ in the meantime, it'll help us get there!

sjackman:
> It's true that use of shared libraries will decrease the footprint, but in this case by relatively small amounts by today's standards (libz.a is 145K while libbz2.a is 88K).

It's not so much the footprint of `libz` and `libbz2` alone, but you're also linking in multiple copies of `glibc`!

I'd appreciate a configure option that disables the use of `-Wl,-Bstatic -Wl,-whole-archive` and `-Wl,-no-whole-archive -Wl,-Bdynamic`.

kwrodarmer:
Notice that `-Bstatic` is enabled for the list of libraries that follow and that `-Bdynamic` returns the linkage to dynamic, which is applied for default gcc linkage such as glibc. The very twisted linking rules were used to build executables and libraries that bring in some libraries statically while leaving others. We don't link statically against glibc.

We can give you an option to override any static linkage, but again please give us some time, and we'll also have to reevaluate the impact.

sjackman:
Sorry. My previous assessment was a red herring. I believe the issue is that you're explicitly adding `-I /usr/include` to the search path. That's a bad idea. `/usr/include` is in the default search path of the compiler. If it's not, there's probably a good reason. Linuxbrew for example doesn't use the system's compiler (`/usr/bin/gcc`), which is often out of date on HPC systems, but uses its own compiler (GCC 5.1) and glibc. The result is that it's using `/usr/include/features.h` when it should be using `~/.linuxbrew/include/features.h`.

sjackman:
It looks as though `sratoolkit` is ignoring the environment variable `CPPFLAGS`. Can you confirm that's the case?

kwrodarmer:
Can you show me a case where we're adding `-I/usr/include` to the search path?

kwrodarmer:
Yes, it's true - we ignore CPPFLAGS.

I have a suggestion - please write to us and tell us what it is that you're trying to modify. I'm not sure everything is open to modification. There are cases where we tune compiler and linker behavior.

sjackman:
`-I/usr/include` is added to the command line when compiling `hdf5arrayfile.pic.o`

https://gist.githubusercontent.com/sjackman/6e716c5a41cfa8a3da44/raw/06.make

```
gcc -c   -DNDEBUG -m64   -DLINUX -DUNIX -D_GNU_SOURCE -D_REENTRANT -D_FILE_OFFSET_BITS=64 -DPKGNAME=linux64 -D_ARCH_BITS=__SIZEOF_POINTER__*__CHAR_BIT__ -DLIBPREFIX=lib -DSHLIBEXT=so  -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/interfaces/override -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/libs/hdf5/linux -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/libs/hdf5/unix -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/libs/hdf5 -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/interfaces -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/interfaces/cc/gcc/x86_64 -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/interfaces/cc/gcc -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/interfaces/os/linux -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/interfaces/os/unix -I/var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/interfaces/ext -I/gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.2/include -I. -I/usr/include -MD -o hdf5arrayfile.pic.o -fPIC -O3 -Wall -Wno-variadic-macros -fno-strict-aliasing -D_LIBRARY /var/tmp/sjackman/sratoolkit--ncbi-vdb20150630-8099-1e8vime/ncbi-vdb-2.5.2/libs/hdf5/hdf5arrayfile.c -D__mod_name__=libs/hdf5 -D__file_name__=hdf5arrayfile -D__file_ext__=c
```

sjackman:
> I have a suggestion - please write to us and tell us what it is that you're trying to modify.

What's the e-mail address? Do you prefer e-mail to GitHub issues? I prefer a searchable issue archive myself.

I'm trying to add `-I ~/.linuxbrew/include` to `CPPFLAGS` to hopefully work around the above issue with `-I/usr/include`.

sjackman:
Unfortunately `-I/linuxbrew/include` doesn't work at all with GCC because it's already in the default `-isystem` path:

```
ignoring duplicate directory "/linuxbrew/include"
  as it is a non-system directory that duplicates a system directory
```

kwrodarmer:
It's fine to use a searchable archive. The request is that you let us know what you're trying to achieve in some way.

The addition of `-I/usr/include` is coming from the configuration of `$HDF5_INCDIR`, which is set through configuration. How are you configuring the build?

Incidentally, it is unlikely that you need hdf5 for any sort of a distribution. We use it internally at NCBI for our loading pipeline to deal with specific files that are sent in HDF5 format, but it is not needed for the external toolkit. We issue a disclaimer stating that the tools and libraries won't be built, but it is not an error. If you are trying to work around HDF5, I suggest you don't add that option.

sjackman:
The above GCC warning message is particularly ironic, because it's mean to prevent adding `/usr/include` to the search path, which is the typical system include directory. =)

sjackman:
It looks as though it's `ngs-sdk` that's failing to configure. The configure line is
`./configure --prefix=/gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.2 --build=/gsc/btl/linuxbrew/Cellar/sratoolkit/2.5.2`
See https://gist.github.com/sjackman/6e716c5a41cfa8a3da44/#file-01-configure-L3-L5
and https://github.com/Homebrew/homebrew-science/blob/master/sratoolkit.rb#L43

sjackman:
> If you are trying to work around HDF5, I suggest you don't add that option.

How do I disable HDF5 support? I'm not explicitly requesting it.

tseemann:
I think the current 2.5.4 resolves this, closing.

PR #2349, opened 2015-06-03
Bandage 0.5.6 (new formula)

tseemann:
COMMENT:
I know this is binary, but I challenge anyone to build it: https://github.com/rrwick/Bandage#installation

SYNOPSIS:
De novo genome assemblers work by building an assembly graph. Different assemblers represent this graph in different ways, but they all contain sequences (which become assembled contigs) and connections between those sequences. In the past, assembly graphs have not been easily accessible, and most users only use the assembled contigs, not the graph. Bandage is a program for visualising assembly graphs. By displaying connections which are not present in the contigs file, Bandage opens up new possibilities for analysing de novo assemblies.


bandage.rb
```diff
@@ -0,0 +1,33 @@
+class Bandage < Formula
+  desc "Bioinformatics Application for Navigating De novo Assembly Graphs Easily"
+  homepage "https://rrwick.github.io/Bandage/"
+  head "https://github.com/rrwick/Bandage.git"
```
sjackman: Please move head after the stable url. It's different than what I used to do, but it's what upstream Homebrew does.

bandage.rb
```diff
@@ -0,0 +1,33 @@
...
+  # tag "bioinformatics"
+  # doi "10.1101/018333"
```
sjackman: Please put doi before tag (alphabetical order).

bandage.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  if OS.linux?
+    depends_on "qt5"
+    depends_on "pcre"
```
sjackman: Please put dependencies in alphabetical order.

bandage.rb
```diff
@@ -0,0 +1,33 @@
...
+    url "https://github.com/rrwick/Bandage/releases/download/v0.5.6/Bandage_Linux_dynamic_v0.5.6.zip"
+    sha256 "229258e452d1715305c504a18e86d8b96e29cb2588d2cf82b3123e82b30e1143"
+  end
+
+  def install
+    if OS.mac?
+      prefix.install "../Bandage.app"
+    else
+      bin.install "Bandage"
+      (share / "bandage").install "sample_LastGraph"
```
sjackman: Please no spaces around `/` (Homebrew style guide).

sjackman:
Hi, Torsten. Bandage is already in `brew cask` for Mac. Is this formula intended primarily for Linux? I think my preference is for this formula to fail on Mac and simply report that the user should use `brew cask install bandage` instead.

```ruby
def install
  odie "Install the Bandage cask using\nbrew cask install bandage" if OS.mac?
  …
end

**Comment by sjackman** [comment]
@Homebrew/science Do you agree with my above comment regarding GUI apps and casks?

**Comment by sjackman** [comment]
http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/2113/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_Homebrew_science_bandage/
```
Error: No such file or directory - ../Bandage.app
```

**Comment by sjackman** [comment]
http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/2113/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_Homebrew_science_bandage___strict/
```
==> audit problems
homebrew/science/bandage:
 * Description is too long. "name: desc" should be less than 80 characters (currently 98).
 * 12: Trailing whitespace was found
```

**Comment by tdsmith** [comment]
I agree that deferring to the cask on OS X seems wise. A Linux-only, binary-only, GUI-app formula seems pretty far from Homebrew's wheelhouse to me and I wouldn't like them to proliferate.

I'd use a LinuxDependency instead of `odie`.

**Comment by tseemann** [comment]
@sjackman - I didn't find the cask via "brew search bandage" ?

@tdsmith - what is "LinuxDependency" - is this a Brew keyword? Can't find docs.

**Comment by tdsmith** [comment]
Oops, I meant LinuxRequirement. There's an example in allpaths-lg.rb.

**Comment by sjackman** [comment]
I've not seen `LinuxRequirement` before. What's the advantage of that over
```ruby
odie "Use `brew cask install bandage`" if OS.mac?
```
For one, I like this error message better.

sjackman:
> I didn't find the cask via "brew search bandage"

Searching `brew cask` is disabled on Linux, because users were confused by it suggesting that you install a cask on Linux, which is Mac-only.

See https://github.com/caskroom/homebrew-cask/blob/master/Casks/bandage.rb

sjackman:
> A Linux-only, binary-only, GUI-app formula seems pretty far from Homebrew's wheelhouse to me

I'm okay with pretty much any science software. I don't seen any reason to be prejudiced against Linux, binary or GUI.

tdsmith:
A failed Requirement should halt the installation before fetching anything or installing any dependencies; maybe it's not important here. You can specify whatever message you want for the Requirement.

sjackman:
Okay. Good to know. Thanks, Tim.

PR #2361, opened 2015-06-05
Update to Cantera 2.2.0

speth:
New upstream release.

Also added a 'devel' spec which will follow the maintenance branch. I'm not sure why the previous one was removed. If I'm misunderstanding what the --devel option is for, this can be dropped.

### DIFF
--- a/cantera.rb
+++ b/cantera.rb
@@ -1,15 +1,12 @@
 require 'formula'
 
 class Cantera < Formula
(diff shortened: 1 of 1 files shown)


tdsmith:
The `devel` block should point to a particular versioned prerelease, not a branch.

tdsmith:
Thanks for the pull request! We're grateful for your support.

PR #2365, opened 2015-06-06
vcflib 20150603 (update)

tseemann:
The manual pull-request patch is no longer needed - @ekg has accepted https://github.com/ekg/vcflib/pull/84


tdsmith:
Thanks!

PR #2366, opened 2015-06-06
OpenCV 3.0.0 stable

guoxiao:


tdsmith:
:+1:, thanks!

makmanalp:
Agh! 3.0 has some massive API changes compared to 2.4, I pulled this and it broke everything. A lot of people are also going to want 2.4 to build other things until software catches up with the API changes. Should this be versioned to a separate recipe or something perhaps?

tdsmith:
Ow, good call; breaking dependent formulae is bad and I really should have remembered to think about it. Reverted for now (in e186520, 108d3019). Sorry for the trouble.

makmanalp:
No worries! For others, here's how I worked around it for now: 

```bash
cd /usr/local/Library/Taps/homebrew/homebrew-science
git checkout fb5dbc3bcac00b6ba52ac6580c8e018925a9bbaa opencv.rb
brew uninstall opencv
brew install opencv
```

tdsmith:
`brew uses opencv` (will need a revision bump): butterflow, cvblob, insighttoolkit, openalpr, openbr, shogun, visp.

Whether we need an `opencv3` or whether we advise people who want the 2.x API to use `brew pin` should depend on how extensive the changes are.

guoxiao:
Here is the transition guide from 2.4 to 3.0 http://docs.opencv.org/master/db/dfa/tutorial_transition_guide.html

makmanalp:
Here is the opencv 3.0 transition guide btw: http://docs.opencv.org/master/db/dfa/tutorial_transition_guide.html

makmanalp:
On a positive note, the 3.0 recipe built great on mavericks :)

tdsmith:
Some porting effort will be necessary; all the dependents I tried to bump failed: http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/2160/version=yosemite/

Kinda thinking the easiest path is a new opencv3 formula but I'll wait for @homebrew/science to weigh in before I make a bigger mess :)

autosquid:
Is it possible to reuse the "devel" option for early birds?

guoxiao:
I support the separation of opencv 3 and opencv 2 into two formulas.
Since 3 is now stable, it's better to keep `opencv` formula using 3, and add a `opencv2` formula to keep back compatibility. We can change these formulas to depend on `opencv2` before the upstream support 3.

tseemann:
Will Brew ever support (keg only) dependencies with version ranges?

tdsmith:
> Will Brew ever support (keg only) dependencies with version ranges?

Probably not, because we only maintain specific versions of each package and adding version ranges (instead of always building against a single, known-to-work version) just adds entropy.

autosquid:
The choice can simply be left to user: if you choose opencv3, you have to solve dependencies yourself; or just stay with opencv 2.   An option like exsiting `dev` should be supported. Any choice, plz. do it hurry., we need opencv 3.

iml:
The release notes say
> The release is pretty much compatible with 2.4.x, but there are some notable differences, which are described in the still-updated 2.4=>3.0 transition guide: http://docs.opencv.org/master/db/dfa/tutorial_transition_guide.html in particular, we have to remove some obsolete/unstable algorithms, functions, we moved some other stuff between modules or to opencv_contrib.

Upstream seems to suggest the transition is not so hard as to warrant a forward-versioned formula (a la python3 or root6). Since this is going from devel to stable, I'd suggest keeping this as-is and spin off an `opencv2` formula. 

On the other hand, this would require rebuilding all the formulae that currently depend on `opencv`, as well as considering how the two versions conflict when linked (i.e. should `opencv2` be keg-only).

iml:
Hmm, there are also some things in osgeo/osgeo4mac that depend on `opencv`.

autosquid:
So, are you trying to make a patch for every package depending on OpenCV? @iml   Is there a way to make clear how many packages are depending on OpenCV2? And how many can work with OpenCV3?

iml:
Not patched, but each rebuilt against a new opencv2 package. That's easy if there only a few, but an `opencv3` formula as @tdsmith suggested originally would be least painful.

You can see with `brew uses opencv` how many packages depend on that formulae. How many are compatible with opencv3 is an open question. If the dependents all failed as above, perhaps none?

autosquid:
@iml  Oh, sorry for missing info. from @tdsmith .
It seems a compromising is to include major version number to original package name, i.e. maintain a new package named OpenCV3. Then those packages depending on OpenCV 2 will still work happily.  And users of OpenCV3 can manager their build themselves according to last section of [transition guild](http://docs.opencv.org/master/db/dfa/tutorial_transition_guide.html) even with both OpenCV and OpenCV3 installed.
I hope I am making myself clear.

tdsmith:
I agree with you in principle, Ian, but I'm worried about the mass of third-party stuff expecting opencv 2 that depends on homebrew/science/opencv, either in taps or informally, and I think we can expect that most/all of that will break against opencv 3. :/ I think that for a while to come, most people who want the opencv formula will be trying to build software that expects 2.x.

I would like to add a keg-only opencv3 formula; I think the keg-only semantics are going to be easier to figure out for the more adventurous people who are already using the new API.

I would also like to add `opencv2` as an alias for `opencv`. I'm not sure we can do that right now; I'll investigate. I don't think we should block on that.

makmanalp:
Second @tdsmith here - you can't just bump a dependency an API-breaking major revision up and hope that everyone else does so immediately!

In reality, maybe this is commentary on how package dependencies are specified in homebrew - you should be able to pin dependencies to be within the same major or minor or bugfix revisions, so packages can be bumped up independently of everything else. Or maybe it is that way, but it's a matter of getting package maintainers to follow that best practice?

iml:
We can only pin dependencies to formulae, rather than versions. This is actually a broader issue than just major bumps in dependencies. Even a minor release can change the name of a linked library, requiring repackaging of dependent formulae. For major releases, API breakage needs to be considered but the canonical formula should eventually track the stable upstream release. Notwithstanding the "pretty much compatible" statement, I agree with a keg-only `opencv3` at this point.

PR #2370, opened 2015-06-06
opencv 3.0.0 and dependents

tdsmith:
Change borrowed from #2366; bumps revisions for dependent packages. Let's see whether they build. :)


tdsmith:
"No."

PR #2371, opened 2015-06-07
ACE 20150501 (new formula)

tseemann:
The most ubiquitous platform, Illumina, mainly suffers from substitution errors. We have developed a tool, ACE, based on K-mer tries to correct such errors. On real MiSeq and HiSeq Illumina archives, ACE yields higher gains in terms of coverage depth, outperforming state-of-the-art competitors in the majority of cases.


tseemann:
This build error is old - can someone trigger the TestBot for me?
```
* Description is too long. "name: desc" should be less than 80 characters (currently 98)."
```

tseemann:

PR #2387, opened 2015-06-09
HMMER 3.1b2 (add real test + install docs)

tseemann:
Same classic formula, new tasty test.


hmmer.rb
```diff
@@ -26,11 +28,12 @@ def install
     system "make"
     system "make", "check" if build.with? "check"
     system "make", "install"
-
-    share.install "tutorial"
+    
+    doc.install "Userguide.pdf", "README", "COPYRIGHT", "LICENSE"
```
tdsmith: FWIW, all of these files except Userguide.pdf will be copied to `prefix` automatically without this line. (The auto-installed "metafiles" meet these criteria: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/metafiles.rb)

tseemann: Ok thanks for the tip!
I'll fix the whitespace audit failure.

tdsmith:
This is great; thanks a lot!

PR #2388, opened 2015-06-09
Upgrade HDF5 to version 1.8.15-patch1

barrywardell:


tdsmith:
It's mis-parsing the version; can you add `version "1.8.15.1"`?

iml:
Might also need to tweak the test.

barrywardell:
I have added the version number and updated the test. brew audit reports that a non-library file was installed (/usr/local/Cellar/hdf5/1.8.15-patch1/lib/libhdf5.settings). The file contains a summary of the configuration options used during the build. What is the recommended way to correct this?

iml:
Just ignore that warning; the file is needed there.

iml:
@tdsmith will the `-patch1` conflict with the bottle or formula regex? I think this will also require revisioning dependents since the library has gone from `libhdf5.9.dylib` –>  `libhdf5.10.dylib` :frowning:

tdsmith:
It looks okay. You're right about the dependents!

@barrywardell, can you look at the output of `brew uses hdf5` and increment the revision (or add `revision 1`) of all of the packages that have a mandatory or recommended dependency on hdf5?

barrywardell:
@tdsmith I have bumped the revision of all dependents of hdf5.

@iml The `-patch1` was previously used with 1.8.10-patch1 (see 41a741028aa99806a2a428adc02cb92aae9e6a45), so I assumed it would still be OK here.

dpo:
@barrywardell Oops you must have rebased your branch. Would you mind cleaning it up or just re-submitted your updated `hdf5` in a new PR?

barrywardell:
@dpo I have now rebased against the current master. I'm not quite sure why the tests failed, but it looks like it is in the dependencies which had a revision bump.

dpo:
@barrywardell Thanks for the update! What I meant is that your PR shows 39 files with changes. Somehow you must have merged upstream updates into your local branch. It should only show one file with changes: `hdf5.rb`.

barrywardell:
@dpo Ah, those files are all of the dependencies of hdf5 which have a revision bump because the library version changed - see https://github.com/Homebrew/homebrew-science/pull/2388#issuecomment-110569515 by @tdsmith above.

dpo:
Closing as we're now at 1.8.16.

PR #2399, opened 2015-06-11
biopp revision

kgori:
Installs bppqt library, at the request of the bio++ developers. Adds 'recommended' qt dependency.
May need an explicit revision statement in the formula.


biopp.rb
```diff
@@ -42,7 +45,12 @@ class Biopp < Formula
     sha256 "c7ec73a5af84808362f301479c548b6a01c47a66065b28a053ff8043409e861a"
   end
 
-  depends_on "cmake" => :build
+  if build.with? "qt"
+    resource "bppqt" do
+      url "http://biopp.univ-montp2.fr/repos/sources/bpp-qt-2.2.0.tar.gz"
+      sha256 "9662f66bc3491d8e128263f6bd91fcdbecdb375ec9f24519f44855cdcdb9d553"
+    end
+  end
```
tdsmith: You can remove the `if` wrapper; it's only downloaded if it's actually staged.

kgori:
I removed the redundant if statement, and all tests pass. Can this be merged?

tdsmith:
Pulled with some small changes to the test, thanks for the ping! You're welcome to comment after you've pushed changes that address feedback; we get notifications for comments but not new commits.

PR #2402, opened 2015-06-12
Add the Caffe deep learning framework

pwnall:
[Caffe](http://caffe.berkeleyvision.org/) is a deep learning framework released by Berkeley's Vision Group. It's not 1.0 yet (though it has a RC out), so this recipe is HEAD only.

The recipe doesn't cover Python or MATLAB support. That will hopefully happen in the future as people start using it.

### DIFF
--- a/caffe.rb
+++ b/caffe.rb
@@ -1,0 +1,88 @@
+class CudaRequirement < Requirement
+  build true
+  fatal true
+
+  satisfy { which "nvcc" }
+
+  cask "cuda"
+
+  env do
+    # Nvidia CUDA installs (externally) into this dir (hard-coded):
+    ENV.append "CFLAGS", "-F/Library/Frameworks"
+    # # because nvcc has to be used
+    ENV.append "PATH", which("nvcc").dirname, ":"
+  end
+
+  def message
+    <<-EOS.undent
+      To use this formula with NVIDIA graphics cards you will need to
+      download and install the CUDA drivers and tools from nvidia.com.
+
+          https://developer.nvidia.com/cuda-downloads
+
+      Select "Mac OS" as the Operating System and then select the
+      "Developer Drivers for MacOS" package.
+      You will also need to download and install the "CUDA Toolkit" package.
+
+      The `nvcc` has to be in your PATH then (which is normally the case).
+
+  EOS
+  end
+end
+
+class Caffe < Formula
+  desc "A deep learning framework"
+  homepage "http://caffe.berkeleyvision.org/"
+  url "https://github.com/BVLC/caffe/archive/rc2.tar.gz"
+  version "rc2"
+  sha256 "55c9c20870b30ce398e19e4f1a62ade1eff08fce51e28fa5604035b711978eec"
+
+  head "https://github.com/BVLC/caffe.git"
+
+  option "with-openblas", "Use OpenBLAS instead of Accelerate"
+  option "with-cudnn", "Build with cuDNN support"
+  option :cxx11
+
+  depends_on "cmake" => :build
+  depends_on "boost"
+  depends_on "gflags"
+  depends_on "glog"
+  depends_on "hdf5"
+  depends_on "leveldb"
+  depends_on "lmdb"
+  depends_on "openblas" => :optional
+  depends_on CudaRequirement => :optional
+  depends_on "opencv3"
+  depends_on "protobuf"
+  depends_on "snappy"
+  depends_on "szip"
+
+  def install
+    ENV.cxx11 if build.cxx11?
+
+    args = std_cmake_args
+    args << "-DBLAS=open" if build.with?("openblas")
+    if build.with?("cudnn")
+      args << "-DUSE_CUDNN=1"
+    else
+      args << "-DCPU_ONLY=1"
+    end
+
+    if ENV.compiler == :clang && !build.bottle?
+      args << "-DENABLE_SSSE3=ON" if Hardware::CPU.ssse3?
+      args << "-DENABLE_SSE41=ON" if Hardware::CPU.sse4?
+      args << "-DENABLE_SSE42=ON" if Hardware::CPU.sse4_2?
+      args << "-DENABLE_AVX=ON" if Hardware::CPU.avx?
+    end
+
+    mkdir "macbuild" do
+      system "cmake", "..", *args
+      system "make"
+      system "make", "install"
+    end
+  end
+
+  test do
+    system bin/"caffe"
+  end
+end


caffe.rb
```diff
@@ -0,0 +1,48 @@
+class Caffe < Formula
+  homepage "http://caffe.berkeleyvision.org/"
+  head "https://github.com/BVLC/caffe.git"
+  revision 1
+
+  option "with-openblas", "Use OpenBLAS instead of ATLAS"
```
tdsmith: Does this use ATLAS or Accelerate on OS X?

caffe.rb
```diff
@@ -0,0 +1,48 @@
+class Caffe < Formula
+  homepage "http://caffe.berkeleyvision.org/"
+  head "https://github.com/BVLC/caffe.git"
+  revision 1
```
tdsmith: No revision is necessary for new formulae.

caffe.rb
```diff
@@ -0,0 +1,48 @@
...
+  depends_on "leveldb"
+  depends_on "gflags"
+  depends_on "glog"
+  depends_on "szip"
+  depends_on "lmdb"
+  depends_on "hdf5"
+  depends_on "opencv"
+  depends_on "protobuf"
+  depends_on "boost"
+  depends_on "openblas" if build.with?("openblas")
```
tdsmith: Use `depends_on "openblas" => :optional`

caffe.rb
```diff
@@ -0,0 +1,48 @@
+class Caffe < Formula
+  homepage "http://caffe.berkeleyvision.org/"
+  head "https://github.com/BVLC/caffe.git"
+  revision 1
+
+  option "with-openblas", "Use OpenBLAS instead of ATLAS"
+  option "without-cudnn", "Build without cuDNN support"
+  option :cxx11
+
+  depends_on "cmake"      => :build
```
tdsmith: Nits: you don't need to align things vertically; if you can alphabetize the dependencies that would be great

tdsmith: (though keeping build dependencies at the top is good)

caffe.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    mkdir "macbuild" do
+      system "cmake", "..", *args
+      system "make"
+      system "make", "install"
+    end
+  end
+
+  test do
+    system bin/"caffe"
```
tdsmith: Can you run some trivial input though caffe?

caffe.rb
tdsmith: It uses Accelerate: http://caffe.berkeleyvision.org/install_osx.html -- can you change this text to reflect that?

tdsmith:
I think in `-science` sometimes it's okay for the `stable` spec to reflect an rc release if the project hasn't had a real stable release yet; I'd lean towards doing that here, but I'll let @sjackman tell me if I'm wrong. :)

caffe.rb
```diff
@@ -0,0 +1,48 @@
...
+  depends_on "boost"
+  depends_on "openblas" if build.with?("openblas")
+
+
+  def install
+    ENV.cxx11 if build.cxx11?
+
+    args = std_cmake_args
+    args << "-DBLAS=open" if build.with?("openblas")
+    args << "-DCPU_ONLY=1" unless build.with?("cudnn")
```
tdsmith: Use `if build.without?` instead of `unless build.with?`

caffe.rb
```diff
@@ -0,0 +1,48 @@
+class Caffe < Formula
+  homepage "http://caffe.berkeleyvision.org/"
+  head "https://github.com/BVLC/caffe.git"
+  revision 1
+
+  option "with-openblas", "Use OpenBLAS instead of ATLAS"
+  option "without-cudnn", "Build without cuDNN support"
```
tdsmith: This should use a CudaRequirement like beagle.rb does. In order to avoid messing around with stdlibs, I think it also make sure that CUDA is at least version 7.

caffe.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  depends_on "cmake"      => :build
+  depends_on "snappy"
+  depends_on "leveldb"
+  depends_on "gflags"
+  depends_on "glog"
+  depends_on "szip"
+  depends_on "lmdb"
+  depends_on "hdf5"
+  depends_on "opencv"
```
tdsmith: It says it can use opencv 3 so let's depend on the new `opencv3` formula.

sjackman:
> I think in -science sometimes it's okay for the stable spec to reflect an rc release if the project hasn't had a real stable release yet; I'd lean towards doing that here, but I'll let @sjackman tell me if I'm wrong. :)

Fine by me.

caffe.rb
pwnall: Sorry, I thought Accelerate is just OSX's fancy name for ATLAS.

caffe.rb
pwnall: I'm looking into that right now. FWIW, the CUDA sdk is available as a cask.
https://github.com/caskroom/homebrew-cask/blob/master/Casks/cuda.rb

Are we interested in linking to that instead of telling people to do manual installation?

tdsmith: That would be good! Adding `cask "cuda"` to the Requirement will add some language recommending the cask IIRC.

dpo:
@BrewTestBot test this please

caffe.rb
pwnall: I asked the authors, and they said that I should run `make runtest`. Is that adequate for testing, or are you looking for something along the lines of a quick smoke test?

tdsmith: Quick smoke test; the `test` block is run after the source tarball is cleaned up. Running `make runtest` in the `install` method is probably a good idea, though. (If it takes a really long time, you can add a "with-tests" option, and then run the tests `if build.with?("tests") || build.bottle?`.

caffe.rb
```diff
@@ -0,0 +1,88 @@
...
+  depends_on "protobuf"
+  depends_on "snappy"
+  depends_on "szip"
+
+  def install
+    ENV.cxx11 if build.cxx11?
+
+    args = std_cmake_args
+    args << "-DBLAS=open" if build.with?("openblas")
+    if build.with?("cudnn")
```
tdsmith: This is probably "cuda" now, not "cudnn"

L2G:
I noticed the following in the CMake output:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    ENABLE_AVX
    ENABLE_SSE41
    ENABLE_SSE42
    ENABLE_SSSE3
```

...so I'm not sure whether there are other option names that should be used instead.

ahundt:
Any luck with this? I'm interested in using it

dpo:
There still isn't a stable release of Caffe and I wasn't able to build it locally so I'm closing this. Please open a new pull request if still interested and someone has a working formula. Thanks.

PR #2412, opened 2015-06-15
kallisto: fix sha256 mismatch.

ntkme:
This PR fixes the SHA256 mismatch.


sjackman:
Thanks for your contribution, @ntkme

Issue #1593, opened 2014-12-30
Formula that fails_with :clang and depends on C++ library built with clang fail


sjackman:
# Affected formula

Formula | Dependency | Issue
---|---|---
abyss | google-sparsehash | #2332
amos | boost | #485
augustus | boost | #1373
blasr | hdf5 | #2415
bless | google-sparsehash | #1839
diamond | boost | #1909
garli | ncl | #1538
htqc | boost | #1595
meraculous | boost | #1695
mira | boost | #1148
quorum | jellyfish | #1661
sailfish | boost | #1555

sjackman:
@Homebrew/owners These formula fail to build with clang, and so are built with GCC by default, but they have C++ dependencies (libraries) that are built with clang by default. These formula then get link errors and fail to build. Is there a solution to this issue?

mikemcquaid:
Given what you've said: I don't think so, no.

sjackman:
Has this issue not come up in Homebrew/homebrew?

mistydemeo:
We have a policy of no longer accepting software that doesn't build with clang.

mikemcquaid:
At this point software that doesn't build with `clang` can't really be considered to be ported to OSX.

sjackman:
Hmm. There's too much important scientific software that doesn't build with `clang` out of the box to adopt that same policy in Homebrew/science, I don't have the time to port it myself. How would you feel about Homebrew/science packaging a `keg_only` version of boost `boost-gcc` that is built with GCC? I don't think it's a great solution, but it's an option.

mikemcquaid:
@sjackman If you do that for one package you're going to end up doing it for a bunch. I think it's a bad call, personally. As I said previously: if stuff doesn't work with `clang` (let alone `libc++`) it can't really be considered to be working on OS X. If you can workaround that within a formula or it can build with `gcc` and still link OK that makes sense otherwise I'd just leave it (or put it in a `homebrew-science-gcc` tap)

sjackman:
This issue is resolved now that `/usr/bin/clang` and `gcc-5` are ABI compatible on recent versions of Mac OS.

mikemcquaid:
> This issue is resolved now that /usr/bin/clang and gcc-5 are ABI compatible on recent versions of Mac OS.

@sjackman Does that mean the standard library checks are no longer needed as-of GCC-5 and recent Clangs? Thanks!

sjackman:
[...]

PR #2415, opened 2015-06-18
blasr 2.1: Add desc and tag "bioinformatics"

sjackman:


sjackman:
blasr fails to build with clang but has a C++ hdf5 dependency, which means bottles can't be built. See #1593

PR #2425, opened 2015-06-19
r8s 1.8 (new formula)

jonchang:
r8s estimates absolute rates ("r8s") of molecular evolution and divergence times on a phylogenetic tree.


tseemann:
@jonchang Could you please use ```(share/"r8s").install``` instead so the data is kept neatly separate from the general folder?
This would require modifying the test block as well.

jonchang:
@tseemann fixed

iml:

PR #2427, opened 2015-06-20
YARP 2.3.64 (moved from homebrew-x11)

kt10aan:
This is the formula for YARP, v.2.3.64.
YARP is a middleware for robots and is used mainly in academia. (See http://yarp.it and http://icub.org)

The formula was in the main Homebrew repository. Then, it was moved to homebrew-x11, along with all other formulas that had an X11 dependency. 

Latest version of YARP switched to Qt5, so it doesn't have an X11 dependency anymore.
Instead of moving it back to main Homebrew, we thought it would fit better in the homebrew-science repository.

The formula was just updated extensively in homebrew-x11 so hopefully, there should not be any issue with it.

Also see: https://github.com/Homebrew/homebrew-x11/pull/85

### DIFF
--- a/yarp.rb
+++ b/yarp.rb
@@ -1,0 +1,79 @@
+class Yarp < Formula
+  desc "Yet Another Robot Platform"
+  homepage "http://yarp.it"
+  url "https://github.com/robotology/yarp/archive/v2.3.64.tar.gz"
+  sha256 "3882b38c39ec9c5fdd06c68f3a4ad21da718bd2733ed7d6a5fb78d9f36dad6b2"
+  head "https://github.com/robotology/yarp.git"
+
+  option "with-qt5", "Build the Qt5 GUI applications"
+  option "with-yarprun-log", "Build with yarprun_log support"
+  option "with-opencv", "Build the opencv_grabber device"
+  option "with-lua", "Build with Lua bindings"
+  option "with-python", "Build with Python bindings"
+  option "with-serial", "Build the serial/serialport devices"
+
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+  depends_on "ace"
+  depends_on "gsl"
+  depends_on "sqlite"
+  depends_on "readline"
+  depends_on "jpeg"
+  depends_on "homebrew/science/opencv" => :optional
+  depends_on "qt5" => :optional
+  depends_on "lua" => :optional
+  depends_on "python" => :optional
+
+  if build.with?("lua") || build.with?("python")
+    depends_on "swig"
+  end
+
+  def install
+    args = std_cmake_args + %W[
+      -DCREATE_LIB_MATH=TRUE
+      -DCREATE_DEVICE_LIBRARY_MODULES=TRUE
+      -DCREATE_OPTIONAL_CARRIERS=TRUE
+      -DENABLE_yarpcar_mjpeg_carrier=TRUE
+      -DENABLE_yarpcar_rossrv_carrier=TRUE
+      -DENABLE_yarpcar_tcpros_carrier=TRUE
+      -DENABLE_yarpcar_xmlrpc_carrier=TRUE
+      -DENABLE_yarpcar_bayer_carrier=TRUE
+      -DENABLE_yarpcar_priority_carrier=TRUE
+      -DCREATE_IDLS=TRUE
+    ]
+
+    args << "-DCREATE_GUIS=TRUE" if build.with? "qt5"
+    args << "-DENABLE_YARPRUN_LOG=ON" if build.with? "yarprun-log"
+    args << "-DENABLE_yarpmod_opencv_grabber=ON" if build.with? "opencv"
+
+    if build.with? "lua"
+      args << "-DYARP_COMPILE_BINDINGS=ON"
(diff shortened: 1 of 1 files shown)


DomT4:
Can ignore the failure - It won't be true if the Science team want this formula because I'll remove it from the X11 tap:

```
==> brew audit Homebrew/science/yarp --strict				 FAILED
==> brew style yarp

Error: 1 problem in 1 formula

==> audit problems
homebrew/science/yarp:
 * Formula name conflicts with homebrew/x11/yarp
```

kt10aan:
Added bash-completion support in the YARP formula.

yarp.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+  depends_on "ace"
+  depends_on "gsl"
+  depends_on "sqlite"
+  depends_on "readline"
+  depends_on "jpeg"
+  depends_on "homebrew/science/opencv" => :optional
+  depends_on "qt5" => :optional
```
dpo: This is redundant with [line 8](https://github.com/Homebrew/homebrew-science/pull/2427/files#diff-f4237c136cd465bbf7eb03cb53ea74c8R8)

yarp.rb
```diff
@@ -0,0 +1,70 @@
...
+  option "with-serial", "Build the serial/serialport devices"
+
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+  depends_on "ace"
+  depends_on "gsl"
+  depends_on "sqlite"
+  depends_on "readline"
+  depends_on "jpeg"
+  depends_on "homebrew/science/opencv" => :optional
```
dpo: This is redundant with [line 10](https://github.com/Homebrew/homebrew-science/pull/2427/files#diff-f4237c136cd465bbf7eb03cb53ea74c8R10) (unless the documentation string is important)

yarp.rb
```diff
@@ -0,0 +1,70 @@
...
+  depends_on "qt5" => :optional
+  depends_on "lua" => :optional
```
dpo: Redundant with [line 11](https://github.com/Homebrew/homebrew-science/pull/2427/files#diff-f4237c136cd465bbf7eb03cb53ea74c8R11)

yarp.rb
```diff
@@ -0,0 +1,70 @@
...
+      args << "-DENABLE_yarpmod_serialport=ON"
+    end
+
+    system "cmake", *args
+    system "make", "install"
+    bash_completion.install "scripts/yarp_completion"
+  end
+
+  def caveats
+    <<-EOS.undent
```
dpo: I tend to prefer `def caveats; <<-EOS.undent` on one line

kt10aan:
Well, I would rather keep the documentation strings for the options. Can I just delete the 3 ``depends_on ... => :optional`` lines?
I changed the ``def caveats; EOS...`` in one line. The commit was squashed.

iml:
You can't remove the dependencies, but the options are fine as they just replace the default wording of `Build with(out) X support` given a dependency X.

kt10aan:
So, the option lines can be kept, I assume.

dpo:
I guess at the very least the dependencies should say `depends_on "qt5" if build.with? "qt5"` but I feel the message don't add much to justify the redundancy.

iml:
> I guess at the very least the dependencies should say `depends_on "qt5" if build.with? "qt5"` but I feel the message don't add much to justify the redundancy.

That isn't necessary, you can change the option description while keeping the `:optional`, `:recommended` DSL. The octave formula does this a lot; maybe a bit verbose but it doesn't require that sort of redundancy in the class body.

kt10aan:
The description for the qt5 option is needed in my opinion, due to the nature of the software.
Is there a unanimous opinion if the option description lines are redundant or not? :) 

By the way, slightly off-topic: since I have contributed to different homebrew sub-repositories, it appears to me that there is not a common style guideline across all homebrew "official" repos (except the ones that are checked by ``brew audit``. Wouldn't it be better (and faster for everyone) if the homebrew people would decide on extending these style guidelines and integrate them in the ``brew audit/style`` ?

dpo:
In my opinion, there's very little benefit to having `Build the Qt5 GUI applications` instead of `Build with qt5 support`, and even less to `Build with Lua bindings` instead of `Build with lua support`. Some of the `octave` options are redundant too, if you ask me.

kt10aan:
There is some benefit from having ``Build with the Qt5 GUI applications``, compared to ``Build with qt5 support``. Actually, they are very different. 
Also, the ``lua bindings`` are different from ``lua support`` as the YARP users expect the terminology ``bindings``. That is the term that for many years is mentioned in manuals, mailing lists, etc when talking and asking about, well.. bindings :)

In any case, I do not see a __unanimous__ opinion, against ``option`` lines.

kt10aan:
Any update on this? :)

francesco-romano:
Any update on this?

francesco-romano:
We create a "personal" tap in https://github.com/robotology/homebrew-cask, so this PR can be closed.
As soon as we update the formula we will also delete it from `x11`.
Thanks guys!

@dpo

PR #2434, opened 2015-06-22
r: fix build with Quartz-only cairo

tdsmith:
Building without pango yields an error because then R uses freetype, and
the freetype headers export a TYPEOF macro that conflicts with the
Rinternals.h definition. Pango probably looks better anyway.

Bumping the revision because R was bottled without cairo graphics
support.


tdsmith:
Bug report filed against freetype: https://savannah.nongnu.org/bugs/index.php?45376

tdsmith:
@Homebrew/science: Does anyone feel strongly that R should have X11 support on OS X by default? R works fine without X (with Quartz and cairo graphics) as long as code doesn't explicitly ask for the `X11()` graphics device, which isn't portable to Windows anyway.

tdsmith:
Text in SVG output is totally FUBAR using pango, so we should probably try to hack around the fontconfig TYPEOF :/

sjackman:
It looks fine to me, but I haven't tested it. I have no strong feelings about whether R should include X11 support by default. I don't use R with X11 on Mac.

tdsmith:
Thanks for the feedback, Shaun! I fixed the troublesome freetype macro in Homebrew/homebrew#41020 so we don't need to worry about that anymore.

PR #2444, opened 2015-06-24
prank 140603 (update formula)

ge11232002:
Download the source code from http://wasabiapp.org and make github repo
as devel branch.


tseemann:
@ge11232002 Thanks for the update. 
Can you please change the SHA1 to SHA256, and also add a "desc" description tag at the top, and squash into one commit?

tseemann:
The `desc` will need to be < 73 characters long.
Can you please squash all your commits into one single commit?
Thank you.

tseemann:
Thanks @ge11232002

PR #2453, opened 2015-06-25
graph-tool 2.2.43

saketkc:


tseemann:
```
        ==> Installing graph-tool from homebrew/homebrew-science
/usr/bin/env python -c import matplotlib
/usr/bin/env python -c import numpy
/usr/bin/env python -c import scipy
graph-tool: Unsatisfied dependency: matplotlib
Homebrew does not provide Python dependencies; install with:
  pip install matplotlib
graph-tool: Unsatisfied dependency: numpy
Homebrew does not provide Python dependencies; install with:
  pip install numpy
graph-tool: Unsatisfied dependency: scipy
Homebrew does not provide Python dependencies; install with:
  pip install scipy
Error: Unsatisified requirements failed this build.
```

sjackman:
Ideally, yes, but you can go on faith if you like. The package does build on Mountain Lion, which has those dependencies. That's good enough for me.
http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/2358/

count0:
A newer version has been released. See #2477

PR #2472, opened 2015-06-29
vcflib 1.0.0-rc0 (we have a tagged release!)

tseemann:


tseemann:
```
        Error: 1 problem in 1 formula
homebrew/science/vcflib:
 * Stable: Git should specify :revision when a :tag is specified.
```

iml:
Git tags can actually be moved to a different commit, whereas the sha1 of a commit cannot be changed.

iml:
Should be `:revision => "8ac9fd517579134ef3b9797714d20c9c99c18ec6"`

tseemann:
@iml Thanks for explaining the movable tag thing - I did not realise that! I'll patch.

PR #2473, opened 2015-06-29
butterflow: fix failure to resolve python modules

dthpham:
The bottled butterflow package is broken (import errors) because it's not picking up the `cv2.so` extension module located in `$(brew --prefix)/lib/python2.7/site-packages` (see: dthpham/butterflow#10). This is fixed by adding that directory to `PYTHONPATH`.


tdsmith:
Good catch; this had worked for me because my system Python was configured to search Homebrew's site-packages. Let's use:
```ruby
ENV.prepend_path "PYTHONPATH", Formula["opencv"].opt_lib/"python2.7/site-packages"
```
instead -- we don't want to create the path if it doesn't exist and we only want to pick up opencv.

dthpham:
Thanks, that's a much better approach. I have updated the formula with that path.

tdsmith:
@BrewTestBot test this please

tdsmith:
@BrewTestBot test this please

tdsmith:
Thanks for the PR!

Issue #2346, opened 2015-06-02
pcl --HEAD formula is broken on OS X Yosemite 


jmtatsch:
any ideas how to fix this?

---
brew install homebrew/science/pcl --HEAD -v
==> Installing pcl from homebrew/homebrew-science
==> Cloning https://github.com/PointCloudLibrary/pcl.git
git --git-dir /Library/Caches/Homebrew/pcl--git/.git status -s
Updating /Library/Caches/Homebrew/pcl--git
git config remote.origin.url https://github.com/PointCloudLibrary/pcl.git
git config remote.origin.fetch +refs/heads/master:refs/remotes/origin/master
git fetch origin
git checkout -f master --
Already on 'master'
Your branch is up-to-date with 'origin/master'.
git reset --hard origin/master
HEAD is now at b60e20d Merge pull request #1239 from v4hn/point-normal-might-fail
==> Checking out branch master
==> Patching
patching file cmake/pcl_find_cuda.cmake
Hunk #2 succeeded at 45 (offset 8 lines).
[...]

PR #2482, opened 2015-07-03
Fixes PCL --HEAD

fran6co:
Fixes https://github.com/Homebrew/homebrew-science/issues/2346. PCL now expects QT5 by default, so we need to tell him to use QT4. Need to investigate QT5 support in HEAD.

### DIFF
--- a/pcl.rb
+++ b/pcl.rb
@@ -1,16 +1,14 @@
-require "formula"
-
 class CudaRequirement < Requirement
   build true
   fatal true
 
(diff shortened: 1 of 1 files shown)


fran6co:
Added some audit changes too

tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this one.

fran6co:
Thanks for keeping up with this awesome project!

PR #2483, opened 2015-07-03
Update nextflow to version 0.14.3

pditommaso:


tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this one. Just for reference, our preferred commit message header is like "somepackage 1.2.3" for version bumps like this.

PR #2489, opened 2015-07-04
Snippy 2.5 (new formula)

tseemann:
Rapid multicore bacterial variant calling between reference and NGS reads.


tseemann:
@iml @sjackman @dpo 
This formula should fail because it depends on 'freebayes' which is not in homebrew?

tdsmith:
It can't be tested because you haven't provided the Perl libraries.

```
==> SKIPPING: homebrew/science/snippy
Unsatisfied dependency: Bio::Perl
Homebrew does not provide Perl dependencies; install with:
  cpan -i Bio::Perl
Unsatisfied dependency: SVG::Graph
Homebrew does not provide Perl dependencies; install with:
  cpan -i SVG::Graph
```

tseemann:
@tdsmith How did you get that report? Is that anywhere in the "Details" link in the top right of the pull request success box?

tdsmith:
If you go to the Jenkins report for this PR, click on one of the build
hosts (one of the OS names), and then click "console," you'll see it there.

Tim
Sent from mobile

iml:
It skips builds with unfulfilled module dependencies, but not normal dependencies AFAIK. We have to rely on manual testing in such cases, since it's not realistic to package massive resources like BioPerl (and its dependencies) for each formula.

sjackman:
I have considered packaging `bioperl` for Homebrew (like `numpy` and [`scipy`](https://github.com/Homebrew/homebrew-python/blob/master/scipy.rb)), but that only help the formula that depend on bioperl and no other PERL packages, which I suspect is not many. Other smaller dependencies could be vendoured in though.

tseemann:
@sjackman Bioperl would be very useful to have packaged, but I still haven't got a proper understanding of how Perl module paths etc fit in with Brew.

sjackman:
I believe (I'm no expert here) that Perl packages to be used by other formula go in `lib/perl5/site_perl` and private Perl packages go wherever in `libexec`.

tseemann:
I am closing this. The formula will be kept up to date in my tap: `tseemann/homebrew-bioinformatics-linux`.

PR #2503, opened 2015-07-10
ea-utils 1.1.2-806 (new formula)

lparsons:


tdsmith:
Audit failure:
```
==> audit problems
homebrew/science/ea-utils:
 * Description is too long. "name: desc" should be less than 80 characters (currently 87).
```

ea-utils.rb
```diff
@@ -0,0 +1,50 @@
...
+  homepage "https://code.google.com/p/ea-utils/"
+  url "https://drive.google.com/uc?export=download&id=0B7KhouP0YeRAc2xackxzRnFrUEU"
+  version "1.1.2-806"
+  sha256 "31f059ac2d6282d8e0e25047fc6aa5448f8c41de6067eaf6b456054cd53c782c"
+
+  # Patch from upstream issue: https://code.google.com/p/sparsehash/issues/detail?id=99
+  patch :DATA
+
+  def install
+    system "PREFIX=#{prefix}",  "make",  "install"
```
tdsmith: The "PREFIX=" bit should be the last argument; this command line isn't interpreted by a shell, so you can't set environment variables like this, but make will understand the same syntax as an argument.

ea-utils.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  test do
+    system "#{bin}/sam-stats", "--help"
```
tdsmith: Cam you give sam-stats a sample FASTA input here?

tdsmith: er, sam input

tdsmith: (maybe fastq-stats is easier?)

lparsons:
Thanks for the feedback, I believe this updated pr addresses the issues.

lparsons:
Ooops, sorry, I was missing a dependency. That's been fixed and tests pass now.

PR #2504, opened 2015-07-11
Superlu_dist 4.1 (updated)

davydden:

PR #2507, opened 2015-07-11
Trilinos 12.0.1 (update)

davydden:
pull after https://github.com/Homebrew/homebrew-science/pull/2504

### DIFF
--- a/trilinos.rb
+++ b/trilinos.rb
@@ -1,7 +1,8 @@
 class Trilinos < Formula
+  desc "Algorithms for the solution of large-scale, complex multi-physics engineering and scientific problems"
   homepage "http://trilinos.sandia.gov"
-  url "http://trilinos.org/oldsite/download/files/trilinos-11.12.1-Source.tar.bz2"
-  sha256 "a41539414529c65905260b3befe3aee4f1dd1015ff719f0f6b5a10902d576fda"
+  url "https://trilinos.org/oldsite/download/files/trilinos-12.0.1-Source.tar.bz2"
+  sha256 "cab674e88c8ca2d2c54176af60030ed28203c0793f3c64c240363dbe7fa46b99"
   head "https://software.sandia.gov/trilinos/repositories/publicTrilinos", :using => :git
 
   option "with-teko",  "Enable the Teko secondary-stable package"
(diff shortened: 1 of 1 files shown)


trilinos.rb
```diff
@@ -150,7 +156,7 @@ def install
     # we still need to provide METIS_INCLUDE_DIRS so that metis.h is picked up on Linuxbrew.
     if build.with? "metis"
       args << "-DTPL_ENABLE_METIS:BOOL=ON"
-      args << "-DMETIS_LIBRARIES=#{Formula["metis"].opt_lib}/libmetis.a"
+      args << "-DTPL_METIS_LIBRARIES=#{Formula["metis"].opt_lib}/libmetis.dylib"
```
dpo: Linux has no dylibs.

trilinos.rb
```diff
@@ -170,7 +176,7 @@ def install
     if build.with? "parmetis"
       # Ensure CMake picks up METIS 5 and not METIS 4.
       args << "-DTPL_ENABLE_ParMETIS:BOOL=ON"
-      args << "-DParMETIS_LIBRARIES=#{Formula["parmetis"].opt_lib}/libparmetis.a;#{Formula["metis"].opt_lib}/libmetis.a"
+      args << "-DTPL_ParMETIS_LIBRARIES=#{Formula["parmetis"].opt_lib}/libparmetis.a;#{Formula["metis"].opt_lib}/libmetis.dylib"
```
dpo: Here you forgot to change libparmetis, but again Linux has no dylibs.

trilinos.rb
```diff
@@ -226,30 +234,86 @@ def install
 
 __END__
 diff --git a/cmake/TPLs/FindTPLUMFPACK.cmake b/cmake/TPLs/FindTPLUMFPACK.cmake
-index e26fc7e..f0b7ab6 100644
+index 963eb71..998cd02 100644
```
dpo: Were the patches submitted upstream?

davydden: yes

trilinos.rb
davydden: i'll test it later on linux to see if it's needed at all...

trilinos.rb
davydden: i didn't forgot to change libparemtis, it's indeed `libparmetis.a` which is installed

davydden:
tested on Linuxbrew, all good for me.

PR #2508, opened 2015-07-11
Scotch 6.0.4 (fix)

davydden:
update download link and sha. PATCH is removed as it is already applied in this official download link.

Unfortunately, there is a compilation error: 
https://gforge.inria.fr/tracker/index.php?func=detail&aid=19407&group_id=248&atid=1079

### DIFF
--- a/scotch.rb
+++ b/scotch.rb
@@ -1,8 +1,9 @@
 class Scotch < Formula
+  desc "A software package for graph and mesh/hypergraph partitioning, graph clustering, and sparse matrix ordering"
   homepage "https://gforge.inria.fr/projects/scotch"
-  url "https://gforge.inria.fr/frs/download.php/file/34099/scotch_6.0.4.tar.gz"
-  sha256 "6461cc9f28319a9dbe6cc10e28c0cbe90b4b25e205723c3edcde9a3ff974d6d8"
-  revision 1
+  url "https://gforge.inria.fr/frs/download.php/file/34618/scotch_6.0.4.tar.gz"
+  sha256 "f53f4d71a8345ba15e2dd4e102a35fd83915abf50ea73e1bf6efe1bc2b4220c7"
+  # revision 1
 
   bottle do
     root_url "https://homebrew.bintray.com/bottles-science"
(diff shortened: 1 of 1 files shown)


davydden:
adding `-DCOMMON_PTHREAD` solves the issue. The problem is that there is
```
typedef struct ThreadGroupHeader_ {
#if ((defined COMMON_PTHREAD) || (defined SCOTCH_PTHREAD))
  int                       flagval;              /*+ Thread block flags       +*/
  size_t                    datasiz;              /*+ Size of data array cell  +*/
  int                       thrdnbr;              /*+ Number of threads        +*/
  ThreadLaunchStartFunc     stafptr;              /*+ Pointer to start routine +*/
  ThreadLaunchJoinFunc      joifptr;              /*+ Pointer to join routine  +*/
  ThreadBarrier             barrdat;              /*+ Barrier data structure   +*/
#endif /* ((defined COMMON_PTHREAD) || (defined SCOTCH_PTHREAD)) */
} ThreadGroupHeader;
```
so if `COMMON_PTHREAD` or `SCOTCH_PTHREAD` are not defined, then this `struct` does not contain anything. Then accessing its attributes 
```
#ifdef GRAPHCOARSENTHREAD
  thrdnbr                  =
  coarptr->thrddat.thrdnbr = SCOTCH_PTHREAD_NUMBER; /* Required by graphMatchInit */
#else /* GRAPHCOARSENTHREAD */
  coarptr->thrddat.thrdnbr = 1;
#endif /* GRAPHCOARSENTHREAD */
```
gives the error, of course. 

Why it did not happen before - i don't know. I could speculate that there was a bug in configure system introduced between the current 6.0.4. and the original `6.0.4` release.

davydden:
@dpo i see that you have manually specified those `cflags` in https://github.com/Homebrew/homebrew-science/pull/1681/ i wonder if you recall how you decided on them?

dpo:
I seem to recall it was another bug of the same sort. Your latest version builds for me. Pull as is?

davydden:
i would say so... if there will be some resolution or comments from the authors in that bug error, i will update the formula.

PR #2511, opened 2015-07-12
Hypre 2.10.0b (update)

davydden:


hypre.rb
```diff
@@ -1,9 +1,8 @@
 class Hypre < Formula
   homepage "http://computation.llnl.gov/casc/hypre/software.html"
-  url "http://ftp.mcs.anl.gov/pub/petsc/externalpackages/hypre-2.9.0b.tar.gz"
-  mirror "ftp://ftp.mirrorservice.org/sites/distfiles.gentoo.org/distfiles/hypre-2.9.0b.tar.gz"
-  sha1 "aaf8354437be38aa2fbcc944ab64d464a5b39207"
-  revision 2
+  url "http://ftp.mcs.anl.gov/pub/petsc/externalpackages/hypre-2.10.0b.tar.gz"
+  mirror "ftp://ftp.mirrorservice.org/sites/distfiles.gentoo.org/distfiles/hypre-2.10.0b.tar.gz"
+  sha1 "ec664bb0e3358ff4ff39fa48d10896e497dddcdb"
```
dpo: We should use the `sha256` now.

hypre.rb
```diff
@@ -1,9 +1,8 @@
 class Hypre < Formula
   homepage "http://computation.llnl.gov/casc/hypre/software.html"
```
dpo: Do you want to add a `desc` field?

PR #2514, opened 2015-07-13
root6: add option to compile with RooFit/Minuit2

anuwish:
Added options to enable RooFit and Minuit2 support when compiling ROOT6.

### DIFF
--- a/root6.rb
+++ b/root6.rb
@@ -23,6 +23,9 @@ class Root6 < Formula
   conflicts_with "root"
 
   needs :cxx11
+  
+  option "with-minuit2", "Compile with Minuit2 support"
+  option "with-roofit",  "Compile with RooFit"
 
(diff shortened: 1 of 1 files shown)


tdsmith:
Should we build these by default?

anuwish:
Personally, I would be in favour of that. Though, following the ROOT developer's decision to not have RooFit and Minuit2 compiled by default, I think it's more "natural" to have them optional as well.

tdsmith:
Let's do it, then. Root's huge so if these just make extra libraries available, I'm in favor of forcing these on so that they're built into the precompiled bottles, unless you know a reason someone might not want them.

anuwish:
Ok, so I'll remove the "with-xyz" options and just add the additional flags for RooFit and Minuit2.

tdsmith:
Thanks for the pull request!

PR #2524, opened 2015-07-16
Genometools 1.5.6 (expose python library)

andrewjpage:
Install the python bindings for genometools

### DIFF
--- a/genometools.rb
+++ b/genometools.rb
@@ -22,6 +22,7 @@ class Genometools < Formula
   option "with-hmmer", "Build with HMMER (to enable protein domain search functionality in the ltrdigest tool)"
 
   depends_on "pkg-config" => :build
+  depends_on :python => :recommended unless OS.mac? && MacOS.version >= :lion
 
   if build.with? "pangocairo"
     depends_on "cairo"
@@ -40,9 +41,18 @@ class Genometools < Formula
     system "make", "install", *args
 
     prefix.install bin/"gtdata"
+
+    if build.with? "python"
+      cd "gtpython" do
+        inreplace "gt/dlload.py", "gtlib = CDLL(\"libgenometools\" + soext)", "gtlib = CDLL(\"#{lib}/libgenometools\" + soext)"
+        system "python", *Language::Python.setup_install_args(prefix)
+        system "python", "-m", "unittest", "discover", "tests"
+      end
+    end
   end
 
   test do
     system "#{bin}/gt", "-test"
+    system "python", "-c", "from gt import *" if build.with? "python"
   end
 end


genometools.rb
```diff
@@ -22,6 +22,7 @@ class Genometools < Formula
   option "with-hmmer", "Build with HMMER (to enable protein domain search functionality in the ltrdigest tool)"
 
   depends_on "pkg-config" => :build
+  depends_on :python
```
tdsmith: Use `depends_on :python if OS.linux? || MacOS.version <= :snow_leopard` to allow us to bottle this without creating an implicit dependency on Homebrew python.

genometools.rb
```diff
@@ -40,9 +41,18 @@ def install
     system "make", "install", *args
 
     prefix.install bin/"gtdata"
+
+    version = Language::Python.major_minor_version "python"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python#{version}/site-packages"
```
tdsmith: This path isn't used so you can probably get rid of this line.

genometools.rb
sjackman: Minor style issue, but I prefer the following logic:
```ruby
depends_on :python unless OS.mac? && MacOS.version >= :lion
```
Read it as "Mac OS versions Lion and above provide Python. Everyone else needs :python".

sjackman: I'd suggest
```ruby
depends_on :python => :recommended unless …
```

genometools.rb
```diff
@@ -40,9 +41,18 @@ def install
...
+    cd "gtpython" do
+      inreplace "gt/dlload.py", "gtlib = CDLL(\"libgenometools\" + soext)", "gtlib = CDLL(\"#{lib}/libgenometools\" + soext)"
+      system "python", *Language::Python.setup_install_args(prefix)
+      system "python", "-m", "unittest", "discover", "tests"
+    end
```
sjackman:
```ruby
end if build.with? "python"
```

genometools.rb
```diff
@@ -40,9 +41,18 @@ def install
...
   end
 
   test do
     system "#{bin}/gt", "-test"
+    system "python", "-c", "from gt import *"
```
sjackman: … if build.with? "python"

genometools.rb
tdsmith: In core we're preferring
```
if python
  cd foo
  end
end
```
over suffixing an if onto a block; I think the longer form is clearer, but I don't especially mind.

sjackman: I'm happy going with core for this style. Linuxbrew is a different case because I have to keep the diff small to prevent merge conflicts.

andrewjpage:
Thanks for the feedback. I've made the suggested changes.

sjackman:
Please squash both commits into a single commit.
See https://github.com/Homebrew/homebrew/blob/master/CONTRIBUTING.md
and https://github.com/Homebrew/homebrew-science/blob/master/CONTRIBUTING.md

andrewjpage:
Thanks, I've squashed it down to a single commit.

sjackman:
The commit log message is `install genometools python library`
it should be `genometools 1.5.6: Install Python library

sjackman:
Thanks for your contribution, Andrew.

genometools.rb
```diff
@@ -40,9 +41,18 @@ def install
...
+        inreplace "gt/dlload.py", "gtlib = CDLL(\"libgenometools\" + soext)", "gtlib = CDLL(\"#{lib}/libgenometools\" + soext)"
+        system "python", *Language::Python.setup_install_args(prefix)
+        system "python", "-m", "unittest", "discover", "tests"
+      end
+    end
   end
 
   test do
     system "#{bin}/gt", "-test"
+    system "python", "-c", "from gt import *" if build.with? "python"
```
sjackman: This line does not work because `build.with? …` is always false within `test`.

PR #2530, opened 2015-07-18
Transition from lesstif to openmotif

mistydemeo:
We're currently transitioning all formulae from the official taps that use lesstif to openmotif. homebrew-x11 and core are both done, so these two formulae are the last two remaining.

Lesstif is no longer maintained, and conflicts with the LGPL openmotif; at this point I believe it's better to have everything build against the same motif library.


mistydemeo:
@BrewTestBot test this please

PR #2531, opened 2015-07-20
bioformats-cpp 5.1.3 (update)

rleigh-dundee:
Update the source file being built; no other formula changes.


tdsmith:
Thanks for the PR!

PR #2565, opened 2015-07-25
fwdpp 0.3.5

vancleve:


fwdpp.rb
```diff
@@ -1,7 +1,7 @@
 class Fwdpp < Formula
   homepage "https://molpopgen.github.io/fwdpp/"
```
dpo: Could you please add a `desc` field with a brief description?

dpo:
Thank you!

PR #2280, opened 2015-05-19
nonpareil 2.4.01

lmrodriguezr:

Issue #2559, opened 2015-07-24
Homebrew should support formulae that install R packages


lmrodriguezr:
The Nonpareil bottle doesn't install the R Nonpareil package.

tdsmith:
As discussed at https://github.com/Homebrew/homebrew-science/pull/2567#issuecomment-125051191, this is because the nonpareil formula calls `R CMD INSTALL`, which has the effect of writing outside of nonpareil's prefix, so it isn't picked up in the bottle (and breaks Homebrew policy).

Fixing R packaging in Homebrew will be a lot of work. We can plausibly remove the nonpareil bottle in the interim if the package isn't useful without the R component.

lmrodriguezr:
I think removing the bottle is the most sensible solution. Compilation time and resources for Nonpareil are small, and that way the R package can still be installed.

Out of curiosity, is there a better way to do it? I could write a patch for the Nonpareil installer, but I cannot think of a way to fix this that doesn't involve writing outside of nonpareil's prefix.

tdsmith:
> Out of curiosity, is there a better way to do it?

I expect that there is. We configure Python so that its default module search path is relative to HOMEBREW_PREFIX and not Python's private prefix, so that it searches in e.g. /usr/local/lib/python2.7/site-packages and not /usr/local/Cellar/python/2.7.10/lib/python2.7/site-packages. This means that if a formula installs /usr/local/Cellar/some_formula/1.2.3/lib/python2.7/site-packages/foo.py, it gets symlinked into /usr/local/lib/python2.7/site-packages, and python finds it just fine.

"A lot of work" was probably an overstatement; someone needs to figure out which switches to feed R at build time and what we can do to influence INSTALL's behavior, and we need to make a quick change to the Homebrew symlinking code. Depending on R's release schedule, it should make sense to make those changes with the next R minor release since packages don't normally persist across minor releases.

tdsmith:
Minor releases are scheduled for spring: https://developer.r-project.org/ so we probably don't want to wait that long.

Doing this on a micro release would also make a lot of sense, since, right now, packages are installed to R's private prefix by default and are lost on any version increment.

dpo:
At the moment, uninstalling `nonpareil` doesn't uninstall its R bindings.

tdsmith:
[...]

PR #2567, opened 2015-07-26
nonpareil v2.4.01

lmrodriguezr:
Closes #2280.

### DIFF
--- a/nonpareil.rb
+++ b/nonpareil.rb
@@ -7,6 +7,12 @@ class Nonpareil < Formula
   sha256 "ca5955e877098ed4a679404ac87635e28a855d15d6970ca51a6be422266c0999"
   head "https://github.com/lmrodriguezr/nonpareil.git"
 
+  bottle do
+    cellar :any
+    sha1 "9fa1b02ccc8d561c9586b07d854c71227a77739e" => :yosemite
+    sha1 "cd8a9999c84c41fc39aaa7f62cc1ab44a3a26300" => :mavericks
+    sha1 "11adac7a92b44e5fd171bc1684a8b76fdceb848f" => :mountain_lion
+  end
   depends_on "r"
   depends_on :mpi => [:cxx, :optional]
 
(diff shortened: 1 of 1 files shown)


dunn:
Looks good but for a nitpick from the bot:
```
homebrew/science/nonpareil:
 * 5: Trailing whitespace was found
 * 8: Trailing whitespace was found
```

dpo:
Does this also fix #2559?

lmrodriguezr:
This change removes the bottle because it's a different base version, but I don't know if an eventual (new) bottle would have the same problem. Is there a way to (1) add libraries in other packages from within a bottle, or (2) avoid bottle creation?

dpo:
Where is the r package supposed to be installed?

lmrodriguezr:
That's directly controlled by R. This formula requires the r formula, but it could be superseded by another R executable happening first in the PATH. Even if the r formula is being used, it can go to the Cellar or to a folder like `~/Library/R/3.2/library` (if it exists).

This is the line installing the R library: `$(R) CMD INSTALL utils/Nonpareil` (https://github.com/lmrodriguezr/nonpareil/blob/master/Makefile#L45).

dpo:
In that case the r package probably ends up somewhere in r's folder structure or a user folder and there's no reason it should  be part of the `nonpareil` bottle. Perhaps it should be made `:optional`.

lmrodriguezr:
But the nonpareil results are nearly meaningless without the R package, so a regular user experience installing Nonpareil from Homebrew would involve manually downloading the code from github just to install the R package. Alternatively, it would involve using the `--build-from-source` flag, in which case not bottling nonpareil would be a better solution than telling the users to build from source in the formula caveats.

sjackman:
> This change removes the bottle because it's a different base version

Don't remove the bottle lines. Bottles are rebuilt automatically by the BrewTestBot.

lmrodriguezr:
So should I revert the bottle removal and continue the bug discussion over at #2559 ?

dpo:
Yes.

Issue #2557, opened 2015-07-23
root fails to build on 10.10.4


lucasc896:
MacBook-Pro:~ chrislucas$ brew install --with-x11 root
==> Installing root from homebrew/homebrew-science
==> Installing dependencies for homebrew/science/root: openssl, xrootd, gsl
==> Installing homebrew/science/root dependency: openssl
==> Downloading https://homebrew.bintray.com/bottles/openssl-1.0.2d_1.yosemite.bottle.tar.gz
######################################################################## 100.0%
==> Pouring openssl-1.0.2d_1.yosemite.bottle.tar.gz
==> Caveats
A CA file has been bootstrapped using certificates from the system
keychain. To add additional certificates, place .pem files in
  /usr/local/etc/openssl/certs

and run
  /usr/local/opt/openssl/bin/c_rehash

This formula is keg-only, which means it was not symlinked into /usr/local.

OS X already provides this software and installing another version in
parallel can cause all kinds of trouble.

Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries

Generally there are no consequences of this for you. If you build your
own software and it requires this formula, you'll need to add to your
build variables:

    LDFLAGS:  -L/usr/local/opt/openssl/lib
    CPPFLAGS: -I/usr/local/opt/openssl/include

==> Summary
🍺  /usr/local/Cellar/openssl/1.0.2d_1: 464 files, 18M
==> Installing homebrew/science/root dependency: xrootd
==> Downloading https://homebrew.bintray.com/bottles/xrootd-4.1.1.yosemite.bottle.1.tar.gz
######################################################################## 100.0%
==> Pouring xrootd-4.1.1.yosemite.bottle.1.tar.gz
🍺  /usr/local/Cellar/xrootd/4.1.1: 214 files, 7.4M
==> Installing homebrew/science/root dependency: gsl
==> Downloading https://homebrew.bintray.com/bottles/gsl-1.16.yosemite.bottle.2.tar.gz
######################################################################## 100.0%
==> Pouring gsl-1.16.yosemite.bottle.2.tar.gz
🍺  /usr/local/Cellar/gsl/1.16: 245 files, 8.2M
==> Installing homebrew/science/root
==> Downloading ftp://root.cern.ch/root/root_v5.34.26.source.tar.gz
######################################################################## 100.0%
[...]

PR #2569, opened 2015-07-26
Update root version - v5-34-32

lucasc896:
Updated root version as per recommendation of the issue:
https://github.com/Homebrew/homebrew-science/issues/2557

Root failed to build for the previous version. v5-34-32 is the latest stable build of root5.

I've added a description (taken from root.cern.ch), updated the version, removed the mirror (which no longer exists) and updated to the recommended hash algorithm, sha256.

### DIFF
--- a/root.rb
+++ b/root.rb
@@ -1,9 +1,9 @@
 class Root < Formula
+  desc "Root object oriented framework for large scale data analysis"
   homepage "http://root.cern.ch"
-  version "5.34.26"
-  sha1 "f9013c37c37946b79dce777d731ccb64e5b28bb8"
+  version "5.34.32"
+  sha256 "939c7592802a54b6cbc593efb6e51699bf52e92baf6d6b20f486aaa08480fc5f"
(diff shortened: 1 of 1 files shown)


root.rb
```diff
@@ -1,9 +1,9 @@
 class Root < Formula
+  desc "Root object oriented framework for large scale data analysis"
   homepage "http://root.cern.ch"
-  version "5.34.26"
-  sha1 "f9013c37c37946b79dce777d731ccb64e5b28bb8"
+  version "5.34.32"
+  sha256 "939c7592802a54b6cbc593efb6e51699bf52e92baf6d6b20f486aaa08480fc5f"
   url "ftp://root.cern.ch/root/root_v#{version}.source.tar.gz"
-  mirror "http://ftp.riken.jp/pub/ROOT/root_v#{version}.source.tar.gz"
   head "https://github.com/root-mirror/root.git", :branch => "v5-34-00-patches"
```
dpo: Does `head` need to be updated?

lucasc896: I believe the v5-34-00-patches branch is still the head, with commits still being made.

PR #2571, opened 2015-07-27
bump to most recent Astrometry.net release

dstndstn:


astrometry-net.rb
```diff
@@ -2,8 +2,8 @@
 
 class AstrometryNet < Formula
   homepage "http://astrometry.net"
-  url "http://astrometry.net/downloads/astrometry.net-0.50.tar.gz"
-  sha1 "fcf6332b603e1af8f25ce7d6996b5a4b0eeaa687"
+  url "http://astrometry.net/downloads/astrometry.net-0.56.tar.gz"
+  sha1 "7d9bf05398944e0e266e60fe5eaf59791186cb74"
```
dpo: Could you please use `sha256` instead of `sha1`? You can find out the hash with `shasum -a 256`. Thanks!

dpo: A `desc` field would also be great!

PR #2572, opened 2015-07-27
insighttoolkit 4.8.0

iMichka:
This also brings back the --with-python option on OS X, after ~1 year of work from different people from the ITK community :)


insighttoolkit.rb
```diff
@@ -1,7 +1,7 @@
 class Insighttoolkit < Formula
   homepage "http://www.itk.org"
-  url "https://downloads.sourceforge.net/project/itk/itk/4.7/InsightToolkit-4.7.2.tar.gz"
-  sha1 "9f05222b79682a438799ebf90cf6e2db2fe94535"
+  url "https://downloads.sourceforge.net/project/itk/itk/4.8/InsightToolkit-4.8.0.tar.gz"
+  sha1 "0ec3f722377c97e1696d89b757b34ae02f7f98c5"
```
dpo: We use the `sha256` now. It would also be great if you could add a `desc` field, with a brief description of the software.

iMichka:
Done. I also enabled the Review module by default. SimpleITK needs ITK compiled with the review module ... when SimpleITK is installed it will re-compile ITK and not use the ITK bottes, which is bad. Compilation time should be 1 or 2 minutes longer (the whole thing compiled in 9,2 min on my macbook pro).

tdsmith:
This looks reasonable, thanks!

It looks like you work on ITK? This is fine for now but I'd prefer if the Python modules were linked with `-undefined dynamic_lookup` instead of with `-lpython`; then, we can bottle against system Python 2.7 and get an extension module that works with any binary-compatible Python 2.7 (like Homebrew's).

While I'm looking at this, is insighttoolkit really only a build dependency of simpleitk?

iMichka:
Yes I work on ITK regularly, but as it's a huge project I do not know all the details.

I think using ```-undefined dynamic_lookup``` sounds reasonable. I'll need to implement this and test it a little bit. This could then be added for the next ITK release (4.8.1). If we could have ITK bottles with Python wrapping; that would be great. I do not know if homebrew's buildbots will be OK with this though as building time is around 1h.

ITK is not only a build dependency of simpleITK. ITK is the main C++ project. Then comes the ITK Python wrappings, which cover most of the classes in ITK. SimpleITK came long after all this and is another way to access ITK through python (with less classes covered and less functionalities).

In theory simpleITK can be built using the Superbuild mode, where it builds ITK internally, so we could remove the dependency here. I never tried to implement this in homebrew and don't know if this would be  wanted.

tdsmith:
Wow, the python bindings add 50 minutes?

Pulled this in 39c3ea863; thanks for the PR! I also fixed up the simpleitk dependency on insighttoolkit in 2fba474.

iMichka:
Yes the wrapping building is quite long. One of ITK's dependencies for this is CastXML, which I hope we can soon provide it through homebrew (I am preparing a formula for this tool). Once this is done, we may gain 15 to 20 minutes. Thanks for the merge. I'll also test the ```-undefined dynamic_lookup``` flag.

iMichka:
Hi again. I looked at the ```-undefined dynamic_lookup```. Could you help me here ? I did not really understand what you wanted to achieve. I looked at build logs on OS X 10.10 for ITK 4.8 and could find no trace of the ```-lpython``` flag.

I tried to use the ```-undefined dynamic_lookup``` flag. ITK worked with system python but I got the classical error (```Fatal Python error: PyThreadState_Get: no current thread Abort trap: 6```) with brewed python; (which happens if you have a python library mismatch). Maybe I need also need to tell cmake to use the system's python paths (```DPYTHON_LIBRARY```and ```DPYTHON_INCLUDE_DIR```) ?

I was not able to find some good documentation about ```-undefined dynamic_lookup``` (expect some discussions on stack overflow ...). Are there some good explanations about this somewhere ?

tdsmith:
Perhaps Python is linked with `-framework Python`? What purpose is `"-DPYTHON_LIBRARY='#{%x(python-config --prefix).chomp}/lib/libpython2.7.dylib'"` serving? Do you have a copy of the build logs you can share? I admit I haven't actually tried to build this myself; I assumed because of that argument that cmake must be using it directly.

The goal is to build an extension module .so that isn't linked to a particular Python framework, i.e. `otool -L module.so` won't show a link to anything named "python".

iMichka:
Okay I think I got it. ```PYTHON_LIBRARY``` was passed to the linker. I removed it and added the ```-undefined dynamic_lookup``` flag, and it worked. ```otool``` does no more show the python library. And itk can be imported and used from OS X's python or homebrew's python.

I made a patch which can be reviewed here: http://review.source.kitware.com/#/c/20072/1

If everything works out, we will merge this for the next release (4.8.1) which will come out soon.

iMichka:
Hi again. Could you have a look at the discussion on the patch ? We found following out:

```
Dlul:~ michkapopoff$ otool -L /usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/Python
/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/Python:
	/usr/local/Frameworks/Python.framework/Versions/2.7/Python (compatibility version 2.7.0, current version 2.7.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
```

The install_names seem to not match. Not sure how that influences our problem here.

PR #2585, opened 2015-07-30
libfolia: new formula

dpo:
Closes #1470.


dpo:
@BrewTestBot test this please

PR #2586, opened 2015-07-30
bowtie2: configurable TBB support

chapmanb:
The new support for TBB in bowtie2 breaks builds on some linuxbrew
machines for two reasons:

- If the brew prefix is not /usr/local, the build fails to find TBB headers and libraries.
- TBB does not build cleanly with some older gcc versions.

This fixes the first issue and provides a '--without-tbb' flag to skip
TBB in cases where we have build issues.


tseemann:
Thanks Brad.

Issue #2577, opened 2015-07-28
Building gal-sim on 10.10.4 and it can't find FFTW3


josePhoenix:
```
Checking for C++ header file fftw3.h... yes
Checking for correct FFTW linkage... 
Error: fftw file failed to link correctly
Check that the correct location is specified for FFTW_DIR

Please fix the above error(s) and rerun scons.
Note: you may want to look through the file INSTALL.md for advice.
Also, if you are having trouble, please check the INSTALL FAQ at 
   https://github.com/GalSim-developers/GalSim/wiki/Installation%20FAQ
```

I can't find anything relevant in the GalSim installation FAQ, and I can't figure out how to force HomeBrew to force scons to force the compiler to pass `-L/absolute/path/to/fftw/lib`.

Here's a [gist](https://gist.github.com/josePhoenix/b3ba45d6df08a10c4727) with:

  * [the output as seen in the terminal](https://gist.github.com/josePhoenix/b3ba45d6df08a10c4727#file-gistfile1-txt)
  * [the scons output](https://gist.github.com/josePhoenix/b3ba45d6df08a10c4727#file-01-scons)
  * [the scons config log](https://gist.github.com/josePhoenix/b3ba45d6df08a10c4727#file-config-log)

I tried adding `ENV['LDFLAGS']='-L/Users/jlong/homebrew/Cellar/fftw/3.3.4_1/lib'` to the `gal-sim.rb` formula, but that didn't appear to be passed to the scons configure process, even though you can see it in the `LDFLAGS` in the scons output.

dpo:
It seems `fftw` was installed earlier. Could you try `brew reinstall fftw` and then `brew install gal-sim`?

josePhoenix:
That's one of the things I tried, but the error is the same. I'm not familiar with how Homebrew exposes the include files / libraries to the build process, but `-Iinclude` seems to be enough to find fftw3 while `-Llib` doesn't find the library.

dstndstn:
As an aside, I just updated the GalSim formula to pick up the new GalSim version v1.3.0 -- that formula is here, and there is a pull request to update homebrew-science.  It might be worth checking whether version 1.3.0 builds correctly...
https://github.com/camphogg/homebrew-science/blob/bump-gal-sim/gal-sim.rb
https://github.com/Homebrew/homebrew-science/pull/2587


Anyway, could you please send the output from running:

ls -l /usr/local/lib/libfftw3*

[...]

PR #2588, opened 2015-07-30
GalSim: fix build when homebrew is not in /usr/local or other default search path

dstndstn:
As reported in #2577 ; this version is reported to work by OP

### DIFF
--- a/gal-sim.rb
+++ b/gal-sim.rb
@@ -31,7 +31,7 @@ class GalSim < Formula
       args << "WITH_OPENMP=true"
     end
     scons *args
-    scons "install", "PREFIX=#{prefix}", "PYPREFIX=#{lib}/python#{pyver}"
+    scons "install", "PREFIX=#{prefix}", "PYPREFIX=#{lib}/python#{pyver}", "EXTRA_INCLUDE_PATH=#{HOMEBREW_PREFIX}/include", "EXTRA_LIB_PATH=#{HOMEBREW_PREFIX}/lib", "EXTRA_PATH=#{HOMEBREW_PREFIX}/bin"
   end
 
   def caveats; <<-EOS.undent
--- a/gal-sim.rb
+++ b/gal-sim.rb
@@ -30,8 +30,8 @@ class GalSim < Formula
       end
       args << "WITH_OPENMP=true"
     end
-    scons *args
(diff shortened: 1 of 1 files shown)


gal-sim.rb
```diff
@@ -30,7 +30,7 @@ def install
       end
       args << "WITH_OPENMP=true"
     end
-    scons *args
+    scons "EXTRA_INCLUDE_PATH=#{HOMEBREW_PREFIX}/include", "EXTRA_LIB_PATH=#{HOMEBREW_PREFIX}/lib", "EXTRA_PATH=#{HOMEBREW_PREFIX}/bin", *args
```
dpo: Why not just add `Formula["fftw3"].opt_include`, etc., instead of `HOMEBREW_PREFIX`?

dstndstn:
That might be an option.  There are scons variables for FFTW, TMV, and Boost paths.  Presumably scons can find the correct python include path.  (And the fftw path should be found with pkg-config anyway.)  I'm not sure I prefer this, though -- since *all* the homebrew headers are installed in a nonstandard place, to me it makes sense to tell scons that.  That seems preferable to telling it that FFTW, TMV, and Boost are each to be found in /path/to/homebrew.

(Also, I don't have a strange homebrew install like this for testing, and I'm not sure the OP is available for testing.  But that's just an excuse!)

thx

dpo:
There's a slight inconsistency about this formula. The formula name is `GalSim` which means that the destination folder in the cellar is `gal-sim`. Yet `scons install` installs stuff in `share/galsim`, not in `share/gal-sim`. The [Github site](https://github.com/GalSim-developers/GalSim) is called `GalSim`, not `Gal-Sim`, so shouldn't we really rename this formula `Galsim` and the formula file `galsim.rb`.

If I try to install the example scripts (which aren't installed by default) using `pkgshare.install`, they end up in `share/gal-sim`, so we have both `share/galsim` and `share/gal-sim`...

gal-sim.rb
dpo: This works for me: https://gist.github.com/dpo/931cc097d7385194edac
I also added tests and installed the examples.

iml:
Might want to wait for the formula renames work to be merged before renaming, otherwise it will strand users with the old formula installed. Could use `(share/"galsim")` in the meantime.

dstndstn:
Hmm, I wanted to call the formula galsim when I submitted it, but somehow I
thought the naming rules were such that CamelCase names got converted to
dashes.  I no longer see a rule like that, so maybe I was mistaken.  Also
"brew create" with the download URL creates a formula named galsim.rb.  I
would suggest that you submit a pull request to make that name change :)

PR #2589, opened 2015-07-30
Added BioInterchange formula.

indiedotkim:
I had previously submitted a [similar pull request](https://github.com/Homebrew/homebrew-binary/pull/255) for Homebrew-binary.

@sjackman and @mistydemeo suggested that Homebrew-science might be better suited. So, here I am submitting a BioInterchange formula for binaries that work under OS X and Linux.

To be honest, I could not figure out how to run the test section. It *should* work, but I do not know how to run it.

Thanks!


biointerchange.rb
```diff
@@ -0,0 +1,24 @@
+class Biointerchange < Formula
+  desc "BioInterchange: a tool for scaling and integrating genomics data"
+  homepage 'https://www.codamono.com'
```
sjackman: Please use double quotes `"` throughout. Also add
```ruby
# tag "bioinformatics"
```

biointerchange.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  if OS.mac?
+    version '2.0.1+74'
+    url 'https://www.codamono.com/brew/osx-10-10/biointerchange-2.0.1+74.tar.gz'
+    sha1 '1d33ec6757e37ccbae860f57eeb2a1e5f6bb2f76'
```
sjackman: Please use `sha256` rather than `sha1`, which is deprecated.

biointerchange.rb
```diff
@@ -0,0 +1,24 @@
...
+  elsif OS.linux?
+    version '2.0.1+75'
+    url 'https://www.codamono.com/brew/debian-8-amd64/biointerchange-2.0.1+75.tar.gz'
+    sha1 'fa2cc1414c823cb95c34298206a819f122884d77'
+  end
+
+  def install
+    bin.install 'biointerchange'
+    doc.install 'license.txt'
```
sjackman: Is there a README or a manual to install?

biointerchange.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  test do
+    assert_match "#{version}", shell_output("biointerchange -v", 0)
```
sjackman: You can omit `, 0`. It's the default value.

biointerchange.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+end
+
```
sjackman: Please delete this blank line.

indiedotkim:
@sjackman Looks better now?

indiedotkim:
@sjackman Oh -- and there is no README as such. Documentation can be found online.

sjackman:
Yep, looks good. It's a bit odd that Mac and Linux have different versions. Could they be the same? It'd also be nice if the version number format were such that it were automatically detected by Homebrew. It's probably the `+` that causing it get confused. Would you consider replacing the `+` with a `.`? See [Semantic Versioning](http://semver.org/) for good recommendations for version numbers.

Please squash the two commits into a single commit.

indiedotkim:
@sjackman I am using the number behind the "+" to indicate the build and it is different for OSX/Linux because I am changing a constant in the actual C code between the different builds. The "+NUMBER" notation itself is already taken from Semantic Versioning: see point "10. Build metadata [...]".

Squashing now...

sjackman:
Well there you are. I guess I never read that far done. =)

sjackman:
I'm ignoring this:
```
==> audit problems
homebrew/science/biointerchange:
Length is calculated as homebrew/science/biointerchange + desc. (currently 97)
```

sjackman:
Homebrew normally hosts bottles (binary packages) on BinTray. Would you like that, or not?

sjackman:
See https://bintray.com/homebrew/bottles-science/biointerchange#files
I'll take these down if you prefer.

indiedotkim:
@sjackman Thanks! I will fix the description with the next release. I pondered about Bintray... could you remove them for now? It is easier for me to track downloads/usage when I host the binaries myself. I might change my opinion about that in the future, but for now, I would rather have them coming from my server.

sjackman:
Done, and merged. The preferred commit message is `biointerchange 2.0.1+74 (new formula)`

indiedotkim:
@sjackman Awesome! Thanks!

So, the next update would need to have a commit message like "biointerchange 2.0.2+XX"?

sjackman:
Yep, perfect.

mistydemeo:
:sparkles:

Issue #2593, opened 2015-08-01
pcl install failed


ahundt:
I ran:

    brew install pcl --with-examples --with-openni2 --HEAD

[Here is the gist log](https://gist.github.com/3a9db53e17117f41be4d)

ahundt:
apparently this is due to qt5

```

CMake Error at apps/modeler/CMakeLists.txt:100 (QT4_WRAP_UI):
  Unknown CMake command "QT4_WRAP_UI".

```

ahundt:
I've confirmed [pcl supports qt5](https://github.com/PointCloudLibrary/pcl/blob/master/CMakeLists.txt#L371)

ahundt:
[This line has to change based on Qt 4 or 5](https://github.com/Homebrew/homebrew-science/blob/master/pcl.rb#L100)

tdsmith:
This is because cmake is preferring qt5 and searching /usr/local/bin explicitly and you have qt5 linked, which is an unsupported configuration: https://gist.github.com/anonymous/3a9db53e17117f41be4d#file-doctor-out-L1077-L1090

I recommend unlinking qt5 before installing pcl.

tdsmith:
(with `brew unlink qt5`)

ahundt:
I have other dependencies that require qt5, unfortunately. I might have fixed it with the pull request, although [i'm not sure if line 155 is right](https://github.com/ahundt/homebrew-science/commit/fd1d3e65b107231fad4a1a458e639d9022fcb6d4#diff-b4c927d4ea039e97b99094678c67fe73R155).

tdsmith:
Unlinking qt5 from /usr/local will not uninstall it and will not affect builds of Homebrew formulas; `brew info qt5` provides advice for building external software against an unlinked qt5.

ahundt:
I'm a developer and when I use qt, I use qt5. So unfortunately it would adversely affect me to unlink. :-)

tdsmith:
You could of course relink it after installing pcl.

ahundt:
I think with these commits it does the right thing

ahundt:
There is also an issue where head and the current specified version differ, so the patch needs to be changed depending on the version. Is that currently possible with homebrew?

I've also created a related [issue to pull this upstream into pcl](https://github.com/PointCloudLibrary/pcl/issues/1288).

ahundt:
I've re-patched it again to handle another small bug. Should be good now.

[...]

PR #2595, opened 2015-08-01
#2593 pcl install failed, adding qt5 support

ahundt:
https://github.com/Homebrew/homebrew-science/issues/2593

### DIFF
--- a/pcl.rb
+++ b/pcl.rb
@@ -66,8 +66,8 @@ class Pcl < Formula
   option "with-examples", "Build pcl examples."
   option "without-tools", "Build without tools."
   option "without-apps", "Build without apps."
-  option "with-qt5", "Build with Qt5"
-  option "without-qt", "Build without Qt4"
+  depends_on "qt" => :recommended
+  depends_on "qt5" => :optional
 
   depends_on "cmake" => :build
   depends_on "pkg-config" => :build
@@ -79,8 +79,7 @@ class Pcl < Formula
 
   depends_on "qhull"
   depends_on "libusb"
-  depends_on "qt" => :recommended
-  depends_on "qt5" => :optional
+
   if build.with?("qt") || build.with?("qt5")
     depends_on "sip" # Fix for building system
     depends_on "pyqt" # Fix for building system
--- a/pcl.rb
+++ b/pcl.rb
@@ -66,8 +66,8 @@ class Pcl < Formula
   option "with-examples", "Build pcl examples."
   option "without-tools", "Build without tools."
   option "without-apps", "Build without apps."
-  depends_on "qt" => :recommended
-  depends_on "qt5" => :optional
+  option "with-qt5", "Build with Qt5"
+  option "without-qt", "Build without Qt4"
 
   depends_on "cmake" => :build
   depends_on "pkg-config" => :build
@@ -79,7 +79,8 @@ class Pcl < Formula
 
   depends_on "qhull"
   depends_on "libusb"
-
+  depends_on "qt" => :recommended
+  depends_on "qt5" => :optional
   if build.with?("qt") || build.with?("qt5")
     depends_on "sip" # Fix for building system
     depends_on "pyqt" # Fix for building system
--- a/pcl.rb
+++ b/pcl.rb
@@ -192,11 +192,11 @@ index 2f0425e..0675a55 100644
 -endif()
 -
  if(MSVC11)
-   # Setting this to true brakes Visual Studio builds.
-   set(CUDA_ATTACH_VS_BUILD_RULE_TO_CUDA_FILE OFF CACHE BOOL "CUDA_ATTACH_VS_BUILD_RULE_TO_CUDA_FILE")
+ 	# Setting this to true brakes Visual Studio builds.
+ 	set(CUDA_ATTACH_VS_BUILD_RULE_TO_CUDA_FILE OFF CACHE BOOL "CUDA_ATTACH_VS_BUILD_RULE_TO_CUDA_FILE")
 @@ -47,10 +37,5 @@ if(CUDA_FOUND)
-   include(${PCL_SOURCE_DIR}/cmake/CudaComputeTargetFlags.cmake)
-   APPEND_TARGET_ARCH_FLAGS()
+ 	include(${PCL_SOURCE_DIR}/cmake/CudaComputeTargetFlags.cmake)
+ 	APPEND_TARGET_ARCH_FLAGS()
 
 -  # Send a warning if CUDA_HOST_COMPILER is set to a compiler that is known
 -  # to be unsupported.
--- a/pcl.rb
+++ b/pcl.rb
@@ -66,6 +66,8 @@ class Pcl < Formula
   option "with-examples", "Build pcl examples."
   option "without-tools", "Build without tools."
   option "without-apps", "Build without apps."
+  option "with-qt5", "Build with Qt5"
+  option "without-qt", "Build without Qt4"
 
   depends_on "cmake" => :build
   depends_on "pkg-config" => :build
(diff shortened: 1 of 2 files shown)


tdsmith:
The bot choked on the first commit; kicked off a rebuild just now.

ahundt:
yeah I had squashed it down to one commit for simplicity, guess it didn't handle that well?

pcl.rb
```diff
@@ -78,10 +80,15 @@ class Pcl < Formula
   depends_on "qhull"
   depends_on "libusb"
   depends_on "qt" => :recommended
-  if build.with? "qt"
+  depends_on "qt5" => :optional
```
dpo: This is redundant with line 69. Similarly, line 70 should just be `depends_on "qt" => :recommended`.

ahundt: Sorry but I don't totally understand what to do here per your suggestion. I don't know ruby very well so I just modified what's here in a way I knew would work. Could you elaborate on how to make your suggested improvement?

dpo: Sorry, I'm saying that `option "with-qt5", "Build with Qt5"` and `depends_on "qt5" => :optional` are redundant. I'd rather only keep the latter. Similarly, `option "without-qt", "Build without Qt4"` and `depends_on "qt" => :recommended` are also redundant.

ahundt:
I believe I have made the change suggested by @dpo. Is anything else required to merge?

ahundt:
@dpo I apologize but it seems I misunderstood the changes you advised and broke the build when attempting to make the change you suggested. Could you perhaps give me a patch to apply or explain again?

dpo:
Your changes look fine (though the `depends_on` should have stayed where they were). It seems there was a bot issue. I'll kick off a new build.

ahundt:
hmm I'm having a tough time getting back to a version that works

dpo:
Here's the error
```
-- Installing: /usr/local/Cellar/cmake/3.3.1/share/cmake/Modules/FindCMinpack.cmake
CMake Error at cmake/cmake_install.cmake:43 (file):
  file INSTALL cannot copy file
  "/tmp/cminpack20150827-88183-1dt9mst/cminpack-1.3.4/cmake/FindCMinpack.cmake"
  to "/usr/local/Cellar/cmake/3.3.1/share/cmake/Modules/FindCMinpack.cmake".
  cmake_install.cmake:47 (include)

make: *** [install] Error 1
==> Sandbox profile:
(version 1)
(debug deny) ; log all denied operations to /var/log/system.log
(allow file-write* (subpath "/private/tmp"))
(allow file-write* (subpath "/private/var/tmp"))
(allow file-write* (regex #"^/private/var/folders/[^/]+/[^/]+/[C,T]/"))
(allow file-write* (subpath "/private/tmp"))
(allow file-write* (subpath "/Library/Caches/Homebrew"))
(allow file-write* (subpath "/Users/brew/Jenkins/science/version/mavericks/logs/cminpack"))
(allow file-write* (subpath "/Users/brew/Library/Developer/Xcode/DerivedData"))
(allow file-write* (subpath "/usr/local/Cellar/cminpack"))
(allow file-write* (subpath "/usr/local/etc"))
(allow file-write* (subpath "/usr/local/var"))
(allow file-write*
    (literal "/dev/ptmx")
    (literal "/dev/dtracehelper")
    (literal "/dev/null")
    (regex #"^/dev/fd/[0-9]+$")
    (regex #"^/dev/ttys?[0-9]*$")
    )
(deny file-write*) ; deny non-whitelist file write operations
(allow default) ; allow everything else
```
I'm afraid I can't help much. `pcl` doesn't even begin to build on my machine.

ahundt:
I've identified the problem, there needs to be a way to make the following directory writable:
`/usr/local/Cellar/cmake/3.3.1/share/cmake/Modules/`

It isn't in the list of `allow file-write*` in the comment you posted. Should I create a separate issue to ask about this?

tdsmith:
@BrewTestBot test this please

ahundt:
@xu-cheng seemed to explain what to do in https://github.com/Homebrew/homebrew-science/issues/2696, could I possibly get assistance with pushing this through so the issue can be merged?

ahundt:
Ok it seems since #2701 was fixed, @tdsmith tried re-running this to see if it passed since no files from a dependency would be installed in the wrong place. Unfortunately there is still a problem that remains and needs to be investigated/resolved.

ahundt:
hmm seems to still be failing on cminpack if I'm reading the build errors correctly, should #2701 be reopened?

dunn:
Just an audit failure as far as I can tell:
```
        Error: 13 problems in 1 formula
homebrew/science/pcl:

File: pcl.rb [hunk 2]
```diff
@@ -160,8 +179,8 @@ def install
 +++ b/cmake/pcl_find_cuda.cmake
 @@ -1,16 +1,6 @@
  # Find CUDA
- 
- 
+
```
dunn: Keep the whitespace in the patch, removing it can sometimes prevent it from applying.

ahundt: oh whoops removed those tabs too, well let's see if it applies correctly in the build. I can put the tabs from the patch back if needed

dunn: Oh, if some of the tabs are in the patch then leave those, yeah!

ahundt: seems to have worked!

pcl.rb
```diff
@@ -173,16 +192,16 @@ def install
 -endif()
 -
  if(MSVC11)
- 	# Setting this to true brakes Visual Studio builds.
- 	set(CUDA_ATTACH_VS_BUILD_RULE_TO_CUDA_FILE OFF CACHE BOOL "CUDA_ATTACH_VS_BUILD_RULE_TO_CUDA_FILE")
+   # Setting this to true brakes Visual Studio builds.
```
dunn: Sorry, I meant leave the inline patch as-is.

ahundt: yeah I knew that but I had already committed the change, seems to work anyway so probably good

ahundt:
What's left to merge this? Thanks!

pcl.rb
dunn: The patch is only for HEAD builds, so the bot won't use it.  Can you try building HEAD locally and see if the patch applies?

ahundt: It seems that even if I revert back to the tabs in the patch it doesn't apply. I reverted the specific lines and ran the following:

```
± brew reinstall pcl --with-qt5 --with-openni2 --with-examples
==> Reinstalling homebrew/science/pcl with --with-qt5, --with-openni2, --with-examples
==> Cloning https://github.com/PointCloudLibrary/pcl.git
Updating /Library/Caches/Homebrew/pcl--git
==> Checking out branch master
==> Patching
patching file cmake/pcl_find_cuda.cmake
Hunk #2 FAILED at 37.
1 out of 2 hunks FAILED -- saving rejects to file cmake/pcl_find_cuda.cmake.rej
Error: Failure while executing: /usr/bin/patch -g 0 -f -p1
```

indianajohn-aemass:
I'm just popping in to request that this patch gets merged. A bug in Qt4 that will never be fixed ( https://bugreports.qt.io/browse/QTBUG-40415 ) has forced me to use QT5 with a PCL application. This necessitates that I build PCL with QT5, and at the moment I'm maintaining my own patches. This request being merged would have made my life a lot easier.

pcl.rb
dpo: The above options are still redundant.

ahundt:
I'm kinda stuck myself on this, I think an actual maintainer of homebrew-science is needed to finish this up.

pcl.rb
dunn: @ahundt did you address this?

ahundt:
does 4d8bf6f mean this pull request should be closed?

PR #2603, opened 2015-08-02
gap: revision 1

alexeymuranov:
* Install `InstPackages.sh` script even if no option is given to run it.

* Rename the option `--with-InstPackages` to `--try-building-packages`.

* Rewrite caveats, some comments, etc.

* Add testing GMP.


gap.rb
```diff
@@ -5,21 +5,15 @@ class Gap < Formula
   version "4.7.8"
   sha256 "d1643d0258a0cb037acbe132aacf888bc2b832afb9c4a284d27310c5ad07233e"
 
-  bottle do
-    cellar :any
-    revision 1
-    sha256 "9cc07fdd9990c6e04f43862718bd90998bc0ba9db750c7cfdb05c8262f00732e" => :yosemite
-    sha256 "f35ba8cd877309d5cea6017f01212ad51dd1204c55041e46b335388ed3c4aedd" => :mavericks
-    sha256 "6606679d504c8bf527afe5f97177a35e9bfe1381fa207d6837374b2f5de19fd0" => :mountain_lion
-  end
```
dpo: Please don't touch the `bottle` section. It will be replaced by the build bot.

gap.rb
```diff
@@ -5,21 +5,15 @@ class Gap < Formula
...
+  # NOTE: the archive contains the [GMP library](http://gmplib.org) under
   #   `extern/`, which is not needed if it is already installed (for example,
   #   with Homebrew), and a number of GAP packages under `pkg/`, some of
-  #   which need to be built.
+  #   which need to be built
 
-  option "with-InstPackages",
-         "Try to build included packages using InstPackages script"
+  option "run-InstPackages",
+         "Try to build included packages by running InstPackages.sh script"
```
dpo: Please keep options to `with` or `without`.

alexeymuranov: I have changed [the way `InstPackages.sh` is installed](https://github.com/alexeymuranov/homebrew-science/blob/gap-options/gap.rb#L95-L98), now the installation is always `with-InstPackages`.  The option is to run it automatically.  Maybe it can be renamed to `--with-running-InstPackages`, but it would be weird IMO.

IMO, since this option is not about enabling or disabling some component, but about executing an optional step during the installation at user's risk, it does not quite fit into `with`/`without` dichotomy.

alexeymuranov: A quote from Formula Cookbook:

> Option names should be prefixed with one of the words `with`, `without`, `no`, or a verb in the imperative tense describing the action to be taken.

alexeymuranov: The future plan is to make package compilation work for all packages and then run it by default.  Then an option might need to be added to skip compiling the packages.  IMO it could be called `--no-package-building` or `--skip-package-building`.

gap.rb
```diff
@@ -13,13 +17,13 @@ class Gap < Formula
...
+  # NOTE: the archive contains the [GMP library](http://gmplib.org) under
   #   `extern/`, which is not needed if it is already installed (for example,
   #   with Homebrew), and a number of GAP packages under `pkg/`, some of
-  #   which need to be built.
+  #   which need to be built
 
-  option "with-InstPackages",
-         "Try to build included packages using InstPackages script"
+  option "try-building-packages",
+         "Try to build included packages by running InstPackages.sh script"
```
alexeymuranov: I have changed [the way `InstPackages.sh` is installed](https://github.com/alexeymuranov/homebrew-science/blob/gap-options/gap.rb#L95-L98), now the installation is always `with-InstPackages`.  The option is to run it automatically.  Maybe it can be renamed to `--with-running-InstPackages`, but it would be weird IMO.

IMO, since this option is not about enabling or disabling some component, but about executing an optional step during the installation at user's risk, it does not quite fit into `with`/`without` dichotomy.

The future plan is to make package compilation work for all packages and run it by default.  Then an option might need to be added to skip compiling the packages.  IMO it could be called `--no-package-building` or `--skip-package-building`.

gap.rb
```diff
@@ -86,47 +88,82 @@ def install
...
+    if build.include? "try-building-packages"
       <<-EOS.undent
-        If the formula is installed without the `--with-InstPackages' option,
-        some packages in:
+        When the formula is installed with `--try-building-packages' option,
+        the script is executed by Homebrew during the installation process.
+        However, currently certain packages fail to be built this way.
+        Probably this is caused by undeclared dependencies and may be
+        considered a bug of this formula (thus it might be fixed in future
+        revisions of the formula).
```
dpo: This is quite unsettling and makes me think this formula isn't ready. Could we try to figure out what dependencies are needed? If the packages fail to build in this way, they won't build either when the user tries by hand.

alexeymuranov: I thought that `superenv` does not allow to use undeclared dependencies even if they are installed.  Maybe i have misunderstood it.  In any case, i thought that building packages by hand the user has more flexibility.

I have asked about dependencies on GAP Support mailing list, i hope the discussion there will continue.

It seems that some people already started using this formula.

CC @alex-konovalov.

alex-konovalov: @dpo Thank you for reviewing this. Let me add some more comments as I am preparing the upstream [GAP distribution archives](http://www.gap-system.org/Releases/index.html). GAP is redistributed with 119 packages (user-contributed extensions). Many of them use only GAP code and will work straightforwardly, but some (about 20) require compilation. We advise our users to use `InstPackages.sh` script available [here](http://www.gap-system.org/Download/install.html) which tries to compile as much of those packages as possible. It works for the majority of widely used and regularly maintained packages but may fail for some others because of various reasons (outdated code not working with modern compiler; works for Linux but not on OS X; sometimes indeed dependencies on other software, which may be not provided in homebrew). External software on which some GAP packages depend include:
- [Singular](http://www.singular.uni-kl.de/)
- [4ti2](http://www.4ti2.de/)
- [PARI/GP](http://pari.math.u-bordeaux.fr/)

but none of these three seems to be available via Homebrew.

The  `InstPackages.sh` script is updated sometimes in new GAP releases, so it looks optimal to me if `gap.rb` just uses that script, while we will work on its improvement in the upstream project, for the benefit of other users too.

CC @alexeymuranov

alex-konovalov: P.S. So if e.g. [the package requiring 4ti2](http://wwwb.math.rwth-aachen.de/~gutsche/gap_packages/4ti2Interface/) will fail to build, the user may install 4ti2 themselves, and then try to rebuild 4ti2interface by hand.

dpo: @alexeymuranov You're right that the formula can't used undeclared dependencies. What I had in mind is collecting the dependencies and brewing them too so as to automate the entire process.

@alex-konovalov Thanks for the additional explanations! So wouldn't it make sense to either have Homebrew formulas for those dependencies or declare them in the formula as resources and build them (or a mixture of both)? The `InstPackages.sh` script essentially just runs `configure` in a bunch of subfolders. The entire process could be more robust if the 20-or-so packages that require compilation had their own formula?!

gap.rb
iml: The with/without is obligatory and part of the options/dependency resolution DSL; all other patterns have been deprecated. Personally, I don't see the value in an option that executes a script which the user can run at any later time, especially if it isn't expected to work.

alex-konovalov: For GAP, the big goal of providing GAP via Homebrew is the ability to shorten GAP installation just to one-line command `brew install gap`, which will greatly ease its installation for GAP beginners. It is a common oversight to omit the step of building packages even although it is documented, and there are some packages there which significantly extend GAP functionality.

The script is expected to work in a sense that failing to build a package does not terminate its execution - it will continue with the next package. It is regularly used in our daily/weekly regression tests. So I think it's fine to take it "as is" as in this PR.

gap.rb
alexeymuranov: @dpo, having formulas for package dependencies surely seems to be the correct way to deal with this.  However, should a missing dependency for a user-contributed package block the potential users from installing the core system and other packages?

alexeymuranov: Maybe a clean solution is to remove those packages that cannot be compiled automatically, unless the user gives an option, for example, `--with-extra-unbuilt-packages`?

alexeymuranov: @alex-konovalov, after thinking more about this, building packages using a downloaded script that is not distributed together with the packages and does not even have versions looks like a bad idea to me.  This script seems to be just an indication for manual building.  More importantly, if the script is updated but left at the same address, this will break the formula installation (from source).

I do not see a better solution to this for now than to include all package building instructions into the formula.  This will allow more flexibility in dealing with packages that cannot be built, but it will add more than 100 lines to the formula file, and maintaining the formula will become more difficult.

Any ideas?

dpo: I don't think we can accept such a formula. The best solution would be to have a separate formula for each dependency. After all, managing dependencies is what Homebrew does. Then it might be preferable to maintain your own tap for `gap`.

PR #2045, opened 2015-04-04
TrimAL 1.4.1 (new formula)

tseemann:
trimAl is a tool for the automated removal of spurious sequences or poorly aligned regions from a multiple sequence alignment Access trimAl publication

trimAl can consider several parameters, alone or in multiple combinations, in order to select the most-reliable positions in the alignment. These include the proportion of sequences with a gap, the level of residue similarity and, if several alignments for the same set of sequences are provided, the consistency level of columns among alignments. Moreover, trimAl is able to manually select a set of columns to be removed from the alignment.

Additionally, trimAl implements a series of automated algorithms that apply different thresholds, based on the characteristics of each alignment, to be used so that the signal-to-noise ratio after alignment trimming phase is optimized.

PR #2609, opened 2015-08-03
trimal: use pkgshare, add desc

iml:


iml:
10.10 and 10.9:
```
Testing homebrew/science/trimal
==> trimal 2>&1
Error: homebrew/science/trimal: failed
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/minitest/unit.rb:302:in `=~'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/minitest/unit.rb:302:in `assert_match'
/usr/local/Library/Taps/homebrew/homebrew-science/trimal.rb:26:in `block in <class:Trimal>'
/usr/local/Library/Homebrew/formula.rb:836:in `block in run_test'
/usr/local/Library/Homebrew/formula.rb:832:in `run_test'
/usr/local/Library/Homebrew/test.rb:28:in `block in <main>'
/usr/local/Library/Homebrew/test.rb:27:in `<main>'

tseemann:
@iml This issue has happened previously when I first wrote the formula:
https://github.com/Homebrew/homebrew-science/pull/2045

@sjackman wrote: *My guess is that it's specific to the version of Ruby. It works for me. It's almost certainly caused by the accents in the author's name and the region being set to US-ASCII instead of UTF-8*

sjackman:
This issue is probably fixable by changing the encoding from `US-ASCII` to `UTF-8` before using `assert_match`. I'm also totally fine ignoring it and pulling it as is. Your call.

iml:
I can't reproduce it locally so not going to bother with the encoding.

sjackman:
It only affects old versions of Ruby.

Issue #2613, opened 2015-08-04
graph-tool fails to import on python3


itcarroll:
I believe it's linking to libboost_python.dylib rather than libboost_python3.dylib. Gist logs are [here](https://gist.github.com/anonymous/1773277467c9bfe25759), the error is below. The import is successful if I manually symlink /usr/local/lib/libboost_python.dylib to the the python3 version.

    Python 3.4.3 (default, Jul 13 2015, 12:18:23) 
    [GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import graph_tool
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/local/lib/python3.4/site-packages/graph_tool/__init__.py", line 102, in <module>
        dl_import("from . import libgraph_tool_core as libcore")
      File "/usr/local/lib/python3.4/site-packages/graph_tool/dl_import.py", line 57, in dl_import
        exec(import_expr, local_dict, global_dict)
      File "<string>", line 1, in <module>
    ImportError: dlopen(/usr/local/lib/python3.4/site-packages/graph_tool/libgraph_tool_core.so, 8): Symbol not found: _PyClass_Type
      Referenced from: /usr/local/lib/libboost_python.dylib
      Expected in: flat namespace
     in /usr/local/lib/libboost_python.dylib
    >>>

tdsmith:
Closed by #2616.

PR #2616, opened 2015-08-04
graph-tool: use libboost_python3

itcarroll:
There may be a more appropriate way to link with libboost_python3 when necessary, but this worked for me. Addresses #2613.


itcarroll:
@tdsmith You seem to be one of the experts on boost-python dependency. I'm not sure if this relates to the test failure, but I'm getting that `brew info graph-tool --with-python3` seems insensitive to whether or not boost-python was poured from bottle (i.e. not using `with-python3`).

tdsmith:
What are you seeing and what did you expect to see?

itcarroll:
With the bottled version of boost-python, `brew info graph-tool --with-python3` shows the dependency for boost-python as satisfied. I expect it to show the dependency is not satisfied with this option, because `brew install graph-tool --with-python3` depends on the boost-python3 library.

tdsmith:
Ah, the `info` icon isn't that smart. The correct behavior should happen on `install`. The test failure looked like a bot error; I wouldn't have expected this change to affect the build since it's only activated by an optional dependency.

itcarroll:
You're right, `install` catches the option.

itcarroll:
I think the test result is a bot fail. Compile error is:

    fatal error: error in backend: IO failure on output stream.

Can't tell where to file an issue...

tdsmith:
I think the bot is running out of disk space.

tdsmith:
Pulled; thanks very much for your contribution to Homebrew!

ToferC:
Sorry - I just tried to install graph-tools via Homebrew and hit this issue with Python 3.  I've upgraded all packages and am using boost 1.58.0 and graph-tool 2.2.44 on a mac OS X Mavericks.

Is the fix above part of keg/science yet?  If not, can you advise how I can get the build working?  Thanks!

This is the error:

Python 3.4.2 (default, Oct 30 2014, 21:11:49) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.54)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import graph_tool
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.4/site-packages/graph_tool/__init__.py", line 102, in <module>
    dl_import("from . import libgraph_tool_core as libcore")
  File "/usr/local/lib/python3.4/site-packages/graph_tool/dl_import.py", line 57, in dl_import
    exec(import_expr, local_dict, global_dict)
ImportError: dlopen(/usr/local/lib/python3.4/site-packages/graph_tool/libgraph_tool_core.so, 8): Symbol not found: _PyClass_Type
  Referenced from: /usr/local/lib/libboost_python.dylib
  Expected in: flat namespace
 in /usr/local/lib/libboost_python.dylib

Python 2.7 works fine.

itcarroll:
The fix was integrated. Did you use the --with-python3 flag? I'm not sure whether you have to manually re-install dependencies (like boost-python) --with-python3 or whether brew does that for you.

sandrofsousa:
I'm facing the same problem even though using the last formulae. The correction commented here is in the script but the ```libboost_python3.dylib``` is missing. Looks like homebrew is ignoring the ```--with-python3``` option.

itcarroll:
@sandrofsousa I think your assessment is correct (that with-python3 isn't getting passed to the boost-python dependency) but I don't know enough ruby to confirm/fix. If you reinstall `boost-python --with-python3 --c++11` and then `graph-tool --with-python3`, does it work?

mkhorton:
Just had this issue myself, can confirm that itcarrol's reinstallation instructions work for anyone else who has this issue (on OS X 10.11.6, Python 3.5.2)

PR #2618, opened 2015-08-04
nlopt: add lib name links for C programs

ge11232002:
This homebrew nlopt library is compiled with C++ with lib names which
are not recognised by some C programs. Add links like libnlopt_cxx.dylib
to libnlopt.dylib

Add the "desc" tag.


iml:
Thanks for your contribution!

PR #2625, opened 2015-08-05
Update gromacs.rb to latest stable version (5.0.6)

nhstanley:
Previous version was not linking gmx binary to the path. Should now
build, test, and install Gromacs suite (test on by default) and link all
binaries to the path.

Also note: While the project preferred method for downloading Gromacs is
via a box.com link (see http://www.gromacs.org/downloads), Homebrew does
not seem able to unpack files downloaded via this method. Direct FTP
download was therefore maintained in this update. Requires further
looking into.

### DIFF
--- a/gromacs.rb
+++ b/gromacs.rb
@@ -3,8 +3,9 @@ require "formula"
 class Gromacs < Formula
   homepage "http://www.gromacs.org/"
   version "5.0.6"
+  desc "GROMACS is a versatile package for performing molecular dynamics calculations, i.e. simulate the Newtonian equations of motion for systems with hundreds to millions of particles, primarily proteins, lipids, and nucleic acids."
   url "ftp://ftp.gromacs.org/pub/gromacs/gromacs-5.0.6.tar.gz"
-  mirror "http://fossies.org/linux/privat/gromacs-5.0.6.tar.gz"
+  mirror "https://fossies.org/linux/privat/gromacs-5.0.6.tar.gz"
   sha256 "e07e950c4cd6cb84b83b145b70a15c25338ad6a7d7d1a0a83cdbd51cad954952"
 
   deprecated_option "with-x" => "with-x11"
@@ -26,13 +27,11 @@ class Gromacs < Formula
     args << "-DGMX_DOUBLE=ON" if build.include? "enable-double"
     args << "-DGMX_X11=ON" if build.with? "x11"
     args << "-DGMX_CPU_ACCELERATION=None" if MacOS.version <= :snow_leopard
-    args << "-DREGRESSIONTEST_DOWNLOAD=ON" unless build.with? "without-check"
+    args << "-DREGRESSIONTEST_DOWNLOAD=ON" if build.with? "check"
 
     inreplace "scripts/CMakeLists.txt", "BIN_INSTALL_DIR", "DATA_INSTALL_DIR"
 
-    system "mkdir", "build"
-
-    cd "build" do
+    mkdir "build" do
       system "cmake", "..", *args
       system "make"
       system "make", "check" if build.with? "check"
--- a/gromacs.rb
+++ b/gromacs.rb
@@ -40,8 +40,8 @@ class Gromacs < Formula
 
     # This is a really hacky solution, but seems needed to pass Homebrew build test
     # Doesn't seem to affect command line completion of built package
-    rm "/usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion-gmx.bash"
-    rm "/usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion.bash"
+    rm "#{bin}/gmx-completion-gmx.bash"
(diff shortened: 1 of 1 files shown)


gromacs.rb
```diff
@@ -2,9 +2,10 @@
 
 class Gromacs < Formula
   homepage "http://www.gromacs.org/"
-  url "ftp://ftp.gromacs.org/pub/gromacs/gromacs-4.6.5.tar.gz"
-  mirror "https://fossies.org/linux/privat/gromacs-4.6.5.tar.gz"
-  sha1 "6bf86bb514e5488bda988d5b0e98867706d4ecd4"
+  version "5.0.6"
```
dpo: Would you mind adding a `desc` field with a brief description of the software?

gromacs.rb
```diff
@@ -25,19 +26,22 @@ def install
     args << "-DGMX_DOUBLE=ON" if build.include? "enable-double"
     args << "-DGMX_X11=ON" if build.with? "x11"
     args << "-DGMX_CPU_ACCELERATION=None" if MacOS.version <= :snow_leopard
+    args << "-DREGRESSIONTEST_DOWNLOAD=ON" unless build.with? "without-check"
```
dpo: `build.without? "check"`

gromacs.rb
```diff
@@ -25,19 +26,22 @@ def install
...
 
     inreplace "scripts/CMakeLists.txt", "BIN_INSTALL_DIR", "DATA_INSTALL_DIR"
 
-    cd "src" do
+    system "mkdir", "build"
+
+    cd "build" do
```
dpo: You can use the shortcut `mkdir "build" do`

dpo:
Other audit problems:
```
 * Stable: version 5.0.6 is redundant with version scanned from URL
 * Stable: Please use https:// for http://fossies.org/linux/privat/gromacs-5.0.6.tar.gz
 * Non-executables were installed to "/usr/local/Cellar/gromacs/5.0.6/bin"
The offending files are:
  /usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion-gmx.bash
  /usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion.bash
```

nhstanley:
Hi, this is my first Formula so please bear with me. I tried to make a few changes but I'm struggling to understand why the other errors are considered failures.

    homebrew/science/gromacs:
    * Stable: version 5.0.6 is redundant with version scanned from URL
    * Non-executables were installed to "/usr/local/Cellar/gromacs/5.0.6/bin"
    The offending files are:
    /usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion-gmx.bash
    /usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion.bash

Should I just `rm` the completion files before `make install` and take out the `version` variable altogether?

nhstanley:
I've tried to resolve the issues, but it seems the tests are still failing. Unfortunately, I don't see a specific actionable problem. The only clue is in the log of the test, which has

`FATAL: Could not checkout 10573ac33e8b5eea1d910660abe0594183b05a46`

Is this an issue with the formula, or something else? Any help you could provide would be greatly appreciated.

dpo:
Bot failure. I'll kick off another build.

gromacs.rb
```diff
@@ -25,19 +26,25 @@ def install
...
       ENV.deparallelize
       system "make", "install"
     end
 
-    bash_completion.install "scripts/completion.bash" => "gromacs-completion.bash"
-    zsh_completion.install "scripts/completion.zsh" => "_gromacs"
+    # This is a really hacky solution, but seems needed to pass Homebrew build test
+    # Doesn't seem to affect command line completion of built package
+    rm "/usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion-gmx.bash"
+    rm "/usr/local/Cellar/gromacs/5.0.6/bin/gmx-completion.bash"
```
iml: Replace `"/usr/local/Cellar/gromacs/5.0.6/bin"` with `"#{bin}"`

nhstanley:
Thanks for the critique @iml . I updated the formula, but looks like the build bot failed again. 
`FATAL: Could not checkout 3267cdd3753c1d68e57417006c6bec4483495695`
Can we get another build?

dpo:
Done

dpo:
The caveats say
```
==> Caveats
GMXRC and other scripts installed to:
  /usr/local/share/gromacs

Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completion has been installed to:
  /usr/local/share/zsh/site-functions
```
I've no idea why the gmx completion stuff ends up in `bin`. Perhaps we should simply install it to `bash_completion` as well?

nhstanley:
I made the change to install them into `bash_completion`, which is cleaner than just `rm`-ing the files, and it still builds and links everything properly.

PR #2626, opened 2015-08-06
Opencv3 with openni2

MrDuncanM:
Small update to add --with-openni2.


tdsmith:
Can we do this by setting environment variables or cmake variables instead of with an inreplace?

MrDuncanM:
Why yes it is. I have made the modification and added the gphoto2 option.

dpo:
Thanks! Would you please squash your commits?

PR #2628, opened 2015-08-07
Update links-scaffolder.rb

warrenlr:
updated DOI
updated release to v1.5.1 : bug fix, input Bloom filter FPR
updated sha256 sum


tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this one.

PR #2630, opened 2015-08-07
Petsc 3.6.1 (update)

davydden:
update to 3.6.1. Propagate openblas dependecy to depends_on formule.

### DIFF
--- a/mumps.rb
+++ b/mumps.rb
@@ -4,6 +4,7 @@ class Mumps < Formula
   url "http://mumps.enseeiht.fr/MUMPS_5.0.1.tar.gz"
   mirror "http://graal.ens-lyon.fr/MUMPS/MUMPS_5.0.1.tar.gz"
   sha256 "50355b2e67873e2239b4998a46f2bbf83f70cdad6517730ab287ae3aae9340a0"
+  revision 1
 
   bottle do
     cellar :any
@@ -14,7 +15,7 @@ class Mumps < Formula
 
   depends_on :mpi => [:cc, :cxx, :f90, :recommended]
   if build.with? "mpi"
-    depends_on "scalapack" => (build.with? "openblas") ? ["with-openblas"] : :build
+    depends_on "scalapack" => (build.with? "openblas") ? ["with-openblas"] : []
   end
   depends_on "metis"    => :optional if build.without? "mpi"
   depends_on "parmetis" => :optional if build.with? "mpi"
@@ -95,7 +96,7 @@ class Mumps < Formula
                     "FC=#{ENV["MPIFC"]} -fPIC",
                     "FL=#{ENV["MPIFC"]} -fPIC",
                     "SCALAP=-L#{Formula["scalapack"].opt_lib} -lscalapack",
-                    "INCPAR=",  # Let MPI compilers fill in the blanks.
+                    "INCPAR=", # Let MPI compilers fill in the blanks.
                     "LIBPAR=$(SCALAP)"]
     else
       make_args += ["CC=#{ENV["CC"]} -fPIC",
@@ -109,7 +110,7 @@ class Mumps < Formula
       make_args << "LIBBLAS=-lblas -llapack"
     end
 
-    ENV.deparallelize  # Build fails in parallel on Mavericks.
+    ENV.deparallelize # Build fails in parallel on Mavericks.
 
     # First build libs, install them, and then link example programs.
     system "make", "alllib", *make_args
@@ -121,7 +122,7 @@ class Mumps < Formula
       s.change_make_var! "libdir", lib
     end
 
-    system "make", "all", *make_args  # Build examples.
+    system "make", "all", *make_args # Build examples.
 
     if build.with? "mpi"
       include.install Dir["include/*"]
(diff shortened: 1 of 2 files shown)


tdsmith:
Do you get crashes if the dependencies don't agree about the blas?

tdsmith:
I guess this is what you're alluding to in #2631. I think this seems fine but what's the failure mode, out of curiosity?

davydden:
it's manly for `linuxbrew`. Imagine what happens if the user writes `brew install slepc` there - it will fail miserably.
In either case, it's good idea to use the same Blas/lapack throughout.

dpo:
I can see it's a good idea to use the same BLAS through the entire chain, but would it really cause a crash (unless one BLAS was 32bit and another 64bit)? Does it crash on Linux?

davydden:
it's not about crashing, it's about not being able to compile on `linuxbrew`. Look at https://github.com/Homebrew/homebrew-science/blob/master/superlu_dist.rb#L31 for example where `superlu_dist` would try to link against `-framework Accelerate` unless it's installed `--with-openblas`.

dpo:
Ok, but that's just because `superlu_dist.rb` isn't ready for Linux. It needs to be changed as in https://github.com/Homebrew/homebrew-science/blob/master/scalapack.rb#L26. Same with `superlu.rb`.

davydden:
To assume that there are `libblas` and `liblapack` (hardcoding `-lblas` and `-llapack`) is not a good/robust way to do it https://github.com/Homebrew/homebrew-science/blob/master/scalapack.rb#L30 . Please, see the discussion here https://github.com/Homebrew/linuxbrew/issues/504 . In a nutshell, there could be numerous different system/vendor blas/lapack.s, each with a different set of libraries files to be linked
against, such as `-L/soft/com/packages/intel/13/079/mkl -lmkl_intel_ilp64 -lmkl_core -lmkl_intel_thread -lpthread -lm`.

dpo:
I understand but all Linux distributions have `libblas` and `liblapack` as the standard BLAS option, just like OSX has the `Accelerate` framework. I don't think we support linking against any other BLAS.

davydden:
could we perhaps merge it as the PR does not break anything but makes the formula more consistent in terms of a blas choice? Moreover it has a newer version.

dpo:
I want to make sure it doesn't break `getdp` and `gmsh` (it has in the past).

davydden:
i have `gmsh` installed
```
$ brew info gmsh
davydden/science/gmsh: stable 2.10.0 (bottled), HEAD
Gmsh is a 3D grid generator with a build-in CAD engine.
http://geuz.org/gmsh
/usr/local/Cellar/gmsh/2.9.3 (219 files, 58M)
  Built from source with: --with-fltk --with-petsc --with-slepc
/usr/local/Cellar/gmsh/2.10.0 (221 files, 58M) *
  Built from source with: --with-fltk --with-petsc --with-slepc
```

davydden:
`getdp` installs fine but a test fails:
```
The following tests FAILED:
	  1 - ../demos/magnet.pro (OTHER_FAULT)
```
I am not using this software, but i would notify the developers so that they can investigate it. Perhaps it was failing even before `3.6.1`.

dpo:
I get the same on OSX 10.9. @schoeps Any idea why this test might fail?

dpo:
I confirm that the same test used to fail before the PETSc upgrade.

davydden:
@BrewTestBot test this please

schoeps:
@dpo the problem were the tests of getdp. The good news: getdp seems to work with the newest petsc, too :) I will submit a fix.

davydden:
@dpo do i need to bump `mumps` to avoid using old bottles with the new `openmpi`?

dpo:
@BrewTestBot test this please

davydden:
I think it's better to merge Scalapack first (make it dynamic again), then PETSc should be good for bottles.

davydden:
Never-mind, just saw that u merged it already.

davydden:
Mumps tests on Mavericks and Mountain Lion can't find Scalapack:
```
==> mpirun -np 2 /usr/local/Cellar/mumps/5.0.1_1/share/mumps/examples/ssimpletest < /usr/local/Cellar/mumps/5.0.1_1/share/mumps/examples/input_simpletest_real
dyld: Library not loaded: /usr/local/lib/libscalapack.dylib
  Referenced from: /usr/local/Cellar/mumps/5.0.1_1/share/mumps/examples/ssimpletest
  Reason: image not found
dyld: Library not loaded: /usr/local/lib/libscalapack.dylib
  Referenced from: /usr/local/Cellar/mumps/5.0.1_1/share/mumps/examples/ssimpletest
  Reason: image not found
```

tdsmith:
I suppose it's not a :build dependency, then.

tdsmith:
You can replace `:build` with an empty list (`[]`) in line 18 of mumps.rb.

davydden:
agreed, will do.

davydden:
@BrewTestBot test this please

davydden:
don't understand Jenkins output, don't think it even tried to build PETSc, the only thing I could dig is `Failed to determine (log)`.

dpo:
Bot hiccup. I kicked off a new build.

davydden:
PETSc website is down... Bot couldn't download the tar file.

davydden:
@BrewTestBot test this please

davydden:
Bot hiccup again...

davydden:
We are good.

dpo:
Thanks! On the with the rest.

PR #2647, opened 2015-08-13
kat 2.0.1

maplesond:


maplesond:
There are a couple of issues here.  First of all I need C++11 support.  How do I go about ensuring a suitable c++ compiler is available?  Second, I need to get the correct hashes for the bottles.

dpo:
You can use `needs :cxx11` to request an appropriate compiler. You need not touch the `bottle` section of the formula. It will be automatically updated by the brew bot.

kat.rb
```diff
@@ -17,23 +17,21 @@ class Kat < Formula
...
+  needs :cxx11
+
   depends_on "pkg-config" => :build
   depends_on "boost"
   depends_on "gnuplot"
-  depends_on "jellyfish-1.1"
-  depends_on "seqan"
 
   def install
     ENV.libstdcxx if ENV.compiler == :clang && MacOS.version >= :mavericks
```
dpo: Should be `libcxx` instead of `libstdcxx`.

dpo: next build error:
```
jellyfish_helper.cc:47:10: fatal error: 'bits/stl_vector.h' file not found
```

maplesond:
I suspect the reason 'bits/stl_vector.h' is not present is due to an older than expected version of GCC on the test systems.  I oldest version I've tested on is 4.9.1.  So it would be good add some additional requirements for the C++ libraries and/or compiler versions.  I'm not sure how to do that in homebrew however.

dpo:
I see. You need a `fails_with` block. Grepping through existing formulas should show you how it's used.

maplesond:
Need to modify source code to make it more portable for C++11 compilers on mac.  Will close this case and reopen a new one targetting the new version of KAT.

PR #2649, opened 2015-08-14
openbr: head

dpo:
It seems `openbr` is now head only despite the fact that they have releases. See
https://github.com/biometrics/openbr/issues/410
https://github.com/biometrics/openbr/issues/411

Is there still interest in keeping this here, or should it go to homebrew/head-only?

Closes #2093.

### DIFF
--- a/openbr.rb
+++ b/openbr.rb
@@ -1,10 +1,10 @@
-require "formula"
-
 class Openbr < Formula
+  desc "Open Source Biometrics, Face Recognition"
   homepage "http://www.openbiometrics.org/"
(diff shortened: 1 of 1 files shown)


dpo:
@sjackman @iml @tseemann

sjackman:
My inclination would be to keep it here, and to petition upstream to tag a stable release.

dpo:
@BrewTestBot test this please

dpo:
Apparently, [there will be a release soon](https://github.com/biometrics/openbr/issues/410#issuecomment-137576443), which is good news.

sjackman:
openbr 1.0.0 and 1.1.0 have been released. See https://github.com/biometrics/openbr/releases

sjackman:
There's a tagged release for 0.5.0 as well.

PR #2652, opened 2015-08-16
r 3.2.2

tdsmith:
Creates site-library in HOMEBREW_PREFIX (#2559)

Also fixes a bug someone mentioned on IRC wrt -L paths memorized from gfortran.


tdsmith:
install.packages is broken on Linuxbrew but I assume it was broken before? Let's deal with it later.

tdsmith:
We should also pass `rhome=#{prefix}` to configure to get a saner directory layout on Linux.

PR #2657, opened 2015-08-16
biointerchange 2.0.2

indiedotkim:
Fixed a memory leak. License update.


biointerchange.rb
```diff
@@ -1,16 +1,16 @@
...
   homepage "https://www.codamono.com"
   # tag "bioinformatics"
 
   if OS.mac?
-    version "2.0.1+74"
-    url "https://www.codamono.com/brew/osx-10-10/biointerchange-2.0.1+74.tar.gz"
-    sha256 "ed1864b13aaf23518d32867683bfcc9bc8c15902c12b549433832e83b43926a6"
+    version "2.0.2+83"
+    url "https://www.codamono.com/brew/osx-10-10/biointerchange-2.0.2+83.tar.gz"
+    sha256 "fb70c5d537c08cf860a270a156cc1c00e13ec095"
```
dpo: This isn't the `sha256`.

indiedotkim:
As @dpo pointed out: sha256 is wrong. My bad! Build script is still using sha1! Going to fix it...

indiedotkim:
SHA's fixed. Commit squashed.

dpo:
Looking good. Thanks!

PR #2660, opened 2015-08-17
Octave 4.0 (initial version)

schoeps:
This is a first formula that allows to compile Octave 4.0. It is based on the work of I assume that many changes will be necessary before it can be merged into base but let's try...


octave.rb
```diff
@@ -1,137 +1,116 @@
...
-      url "https://gist.github.com/tchaikov/6ce5f697055b0756126a/raw/4fc94a1fa1d5b032f8586ce3ab0015b04351426f/octave-clang3.5-fix.patch"
-      sha1 "6e5c0d8f6b07803152c8a1caad39a113fc8b8d0a"
-    end
-  end
-
-  bottle do
-    sha256 "98d8a35cd7c4e20e13b77e6999bc75a39ef52f8e5fa6deafd80f60e10c8db8d0" => :yosemite
-    sha256 "0a28985e322d56002755469a22cc5ae3c487c191d35c8b5b3b25c34abcc0b2cb" => :mavericks
-    sha256 "35344dafa1cb05d8d7d1c77c469e012d404b0a5062c840ed062570c94fce3d2b" => :mountain_lion
-  end
```
dpo: There's no need to remove the `bottle` section. It will be automatically updated.

octave.rb
```diff
@@ -1,137 +1,116 @@
...
-  # Allows mkoctfile to process "-framework vecLib" properly.
-  # See: https://savannah.gnu.org/bugs/?42002
-  patch do
-    url "https://savannah.gnu.org/patch/download.php?file_id=31072"
-    sha1 "19f2dcaf636f1968c4b1639797415f83fb21d5a3"
-  end
 
-  # Allows other software to interface Octave and find its headers.
-  # See: https://github.com/shogun-toolbox/shogun/issues/2396
   patch :DATA
```
dpo: Do you have a reference for the patch that's left?

octave.rb
```diff
@@ -1,137 +1,116 @@
...
     args = ["--prefix=#{prefix}"]
+    args << "--enable-dependency-tracking"
+    args << "--enable-link-all-dependencies"
+    args << "--enable-shared"
 
-    args << "--with-blas=-L#{Formula["openblas"].opt_lib} -lopenblas" if build.with? "openblas"
+    # arguments, enabled by default
     args << "--disable-docs"     if build.without? "docs"
-    args << "--enable-jit"       if build.with? "jit"
+    args << "--disable-java"     if build.without? :java
```
dpo: `:java` should be `"java"`

dpo:
That was easy! ;-)

@mcg1969

mcg1969:
Very cool! I'm following the work you're doing on octave-dev too. Thank you!

schoeps:
I have included dpo's comments. Still open:
- some small GUI errors need to be fixed 
- does this formula work also on Mac OS X < 10.10?
- I need to check if the documentation is proper (there are issues with osmesa)

dpo:
Thanks. The previous build succeeded on 10.8, 10.9 and 10.10.

octave.rb
```diff
@@ -22,116 +12,111 @@ class Octave < Formula
...
     depends_on "freetype"
+    depends_on "librsvg"      => :build
   end
-  depends_on "llvm"           if build.with? "jit"
-  depends_on "curl"           if build.with?("curl") && MacOS.version == :leopard
-  depends_on :java            => :recommended
 
-  depends_on "gnuplot"       => [:recommended, build.with?("gui") ? "with-qt" : ""]
+  # recommended dependencies
   depends_on "suite-sparse421" => :recommended
```
davydden: could you pls try if the head version of suite-sparse (4.4.4) works?

dpo: If I remember well we tried a few weeks ago and it failed. The SuiteSparse API has changed.

davydden: sure, but that was for Octave `3.8`, i suppose? I just hoped it may already ok in 4.0.

dpo: No, we tried with 4.0. Things might have changed since.

davydden: oh, i see, pity.

schoeps: I recently tried to compile the latest suite-sparse with the development branch of octave ("--HEAD") and it did not work either, i.e., there is no easy backport possible. For now, I suggest to create a formula for Octave 4.0 with the old suite sparse and let's hope that the problem resolves itself soon...

mrn-aglic:
Any estimate on when this might get merged? Looking forward to Octave 4!

schoeps:
You can help by trying this formula and reporting problems: brew install https://raw.githubusercontent.com/Homebrew/homebrew-science/e1cc7c9d89cf663eb9742f64ee895bfbbfe5f80f/octave.rb

mrn-aglic:
I got the following:
brew install https://raw.githubusercontent.com/Homebrew/homebrew-science/e1cc7c9d89cf663eb9742f64ee895bfbbfe5f80f/octave.rb
Error: No available formula for suite-sparse421 (dependency of octave)
==> Searching formulae...
==> Searching taps...
homebrew/science/suite-sparse421

I don't know why suite-sparse421 wasn't found, it obviously exists in the repo... brew update and brew upgrade didn't fix the issue.

schoeps:
Indeed dependecy tracking does not work automatically if you install via an url. please install dependencies manually, e.g., brew install suite-sparse421.

mrn-aglic:
Ok, I didn't tap homebre/science, I assumed it would find suite-sparse421 by its own... installing like:
brew install homebrew/science/suite-sparse421
worked.

after that I was managed to install octave 4 the way you said. 

Seems to work for now... tried on some simple expression, like matrix multiplication.

mrn-aglic:
octave crashes when dragging GUI windows

davydden:
i installed it fine. Does not crash for me so far, tried plotting some functions. so far so good.

mrn-aglic:
Tried reinstalling... stil crashes when I drag the command line window and then attempt to place it back...

davydden:
oh, that's what you mean. this crashes for me as well.

schoeps:
ok. please do more tests! I am not sure if I can easily fix this (the bug might be either in octave, qt, qscintilla, ...). If there are many bugs like this, I should probably create a cli-only version by default .

whistlerbrk:
I was able to successfully install. The gui is a bit crashtastic, but `--no-gui` is working great for me.

I was earlier troubleshooting installation and discovered the latest python and a pip installed hggit were necessary for me as well.

schoeps:
New version; I have disabled the gui by default. This is probably the best choice for now. 

@whistlerbrk python and hggit should not be necessary neither for 4.0 nor for HEAD. Can you find out for what they were needed exactly?

mrn-aglic:
Install link for the new version?

schoeps:
@mrn-aglic 
`brew install https://raw.githubusercontent.com/schoeps/homebrew-science/octave/octave.rb`

(this should always point to the latest version)

dunn:
Just a minor failure:
```
homebrew/science/octave:
 * 181: Trailing whitespace was found
 * 191: Trailing whitespace was found
```

schoeps:
What do you think? Are we ready? Or should we create (a) an "octave4.rb" or (b) rename the current octave to octave38?

mrn-aglic:
I think there should be an octave4.rb, since we don't know (at least I don't) what's causing the problems with the GUI and can/when it will be resolved.

schoeps:
I disabled the GUI by default. In other words the resulting binary should not be more unreliable than 3.8.

dpo:
A related question is: how do packages that depend on `octave` behave?
```
$ brew uses octave
homebrew/science/dynare		  homebrew/science/nlopt	    homebrew/science/swrcfit
homebrew/science/flann		  homebrew/science/shogun
```

schoeps:
@dpo I am not using any of those. What is the typical procedure at homebrew? I can check if the build but that's all I can do. However, we could still follow the "octave4.rb" route.

dpo:
@schoeps I'm happy to check if they build too. Beyond building and running the default tests, there's not much more we can do unless we're users of those libraries. I added the comments mostly as a reminder.

tdsmith:
Let's keep this in `octave.rb`.

dpo:
Many thanks!

PR #2664, opened 2015-08-18
clfft 2.6.1

John-Colvin:
A few extra changes:
* Don't build the examples any more, otherwise they get installed in "${bin}/examples"
* actually test the client

### DIFF
--- a/clfft.rb
+++ b/clfft.rb
@@ -1,7 +1,14 @@
 class Clfft < Formula
+  desc "FFT functions written in OpenCL"
   homepage "https://github.com/clMathLibraries/clFFT/"
-  url "https://github.com/clMathLibraries/clFFT/archive/v2.4.tar.gz"
-  sha256 "d77506af774bbe8ccf4226a58e623c8a29587edcf02984e72851099be0efe04b"
+  url "https://github.com/clMathLibraries/clFFT/archive/v2.6.1.tar.gz"
+  sha256 "2b5b15b903baeef4dcea6bb8efbe6aba284510148c04d20f9b151a94ae71c050"
+
+  patch do
+    # New timer implementation for OS X
+    url "https://github.com/clMathLibraries/clFFT/commit/e4063ef4541cd34951b25b38d429f199a15f8e14.diff"
+    sha256 "6c46e429d368a95afa030f6865c392ee0a76d4fa8c7fb55daef88595fb7a5b72"
+  end
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


dpo:
Would it make sense to ask `cmake` to install the binaries to `libexec` and only link the actual executables in `bin`?

John-Colvin:
why? I thought ```libexec``` was for executable dependencies that aren't meant for using directly?

dpo:
It allows you to install the examples but only link actual executables in `bin` to avoid `brew audit` complaining. Several formulas use this trick, especially to keep `lib` clean. See, e.g., `asl.rb`.

John-Colvin:
Ok, but why bother building the examples anyway? From an installation standpoint, they're just bloat.

dpo:
Wouldn't a user want to look at them, and run them?

John-Colvin:
As far as I can tell, the example sources aren't installed by cmake anyway even if they are built.

dpo:
@BrewTestBot test this please

dpo:
The test is failing.

John-Colvin:
test is a 404

dpo:
@BrewTestBot test this please

John-Colvin:
The problem is this: clFFT is broken on CPUs. The only openCL device on the autotester is a CPU.

dpo:
Thanks for investigating. Should this be reported upstream then?

John-Colvin:
If I remember correctly, it's a relatively well-known long-standing bug in Apple's OpenCL implementation. clFFT are aware of it and I'm sure they've done their fair share of complaining to Apple.

dpo:
Do I need anything special to test this locally? When I run the test on my machine (which has a gpu) I get
```
$ clFFT-client -g -x 192 -y 108 --inLayout 5 --outLayout 3
OPENCL_V_THROWERROR< CLFFT_DEVICE_NOT_AVAILABLE > (456): No devices available
clFFT error condition reported:
OPENCL_V_THROWERROR< CLFFT_DEVICE_NOT_AVAILABLE > (456): No devices available
```

John-Colvin:
That's very odd. You shouldn't need anything extra, assuming your graphics card is supported by Apple's OpenCL implementation.

dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

dpo:
See #3163 for version 2.8.0. Comments welcome.

dpo:
Superseded by #3189.

PR #2668, opened 2015-08-18
bamutil 1.0.13

pjvandehaar:
Switches to an up-to-date download location.  The new location does not include libStatGen in the download, so `make cloneLib` must be used.


pjvandehaar:
Superceded by https://github.com/Homebrew/homebrew-science/pull/2670

PR #2669, opened 2015-08-18
r: hint openssl

### LINKED ISSUE #2667: add openssl CFLAGS, LDFLAGS to R Makeconf
So that e.g. git2r builds against Homebrew openssl and not the system openssl, which will become a problem on 10.11

tdsmith:
Fixes #2667.


tdsmith:
Note that R doesn't have an openssl dependency. I think that doesn't matter!?

PR #2671, opened 2015-08-19
kat 2.1.0

maplesond:


dpo:
@BrewTestBot test this please

tseemann:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

dpo:
@tseemann This seems to include a version of `jellyfish`. Could it depend on the existing `jellyfish` formula?

If not, it needs
```
  fails_with :clang do
    cause "error: variable length array of non-POD element type"
  end
```
or `jellyfish` won't build.

sjackman:
I run into other compilation issues with kat 2.0.8.
See https://github.com/TGAC/KAT/issues/44
and https://github.com/TGAC/KAT/issues/45

sjackman:
Here's the SHA256 of all the recent releases, for the record.
```
  url "https://github.com/TGAC/KAT/releases/download/Release-2.0.1/kat-2.0.1.tar.gz"
  sha256 "8ccb8353529ec89351bb69646a9482f9ed8cc9864cd02ac730b4603357bef22e"

  url "https://github.com/TGAC/KAT/releases/download/Release-2.0.2/kat-2.0.2.tar.gz"
  sha256 "87ce7f330ac7ec63ed8f367badc2d1783e96d710c8a567ffaffe2b6a1833d523"

  url "https://github.com/TGAC/KAT/releases/download/Release-2.0.3/kat-2.0.3.tar.gz"
  sha256 "2fba9c7c6314691c1cfbf551315757334a0a8a805407fb2bcf1e78f53f33d567"

  url "https://github.com/TGAC/KAT/releases/download/Release-2.0.4/kat-2.0.4.tar.gz"
  sha256 "495d23c892ba85b506fb6c6c524b73a1977589a2e564af626304265c0fa18731"

  url "https://github.com/TGAC/KAT/releases/download/Release-2.0.5/kat-2.0.5.tar.gz"
  sha256 "fdf3a8e916871f54f18f7ea70e19e71a33f1dba8fa9a74677ef9dd2c4fe3366d"

  url "https://github.com/TGAC/KAT/releases/download/Release-2.0.6/kat-2.0.6.tar.gz"
  sha256 "fb56d1b0a5c3a87347d7127d5f4b9fc02eacf91b55067e552b0572109f63d993"

  url "https://github.com/TGAC/KAT/releases/download/Release-2.0.8/kat-2.0.8.tar.gz"
  sha256 "6ee1d8e7f810054710ad3dab32650ebecc5c8b9fa19687e06dc9b6b7e2e7f0ef"
```

sjackman:
I needed to add `"--with-boost=#{Formula["boost"].prefix}"` for a system not installed in `/usr/local` but `~/.homebrew`.

tseemann:
KAT1 needed jellyfish 1.1 but I think KAT2 now bundles with it?

sjackman:
Yes, it bundles Jellyfish 2.2.0, which unfortunately does not build on Mac. For Mac the best route forward is probably to disable the bundled version of Jellyfish (and Seqan possibly) and use the Homebrew version.

sjackman:
I loathe bundled dependencies. *loathe*

maplesond:
I understand why this might not be best practice in most cases, unfortunately, it isn't possible to use an external version of jellyfish here.  We made modifications to the jellyfish source code in order for it to work the way we needed.  While KAT V2 can use jellyfish hashes as input, it's generally preferred (for runtime and storage issues) to work directly from the FastA/Qs.  

There was also some inconsistency in Jellyfish between versions (mostly bugs that were subsequently fixed rather than intention API changes) but it caused a problem for users trying to marry compatible jellyfish and KAT versions.  Keeping Jellyfish as an internal dependency removes that complexity for the user completely.

We also have added prefixes to the jellyfish binaries so they can sit side by side with existing jellyfish installations.

I am also not really planning to spend a lot of time supporting the Mac platform, KAT requires a large amount of memory (>20GB) to run for most envisioned use cases with Illumina data and most Mac specs seem limited to 16GB for the time being.  That being said I'll introduce the 2.2.4 changes into the next version, if they are not already there.

sjackman:
Fair enough. Thanks for your quick response, Daniel. Often Mac build issues are actually `clang` build issues and can be tested on Linux as well. Travis CI offers `clang` for Linux and a Mac OS build environment.

sjackman:
I'd like to one day support installing older versions of a formula. To support this, we need a record of all the previous versions in the git commit history. When Homebrew-science has missed one or more upstream updates, I'd like to commit the first revision (2.0.1 in this case) with all the necessary formula fixes, then all the subsequent releases one per commit (that nominally update only the `url` and the `sha256`) then submit the lot as one pull request, which will trigger a bottle build for the only the most recent version. I've posted the URLs and SHA256 above. https://github.com/Homebrew/homebrew-science/pull/2671#issuecomment-209117722

This isn't yet Homebrew-science policy, but I think a good idea. @iml @tseemann Thoughts?

sjackman:
@maplesond Hi, Daniel. Would you like to update this PR for `kat` 2.0.8?

sjackman:
Thanks for the update, Daniel!

Are the Python dependencies new? Python dependencies in this style will unfortunately render it not possible to build and test `kat` with the BrewTestBot. There's lots of packages in that situation, so if that's the case, then so be it, but if they're in fact optional dependencies, it would be great to continue to build bottles for `kat`.

kat.rb
```diff
@@ -18,24 +24,32 @@ class Kat < Formula
...
   end
+  
+  needs :cxx11
 
   depends_on "pkg-config" => :build
   depends_on "boost"
+  depends_on "numpy" => :python
+  depends_on "scipy" => :python
+  depends_on "matplotlib" => :python
+  depends_on "sphinx" => :python
```
tdsmith: Use `depends_on "sphinx-doc" => :build` instead.

sjackman:
Both of these are okay because they're included in Mac OS, I believe:
```
depends_on "numpy" => :python
depends_on "matplotlib" => :python
```
I don't believe `scipy` is included in Mac OS:
```
depends_on "scipy" => :python
```

maplesond:
Hi all,  I'll make the suggested change for sphinx and recommit.  Unfortunately, while the python dependencies are optional, the recommended configuration is to use python (including scipy).

ilovezfs:
```
==> brew audit homebrew/science/kat
==> FAILED
homebrew/science/kat:
  * 27: Trailing whitespace was found
  * 49: Trailing whitespace was found
```

maplesond:
Removed whitespace as requested.

ilovezfs:
`brew audit` thanks you.

sjackman:
Shall we tag this formula as `bottle :unneeded` then?

maplesond:
I'm happy to go along with whatever suggestion you recommend here.  It shouldn't make too much difference from the point of view of KAT as long as these packages are available at runtime.

sjackman:
The above is more of a long-term strategy discussion. What you have now is correct for the status quo of how Python projects are handled.

I belive you should remove the `bottle` section and replace it with `bottle :unneeded`. It hasn't come up before (for me anyway) that a formula has gone from bottled to unbottled. Can someone confirm? @Homebrew/science

tdsmith:
@sjackman I think a "regular" ScipyRequirement would also make a lot of sense. Mike balked at adding a NumpyRequirement to core but I think it would be useful for -science.

sjackman:
I agree. Is it possible to factor out that requirement, or do we have to duplicate it in each formula that needs it? There's only four apparently, which is much fewer than I had expected.
`graph-tool kat nest poretools`

tdsmith:
On Thu, Jun 23, 2016 at 4:13 PM, Shaun Jackman <notifications@github.com>
wrote:

> I agree. Is it possible to factor out that requirement, or do we have to
> duplicate it in each formula that needs it?


I expect we can factor it out, like @dpo is doing for CUDA in #3759.

sjackman:
Oh good. That makes me happy.

sjackman:
@maplesond Daniel, are you interested in taking a stab at factoring out the `ScipyRequirement`? It's not required to getting this PR merged, but would be awesome.

maplesond:
Sure I can take a look.  I'm not too familiar with the details of how homebrew works though, so I'd like to make sure I understand this properly first though before I blindly start changing/adding things.  If I understand correctly it's possible to create a `ScipyRequirement` class that can then be used by the KAT formula and retain the bottling section.   So I presume a `NumpyRequirement` class already exists somewhere that I could use as a template, like the `CudaRequirement` class mentioned in #3759.  If so could you point me to where I can find that please?

sjackman:
The requirement `depends_on "scipy" => :python` instantiates the object `LanguageModuleRequirement`. See https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/language_module_requirement.rb#L3
You'll want something like that, but specific for `python`. It could be specific for `scipy` but you could also generalize it for any python module that's available in `homebrew/python`. Something like the following (totally untested, just a brain dump):

### `Requirements/python_module_requirement.rb`
```ruby
require "requirement"

class PythonModuleRequirement < Requirement
  fatal true

  def initialize(module_name, import_name = nil)
    @module_name = module_name
    @import_name = import_name || module_name
    default_formula "homebrew/python/#{module_name}"
    super([module_name, import_name])
  end

  satisfy(:build_env => false) {
    quiet_system(*%W[/usr/bin/env python -c import\ #{@import_name}])
  }
end
```

maplesond:
Sorry Shaun, I think realistically I'm probably not going to have time to get around to developing the Python requirements functionality.

One (unrelated?) thing I noticed on my system is that I get an error from KAT when it's installed via homebrew:

> Throw location unknown (consider using BOOST_THROW_EXCEPTION)
> Dynamic exception type: > boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::logic_error> >
> std::exception::what: character conversion failed

This seems to be related to boost program_options failing to understand input from the command line.  Could there be some incompatibility issue between the KAT code and the homebrew boost installation? For example, does homebrew boost work with C++11 code?  Or is there some requirement for supporting UTF-8 command line processing?

sjackman:
Hi, Daniel. No worries. I understand time constraints. Which command is failing for you? `kat --version` and `kat --help` work for me. I believe Boost is compiled by default without C++11 support, and I believe that compiling it with C++11 support makes it ABI incompatible with the default install without C++11 support. Does `kat` require C++11, and does it require C++11 features of Boost?

kat.rb
```diff
@@ -19,23 +25,31 @@ class Kat < Formula
...
   end
 
+  needs :cxx11
+
   depends_on "pkg-config" => :build
   depends_on "boost"
+  depends_on "numpy" => :python
+  depends_on "scipy" => :python
+  depends_on "matplotlib" => :python
+  depends_on "sphinx-doc" => :python
```
sjackman: Please put dependencies in alphabetical order (in the future, I'll fix it up this time around)

sjackman:
@BrewTestBot test this please

sjackman:
Merged. Thanks for your contribution, Daniel. Sorry for the extended delay. For the future, the preferred short commit log message is `kat 2.1.0`

sjackman:
Is `sphinx-doc` the name of a Python package? I don't see it on PyPI:
https://pypi.python.org/pypi?%3Aaction=search&term=sphinx&submit=search

sjackman:
I was too quick in merging this PR. It builds on neither Mac nor Linux for me. I get the following error message on both systems:
```
checking for boostlib >= 1.52... yes
checking whether the Boost::Filesystem library is available... yes
configure: WARNING: Could not find a dynamic version of the library!
configure: error: Could not find any version of the library to link to
```
https://gist.github.com/sjackman/0225bb6b531a86c2efcfb00f7d4574cf#file-01-configure-L177-L180

maplesond:
With regards to the command line error on my system, maybe this is just a linuxbrew thing?  Or maybe there is something misconfigured on my system.  If it works well in your build environments and on macs then I'm happy.  When I manually build and install on my system, or on a travis build environment I don't run into this issue.

With regards to boost availability I'm not sure exactly what the issue is here.  Again, this is something that  For my own travis builds, and installing with linuxbrew, I don't have any problem linking with boost (although I do have the command line issue discussed previously).  I have configured KAT to statically link against boost to reduce runtime dependencies but it looks like this shouldn't be a problem as you clearly have the *.a boost libs on your system.   

What you should be seeing is something like this (taken from my travis build):

~~~~
checking for sphinx-build... no
configure: WARNING: Sphinx not detected, wont create documentation.  You can still find the documentation online at: https://kat.readthedocs.org/en/latest/
checking for boostlib >= 1.52... yes
checking whether the Boost::Filesystem library is available... yes
checking for exit in -lboost_filesystem... yes
checking for /usr/local/lib/libboost_filesystem.a... yes
checking whether the Boost::System library is available... yes
checking for exit in -lboost_system... yes
checking for /usr/local/lib/libboost_system.a... yes
checking whether the Boost::Program_Options library is available... yes
checking for exit in -lboost_program_options... yes
checking for /usr/local/lib/libboost_program_options.a... yes
checking whether the Boost::Chrono library is available... yes
checking for exit in -lboost_chrono... yes
checking for /usr/local/lib/libboost_chrono.a... yes
checking whether the Boost::Timer library is available... yes
checking for exit in -lboost_timer... yes
checking for /usr/local/lib/libboost_timer.a... yes
configure: creating ./config.status
~~~~

So there should be no warning of missing dynamic libraries, because you have those.  Is the boost library directory `/gsc/btl/linuxbrew/Cellar/boost/1.60.0_2/lib/` mentioned in LDFLAGS? (Along with the boost include directory on CPPFLAGS?) If not KAT won't be able to find it.  The alternative approach is to use the `--with-boost` option in the KAT configure script.  Using this you can point directly to  `/gsc/btl/linuxbrew/Cellar/boost/1.60.0_2/lib/`.  Hope this helps.

maplesond:
Actually, I think homebrew makes links for all installed libraries into `<homebrew_dir>/lib`, and headers into `<homebrew_dir>/include`.  Providing these directories are mentioned in CPPFLAGS and LDFLAGS then this should be fine.

PR #2676, opened 2015-08-20
r: Linuxbrew fix

dholm:
On 64-bit Linux the libdir will be set to lib64 but the recipe expects
it to match #{lib}.

### DIFF
--- a/r.rb
+++ b/r.rb
@@ -66,6 +66,7 @@ class R < Formula
 
     args = [
       "--prefix=#{prefix}",
+      "--libdir=#{lib}",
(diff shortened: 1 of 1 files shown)


tdsmith:
What flavor of Linux are you running? (I don't think I saw this on Ubuntu trusty.)

Do you know how/if this impacts framework builds on OS X?

dholm:
It's RedHat 6 (at work).

It shouldn't impact OSX (which I use at home) since Homebrew would expect libraries to be in #{lib} anyway.

tdsmith:
Well, on OS X the libraries are installed inside the framework path and then Homebrew symlinks them around. I'll take a look at this later, thanks.

PR #2684, opened 2015-08-23
hwloc: Depends on :x11

dholm:
Build will fail without this dependency since it tries to link with libX11.

### DIFF
--- a/hwloc.rb
+++ b/hwloc.rb
@@ -20,6 +20,7 @@ class Hwloc < Formula
 
   depends_on "pkg-config" => :build
   depends_on "cairo" => :optional
+  depends_on :x11
 
   def install
     system "./autogen.sh" if build.head?


dpo:
The [previous PR](https://github.com/Homebrew/homebrew-science/pull/2681) (with no `:x11` dependency) built correctly on all platforms. The `:x11` dep introduces a build error. `hwloc` thinks it's building against `cairo`:
```
Hwloc optional build support status (more details can be found above):

Probe / display I/O devices: no
Graphical output (Cairo):    yes
XML input / output:          full
Plugin support:              no
```
but during the build:
```
lstopo.c:27:10: fatal error: 'cairo.h' file not found
#include <cairo.h>
         ^
1 error generated.
```

tdsmith:
Can you provide a log of a failure without a :x11 dependency?

dholm:
@tdsmith The output of `brew install -v hwloc --with-cairo` is available here: #2a97667c9cd45845dc1a

hwloc.rb
```diff
@@ -20,6 +20,7 @@ class Hwloc < Formula
 
   depends_on "pkg-config" => :build
   depends_on "cairo" => :optional
+  depends_on :x11
```
dpo: Should this be `depends_on :x11 if build.with? "cairo"` then?

dpo:
@dholm

hwloc.rb
dholm: I wasn't able to reproduce it so hold off on that unless someone else is able to reproduce.

UniqMartin:
PR #3202 addresses the issue without introducing an unnecessary X11 dependency and also bumps the formula to the newest version.

PR #2685, opened 2015-08-24
biointerchange 2.0.3+100

indiedotkim:
Bugfixes and updated features.

Good news: Linux and OS X versions have the same build number now.

Bad news: I do not know how to create bottles.


tdsmith:
I resolved the merge conflict manually but please rebase on origin/master before submitting a pull request.

There's no need to add a bottle if you're distributing binaries so I removed it. (In general, though, leaving the bottle block alone is the right thing to do; our CI helps us update that block when we pull changes.)

indiedotkim:
@tdsmith Thank you! Will rebase!

PR #2688, opened 2015-08-26
nextflow 0.15.3

pditommaso:


dpo:
Seems like the `post-install` is trying to create `~/.nextflow`, which Homebrew isn't allowing:
```
mkdir: /Users/brew/Jenkins/science/version/mavericks/home/.nextflow/framework/0.15.3: Operation not permitted
```

pditommaso:
Is this a new constraint? This formulate always worded in this way in the past.

When launching nextflow the very first time it will download some missing dependencies. So I guess won't solve the problem just pre-creating that folder.  

Any clue how it could be managed by Homebrew?

tdsmith:
It's new enforcement of an old policy. Ideally, nextflow could be convinced to look for these downloads in the Cellar, or else users will just have to deal with this on first run.

pditommaso:
I think the latter it the best option, but as a consequence of that it could no longer be possible to execute as part of [the embedded test](https://github.com/Homebrew/homebrew-science/pull/2688/files#diff-ad1708add11038ef59eb319d119c1183L31)

tdsmith:
$HOME is set to a temporary path during test, so the test might be a little slower but it should be fine if the downloaded resources aren't huge.

pditommaso:
@tdsmith That would work. But if so is trying the following create this directory: 

    /Users/brew/Jenkins/science/version/mavericks/home/.nextflow/framework/0.15.3

And not in the temporary `HOME` folder ?

tdsmith:
That is not a child of the temporary HOME; I guess HOME isn't overridden during post-install.

pditommaso:
@tdsmith I see. Thus it should be enough to remove the `post-install` because it will download the required deps automatically in the `test` step. 

I will give it a try. Tx

pditommaso:
@tdsmith OK, fixed. Thanks!

PR #2689, opened 2015-08-27
phyutility 2.7.1 (new formula)

jonchang:
Phyutility is a command line program that performs simple analyses or modifications on both phylogenetic trees and data matrices.


tdsmith:
Accepted with a few tweaks; thanks for the PR!

PR #2694, opened 2015-08-28
blaze-lib 2.4 (new formula)

hgossler:
Open-source, high-performance C++ math library for dense and sparse
arithmetic

### DIFF
--- a/blaze-lib.rb
+++ b/blaze-lib.rb
@@ -1,0 +1,37 @@
+class BlazeLib < Formula
+  desc "High-performance C++ math library for dense and sparse arithmetic"
+  homepage "https://bitbucket.org/blaze-lib/blaze/"
+  url "https://bitbucket.org/blaze-lib/blaze/downloads/blaze-2.4.tar.gz"
+  sha256 "34af70c8bb4da5fd0017b7c47e5efbfef9aadbabc5aae416582901a4059d1fa3"
+
+  depends_on "boost"
+
+  def install
+    inreplace "Configfile", "CXX=\"g++\"", "CXX=\"#{ENV.cxx}\""
+
+    system "./configure"
+    system "make"
+
+    include.install "blaze"
+    lib.install "lib/libblaze.a"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <blaze/Blaze.h>
+      int main(int argc, char *argv[])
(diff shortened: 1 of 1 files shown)


blaze-lib.rb
```diff
@@ -0,0 +1,37 @@
+class BlazeLib < Formula
+  desc "High-performance C++ math library for dense and sparse arithmetic"
+  homepage "https://bitbucket.org/blaze-lib/blaze/"
+  url "https://bitbucket.org/blaze-lib/blaze/downloads/blaze-2.4.tar.gz"
+  sha256 "34af70c8bb4da5fd0017b7c47e5efbfef9aadbabc5aae416582901a4059d1fa3"
+
+  depends_on "boost"
+
+  def install
+    inreplace "Configfile", "CXX=\"g++\"", "CXX=\"#{ENV.cxx}\""
```
dpo: If this library only builds with `g++`, there should be a `fails_with` block before `install` so Homebrew can use the appropriate compiler.

blaze-lib.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    system "./configure"
+    system "make"
+
+    include.install "blaze"
```
dpo: Is there no `make install`? What about documentation?

blaze-lib.rb
```diff
@@ -0,0 +1,37 @@
...
+    lib.install "lib/libblaze.a"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <blaze/Blaze.h>
+      int main(int argc, char *argv[])
+      {
+        return 0;
+      }
```
dpo: This is the right way to write a test, but how about a slightly more compelling example, such as https://bitbucket.org/blaze-lib/blaze/wiki/Getting_Started#!a-first-example?

hgossler:
Improved the test according to your suggestion. There is no make install, hence the manual copy of the files. The documentation is a separate download

dunn:
Just a nitpick from the audit, now:
```
==> audit problems
homebrew/science/blaze-lib:
Length is calculated as homebrew/science/blaze-lib + desc. (currently 93)
```

hgossler:
Ok, shortened it a bit

dpo:
Thanks! Looks good.

PR #2703, opened 2015-09-02
Update qualimap to version 31-08-15.

roryk:
This fixes a couple showstopper bugs.


tdsmith:
Surely we haven't accepted versions in DD-MM-YY format? :/ Version comparisons are hopeless; this is treated as though it has a major version of 31.

We can switch to the format "20150831" but we'll never be able to use a release version again.

tdsmith:
I see Ian twigged to that here: https://github.com/Homebrew/homebrew-science/pull/2040#issuecomment-89821168

Let's use `version "20150831"` here.

PR #2710, opened 2015-09-04
hyphy : fixes open-mpi dependency for with-mpi option

stevenweaver:


hyphy.rb
```diff
@@ -16,7 +16,7 @@ class Hyphy < Formula
 
   depends_on "openssl"
   depends_on "cmake" => :build
-  depends_on :mpi => :optional
+  depends_on "open-mpi" => "with-mpi"
```
dpo: The first one was the correct way to depend on MPI. If you're observing a build error, it would help if you could post here the output of `brew gist-logs hyphy`. Thanks!

stevenweaver: Hi there,

I understand that there are multiple mpi implementations, but it would be beneficial for brew to either suggest an mpi package to install or halt installation with a warning message.

https://gist.github.com/anonymous/8fdc7acfd69f010efb67

Regards,
Steven

stevenweaver:
Will ask users to ensure an mpi implementation is installed before attempting install.

dpo:
Not sure I understand. `hyphy` builds with or without MPI here. I just tried `brew install hyphy` and `brew install hyphy --with-mpi`. The only difference is that the second has `make MPI` while the first doesn't. The brew dependency `depends_on :mpi => :optional` should install MPI if the user build `--with-mpi`.

dpo:
Your build failed because CMake can't find MPI, yet somehow `brew` thought it was installed:
```
-- Could NOT find MPI_C (missing:  MPI_C_LIBRARIES MPI_C_INCLUDE_PATH) 
-- Could NOT find MPI_CXX (missing:  MPI_CXX_LIBRARIES MPI_CXX_INCLUDE_PATH) 
```
`brew doctor` reports lots of issues, including a rogue `libhyphy`. Could you try `brew uninstall -f openmpi mpich` followed by `brew install hyphy --with-mpi`?

stevenweaver:
Hi Dominique,
Thank you for looking into this. 

The issues in my brew doctor are accounted for. `dlib` is a library I use for opencv scripting, and `hmsbeagle` is for Bayesian computation.
 
I was able to reproduce this error across three different installations of OS X, which resulted in a rash update and pull request.

We can close this pull request, and I can open an issue with Homebrew after further testing. 

Regards,
Steven

dpo:
@tdsmith @xu-cheng Did something change recently with MPI? The error above is odd and I'm also having trouble. I have MPICH installed, but building a formula with a `:mpi` requirement fails because it insists on building `open-mpi`. Any idea what's going on?

xu-cheng:
> Did something change recently with MPI? 

Not that I am aware of. Maybe Tim has better knowledge on this.

dpo:
This is the kind of behavior I'm getting now:
```
$ brew install mumps
==> Installing mumps from homebrew/homebrew-science
==> Installing dependencies for homebrew/science/mumps: open-mpi
==> Installing homebrew/science/mumps dependency: open-mpi
Error: Cannot install open-mpi because conflicting formulae are installed.

  mpich2: because both install mpi__ compiler wrappers

Please `brew unlink mpich2` before continuing.
```
I can't tell what changed. Can anybody reproduce this?

Issue #2148, opened 2015-04-24
trilinos with python (dyld: Symbol not found: _PyBool_Type)


davydden:
just installed default trilinos without disabling python. 

when I run executable that's the error I got:
```
dyld: Symbol not found: _PyBool_Type
```

I never use python, so I have no ideas what/if something has to be set up. `numpy` is installed as a dependency, i use the os-x python
```
$ python --version
Python 2.7.6
```

tseemann:
To whoever understands this, here is some likely relevant lines in the formula:

```
  depends_on :python        => :recommended
  depends_on "homebrew/python/numpy"  if build.with? "python"

  args << onoff("-DTrilinos_ENABLE_PyTrilinos:BOOL=", (build.with? "python"))
```

iml:
>If you create an issue, please also create a Gist by running
>
>```
>brew gist-logs <formula name>
>```

dpo:
@davydden Could you give the shortest Python script that reproduces the issue?

davydden:
it's not a python script, it's running anything that is linked against `trilinos`, even unit tests in any other software such as https://github.com/Homebrew/homebrew-science/pull/1299

davydden:
For a moment i use `trilinos --without-python` to make C++ code run (i don't use python anyhow)

tdsmith:
Can't reproduce this. :( `Epetra_BasicPerfTest_test.exe 1 1 1 1 5 -v` succeeds.

tdsmith:
If you stumble across build logs from a build where this manifests, the gisted logs would help diagnose this.

davydden:
will do that later during the week...

davydden:
i reinstalled trilinos with python and indeed `brew test trilinos` works fine and also i can run executables. 

Could that be that this issue appears when executables are runned within the 
```
def install
...
end
```
block? Somewhat related to super-envirnoment and all that stuff?


That was the original place where I saw the problem, namely when installing deal.II formula which executes a minimum number of test programs during installation https://github.com/Homebrew/homebrew-science/blob/master/dealii.rb#L81

@tdsmith could you please try to reproduce it with deal.II?

dpo:
Sorry, I'm not seeing the error in your logs. Could you point us to it?

Are you observing the crashes with OpenMPI or MPICH2?

davydden:
it's a run-time error of C++ executables built against the trilinos library. 
[...]

PR #2712, opened 2015-09-05
Trilinos 12.2.1 (update)

davydden:
Made python optional.
Enabled ShyLU, Cholmod and Teko by default.
Force C++11 as main packages require it.

- [x] check on linux

### DIFF
--- a/trilinos.rb
+++ b/trilinos.rb
@@ -1,8 +1,8 @@
 class Trilinos < Formula
   desc "Algorithms for the solution of large-scale, complex multi-physics engineering and scientific problems"
   homepage "http://trilinos.sandia.gov"
-  url "https://trilinos.org/oldsite/download/files/trilinos-12.0.1-Source.tar.bz2"
-  sha256 "cab674e88c8ca2d2c54176af60030ed28203c0793f3c64c240363dbe7fa46b99"
+  url "https://trilinos.org/oldsite/download/files/trilinos-12.2.1-Source.tar.bz2"
+  sha256 "4a884fd5eef885815d25fc24c5a4b95e5e67b9eefaa01d61524eef92ae9319fd"
   head "https://software.sandia.gov/trilinos/repositories/publicTrilinos", :using => :git
 
   bottle do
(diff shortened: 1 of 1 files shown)


dpo:
Why make Python `:optional`? Because of `dealii`? Isn't there a patch yet for that bug?!

davydden:
unfortunately not... 

but how about we do some `replace` in trilinos config files which got installed to Cellar to correct this issue? I am not sure if that will break python usage, though...

davydden:
do you use `python` interface of trilinos to test it?

davydden:
to be more specific what @wfspotz suggested here https://github.com/Homebrew/homebrew-science/issues/2148#issuecomment-103614509

> A workaround is to remove libpytrilinos from the export Makefile you are using. 

namely modify `/lib/cmake/Trilinos/TrilinosConfig.cmake` to remove `pytrilinos` from `Trilinos_LIBRARIES` list.

What I don't know is wheter or not it will break any python programs which use trilinos. I have no idea how it works there and wheterh or not they might be using that `.cmake` file to set up something.

davydden:
the current version, namely with 
```
inreplace "#{lib}/cmake/Trilinos/TrilinosConfig.cmake" do |s|
  s.gsub! "PyTrilinos;", "" if s.include? "COMPONENTS_LIST"
end
```
worked fine for me with `dealII` on `10.10.5`. Hopefully there will not be any problems with test bot...

davydden:
@BrewTestBot test this please

wfspotz:
I talked to the build folks not too long ago about making the removal of libpytrilinos from the export Makefiles a part of the build system.  At the very least it is higher on their priority list.

I do this in a build script for a project I work on and it works just fine.  If you want to add this to your build process, it should not create any problems.

davydden:
thanks, Bill! I implemented this strategy here (https://github.com/Homebrew/homebrew-science/pull/2712#issuecomment-138061653) in this PR and it works good so far.

dpo:
@davydden You have a check box in your original message. Are you currently testing on Linux, or is this good to go?

dpo:
@BrewTestBot test this please

davydden:
just tested on Ubuntu 14.04. All good.

davydden:
@BrewTestBot test this please

PR #2714, opened 2015-09-06
opencv modify libstdc++ to libc++

tobefuturer:
opencv2/flann/lsh_index.h need std::unorder_map but libstdc++ not support
libstdc++ only support std::tr1::unorder_map so modify libstdc++ to libc++


tdsmith:
This change requires CUDA 7 which is probably fine but most of this logic can go away now.

tobefuturer:
I see.  I ignore the dependence of CUDA .Thanks for your reminder.

iMichka:
opencv has been migrated to homebrew-core. Any wish - related issue has to be opened there so it can be considered there. Thanks.

PR #2718, opened 2015-09-07
Upgrade superlu to version 5.0

Moisan:
Superlu had a new release, bump the version and a few changes to make audit happy.


superlu.rb
```diff
@@ -1,7 +1,8 @@
 class Superlu < Formula
+  desc "Large, sparse, nonsymmetric linear equations systems"
```
dpo: Thanks! How about "Solution of large, sparse nonsymmetric systems of equations"?

superlu.rb
```diff
@@ -1,7 +1,8 @@
...
   homepage "http://crd-legacy.lbl.gov/~xiaoye/SuperLU/"
-  url "http://crd-legacy.lbl.gov/~xiaoye/SuperLU/superlu_4.3.tar.gz"
-  sha1 "d2863610d8c545d250ffd020b8e74dc667d7cbdd"
+  url "http://crd-legacy.lbl.gov/~xiaoye/SuperLU/superlu_5.0.tar.gz"
+  sha256 "4856fa42b623e7047c1b0532ed0fc35f5dc283a74246a7f25ef8aa5155084ecc"
   revision 2
```
dpo: You can remove the `revision` line when upgrading to a new version.

davydden:
could you please check that formulae which use `superlu` still work (`petsc`, `trilinos`)? Would be a pity to break them...

dpo:
@tdsmith @xu-cheng Shouldn't the bot automate that? Every time a formula is upgraded or bumped, the formulas that `brew --use` it should also be rebuilt, tested and rebottled.

superlu.rb
Moisan: "Solution of large, sparse nonsymmetric systems of equations" is 85 characters and audit restricts it to 80 :). I'm going with "Solve large, sparse nonsymmetric systems of equations".

dpo:
All seems well for me in `armadillo`, `petsc` and `trilinos`.

tdsmith:
> Shouldn't the bot automate that? Every time a formula is upgraded or bumped, the formulas that brew --use it should also be rebuilt, tested and rebottled.

The bot installs and tests dependents which are bottled and have tests defined.

dpo:
Does it? Where are those bottles? I don't see `brew pull --bottle` pull them and they don't seem to be in the list of artefacts. That would be a great side-effect though. Another good reason to add tests to formulae.

iml:
I think Tim means the dependents which have bottles available are tested, but not those which must be build from source.

tdsmith:
They aren't rebuilt, just installed from the bottles that already exist in order to test them; in the common case that the new version of the parent formula is ABI-compatible with the prior version, nothing needs to be rebuilt. If the new version isn't compatible, hopefully the test fails.

davydden:
`petsc` is broken with this superlu (at least on linuxbrew).

dpo:
@davydden All the tests passed for me. That's why I tested locally. I think it would be safer if the bot rebuilt the dependents automatically, but then again, the bot doesn't test on linux. What's the problem exactly?

davydden:
did you test it on linux? There were some compiler errors...

davydden:
nevermind, it strange (worked on ubuntu but didn't on linuxmint).

davydden:
petsc was not reinstalled on my ubuntu, I **do** have the same compiling error there which is most likely related to the change of API in `superlu`:
```
/home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/src/mat/impls/aij/seq/superlu/superlu.c:180:37: error: too few arguments to function 'dgssvx'
     PetscStackCall("SuperLU:dgssvx",dgssvx(&lu->options, &lu->A, lu->perm_c, lu->perm_r, lu->etree, lu->equed, lu->R, lu->C,
```

davydden:
the same should happen when you try to reinstall petsc on os-x with `superlu 5.0`.

davydden:
```
In file included from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/src/mat/impls/aij/seq/superlu/superlu.c:27:0:
/home/davydden/.linuxbrew/opt/superlu/include/superlu/slu_ddefs.h:123:1: note: declared here
 dgssvx(superlu_options_t *, SuperMatrix *, int *, int *, int *,
 ^
In file included from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/include/petscsys.h:1559:0,
                 from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/include/petscis.h:7,
                 from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/include/petscvec.h:9,
                 from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/include/petscmat.h:6,
                 from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/include/petsc/private/matimpl.h:5,
                 from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/include/../src/mat/impls/aij/seq/aij.h:5,
                 from /home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/src/mat/impls/aij/seq/superlu/superlu.c:11:
/home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/src/mat/impls/aij/seq/superlu/superlu.c:205:38: warning: passing argument 17 of 'dgsisx' from incompatible pointer type [-Wincompatible-pointer-types]
                                      &lu->mem_usage, &lu->stat, &sinfo));
                                      ^
/home/davydden/Library/Temp/petsc20150918-26914-16a95wm/petsc-3.6.1/include/petscerror.h:871:64: note: in definition of macro 'PetscStackCall'
 #define PetscStackCall(name,routine) do { PetscStackPush(name);routine;PetscStackPop; } while(0)
                                                                ^
```

davydden:
and so on

dpo:
You're right. It breaks here too. I'm not sure why it succeeded when I last tried. Something else must have happened. This is very unreliable.

@tdsmith `petsc` defines no tests. It should, but even if it did, I'm not sure an incompatibility like this would be caught. Wouldn't it make sense for the bot to rebuild the dependents from source when a formula is updated? In essence, all dependent bottles would also be updated.

tdsmith:
In a world of unlimited CI resources and user bandwidth that would be the most bulletproof thing to attempt but as things stand, rebuilding every formula's entire dependency tree on every PR isn't realistic.

davydden:
how about not the whole dependency tree, but the first lelvel, i.e. if we rebuild A (A->B->C; A->D), then B and D got checked and we hope C is alright?

davydden:
maybe check if there is a major version change (4.9 ->5.0) or minor (4.8->4.9). Do those tests for major only as those usually imply incompatible API, as it is the case here.

dpo:
@davydden We have to add meaningful tests to `petsc`, `slepc`, `dealii`, etc., so they can be run when a dependency is updated. Build-time tests are insufficient.

dpo:
Should be fixed by #2753.

Issue #2724, opened 2015-09-08
test python when appropriate in opencv.rb and opencv3.rb


ahundt:
Under the test section of the `opencv` and `opencv3` formulae we should add code that does the following with python is enabled in addition to the existing c++ test:

```python
import cv2
cv2.__version__
```

ahundt:
not sure how to do this correctly for opencv3.rb, because it may not work correctly by default since the formula is keg-only. Thoughts?

PR #2725, opened 2015-09-08
#2724 test python when appropriate in opencv.rb

ahundt:
#2724

### DIFF
--- a/opencv.rb
+++ b/opencv.rb
@@ -152,5 +152,12 @@ class Opencv < Formula
     EOS
     system ENV.cxx, "test.cpp", "-I#{include}", "-L#{lib}", "-o", "test"
     assert_equal `./test`.strip, version.to_s
+    
+    pythons = []
+    pythons << "python"  if build.with? "python"
+    pythons << "python3" if build.with? "python3"
+    pythons.each do |python|
+      system python, " -c \"import cv2;print(cv2.__version__)\""
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


tdsmith:
FWIW the bot doesn't consider non-default options so they aren't very useful in `test`.

ahundt:
I believe enabling python is one of the default options

tdsmith:
Of course; python3 isn't. Making this non-conditional and only testing python 2 would be great.

opencv.rb
```diff
@@ -152,5 +152,12 @@ def install
     EOS
     system ENV.cxx, "test.cpp", "-I#{include}", "-L#{lib}", "-o", "test"
     assert_equal `./test`.strip, version.to_s
+    
+    pythons = []
+    pythons << "python"  if build.with? "python"
+    pythons << "python3" if build.with? "python3"
+    pythons.each do |python|
+      system python, " -c \"import cv2;print(cv2.__version__)\""
```
tdsmith: This doesn't need to be escaped; `system "python", "-c", "import cv2"`

ahundt: I was also hoping to check that the version string is correct to detect conflicts. Is there a supported use case for testing other versions (more applies to opencv3 actually). For example, can tests be run on personal machines with customized configurations?

tdsmith: You can check the version string using assert_match against `shell_output("python -c '...'")` if you want. sys.path is modified during the test block setup so I don't think it will tell you much about your normal Python environment.

ahundt:
why closed?

tdsmith:
I added a test in e5d02b5.

ahundt:
cool thanks!

PR #2738, opened 2015-09-15
adam 0.17.1

heuermh:
Note the build for ADAM changed significantly between version 0.13.0 and the current version 0.17.1.

The major difference is an external dependency on ```apache-spark``` rather than including all of the Spark jars in the ADAM jar.


iml:
Thanks!

PR #2748, opened 2015-09-17
Bump graph-tool version to 2.7

count0:


count0:
Friendly bump on this. Despite what is shown above, it seems like all tests have passed.

dpo:
Seems like El Capitan doesn't want to build this. I'll try and restart it.

count0:
The build fails for el capitan because of sparsehash. See Homebrew/homebrew#41087.

dpo:
OK, at least now there's an error message.

dpo:
There was an issue uploading the bottles. Trying again.

PR #2749, opened 2015-09-18
netcdf: revision 1 for hdf5

claudiodsf:
Programs linked to `libnetcdf` fail with: 
```
dyld: Library not loaded: /usr/local/lib/libhdf5_hl.9.dylib
  Referenced from: /usr/local/lib/libnetcdf.7.dylib
  Reason: image not found
```
After upgrading `hdf5` to version 1.8.15.

Adding a revision number forces `netcdf` to recompile against the new `hdf5`.

There might be however a problem in the way `netcdf` is compiled, since it should actually link against `libhdf5_hl.dylib` which, in turn, is a symlink to `libhdf5_hl.9.dylib` or `libhdf5_hl.10.dylib`, depending on version.


davydden:
something like `-Wl,-soname` (but for clang) could be missing in netcdf (see http://stackoverflow.com/a/6339947/888478)

iml:
Unfortunately -soname isn't supported by the OS X linker. HDF5 updates shouldn't be pulled without bumping its dependents, per https://github.com/Homebrew/homebrew-science/pull/2388 and prior discussions. The problem then becomes splitting up the PR such that all those can be built under the CI time limit.

dpo:
But isn't this another manifestation of https://github.com/Homebrew/homebrew-science/pull/2718#issuecomment-141569221?

tdsmith:
Mike, Dominyk, and Xu Cheng are working on offloading some of the CI load to Travis, which may allow us to increase the time limit on the bot for bottles.

tdsmith:
install_name is the soname equivalent for darwin; if someone can verify that install_name on darwin and the soname on linux are different and the releases are indeed ABI-compatible then someone should reach out to the hdf5 devs. Otherwise, I don't think Homebrew should do anything except a) exercise vigilance when upgrading libraries like hdf5 that we've observed to have poor upgrade behavior on apparently trivial version changes and b) add tests so the bot can help catch failures.

dpo:
Ok. Independently, I've been trying to implement my suggestion on Travis as well (for OSX, not Linux).

saketkc:
@dpo Is your implementation on travis public? I would like to take another shot at travis for `linuxbrew`

dpo:
@saketkc Here are my current [`.travis.yml`](https://github.com/dpo/homebrew-science/blob/travis-osx/.travis.yml) and [build script](https://github.com/dpo/homebrew-science/blob/travis-osx/travisCI/travis.sh). This is for OSX, so stuff can be installed from bottles when convenient. I'm currently not worrying about the log capacity or time limit. [Here](https://travis-ci.org/dpo/homebrew-science/builds/81195791) is an example log where [`armadillo`] was updated.

I'd love to get this working on Linux too!

tdsmith:
I'd pull this but the bottle job never ran :/ @BrewTestBot test this please

PR #2750, opened 2015-09-18
Trilinos 12.2.1 (linuxbrew fix)

davydden:
Disable OpenMP for now as this leads to compiler errors with deal.ii

### DIFF
--- a/trilinos.rb
+++ b/trilinos.rb
@@ -4,6 +4,7 @@ class Trilinos < Formula
   url "https://trilinos.org/oldsite/download/files/trilinos-12.2.1-Source.tar.bz2"
   sha256 "4a884fd5eef885815d25fc24c5a4b95e5e67b9eefaa01d61524eef92ae9319fd"
   head "https://software.sandia.gov/trilinos/repositories/publicTrilinos", :using => :git
+  revision 1
 
   bottle do
     revision 1
@@ -98,6 +99,11 @@ class Trilinos < Formula
                -DTrilinos_VERBOSE_CONFIGURE:BOOL=OFF
                -DTrilinos_WARNINGS_AS_ERRORS_FLAGS=""]
 
+    # Explicit instantiation will build object files for the Trilinos templated classes with the most common types.
+    # That should speed up compilation time for librareis/driver programs which use Trilinos.
+    # see https://trilinos.org/pipermail/trilinos-users/2015-September/005146.html
+    args << "-DTrilinos_ENABLE_EXPLICIT_INSTANTIATION:BOOL=ON"
+
     # enable tests only when we inted to run checks.
     # that reduced the build time from 130 min to 51 min.
     args << onoff("-DTrilinos_ENABLE_TESTS:BOOL=",  (build.with? "check"))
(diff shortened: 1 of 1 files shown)


trilinos.rb
```diff
@@ -123,7 +124,13 @@ def install
...
 
     args << onoff("-DTPL_ENABLE_MPI:BOOL=",         (build.with? "mpi"))
-    args << onoff("-DTrilinos_ENABLE_OpenMP:BOOL=", (ENV.compiler != :clang))
+    # TODO:
+    # OpenMP leads deal.II to fail with compiler errors in trilinos headers even though trilinos compiles fine
+    # It could be that there is a missing #include somewhere in Trilinos which becames visible when we
+    # try to use it.
+    # For now disable OpenMP:
+    # args << onoff("-DTrilinos_ENABLE_OpenMP:BOOL=", (ENV.compiler != :clang))
+    args << "-DTrilinos_ENABLE_OpenMP:BOOL=OFF"
```
dpo: Does this only happen on Linux? What about gcc on OSX?

davydden: likely to be the same issue. I don't use gcc on OSX, though.

dpo:
There's a problem on line 227 because the `superlu` dep was commented out.

davydden:
It has nothing to do with this PR o Superlu 5.0. There are compiler errors, i notifed developers on the mailing list about two weeks ago, no reply so far...

dpo:
I remember. What I'm saying is that lines with `build.with? "superlu"` should be commented out as well.

dpo:
There's a similar issue with other commented-out dependencies, e.g., `petsc`.

davydden:
Unless you feel strong about it, i would keep it the it is now: when `superlu` is fixed, it's going to be one line to un-comment. Same applies to `petsc`. Everything else will be ready to use.

dpo:
The cleanest would probably be to fix in a separate PR and then rebase this one.

davydden:
With all due respect, i don't think it's a good idea. In addition to `petsc` and `superlu` mentioned above, one would also need to comment all the lines related to `lemon`, `cask`, `qd`, `binutils`, `csparse`. The current state folllows what we agreed in https://github.com/dpo/homebrew-trilinos/blob/master/new-trilinos.rb . Therefore, I will keep this PR as is.

dpo:
I don't know why it wasn't brought up at the time. It just seems inconsistent and it's an easy fix. It's in #2759.

dpo:
You can rebase this against the recent changes in `trilinos.rb`.

trilinos.rb
dpo: Was this reported upstream?

davydden: i still investigate whether or not adding some #include can fix the issue. As soon as I will know more, i'll post to the mailing list

davydden: i reported the issue to the mailing list. Indeed it appears to be a Trilinos bug.

dpo: Thanks!

davydden:
shall we merge this one?

dpo:
There was an issue with the bottles, which is now resolved.

PR #2759, opened 2015-09-21
Trilinos: fix inconsistencies.

dpo:
Some dependencies are currently commented out.
Comment out corresponding code.

### DIFF
--- a/trilinos.rb
+++ b/trilinos.rb
@@ -161,12 +161,12 @@ class Trilinos < Formula
     args << onoff("-DTPL_ENABLE_Matio:BOOL=",       (build.with? "libmatio"))
     args << onoff("-DTPL_ENABLE_yaml-cpp:BOOL=",    (build.with? "yaml-cpp"))
 
-    if (build.with? "suite-sparse") && (build.with? "csparse")
-      args << "-DTPL_ENABLE_CSparse:BOOL=ON"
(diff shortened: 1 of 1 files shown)


trilinos.rb
```diff
@@ -216,19 +216,19 @@ def install
 
     args << onoff("-DTPL_ENABLE_SCALAPACK:BOOL=",   (build.with? "scalapack"))
 
-    args << onoff("-DTPL_ENABLE_SuperLU:BOOL=",     (build.with? "superlu"))
-    args << "-DSuperLU_INCLUDE_DIRS=#{Formula["superlu"].opt_include}/superlu" if build.with? "superlu"
+    args << onoff("-DTPL_ENABLE_SuperLU:BOOL=", "OFF") #   (build.with? "superlu"))
```
davydden: shouldn't here (and elsewhere) be `0`/`false` instead of `OFF`?

dpo:
We use `ON`/`OFF` everywhere: https://github.com/Homebrew/homebrew-science/pull/2759/files#diff-06990b9d06d23ea1ef3926d6f3944c26R77
Has anything changed in that respect?

trilinos.rb
dpo: Sorry, I misunderstood your comment. Yes, it should be `false`. Thanks.

davydden:
@BrewTestBot test this please

davydden:
failures in `brew doctor`, otherwise pass.

PR #2761, opened 2015-09-21
blasr: build with MPI if HDF5 was.

dpo:


dpo:
@sjackman @tseemann Is this reasonable? I'm getting undefined MPI symbols when I try to build `blasr` because my `hdf5` was built `--with-mpi`.

dpo:
@BrewTestBot test this please

sjackman:
Hmm. I believe the Homebrew policy is that the default options of a package should not depend on the state of the system. I think this should instead print an error message if `hdf5` was built with `mpi` but `blasr` is not being built with `mpi`.

`blasr` itself does not use `hdf5` at all, does it? It's frustrating that the build options of a package that uses `hdf5` should change depending on how `hdf5` was built.

sjackman:
I'm :-1: on this particular patch, because I don't think the `blasr` formula should know anything about `mpi`, but the problem does need to be addressed.

iMichka:
It seems that there is no consensus on this for moment. I do not know if it's still open for discussion, but I will just close this PR for the time being, to clean up a little bit the PR list.

BTW blasr can't be build anymore, see #4254

PR #2768, opened 2015-09-22
root6 6.04.02

kratsg:
https://root.cern.ch/root/html604/notes/release-notes.html#patch-releases

Primarily, there were quite a few bugs targeting the fact that users had to downgrade Xcode in order to work-around 6.04.00. 6.04.02 is a very stable and more commonly used version for right now.


dpo:
Audit error:
```
==> brew audit homebrew/science/root6
==> FAILED
Error: 1 problem in 1 formula
homebrew/science/root6:
 * Emacs Lisp files were linked directly to /usr/local/share/emacs/site-lisp

This may cause conflicts with other packages; install to a subdirectory instead, such as
/usr/local/Cellar/root6/6.04.02/share/emacs/site-lisp/root6

The offending files are:
  /usr/local/Cellar/root6/6.04.02/share/emacs/site-lisp/root-help.el
```

kratsg:
Hi @dpo - I do not understand the error. I cannot reproduce it locally.

```
/usr/local/Library/Taps/homebrew/homebrew-science (bumpROOT6)$ brew audit homebrew/science/root6
```

dpo:
Try `brew audit --strict root6` after you've built and installed it. I'm getting other audit warnings too:
```
homebrew/science/root6:
 * `mirror` (line 6) should be put before `version` (line 4)
 * `caveats method` (line 78) should be put before `test block` (line 63)
Length is calculated as homebrew/science/root6 + desc. (currently 82)
 * Please remove the indefinite article "An" from the beginning of the description

```

kratsg:
```
Lord Stark:~$ brew audit --strict root6
==> brew style root6
Error: The 'rubocop' gem is installed but couldn't find 'rubocop' in the PATH:
/Users/kratsg/.gem/ruby/2.0.0/bin:/Users/kratsg/mongodb/bin:/usr/local/heroku/bin:/Users/kratsg/.rvm/gems/ruby-2.1.4/bin:/Users/kratsg/.rvm/gems/ruby-2.1.4@global/bin:/Users/kratsg/.rvm/rubies/ruby-2.1.4/bin:/usr/local/texlive/2013/bin/x86_64-darwin:/usr/local/bin:/usr/local/share/npm/bin:/usr/local/Cellar/root/HEAD/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/opt/X11/bin:/usr/texbin:/Users/kratsg/.rvm/bin:/usr/local/Library/ENV/scm
```

I'm starting to get super confused by why I can't seem to audit. It looks like it has something to do with my ruby version, but `brew doctor` doesn't show me anything.

kratsg:
This has two problems left that I do not understand personally

```
==> brew style root6
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.6-compliant syntax, but you are running 2.1.4.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.

==> audit problems
homebrew/science/root6:
 * Stable: version v.source should not have a leading 'v'
 * Stable: version (v.source) is set to a string without a digit

```

not sure what needs to be done.

kratsg:
It turns out those complaints are due to the `version` variable being specified after the `url` and `mirror`. If I move the `version` variable before those lines, I get the following output:

```
Lord Stark:/usr/local/Library/Taps/homebrew/homebrew-science (bumpROOT6)$ brew audit --strict root6
==> brew style root6
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.6-compliant syntax, but you are running 2.1.4.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.

==> audit problems
homebrew/science/root6:
 * `mirror` (line 6) should be put before `version` (line 4)
 * Emacs Lisp files were linked directly to /usr/local/share/emacs/site-lisp

This may cause conflicts with other packages; install to a subdirectory instead, such as
/usr/local/Cellar/root6/6.04.02/share/emacs/site-lisp/root6

The offending files are:
  /usr/local/Cellar/root6/6.04.02/share/emacs/site-lisp/root-help.el

```

It doesn't look like `brew audit` is complaining in a consistent manner. Is this a bug with `rubocop`?

dpo:
Thanks for your patience and contribution!

kratsg:
@dpo Thanks! I'm still learning the ins-and-outs of Homebrew, so sorry about that. If you don't mind answering another question:

I wrote a short formula for a package that can be compiled using either `root6` or `root` (eg: root5). In both of these formulas in the `homebrew-science` keg -- they have a line like

```
   conflicts_with "root"
```

How can I get my formula to accept either `root` or `root6` for compilation (eg: auto-detect and use)? The best option I can think of is to write it with `root6` as a recommended dependency (flag) so that the user can switch to the other.

dpo:
You could say something like
```
depends_on "root6" => :recommended
depends_on "root" if build.without? "root6"
```
and in your `install` method, you can use
```
if build.with? "root6"
  ...
else
 ...
end
```

kratsg:
Thanks! Both versions have the same headers/libraries - it is quite literally just differences between then, such as analogous to say... python2.7 and python3. The install method doesn't change - it's just that I need the dependency resolution which comes with a homebrewed root (or root6). I will try that.

veprbl:
Why did you downgrade cmake for "./configure"?

kratsg:
@veprbl I could not get CMake to set the lisp path. `./configure` was a lot more flexible in this regard. If you want to try and get CMake working again with `brew audit` not complaining, be my guest -- but the current version of ROOT6 had issues and it needed to be upgraded.

veprbl:
This is not cool.

Also:
```
# in order to update, simply change version number and update sha256
depends_on "openssl" => :recommended # use homebrew's openssl
# root5 obviously conflicts, simply need `brew unlink root`
# cling also takes advantage
```
These are useless.

```
system "./configure", "--help"
```
How this passed the review?

PR #2771, opened 2015-09-23
Petsc 3.6.1 (fix)

davydden:
Fix a bug introduced during the switch to Superlu43


dpo:
Duh. Sorry about that. But I don't understand how in the world the build succeeded on the bot previously...

davydden:
it succeeded because it built without `superlu`, so that part `if build.with?` never got executed.

dpo:
But `superlu` was `:recommended`?!

davydden:
well, it does not matter. `superlu43` got installed, but it never got into `if build.with? "superlu"` and effectively PETSc was configured without superlu.

dpo:
That's right.

PR #2801, opened 2015-10-01
trinity 2.1.0

saketkc:


saketkc:
@BrewTestBot test this please

saketkc:
@BrewTestBot test this please

tseemann:
@saketkc - does the Console output help figure out what is happening?
http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/3256/console

trinity.rb
```diff
@@ -13,16 +14,13 @@ class Trinity < Formula
     sha256 "f79aeab456fad74f45276ff723e05c14783b81083cc87ff18fa72209e7485ade" => :mountain_lion
   end
 
+  depends_on "ncurses" => :build
```
iml: This is provided by OSX and thus in `homebrew/dupes`

saketkc:
@tseemann Not really. The `ltinfo` seems to be related to `ncurses`. It works fine on my linux machine.

saketkc:
@iml Does the failure sound familiar to you? I might not have access to a Mac to test this out.

iml:
Yes, this is the error:
```
==> brew uses homebrew/science/trinity
Error: No available formula for ncurses
```
It works because you have `homebrew/dupes` tapped, but the bot doesn't and non-core dependencies require the full path. I don't really agree with adding this, but if ncurses is not common on clusters etc that's fine.

saketkc:
```
gcc -g -Wall -O2 -o samtools bam_tview.o bam_plcmd.o sam_view.o bam_rmdup.o bam_rmdupse.o bam_mate.o bam_stat.o bam_color.o bamtk.o kaln.o bam2bcf.o bam2bcf_indel.o errmod.o sample.o cut_target.o phase.o bam2depth.o padding.o bedcov.o bamshuf.o bam_tview_curses.o bam_tview_html.o  libbam.a -Lbcftools -lbcf -ltinfo -lcurses  -lm -lz -lpthread
ld: library not found for -ltinfo
```

This goes fine on linux.

iml:
Looks like an ncurses-related library. Try adding back `depends_on "homebrew/dupes/ncurses" => :build`

saketkc:
Fails irrespectively.

iml:
ncurses needs to be built with the terminfo library, see https://github.com/gittup/ncurses/blob/gittup/INSTALL#L909.

saketkc:
See #2913

PR #2811, opened 2015-10-02
Updated the sha256 hash of the Neuron source tarball to the latest ve…

wvangeit:
The sha256 hash of the Neuron source tarball is out of date, preventing compilation from source. I've update the hash value.


neuron.rb
```diff
@@ -1,7 +1,7 @@
 class Neuron < Formula
   homepage "http://www.neuron.yale.edu/neuron/"
```
dpo: Would you mind adding a `desc` field with a brief description of the software?

wvangeit: I've added a description line.

PR #2818, opened 2015-10-03
snpeff: update to 4.1l

chapmanb:
Upgrade to latest point release which fixes issues with structural
variant annotations.


tseemann:
Thanks Brad.

PR #2832, opened 2015-10-06
dida 1.0.1 (new formula)

sjackman:
Distributed Indexing Dispatched Alignment
DIDA is a novel framework that performs the large-scale alignment tasks by
distributing the indexing and alignment stages into smaller subtasks over a
cluster of compute nodes.


sjackman:
```
checking for g++-5 option to support OpenMP... -fopenmp
checking mpi.h usability... no
checking mpi.h presence... no
checking for mpi.h... no
checking for MPI_Init in -lmpi... no
configure: error: DIDA must be compiled with MPI support.
```

PR #2839, opened 2015-10-06
Update armadillo.rb

seonho:
Update armadillo 6.100.0


dpo:
Duplicate of #2836. Thank you for your submission though!

PR #2765, opened 2015-09-22
vtk 6.3.0

quantumsteve:
updated vtk to version 6.3.0

PR #2842, opened 2015-10-06
VTK --with-qt5 should depend on pyqt5

fran6co:
It should use pyqt5 for qt5


fran6co:
I have no idea why this is failing.

fran6co:
@BrewTestBot test this please

iml:
That's more of a warning; it needs to be patched upstream. Using `depends_on :python` means python bindings in the current bottles are linked against Homebrew python.

dpo:
https://github.com/Homebrew/homebrew-science/pull/2765#issuecomment-142336362

PR #2790, opened 2015-09-26
[PCL] Upstream merged CUDA 7.5 not requiring libstdc++

fran6co:
https://github.com/PointCloudLibrary/pcl/pull/1347

PR #2844, opened 2015-10-06
[PCL] QT5 support for HEAD

fran6co:
This supersedes https://github.com/Homebrew/homebrew-science/pull/2595 and fixes https://github.com/Homebrew/homebrew-science/issues/2593

I fixed some stuff:

  - squashed @ahundt commits
  - removed @csparker247 commit, this should be fixed upstream.. the apps that depend on QT shouldn't build if QT is not present... I'll do a proper fix later on
  - cleaned up style
  - QT5 is only supported by the HEAD version, 1.7.2 doesn't support it
  - should be using pyqt5 when using qt5 (it's causing some problems if using python3)


fran6co:
First thing this change affects HEAD only, for 1.7.2 just do ```--without-qt``` and it's going to build it without qt. In the case of HEAD both of them are optional, so just add ```--with-qt``` or ```--with-qt5``` to select the qt version. For no qt just don't add them.

dpo:
What of #2843? I also did a little bit of cleaning in there.

fran6co:
Oh, sorry.. I didn't see it. Uhm, you are reverting https://github.com/Homebrew/homebrew-science/pull/2790 in yours and it has the disable apps when qt is disabled that I think it's wrong.

dpo:
I guess #2595 hasn't been rebased. Didn't mean to revert #2790.

dpo:
@BrewTestBot test this please

fran6co:
Oh, and the other thing it's missing is the pyqt5 dependency.

PR #2847, opened 2015-10-07
New Travis Config

xu-cheng:
Supporting both OS X and Linux(using Linuxbrew docker image).

cc @Homebrew/science @tdsmith


xu-cheng:
Building result: https://travis-ci.org/xu-cheng/homebrew-science/builds/84053933

tdsmith:
Very slick :+1:

dpo:
A recent update of `hdf5` showed that it's in general not safe to pull a formula without rebuilding its dependents from source. Not all formulas have tests, and even when they do, running the tests doesn't necessarily expose incompatibilities. The net result is a lot of new open issues. I wrote [an atttempt](https://github.com/dpo/homebrew-science/blob/travis-osx/travisCI/travis.sh#L25) at rebuilding the dependents from source. I welcome better solutions, but I think it's an important feature.

xu-cheng:
@dpo Sorry, but I don't think you should rebuild the dependents from source. Not only it doesn't solve the problem(since users are still installing bottle as default), it will make things worse by hiding the the information on which bottles should be rebuilt.

> Not all formulas have tests, and even when they do, running the tests doesn't necessarily expose incompatibilities. 

I think that's the root of your problem. Generally a substantial test other than `--version` or `--help` should be enough to expose the incompatibilities, which is what we are seeing in the core. If that's still not enough, it only means you should add a even more substantial test.

saketkc:
Awesome! Are linux bottles supported too?

xu-cheng:
@saketkc We cannot use travis to create bottles. The reason is travis doesn't support dependent build, and we haven't found a way to safely upload bottles on travis without exposing credential tokens in building env.

saketkc:
@xu-cheng By dependent builds, are you referring to the static/dynamic linking issue? Also, is it not possible to encrypt credentials in `.travis.yml`? http://docs.travis-ci.com/user/encryption-keys/

xu-cheng:
By dependent builds, I mean a downstream build job can be triggered after PR build is finished. (Travis's issue can be found [here](https://github.com/travis-ci/travis-ci/issues/249)).

> Also, is it not possible to encrypt credentials in .travis.yml?

That doesn't work for PR. In more detail, Homebrew requires a rather unique flow to build bottles. The token should not be available on building phase but be available in uploading phase. Hence, we require dependent build and a secure way to store/transport artifacts between these two phases.

dpo:
@xu-cheng The idea of rebuilding dependents was to rebuild bottles, but if that's not possible on Travis, there's no point. I agree that tests are the right way to resolve the issue.

xu-cheng:
@dpo Any objection to merge this.

In midterm, we may leverage travis to test PR. Our own CI infrastructure will then only be used to build bottle after travis is greened. Therefore, it can help to reduce the backlog whenever our CI is down.

dpo:
No, no objections. I think it's great.

sjackman:
Fantastic! Thanks, Xu. :100:

PR #2848, opened 2015-10-07
Updated Neuron formula to allow correctly building with --HEAD and --with-mpi option

wvangeit:
Changes:
- Added patch to Makefile.am. When building with --HEAD, Makefile.am
  patch is applied, otherwise Makefile.in patch.
- To correctly set the version during --HEAD compilation, the .hg folder
  has to be copied to the build directory. Added new download strategy
for this purpose
- Added :cc to mpi dependency, which triggers the download of an mpi
  bottle when building with --with-mpi
- In case of building with --HEAD, autoconf/automake/libtool are added
  as dependency
- In case of building with --HEAD, the script "./build.sh" in the repo
  is executed


wvangeit:
Will create a new PR with also fix for inter-views

PR #2855, opened 2015-10-08
vips 8.1.0

ottob:
simple update


ottob:
Do I need to do anything? I can tell why it fails.

dpo:
@BrewTestBot test this please

PR #2860, opened 2015-10-09
flint 2.5.2

xu-cheng:


saketkc:
`desc` seems to be missing.

xu-cheng:
@saketkc Thanks for catching that.

xu-cheng:
@BrewTestBot test this please

PR #2864, opened 2015-10-09
FStar 0.9.1 (version increment)

beurdouche:
- Update to version 0.9.1
- Remove dependancy in fsharp as the package has been deleted
- Change path to z3 formula that was migrated to core


beurdouche:
See https://github.com/Homebrew/homebrew/issues/44797

PR #2869, opened 2015-10-10
strelka 1.0.14 (new formula)

brunogrande:
added Strelka, somatic variant caller (for SNVs and indels)

### DIFF
--- a/strelka.rb
+++ b/strelka.rb
@@ -7,7 +7,7 @@ class Strelka < Formula
   # Needs Python for testing installation
   depends_on :python => :build
   # Needs rsync for installing to #{prefix}
-  depends_on :rsync => :build
+  depends_on "homebrew/dupes/rsync" => :build
 
   def install
     system "./configure", "--disable-debug",
--- a/strelka.rb
+++ b/strelka.rb
@@ -4,6 +4,7 @@ class Strelka < Formula
   url "ftp://strelka:%27%27@ftp.illumina.com/v1-branch/v1.0.14/strelka_workflow-1.0.14.tar.gz"
   sha256 "b3aeca63dfea7787a4046a0ff747e9fe4a267297a3d0ae5126fab890e7152f39"
 
+  # 
   # Needs Python for testing installation
   depends_on :python => :build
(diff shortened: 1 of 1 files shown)


brunogrande:
## Linux Build
I'm not sure why the Linux build stalls on `make install` for Strelka. Any ideas? I didn't run into this issue when installing on a local Linux machine. I'll test it out on other Linux machines. 

## OS X Build
From what I can tell, the issue with the OS X build is an incompatibility between the version of samtools redistributed with Strelka (v0.1.14, I believe) and clang. I added a fails_with for clang, but installing all of GNU gcc on OS X seems overkill for a minor compiling issue. I've come across potential solutions that might resolve the samtools compilation error. I'll need help though integrating these workarounds into the formula. 

### Potential Workarounds
* Add `-std=gnu89` flag to clang command during compilation ([source 1](https://github.com/smithlabcode/methpipe/issues/16), [source 2](http://clang.llvm.org/compatibility.html#inline))
* Change inline functions to static ([source](https://github.com/lh3/samtools/issues/15))
* If worse comes to worst, install latest GNU gcc

dpo:
Isn't it possible to have `depends_on "samtools"`? The error on 10.9 is quite mysterious. I'll restart that and the Linux builds.

dpo:
```
gcc -g -Wall -O2  -o samtools bam_tview.o bam_plcmd.o sam_view.o bam_rmdup.o bam_rmdupse.o bam_mate.o bam_stat.o bam_color.o bamtk.o kaln.o bam2bcf.o bam2bcf_indel.o errmod.o sample.o cut_target.o phase.o bam2depth.o -Lbcftools  libbam.a -lbcf  -lm -lz
  "_lseek64", referenced from:
      _knet_seek in libbam.a(knetfile.o)
```

jonchang:
@brunogrande 

OS X doesn't have `lseek64`, please check if Homebrew is building on OS X and use `inreplace` to substitute the call to `lseek64` with plain `lseek`. Culprit appears to be in `strelka/redist/samtools-0.1.18_no_tview/knetfile.c`

iMichka:

PR #2873, opened 2015-10-11
petsc: hypre depends on MPI by default.

dpo:


dpo:
This fixes a side effect by which brew rebuilds `hypre` from source, even though `--with-mpi` is a default option.

dpo:
@xu-cheng any ideas why Travis takes hours to kick in?

xu-cheng:
@dpo We noticed that Travis is running jobs in sequence for all repositories in our organization. I have dropped an email to travis support, but no reply so far.

PR #2875, opened 2015-10-12
Octave: Add optional support for audio operations

aharrison24:
As per this stack overflow question: http://stackoverflow.com/q/32444185/4554650

The octave 4.0.0 formula does not bring in libsndfile and portaudio, which are necessary for reading and writing wav files and for playing audio, amongst other things.

For instance, trying to use the 'wavread' function within octave results in the following error:

    error: sndfile not found on your system and thus audioinfo is not functional
    error: called from
        wavread at line 91 column 8

This PR adds a '--with-audio' option to build with the necessary dependencies, for those who want audio capabilities.

### DIFF
--- a/octave.rb
+++ b/octave.rb
@@ -48,6 +48,7 @@ class Octave < Formula
   option "with-jit",               "Use the experimental JIT support (not recommended)"
   option "with-openblas",          "Use OpenBLAS instead of native LAPACK/BLAS"
   option "with-osmesa",            "Use the OSmesa library (not recommended)"
+  option "with-audio",             "Use the sndfile and portaudio libraries for audio operations"
 
   # build dependencies
   depends_on "gnu-sed"        => :build
@@ -58,6 +59,12 @@ class Octave < Formula
   if build.with? "gui"
     depends_on "qscintilla2"
     depends_on "qt"
+  end
+
+  # Audio dependencies
+  if build.with? "audio"
+    depends_on "libsndfile"
+    depends_on "portaudio"
   end
 
   # dependencies needed for development build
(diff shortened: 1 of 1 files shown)


dpo:
@schoeps Any comments on this?

schoeps:
Cool, thanks! Looks great to me. I will do some tests (I never used audio with octave). However, I am preparing a new revision of the octave formula with some other fixes (e.g. regarding building head and with osmesa, possibly new suite-sparse etc). We could either wait a few days or do this merge quickly....

dpo:
I suggest we go incrementally.

schoeps:
I did some checks. Everything seems to work. However, there might be a problem in the future:
```
WARNING:  140: This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h.
```
@aharrison24: do you get that message, too? do you know anything about how audio works?

aharrison24:
I don't have that message in any of my build logs in
~/Library/Logs/Homebrew/octave. The only warnings I see are related to FLTK
and OSMesa not being found. That's probably because I'm on OSX 10.10
(Yosemite). It looks like the deprecation message came in with 10.11:
https://trac.xiph.org/ticket/2210

I'm afraid I don't know anything about how the audio stuff is implemented
in Octave. However it looks like the issue is likely coming via portaudio:
https://www.assembla.com/spaces/portaudio/tickets/218#/activity/ticket:

schoeps:
Thanks. So it seems that portaudio is guilty. Good for octave :)

aharrison24:
Having said that, there are a number of references to Carbon in the Octave
make files. In octave-4.0.0/etc/README.MacOS there is mention of a
configure switch called '--without-framework-carbon'. I have no idea what
it actually does, but is it potentially something useful to have in the
homebrew formula going forward...? Anyway, I'm probably getting off-topic
for this PR!

fulldecent:
Unfortunately, the Port Audio project is not actively maintained 

https://www.assembla.com/spaces/portaudio/tickets/243-portaudio-support-for-os-x-10-11-el-capitan/details

dpo:
@schoeps What's your verdict here?

schoeps:
This was already covered in my recent (accepted) pull request that you beautified (by the way: thanks). This request one can be closed.

dpo:
Great. Thanks!

PR #2877, opened 2015-10-13
Augustus 3.2 (update)

tseemann:


ntkme:
@tseemann #3104.

dpo:
Current formula installs version is 3.2.1.

PR #2878, opened 2015-10-13
metis4: build proper shared lib.

dpo:


dpo:
@BrewTestBot test this please

PR #2885, opened 2015-10-13
opencascade: fix compilation on El capitan

schoeps:
- corrected path
- reduced optimization level -O1


dpo:
Need to bump the revision number.

dpo:
@BrewTestBot test this please

PR #2889, opened 2015-10-14
add BlasRequirement

davydden:
supersedes  https://github.com/Homebrew/homebrew/pull/44848


blastest_single2.rb
```diff
@@ -0,0 +1,25 @@
+require_relative "blas_requirement"
```
dpo: We should probably place source files that aren't formulae in a separate `src` folder.

blas_requirement.rb
```diff
@@ -0,0 +1,124 @@
...
+  # need BLAS with single precision or complex with Fortran =>
+  # use veclibfort
+  default_formula "veclibfort" if OS.mac?
+  # On Linux we can always fallback to openblas
+  default_formula "openblas"   unless OS.mac?
+
+  def initialize(tags = [])
+    # if we are on OSX and need fortran and veclibfort is installed
+    # by default try it. Otherwise do standard "blas;lapack"
+    if tags.include?(:fortran) && OS.mac? && Formula["veclibfort"].installed?
```
dpo: I would expect that the requirement *install* `veclibfort` since its purpose is to solve certain shortcomings of `libblas` and `liblapack`.

davydden: It certainly will! The trick is to not to fail when veclibfort is **already** installed. Otherwise eqch formula which needs fortran single Blas will report 'blas not configured'.

blastest_single2.rb
davydden: Perhaps 'requierements/BlasRequirement' is better as it mimics Homebrew core?

blas_requirement.rb
```diff
@@ -0,0 +1,124 @@
...
+      tmpdir = Pathname.new tmpdir
+      (tmpdir/"blastest.c").write <<-EOS.undent
+        double cblas_ddot(const int, const double*, const int, const double*, const int);
+        int main() {
+          double x[] = {1.0, 2.0, 3.0}, y[] = {4.0, 5.0, 6.0};
+          cblas_ddot(3, x, 1, y, 1);
+          return 0;
+        }
+      EOS
+      success = system "#{ENV["CC"]} #{cflags} #{tmpdir}/blastest.c -o #{tmpdir}/blastest #{ldflags}",
```
dpo: Shouldn't this be `system ENV.cc, cflags, tmpdir/"blastest.c", "-o", etc.` ?

blas_requirement.rb
```diff
@@ -0,0 +1,124 @@
...
+            b(i) = 3.5*i
+            d(i) = 1.0d0*i
+            e(i) = 3.5d0*i
+          end do
+          if (ABS(ddot(3,d,1,e,1)-sdot(3,a,1,b,1))>1E-10) then
+            call exit(1)
+          endif
+          end program test
+        EOS
+        fortran = which(ENV["FC"] || "gfortran")
```
dpo: Shouldn't we fall back on whatever `:fortran` decided the compiler is?

blas_requirement.rb
davydden: Ps. That's, of course, relevant when the user does not specify 'Homebrew_Blaslapack_xxx'

blas_requirement.rb
davydden: Dont know how to get that info here.

blas_requirement.rb
```diff
@@ -0,0 +1,124 @@
...
+      @default_names = "veclibfort"
+      @default_lib   = "#{Formula["veclibfort"].opt_lib}"
+      @default_inc   = "#{Formula["veclibfort"].opt_include}"
+    else
+      @default_names = "blas;lapack"
+      @default_lib   = ""
+      @default_inc   = ""
+    end
+    super(tags)
+  end
```
dpo: Is it possible to have an `:openblas` tag so a formula could say, e.g.,
```
depends_on BlasRequirement => :openblas
```

blas_requirement.rb
davydden: Will try, last time it didn't work

dpo:
Thanks a lot @davydden, that's a lot of work right there!

blas_requirement.rb
dpo: I see. I'll give this a try asap.

blas_requirement.rb
davydden: In what scenario would you do that? I think the point is to hide the choice from the formula and move it to the global environmen settings.

blas_requirement.rb
davydden: U can adopt my modifications if formulae from other PR, for example take scalapack, which AFAIK needed veclibfort. Then try test first and scalapack second.

blas_requirement.rb
dpo: I was thinking of an alternative default that corresponds to a Homebrew formula. Global environment settings override everything, but
```
depends_on BlasRequirement => :fortran
```
defaults to the `veclibfort` formula, while
```
depends_on BlasRequirement => :openblas
```
defaults to the `openblas` formula, both of which are managed by Homebrew. For example, a formula could say
```
depends_on BlasRequirement => ((build.with? "openblas") ? :openblas : :fortran)
```
which seems like a generic way of covering all cases.

blas_requirement.rb
davydden: this
```
success = system ENV["CC"], cflags, tmpdir/"blastest.c", "-o", tmpdir/"blastest", ldflags,
                :err => "/dev/null"
```
does not work. It runs, but does not seem to do what's needed as `success==1` and it gives me `BLAS not configured`.

blas_requirement.rb
davydden: i think it's against the purpose of the class. We don't have anything like that with `depends_on :mpi`. There is no way to define that we need `open-mpi` or `mpich2`. Whichever is installed and passes internal tests in the requirement (which are far simpler for MPI), will be used. We write `depends_on :mpi => :cxx` to tell that we need C++ compiler only. So it's a tag for **what we need** from the requirement, not for the fall-back strategy. From my point of view those two things are logically different and I would like to keep it so.

I believe same should happen here. The default formula for OS-X is always `veclibfort` only because we know for granted that `-lblas -llapack` are there (framework Accelerate), thus if we don't need Fortran single/complex BLAS, it'll be just fine and tests will pass. The tag `:fortran` does not change the `default`, it specifies that we need fortran single BLAS. It just happens that OS-X blas fails with it and we know the simplest cure -> `veclibfort`. But if one day it works without - the requirement will not try installing veclibfort.

So i would say we add somewhere on `Homebrew-science`  
```
In order to compile Blas/Lapack-dependent formulae with Openblas, the user has to first install openblas (brew install openblas) and then define ENV variables as follows
HOMEBREW_BLASLAPACK_NAMES="openblas"
HOMEBREW_BLASLAPACK_LIB="/path/to/homebrew/opt/openblas/lib"
HOMEBREW_BLASLAPACK_INC="/path/to/homebrew/opt/openblas/inc"
```
that's exactly what was tested [here](https://github.com/Homebrew/linuxbrew/pull/592#issuecomment-147381270).

davydden:
i `git mv blas_requirement ...` to a folder, unfortunately `GitHub` collapses all the comments, as it perceives it a new file now...

requirements/blas_requirement.rb
```diff
@@ -0,0 +1,124 @@
...
+            b(i) = 3.5*i
+            d(i) = 1.0d0*i
+            e(i) = 3.5d0*i
+          end do
+          if (ABS(ddot(3,d,1,e,1)-sdot(3,a,1,b,1))>1E-10) then
+            call exit(1)
+          endif
+          end program test
+        EOS
+        fortran = which(ENV["FC"] || "gfortran")
```
davydden: going back to

> Shouldn't we fall back on whatever :fortran decided the compiler is?

this code is exactly what `:fortran` requirement [does](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/fortran_requirement.rb#L11). I just don't know how do we get the result of that environment? 
`ENV.fortran` did not work for me: 
```
Error: undefined method `fortran' for #<Object:0x007fae998b5250>
```

blas_requirement.rb
davydden: Btw, just realised it is NOT possible to do because in tge parent class 'default_formula' is a static attribute (don't recall the ruby terminology). So it's constant for a class as a whole irrespectively of its instances.

dpo: The situation is similar to that in the Python requirement. We could probably just have a `class OpenBlasRequirement < BlasRequirement` that has `openblas` as `default_formula` and sets up the build flags correctly.

I don't think it's against the purpose of the class. On the contrary. It would be really silly to have to say, e.g.,
```
HOMEBREW_BLASLAPACK_NAMES="openblas" HOMEBREW_BLASLAPACK_INC="/usr/local/opt/openblas/include" brew install whatever
```
when `openblas` is already managed by Homebrew.

davydden: doing 

```
HOMEBREW_BLASLAPACK_NAMES="openblas" HOMEBREW_BLASLAPACK_INC="/usr/local/opt/openblas/include" brew install whatever
```
is certainly not the way it's meant to be used. 

These things should go to one's `.bash_profile` (or equivalent) as it is meant to be a **global setting**. At the very least it should be executed inside terminal once before doing `brew install whatever`.

If we follow your approach, assuming some growth in BLAS support inside Homebrew-Science we will have something like:

```
depends_on "atlas" => :optional
depends_on "gotoblas" => :optional
depends_on BlasRequirement unless build.with? "openblas" || build.with? "atlas" || build.with? "gotoblas"
depends_on OpenBlasRequirement if build.with? "openblas"
depends_on AtlasRequirement if build.with? "atlas"
depends_on GoToBlasRequirement if build.with? "gotoblas"
```

I believe it's **too much clutter** and defeats the point of encapsulating the BLAS inside the requirement. At the end of the day, all those different optional packages provide you the same functionality (C/Fortran functions), so these are just different ways to satisfy **the same** requirement to build a given formula. That's why i firmly believe it should be **one requirement**.

If you are Ok with the current logic inside `depends_on :mpi`, i do not understand why you feel so strongly about doing it differently for `blas-lapack`...

ps. Yes, it would be nice to have an easier way to switch between defaults/fallbacks if needed, but the current object-oriented implementation of Homebrew core class `Requirement`, IMHO, does not make it particularly easy thing to do in an **extensible and clutter-less way**.

davydden: here is a solution which hopefully satisfy what i interpret as your concerns: https://github.com/davydden/homebrew-science/commit/948496771a511fa2e619e48f8baa4aed09404229  
Namely, If `openblas` is installed, we use it by default (unless overridden by ENV variables). So one does not even need to specify ENV variables, but just install openblas first and then do normal `brew install whatever`. If one day `atlas` is available, it's a matter of adding another `if Formula["atlas"].installed?` and changing the defaults.

Here is a confirmation that it does work as expected (i.e. link against veclibfort if there is no openblas, otherwise link against openblas)
```
$ brew install blastest_single2
==> Installing blastest_single2 from davydden/homebrew-science
==> Downloading https://gist.githubusercontent.com/davydden/1f9ebf3692beca2438f8/raw/0ba39c8775db3bb21c09c3440bf78ff5f778277d/blas.f90
Already downloaded: /Library/Caches/Homebrew/blastest_single2-1.0.f90
==> /usr/local/bin/gfortran blas.f90 -I/usr/local/opt/veclibfort/include -o blastest_single -L/usr/local/opt/veclibfort/lib -lveclibfort -pthread -lm
🍺  /usr/local/Cellar/blastest_single2/1.0: 2 files, 24K, built in 2 seconds
MacBook-Pro-Denis:homebrew-science davydden$ brew install openblas
==> Installing openblas from davydden/homebrew-science
==> Downloading https://github.com/xianyi/OpenBLAS/archive/v0.2.14.tar.gz
==> Downloading from https://codeload.github.com/xianyi/OpenBLAS/tar.gz/v0.2.14
==> make FC=/usr/local/bin/gfortran libs netlib shared
==> make FC=/usr/local/bin/gfortran tests
==> make PREFIX=/usr/local/Cellar/openblas/0.2.14_1 install
==> Caveats
This formula is keg-only, which means it was not symlinked into /usr/local.

OS X already provides this software and installing another version in
parallel can cause all kinds of trouble.

Generally there are no consequences of this for you. If you build your
own software and it requires this formula, you'll need to add to your
build variables:

    LDFLAGS:  -L/usr/local/opt/openblas/lib
    CPPFLAGS: -I/usr/local/opt/openblas/include

==> Summary
🍺  /usr/local/Cellar/openblas/0.2.14_1: 14 files, 36M, built in 5.6 minutes
MacBook-Pro-Denis:homebrew-science davydden$ brew uninstall blastest_single2
Uninstalling /usr/local/Cellar/blastest_single2/1.0... (2 files, 24K)
MacBook-Pro-Denis:homebrew-science davydden$ brew install blastest_single2
==> Installing blastest_single2 from davydden/homebrew-science
==> Downloading https://gist.githubusercontent.com/davydden/1f9ebf3692beca2438f8/raw/0ba39c8775db3bb21c09c3440bf78ff5f778277d/blas.f90
Already downloaded: /Library/Caches/Homebrew/blastest_single2-1.0.f90
==> /usr/local/bin/gfortran blas.f90 -I/usr/local/opt/openblas/include -o blastest_single -L/usr/local/opt/openblas/lib -lopenblas -pthread -lm
🍺  /usr/local/Cellar/blastest_single2/1.0: 2 files, 20K, built in 2 seconds
```

requirements/blas_requirement.rb
davydden: i guess one can't query results of other dependencies inside. Indeed: `ohai "ENV[FC] inside requirement: " + ENV["FC"]` gives `Error: no implicit conversion of nil into String`.

Whereas inside the formula `    ohai "ENV[FC]: " + ENV["FC"]` gives `==> ENV[FC]: /usr/local/bin/gfortran`.

davydden:
@dpo @tdsmith MKL is now available for free for OS-X. Even though it's not stated here https://software.intel.com/en-us/articles/free_mkl , after registration one can choose OS-X and then MKL to be downloaded.

dpo:
Not on 10.9 though. I've been in touch with Intel about that but it seems it won't happen.

davydden:
interesting. I guess they don't wont to bother with old versions. Another things i noticed is that their MKL is not complete, i.e. `scalapack` and `blacs` are not there.

PR #2894, opened 2015-10-14
fwdpp 0.3.9

vancleve:


vancleve:
The tests appear to be failing on `fetch homebrew/science/fwdpp` for Mavericks and Yosemite where the fetch tries to get a bottle version and fails.

Any ideas how to fix this?

PR #2897, opened 2015-10-15
Bump graph-tool version to 2.10

count0:
Note that since version 2.9 graph-tool actually requires C++14. But I'm not sure that one should replace the cxx11 homebrew option.

### DIFF
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -1,7 +1,7 @@
 class GraphTool < Formula
   homepage "http://graph-tool.skewed.de/"
-  url "http://downloads.skewed.de/graph-tool/graph-tool-2.9.tar.bz2"
-  sha256 "d3df98adc8a7ea6202e270b62c05825a483fc08c51000721356971d76af1146e"
+  url "http://downloads.skewed.de/graph-tool/graph-tool-2.10.tar.bz2"
+  sha256 "1569a0dbb28f5a09efd86dad058b1729a4c1aaed3ca76b99723298871f5be2d4"
 
   head do
     url "https://github.com/count0/graph-tool.git"
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -50,6 +50,10 @@ class GraphTool < Formula
     depends_on "pygobject3" => with_pythons
   end
 
+  fails_with :clang => '3.5' do
+    cause "seems to have insuficient C++14 support"
+  end
+
   def install
     ENV.cxx11
 


dpo:
Is this what you mean:
```
../../../../src/graph/centrality/../graph_adjacency.hh:313:26: error: 'auto' not allowed in lambda parameter
                      [](auto &es){es.shrink_to_fit();});
                         ^~~~
```

count0:
That is right. This is valid C++14, but not C++11. GCC >= 4.9 and Clang >= 3.4 should support it. Which compiler was used in the build?

dpo:
On 10.9, the bot used clang 6.0 build 600. `configure` says
```
checking whether clang++ supports C++14 features by default... yes
```

count0:
Hm... this is probably the culprit. Could you please force "-std=c++14" in the compilation flags? I will correct the configure check in the next release.

dpo:
That doesn't work for me:
```
checking whether the C compiler works... no
configure: error: in `/private/tmp/graph-tool20151016-19381-1tk5728/graph-tool-2.10/build-python-2.7':
configure: error: C compiler cannot create executables
```
I'm not knowledgeable about C++ standards, though.

count0:
Could you please also post the config.log file? The error is probably just a simple typo.

dpo:
Well the error is
```
configure:3831: clang -std=c++14   conftest.c  >&5
error: invalid value 'c++14' in '-std=c++14'
```
Here's the full config.log: https://gist.github.com/e220e404d3f67285b333

count0:
It seems the clang version is slightly older than 3.5. Please just try using "-std=c++1y" instead.

dpo:
Same error. I have the same config as the build bot though, where `configure` seemed to say that C++14 is supported.

count0:
What you mean same error? It could not have possibly been: "error: invalid value 'c++14' in '-std=c++14'"

dpo:
No, I still get `configure: error: C compiler cannot create executables`. Now `config.log` says
```
configure:3809: checking whether the C compiler works
configure:3831: clang -std=c++1y   conftest.c  >&5
error: invalid argument '-std=c++1y' not allowed with 'C/ObjC'
```

count0:
This means you have passed the option to the C compiler, but you need to pass it only to the C++ compiler (clang++).

dpo:
That's what I thought, so I tried
```
ENV.append "CXXFLAGS", "-std=c++1y", "-stdlib=libc++"
```
but there are more errors:
```
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:603:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/tuple:180:73: error: reference to type 'graph_tool::GraphInterface' requires an initializer
             _NOEXCEPT_(is_nothrow_default_constructible<_Hp>::value) : value()
                                                                        ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/tuple:402:13: note: in instantiation of member function 'std::__1::__tuple_leaf<1, graph_tool::GraphInterface &, false>::__tuple_leaf' requested here
            __tuple_leaf<_Ul, _Tl>()...
            ^
```
etc...

dpo:
It builds with gcc-5 in C++14 mode, but not with clang on 10.9.

dpo:
@xu-cheng @tdsmith Here's a formula that must be built with `-std=c++14`. Is there interest for a `:cxx14` mode in Homebrew? If so, is there more to it [than this](https://github.com/dpo/homebrew/commit/10aceda2bf4d609ee637539ce9e459577cbb6362)?

xu-cheng:
a PR on core is welcome.

count0:
@dpo does it compile with clang 3.7?

dpo:
@count0 I don't know. You have to update your PR and we'll see what the build bot says.

count0:
@dpo can I enable cxx14 instead of cxx11, or is this not yet supported?

dpo:
It's not yet supported, you'll have to add `ENV.append "CXXFLAGS", "-std=c++1y", ...` You can require a minimum version of clang using a `fails_with` block.

PR #2900, opened 2015-10-15
Restore galsim v1.3.0

josePhoenix:
It looks like https://github.com/Homebrew/homebrew-science/commit/02b09555d6f7538b4ce45979a391470dffbc5a2d accidentally downgraded galsim from v1.3.0 to v1.2.0.

I also changed the Python install line such that galsim gets installed in `site-packages`. (i.e. It is immediately importable from a brewed Python after `brew install galsim`.) There may be a better way to do this, but a cursory search for best practices was inconclusive. (Presumably a symlink or `.pth` file would work just as well, but I don't know what's preferred.)

### DIFF
--- a/galsim.rb
+++ b/galsim.rb
@@ -1,8 +1,8 @@
 class Galsim < Formula
   desc "The modular galaxy image simulation toolkit"
   homepage "https://github.com/GalSim-developers/GalSim"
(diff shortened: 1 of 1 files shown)


dpo:
Thanks! The old `gal-sim` must have been updated while the rename PR was in review.

dpo:
And sorry about that!

josePhoenix:
No worries, thanks for the quick attention!

PR #2903, opened 2015-10-16
Update ceres-solver to 1.11 release.

alexsmac:


ceres-solver.rb
```diff
@@ -2,8 +2,8 @@
 
 class CeresSolver < Formula
   homepage 'http://ceres-solver.org/'
-  url 'http://ceres-solver.org/ceres-solver-1.10.0.tar.gz'
-  sha1 '340bf0676ed8b1da02a66ee4595923ba9fc75f1f'
+  url 'http://ceres-solver.org/ceres-solver-1.11.0.tar.gz'
+  sha1 '5e8683bfb410b1ba8b8204eeb0ec1fba009fb2d0'
```
dpo: We now use the `sha256`. You can find it out with `shasum -a 256`. Would you mind updating the formula? It would also be great if you could add a `desc` field with a (brief) description of the software. Thanks!

PR #2907, opened 2015-10-18
opencv3: add an option to build with samples

mstfldmr:


mstfldmr:
add an option to build with samples

dpo:
Very sorry for the long wait. Could you add a revision number to the formula?

dpo:
You need to add an option `with-samples` for this to work.

heinemml:
I took a look at the `-D BUILD_EXAMPLES=ON` option. And I have a problem where I'm not sure how to handle.
The examples build fine (some workarounds are needed depending on the other options). But the install routine is only defined for windows. So the examples get thrown away as they are only in the temporary build/bin directory. 

Question is: Does it make sense to add an installer step to the formula? Or is this out of scope for homebrew and should be added upstream?

@dpo any opinion on that?

Independent from that I would take a look at adding `-DINSTALL_C_EXAMPLES=ON` and `-DINSTALL_PYTHON_EXAMPLES=ON` which adds the example sources to the `share` folder.

dpo:
@heinemml It makes sense to me to install the source files for the examples, not so much the compiled binaries. If you'd like, go ahead and open a new PR adding those flags, and installing the samples if that's useful. In your commit message, you can add `Closes #2907` to indicate that your PR supersedes this one.

PR #2910, opened 2015-10-19
qr_mumps: cleaner shared lib and test.

dpo:


qr_mumps.rb
```diff
@@ -100,4 +99,65 @@ def caveats; <<-EOS.undent
...
+        call qrm_adealloc(b)
+        call qrm_adealloc(r)
+        call qrm_adealloc(x)
+        call qrm_spmat_destroy(qrm_mat, all=.true.)
+      end program _qrm_test
+    EOS
+
+    %w[s d c z].each do |p|
+      system "perl #{pkgshare}/examples/#{p}.pl < test.f90 > #{p}test.f90"
+      system "gfortran", "#{p}test.f90", "-o", "#{p}test", "-I#{opt_libexec}/modules", "-L#{opt_lib}", "-lqrm_common", "-l#{p}qrm", "-lgomp"
```
dpo: @xu-cheng How can I access `ENV["FC"]` in the `test` method? For now, it seems to be empty. Thanks!

xu-cheng: Shouldn't you call `ENV.fortran` first?

dpo: Ah, thanks. That's how!

PR #2917, opened 2015-10-19
hypre: explicit version.

### LINKED ISSUE #2916: version extraction fails for Hypre
_From @davydden on October 19, 2015 21:43_

subj.

Here's what Homebrew does
```
==> Downloading http://ftp.mcs.anl.gov/pub/petsc/externalpackages/hypre-2.10.0b-p2.tar.gz
Already downloaded: /Library/Caches/Homebrew/hypre-2.tar.gz
==> ./configure --prefix=/usr/local/Cellar/hypre/2
```
as you see it's `2` instead of `2.10.0b`.

_Copied from original issue: Homebrew/homebrew#45161_

dpo:
Fixes #2916.

### DIFF
--- a/hypre.rb
+++ b/hypre.rb
@@ -4,6 +4,7 @@ class Hypre < Formula
   url "http://ftp.mcs.anl.gov/pub/petsc/externalpackages/hypre-2.10.0b-p2.tar.gz"
   mirror "ftp://ftp.mirrorservice.org/sites/distfiles.gentoo.org/distfiles/hypre-2.10.0b-p2.tar.gz"
   sha256 "76414e693e5381e352e759c851c8cd5969dba7001b1dc153fe0f1ff60a5bb168"
+  version "2.10.0b-p2"
 
   bottle do
     cellar :any_skip_relocation


dpo:
Should we make `mpi` `:recommended` in `sundials`? Currently, it needs to be built from source every time you install `petsc`.

davydden:
yes, i would make it `:recommended`.

davydden:
can you check that installing `dealii` does not invoke re-install `hypre`? (a case of dependency of a dependency).

dpo:
See #2918.

dpo:
I brewed `dealii` and it didn't reinstall `hypre`.

dpo:
The problem occurs in `trilinos`. If I change [this line](https://github.com/Homebrew/homebrew-science/blob/master/trilinos.rb#L61) to just
```
  depends_on "hypre"        => [:recommended]
```
(which is conceptually the same by default), then `hypre` is not rebuilt.

davydden:
regarding `trilinos`, why would `hypre` should be rebuild? The line you quoted was outdated as hypre is built with mpi by default, so i think behaves ok from that point of view.

What bothers me is that in the case of a chain of dependencies (dealii->petsc->hypre) hypre got reinstalled again...

dpo:
Try it. Fetch this PR locally and build `hypre` from source. Then try to install `trilinos` and you'll see `hypre` listed as a missing dependency. Interrupt it, change the line in `trilinos.rb` as I indicated, and try to build `trilinos`. Now `hypre` isn't a missing dependency any more.

hypre.rb
```diff
@@ -4,6 +4,7 @@ class Hypre < Formula
   url "http://ftp.mcs.anl.gov/pub/petsc/externalpackages/hypre-2.10.0b-p2.tar.gz"
   mirror "ftp://ftp.mirrorservice.org/sites/distfiles.gentoo.org/distfiles/hypre-2.10.0b-p2.tar.gz"
```
davydden: i think in order to avoid confusing Homebrew-core regex scripts to extract version we should rather download normall Hypre and add 2 patches from https://bitbucket.org/petsc/pkg-hypre/commits/all .

davydden:
> Try it. Fetch this PR locally and build hypre from source. Then try to install trilinos and you'll see hypre listed as a missing dependency. Interrupt it, change the line in trilinos.rb as I indicated, and try to build trilinos. Now hypre isn't a missing dependency any more.

The following line in trilinos fixes re-build for me:
```
  depends_on "hypre"        => [:recommended] + ((build.with? "mpi") ? [] : ["without-mpi"]) + openblasdep
```
no idea why...

davydden:
same change should happen to deal.II. 

I suppose at some point mpi became default in hypre but was not updated in other formulae.

hypre.rb
davydden: nevermind, i think it should work as it is given the modifications to trilinos and dealii i mentioned.

davydden:
and to petsc.

dpo:
`hypre` as `:mpi` marked as `:recommended`, so that can't be the reason.

davydden:
I think you missed the point - hypre does not have an option '--with-mpi', it has '--without-mpi'. 

Say i have formula1 installed WITHOUT any options. If formula2 wants formula1 --with-option1, it will be re-installed with this option. If formula1 have --with-option1, brew info will tell that it's now installed with this option. Otherwise it wont and therefore incorrect option in formula2 will trigger its reinstall irrespectivly.

You see my point?

dpo:
Yes I see, thank you. I think you're right.

PR #2923, opened 2015-10-22
cantera: sundials builds with mpi by default.

dpo:


ilovezfs:
@dpo why is important for this to have sundials without mpi?

dpo:
I honestly can't recall.

PR #2928, opened 2015-10-23
wcstools 3.9.2

larrybradley:


wcstools.rb
```diff
@@ -2,8 +2,8 @@
 
 class Wcstools < Formula
   homepage "http://tdc-www.harvard.edu/wcstools/"
-  url "http://tdc-www.harvard.edu/software/wcstools/wcstools-3.9.1.tar.gz"
-  sha1 "05d97277fdfcabd40f926b0b7ac66b0f434497f0"
+  url "http://tdc-www.harvard.edu/software/wcstools/wcstools-3.9.2.tar.gz"
+  sha1 "3d86e2480443717d42bb8129be9bcc339ef5ff4e"
```
dpo: Please use the `sha256`. It would also be great if you could add a `desc` field with a brief description of the software. Thanks!

larrybradley: Done!

PR #2929, opened 2015-10-23
travis: Use travis_retry

saketkc:
The failure on `travis` is often random:

1. https://travis-ci.org/saketkc/homebrew-science/builds/87135231
2. https://travis-ci.org/Homebrew/homebrew-science/jobs/87135238

This should possibly avoid such random failures.
Feel free to close, if there is a better workaround


saketkc:
@xu-cheng

xu-cheng:
Sorry, we definitely don't want this which will otherwise use lots of resources on actual failure.

Also the build link you give is failed on sandbox violation.

PR #2931, opened 2015-10-24
P4est 1.1. (bump revision)

davydden:


DomT4:
Aborted this to resolve the side-effects of an earlier CI crash. It's already back in the build queue now, so just a heads up on the abort reason @Homebrew/science.

PR #2924, opened 2015-10-23
Petsc 3.6.3 (update)

davydden:

PR #2939, opened 2015-10-27
slepc: add test do block

davydden:


davydden:
it would be nice if we could first finish with https://github.com/Homebrew/homebrew-science/pull/2924

davydden:
@BrewTestBot test this please

PR #2951, opened 2015-10-30
Added the KiCad EDA Suite.

metacollin:
Hi, this is a formula for the KiCad Electronic Design Automation suite.  It does build a .app, however, there is no stable version of the .app available or otherwise easily obtainable from the official site or anywhere else.  There are only nightly snapshots, but these are missing critical features, namely python scriptability, and also require manual installation of library support files in a somewhat invasive way (copying into the system-wide /Library directory).  There are also several build options that are customizable and very handy (like menu icons) that, for whatever reason, are turned off but only in the OS X version.  

This formula builds the HEAD and stable version without issue, patches one file using inreplace to 'macify' the paths, and stores it in /usr/local/etc as these files ought to persist if there is a reinstall later.  It also means that the homebrewed KiCad will play nicely if there is previous data files or a separate install of the crippled nightly of KiCad already on the users system.  

You may notice a few things that might initially be worrisome, so I'll explain.  Yes, this formula has wx and wxPython as resources, and builds them specially for KiCad.  This is because wx and wxPython have a number of bugs, some common to all platforms, and some extra ones just on OS X, that the KiCad developers uncovered while working on KiCad.  They have submitted bug reports and patches to fix all of them, and the wx team has simply ignored these entirely, saying they aren't going to include them because they are not planning on doing another release of wx past 3.0.2, and instead are focusing on wxPhoenix.  So the bugs are there and will stay there. 

Unfortunately, KiCad can't actually function unless wx actually works correctly, instead of working mostly correctly.  So, and this occurs on every platform, and is part of their actual build scripts, KiCad has to pull in its own copy of the wx source, patch the bugs, and build it just for itself.  In my formula, wx and the necessary wxpython components are staged and linked/bundled into the .app bundle, and only live there, so there are no possible issues that could occur.  It's mostly just a lengthy formula to build.  

There are many electrical engineers who run OS X (myself included) but run KiCad in a virtual machine, usually windows, simply because the OS X support is lackluster and complete lack of stable binary availability. This is by no fault of the developers, their OS X contributors are spread thin and far apart.  They've done a terrific job of the OS X version in spite of this and I can vouch for this, I use it all the tie to produce commercial circuit boards and it runs great.  IF you build it yourself.  And doing that is also very difficult on OS X, and the demand for a good way to get this program is there, but right now, there is no good solution.  

I've maintained a tap for months now, but I finally refactored it into something very nice I think, and I feel is worthy for inclusion in homebrew-science.  

I know .apps are frowned upon, but I think this is the exact sort of circumstance where it would be worth making an exception.  But that's what I think :).

Please consider accepting this formula, but I know whatever the decision is, it is out of good reasons (which I may or may not know as I write this), and I support whatever that decision is - inclusion or rejection.  

Thanks for taking the time to read this lengthy description!


metacollin:
Wow, uh, looks like the build exceeded some time limit on the build bot.  I'm sure it was almost done, I promise! :blush:

metacollin:
Hey, so this formula works.... it's only failing because the build takes too long on travis-ci.

PR #2953, opened 2015-10-30
Telseq 0.0.1 (new formula)

danielecook:


telseq.rb
```diff
@@ -0,0 +1,28 @@
+class Telseq < Formula
+  desc "Program for estimating telomere length from sequence data"
+  homepage "https://github.com/zd1/telseq"
+  # doi "10.1093/nar/gku181"
+  # tag "bioinformatics"
+
+  url "https://github.com/zd1/telseq/archive/0.0.1.tar.gz"
+  sha1 "b502b6695af2ac87c6ad28e77113866b63991441"
```
dpo: Please use the `sha256`.

telseq.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  head "https://github.com/zd1/telseq.git"
+
+  depends_on "gcc"
```
dpo: There's no need to depend on `gcc` but if this doesn't build with `clang`, then there should be a `fails_with` block.

jonchang:
ping @danielecook

iMichka:

PR #2954, opened 2015-10-30
FastFlow 2.0.4 (new formula)

fran6co:


tseemann:
@fran6co - is this a Science specific piece of software? If not, it should be in regular Brew.

fran6co:
I wasn't sure about it, it's a parallel programming framework being used in several scientific projects but it's generic enough to go to standard homebrew I guess.

PR #2955, opened 2015-10-30
gdcm: add openssl option

kkumler:
Building with openssl enabled allows for gdcmanon to operate for anonymization.

### DIFF
--- a/gdcm.rb
+++ b/gdcm.rb
@@ -1,8 +1,9 @@
 class Gdcm < Formula
-  desc "C++ library for DICOM medical files"
+  desc "Grassroots DICOM library and utilities for medical files"
   homepage "http://sourceforge.net/projects/gdcm/"
   url "https://downloads.sourceforge.net/project/gdcm/gdcm%202.x/GDCM%202.4.4/gdcm-2.4.4.tar.bz2"
   sha256 "c5cd2f43a16180f5a9ecd5211bf214971b0620e9d9e027c2e11a89c4ce7d5b1f"
+  revision 1
 
   bottle do
     revision 1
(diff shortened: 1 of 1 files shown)


tseemann:
@kkumler Would you also be able to add a new `desc "__________"` tag at the top also, to help modernize it like you did with the `sha256` ? 
Then squash into a single commit?
Thanks!

kkumler:
@tseemann No problem. Corrected a couple little existing things, found from audit, as well.

gdcm.rb
```diff
@@ -14,15 +13,17 @@ class Gdcm < Formula
   option "with-check", "Run the GDCM test suite"
   depends_on :python => :optional
   depends_on :python3 => :optional
+  option "with-openssl", "Build with openssl support, to enable gdcmanon encryption"
+  depends_on "openssl" => :optional
```
dpo: Is this redundancy necessary? It's enough to specify line 17.

kkumler: Only included to specify what portion of the package actually uses openssl. Is there a preference for specifying those things?

dpo: It's fine if you think the custom help message if useful. But in that case, line 17 should be modified as I indicated above.

dpo:
Sorry to make you wait. Could you please add a revision number to the formula?

kkumler:
@dpo Done.

dpo:
Great! Thanks.

dpo:
Oops, there's been an update of `gdcm.rb` in the interm. Would you mind rebasing your branch against `master`? There will be a conflict or two to resolve. Thanks!

kkumler:
@dpo Yeah, done.

dpo:
Thanks!

PR #2962, opened 2015-11-02
mallet: change to sha256, add head and devel

loganrosen:


tseemann:
@loganrosen Could you also add the new `desc` tag (and `# doi` and '# tag` if possible) then squash into a single commit? Thanks!

loganrosen:
Done! I also added a simple test and did some other style fixes. I couldn't find a DOI that corresponded to the software, though.

tseemann:
@loganrosen thanks! i'm having some issues with the build system so I can't accept it just yet, sorry

PR #2963, opened 2015-11-02
p4est: build debug and release libraries

davydden:
that's handy for deal.II as it can pick up debug/release versions and build itself against the two in debug/release modes.

### DIFF
--- a/p4est.rb
+++ b/p4est.rb
@@ -3,7 +3,7 @@ class P4est < Formula
   homepage "http://www.p4est.org"
   url "http://p4est.github.io/release/p4est-1.1.tar.gz"
   sha256 "0b5327a35f0c869bf920b8cab5f20caa4eb55692eaaf1f451d5de30285b25139"
-  revision 1
+  revision 2
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


dpo:
Strange. It fails everywhere with:
```
configure: error: Fortran 77 compiler cannot create executables
```

dpo:
@davydden Is Java really required to build `open-mpi`? https://travis-ci.org/Homebrew/homebrew-science/jobs/88838235#L245

Your PR builds locally for me. I'm not sure why the bot fails.

davydden:
i don't know why but they declare it as a requirement, see this thread: https://github.com/Homebrew/linuxbrew/issues/365
It has to do with `vtsetup.jar`.

dpo:
Thanks. I didn't remember that. Does it also apply to OSX though?

davydden:
sure, the updated link is in the `Homebrew-core`.

davydden:
@jppelteret we had some problems with it on linux, right? I guess i will close this PR for now...

jppelteret:
@davydden Yes, as far as I can recall there were linking issues on Linux. I think that it was due to the libraries not being symlinked to `lib`, but rather being left in `opt` in their respective `DEBUG` and `FAST` directories.

PR #2964, opened 2015-11-03
Update galsim.rb

kremerj:
the caveat message did not include the version number, executing the proposed command failed to add the homebrew python packages to the default path.


tseemann:
@kremerj The `caveat` seems to apply only to OS X - can you supplement with instructions for Linux users too?

PR #2966, opened 2015-11-03
fwdpp 0.4.1

vancleve:


tseemann:
@vancleve Can you please squash into 1 commit, and also remove "fdwp is a" from the `desc` tag?

vancleve:
Fixed in new pull request #2972

PR #2983, opened 2015-11-06
fwdpp 0.4.2

vancleve:


dpo:
That's a fast release cycle!

vancleve:
ha, yes. I won't try to do every point update but since this one was a but fix, it seemed appropriate.

tseemann:
thank you @vancleve

PR #2988, opened 2015-11-07
miniasm 0.2 (new formula)

tseemann:
Miniasm is a very fast OLC-based de novo assembler for noisy long reads. It takes all-vs-all read self-mappings (typically by minimap) as input and outputs an assembly graph in the GFA format. Different from mainstream assemblers, miniasm does not have a consensus step. It simply concatenates pieces of read sequences to generate the final unitig sequences. Thus the per-base error rate is similar to the raw input reads.

So far miniasm is in very early development stage. It has only been tested on twelve bacterial genomes sequenced with PacBio. Including the mapping step, it takes about 3 minutes to assemble a bacterial genome. Under the default setting, miniasm assembles 5 out of the 12 datasets into a single contig. The 12 data sets are PacBio E. coli sample, ERS473430, ERS544009, ERS554120, ERS605484, ERS617393, ERS646601, ERS659581, ERS670327, ERS685285, ERS743109 and a deprecated PacBio E. coli data set.

Miniasm confirms that at least for high-coverage bacterial genomes, it is possible to generate long contigs from raw PacBio reads without error correction. It also shows that minimap can be used as a read overlapper, even though it is probably not as sensitive as the more sophisticated overlapers such as MHAP and DALIGNER. Coupled with long-read error correctors and consensus tools, miniasm may also be useful to produce high-quality assemblies.


tseemann:
@dpo Anyway we can bypass this for software by @lh3 ?

```
==> audit problems
homebrew/science/miniasm:
 * GitHub repository not notable enough (<10 forks, <10 watchers and <20 stars)
 * GitHub repository too new (<30 days old)
```

tseemann:
```
==> audit problems
homebrew/science/miniasm:
 * GitHub repository too new (<30 days old)
```

Let's wait then.... *sigh*

dpo:
There's an error due to https://github.com/Homebrew/homebrew-science/pull/2988/files#diff-97e39c8f2d7047cb72b7329ea73f3406R16.

sjackman:
@tseemann Would you like to fix up this small issue?

https://travis-ci.org/Homebrew/homebrew-science/jobs/90182938#L1513
```
==> make
.
Error: No such file or directory - misc/demo.sh
==> FAILED
```

PR #2989, opened 2015-11-07
KiCad (new formula)

metacollin:
Pull request to add this formula for the Kicad electronic design automation suite.  It's having some trouble timing out on travis-ci due to the heavy-weight nature of this formula.

There is no available source for the stable version on OS X, and the nightlies do not have python integration.  This formula makes both accessible to OS X users.  Also, it does indeed download and build wx and wxPython as a resource.  This is normal, and is in fact part of the actual KiCad build procedure.  They (the KiCad dev team) uncovered several bugs in wx and wxpython and have submitted patches upstream, but the wx team is not interested in doing another bug fix release, and has turned all their attention towards wxPhoenix.  These bugs prevent KiCad from functioning, so they have no choice but to apply these bug fixes to wx themselves, and build a version of wx that works correctly that is specific to KiCad.  

The patch it downloads is simply the patches included within the KiCad source itself, but concatenated into one file for convenience.  The only patches in my formula that are my own and specific to homebrew are done via inreplace.  Otherwise, all oddities are actually part of the correct build procedure for this program, and not my own modifications.

### DIFF
--- a/kicad.rb
+++ b/kicad.rb
@@ -29,6 +29,7 @@ class Kicad < Formula
   depends_on "python" => :optional
   depends_on "swig" => :build if build.with? "python"
   depends_on "xz"
+  depends_on "glm"
 
   fails_with :gcc
   fails_with :llvm
@@ -52,7 +53,7 @@ class Kicad < Formula
   end
 
   def install
-    ENV["MAC_OS_X_VERSION_MIN_REQUIRED"] = "#{MacOS.version}"
+    ENV["MAC_OS_X_VERSION_MIN_REQUIRED"] = MacOS.version.to_s
     ENV.append "ARCHFLAGS", "-Wunused-command-line-argument-hard-error-in-future"
     ENV.append "LDFLAGS", "-headerpad_max_install_names"
     if MacOS.version < :mavericks
@@ -68,7 +69,7 @@ class Kicad < Formula
     end
 
     resource("wxk").stage do
-      (Pathname.pwd).install resource("wxpatch")
+      Pathname.pwd.install resource("wxpatch")
       safe_system "/usr/bin/patch", "-g", "0", "-f", "-d", Pathname.pwd, "-p1", "-i", "wxp.patch"
 
       mkdir "wx-build" do
@@ -93,7 +94,7 @@ class Kicad < Formula
         ]
 
         system "../configure", *args
-        system "make", "-j#{ENV.make_jobs}"
+        system "make", "-j4"
         system "make", "install"
       end
 
@@ -153,7 +154,7 @@ class Kicad < Formula
       end
 
       system "cmake", "../", *(std_cmake_args + args)
-      system "make", "-j#{ENV.make_jobs}"
+      system "make", "-j4"
       system "make", "install"
     end
   end
@@ -172,9 +173,7 @@ class Kicad < Formula
   end
 
   def caveats
-    s = ""
-    if build.with? "default-paths"
-      s += <<-EOS.undent
+    if build.with? "default-paths" then <<-EOS.undent
 
       KiCad component libraries and preferences are located in:
         #{kicaddir}
@@ -184,8 +183,7 @@ class Kicad < Formula
       within Pcbnew.  It will automatically guide you
       through this process upon first lauch.
       EOS
-    else
-      s += <<-EOS.undent
+    else <<-EOS.undent
 
       KiCad component libraries must be installed manually in:
         /Library/Application Support/kicad
@@ -195,8 +193,6 @@ class Kicad < Formula
           /Library/Application\ Support/kicad
       EOS
     end
-
-    s
   end
 
   test do
--- a/kicad.rb
+++ b/kicad.rb
@@ -1,0 +1,205 @@
+class Kicad < Formula
+  desc "Electronic Design Automation Suite"
+  homepage "http://www.kicad-pcb.org"
+  url "https://launchpad.net/kicad/4.0/4.0.0-rc1/+download/kicad-4.0.0-rc1.tar.xz"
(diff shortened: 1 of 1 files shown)


tseemann:
Can you squash your commits into a single commit with `git rebase -i HEAD~2` for example?

metacollin:
Yes, sorry, I forgot to do this.  I'm not sure what to do about the travis-ci problem, it would appear that the 50 minute build cutoff is a hard limit, and it only builds quickly enough if there are 4 threads (`make -j4`).

metacollin:
OK, all squashed.  Is there anything I should do in regards to travis-ci? Or is it ok if only Brew Test Bot passes in this case? I dunno.

metacollin:
Hi, please don't forget about this.  It works fine, only issue is the 50 minute cutoff on travis-ci.  Commits squashed, it's ready to go :).

kicad.rb
```diff
@@ -0,0 +1,205 @@
...
+  # Patches are the ones that come from the stable release archive of KiCad under the patches directory.
+  resource "wxpatch" do
+    url "https://gist.githubusercontent.com/metacollin/2d5760743df73c939d53/raw/b25008a92c8f518df582ad88d266dcf2d75f9d12/wxp.patch"
+    sha256 "0a19c475ded29186683a9e7f7d9316e4cbea4db7b342f599cee0e116fa019f3e"
+  end
+
+  resource "wxk" do
+    url "https://downloads.sourceforge.net/project/wxpython/wxPython/3.0.2.0/wxPython-src-3.0.2.0.tar.bz2"
+    sha256 "d54129e5fbea4fb8091c87b2980760b72c22a386cb3b9dd2eebc928ef5e8df61"
+  end
```
dpo: There's already a formula for wxPython in Homebrew Core. Wouldn't it make sense to patch that one instead of rebuilding here? Bottles would also save a lot of build time.

metacollin: The wx and wxpython formulas cannot be used.  They absolutely must be patched and built specifically for KiCad.  And you're absolutely right, it's a fairly frustrating situation, but this is not addressable by homebrew and outside the scope of its purpose.  See: http://kicad-pcb.org/contribute/build-osx/

Specifically, this excerpt:

"At the moment you must not use a wxWidgets/wxPython version installed by any package manager.
KiCad on OS X needs overlay support and some other fixes, which are not yet contained in mainline wxWidgets/wxPython sources and builds.
If you don’t use a patched wxWidgets you will get graphical artifacts (something like a magnifying glass effect in the upper left corner of KiCad windows) or KiCad will just crash on start (due to duplicate libraries)."

As for using bottles, I have had a tap for KiCad for a while now (but I've only submitted it now because it had no stable release up until recently).   I actually did exactly what you suggested, which was to include 2 additional formulae, keg-only, which built KiCad specific versions of wx and wxpython.  After that, they were simply reused and it massively shortened build time.  For a while I even provided my own manual bottles.  

I removed this pretty much entirely because I thought homebrew-science would not accept such a setup, as duplicate keg-only formula that are specific to just one other formula would clutter stuff up.  I had named my formulas 'wxkicad' and 'wxkython' (not really ideal, but they were meant to be hidden from the user anyway). 

I would actually love to return this formula to the quicker method.  I wouldn't want to have a formula for each custom version of wx however.  I would ask for some guidance on what you think the acceptable way to handle this is.  I maintained bottles of both wx and wxpython (the kicad-specific patched versions I mean) for 10.6-10.10 at one point, so that solution is well tested and works well.  But these were bottles hosted on my own web space.  Is there a way to 'bottle' a resource and have it built by brew-bot without creating an entire separate formula?

Basically, we're 100% on the same page here, I am just not sure what would be the preferred means of accomplishing this.  But, as a KiCad user and minor contributor, I've been rebuilding dev versions almost every day, and reused the patched wx and wxpython versions every time, and it most definitely is a huge time saver.  

I merged it all into one formula mostly because I thought I needed to for a chance of acceptance.  But the alternative way is definitely already written and sitting on a tap right now :).

kicad.rb
```diff
@@ -0,0 +1,205 @@
...
+
+  resource "kicad-library" do
+    url "https://github.com/KiCad/kicad-library.git"
```
dpo: Could we use a release here?

metacollin: I'm not sure what you're asking.  

1.  The patch touches many wx source files, particularly OS X specific ones.  While these are entirely bug fixes, they are bug fixes that have not been accepted upstream, and won't be.  So patching the wx formula would still require rebuilding it.  

2.  That said, it can be bottled.  I originally had this formula split into two formula, one for the special version of wx, and one for kicad.  This way, wx only had to be built once, and could be bottled.  

3.  Is there a way to have a sub formula like that so it won't pollute the main repo with a keg-only formula whose entire existence is just to support a single different formula?  If so, I would love to do that.  I don't think that is possible though.

4.  What do you mean a release?  The latest wxPython/wxwidgets release is 3.0.2.0, and they have ceased development of both of those and are no longer accepting any upstream patches.  They are devoting all future effort into wxPhoenix, which is intended to one day replace both wxwidgets and wxPython entirely.

To be clear, **patching source files requires rebuilding wx and wxPython, and we must patch it ourselves because they are not accepting any more bug fixes upstream.  KiCad does not work with the unpatched version of wx due to several bugs that will not be fixed up stream.**

dpo: I meant a release of `kicad-library`, as opposed to the git head.

iMichka:
@metacollin Progress on this has been stalled since more than a year now. I see that there is a new version of kicad. Would you be interest on pursuing the effort to package this with the newest version, or can we abandon this?

iMichka:
Closing this as nobody has been working on this. A new pull request can be submitted if somebody wants to pick this up.

adzenith:
For anyone coming here in the future, you can install Kicad with `brew cask install kicad`.

PR #2992, opened 2015-11-08
parmetis: fix linking with gcc 4.8

davydden:
courtesy of @jppelteret


davydden:
something is broken with travis, `openmpi` does not see `fortran` compiler.

dpo:
Sorry for the delay. The formula needs a new revision.

dpo:
I had the same error as on Travis with `open-mpi` earlier today. It was resolved by building from source. Perhaps it simply needs to be rebottled.

PR #2994, opened 2015-11-08
Add mrtrix formula

brainstorm:
How can I generalize this:

lib.install Dir["lib/libmrtrix.dylib"]

So that it works on linux(brew) too (via `.so` files instead?).


mrtrix.rb
```diff
@@ -0,0 +1,25 @@
+require 'formula'
```
dpo: This is outdated. Formulas no longer require this.

mrtrix.rb
```diff
@@ -0,0 +1,25 @@
...
+
+class Mrtrix < Formula
+  homepage 'https://github.com/MRtrix3/mrtrix3/wiki'
+  url 'https://github.com/MRtrix3/mrtrix3/archive/0.3.12.tar.gz'
+  sha1 'de84ed0a7b73526d4077398ada100f9da389768d'
```
dpo: We now use the `sha256`.

mrtrix.rb
```diff
@@ -0,0 +1,25 @@
+require 'formula'
+
+class Mrtrix < Formula
+  homepage 'https://github.com/MRtrix3/mrtrix3/wiki'
```
dpo: Please add a `desc` field with a brief description of the software.

mrtrix.rb
```diff
@@ -0,0 +1,25 @@
...
+  url 'https://github.com/MRtrix3/mrtrix3/archive/0.3.12.tar.gz'
+  sha1 'de84ed0a7b73526d4077398ada100f9da389768d'
+  head 'https://github.com/MRtrix3/mrtrix3'
+  revision 1
```
dpo: A new formula shouldn't have a revision.

mrtrix.rb
```diff
@@ -0,0 +1,25 @@
...
+  def install
+    system "./configure"
+    system "./build"
+
+    bin.install Dir["bin/*"]
+    lib.install Dir["lib/libmrtrix.dylib"]
+  end
+
+  def test
+    system "git clone https://github.com/MRtrix3/testing.git test"
```
dpo: You'll want to do this in `testpath` (grep formulas for `testpath`).

dpo:
You can do something like
```
so = (OS.mac?) ? "dylib" : "so"
lib.install "libsomething.#{so}"
```

mrtrix.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  depends_on "gsl"
+  depends_on "qt5"
+
+  def install
+    system "./configure"
+    system "./build"
+
+    bin.install Dir["bin/*"]
+    lib.install Dir["lib/libmrtrix.dylib"]
```
dpo: There's no need to use `Dir` if you're not globbing.

brainstorm:
Fixed your comments and the ones reported by travis... not sure how to test this package on the best way though since they have a separate repo for testing:

https://github.com/MRtrix3/testing

brainstorm:
Apparently the testing repo requires a symlink to `build` (where mrtrix was built):

```
bash-3.2$ cat testing.log
-------------------------------------------
  Testing MRtrix3 installation
-------------------------------------------

  testing folder: /private/tmp/mrtrix20151109-85105-1lck3in
  MRtrix3 folder: /private/tmp

-------------------------------------------
```

Is there a way to do that in Homebrew post-install?

@MRtrix3, see the "problem" of not integrating "tests" inside the main repo via a subrepo, for instance?

PR #2998, opened 2015-11-10
rstudio-server: kill rserver when script exit

randy3k:
`rserver` process should be killed when script exit.


rstudio-server.rb
```diff
@@ -1,5 +1,6 @@
 class RstudioServer < Formula
   homepage "http://www.rstudio.com"
+  head "https://github.com/rstudio/rstudio.git"
   url "https://github.com/rstudio/rstudio/archive/v0.99.484.tar.gz"
   sha256 "8ca4abccb9b554713077cf1057ac13abadfd7606f22ac3386b2a88a38ae8a427"
```
dpo: Please add a revision number.

randy3k: why do us fix to a revision?

dpo: A revision number is necessary so the bottles can be updated properly. Please add `revision 1` to the header.

randy3k: Updated.

I thought you wanted me to do something like
```
head "https://github.com/rstudio/rstudio.git", :revison => xxxxxx
```

dpo:
Thank you!

PR #3002, opened 2015-11-11
abinit: link against case-sensitive veclibfort.

dpo:


dpo:
@DomT4 This formula depends on MPI indirectly, via `scalapack` and `fftw --with-mpi`. In the logs, we see that `open-mpi` was fetched, but it's not listed as a dependency, and not installed: https://travis-ci.org/Homebrew/homebrew-science/jobs/90459942#L371. The build fails in the same way on both Travis and the brew bot. Do you have any idea why MPI isn't being installed?

DomT4:
No idea, unless `open-mpi` isn't compatible here suddenly for some reason. We've had some weirdness with special dependencies lately, particularly Fortran, so perhaps this is related in some way.

CC @xu-cheng @mikemcquaid @tdsmith in case they have any ideas.

dpo:
Thanks. I get the same behavior locally.

dpo:
@BrewTestBot test this please

PR #3003, opened 2015-11-11
(New Formula) XCDF

metacollin:
This is a very powerful bitpacking library and program with python bindings.  It can compress datasets with excellent compression ratios, superior to general purpose compression algorithms for certain data types (like floating point numbers).  It can also generate histograms from the sets within or without python.  I know they are using it on some of the data the http://hawc-observatory.org/


dpo:
@BrewTestBot test this please

PR #3015, opened 2015-11-13
Add mrtrix formula

brainstorm:


tseemann:
```
==> Downloading https://github.com/MRtrix3/mrtrix3/archive/0.3.12.tar.gz
Already downloaded: /Library/Caches/Homebrew/mrtrix-0.3.12.tar.gz
==> Verifying mrtrix-0.3.12.tar.gz checksum
tar xf /Library/Caches/Homebrew/mrtrix-0.3.12.tar.gz
==> ./configure
.
==> ./build
...
==> git clone https://github.com/MRtrix3/testing
.
Error: wrong number of arguments (1 for 2..3)
Please report this bug:
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/fileutils.rb:343:in `ln_s'
/usr/local/Library/Taps/homebrew/homebrew-science/mrtrix.rb:17:in `install'
/usr/local/Library/Homebrew/build.rb:129:in `block in install'
/usr/local/Library/Homebrew/formula.rb:906:in `block in brew'
/usr/local/Library/Homebrew/formula.rb:1478:in `block in stage'
/usr/local/Library/Homebrew/resource.rb:91:in `block in unpack'
/usr/local/Library/Homebrew/resource.rb:88:in `unpack'
/usr/local/Library/Homebrew/resource.rb:81:in `stage'
/usr/local/Library/Homebrew/formula.rb:1469:in `stage'
/usr/local/Library/Homebrew/formula.rb:902:in `brew'
/usr/local/Library/Homebrew/build.rb:107:in `install'
/usr/local/Library/Homebrew/build.rb:176:in `<main>'
```

mrtrix.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  depends_on "gsl"
+  depends_on "qt5"
+
+  def install
+    system "./configure"
+    system "./build"
+
+    # Fetch and prepare testsuite beforehand
+    system "git", "clone", "https://github.com/MRtrix3/testing"
```
dpo: This fetches the test data into the temporary build dir, which is destroyed when `install` exits. You want to either install `testing` (`pkgshare` would be a good place), or do this inside the `test` method. In this case, you'll have to fetch inside a `cd testpath ... end` block

brainstorm: @dpo, thanks for the pointers on that. The problem I'm facing here is that the independent `testing` repo I'm fetching does rely on <a href="https://github.com/MRtrix3/testing">`build` products</a>:

```
create a symbolic link to the build script in your main MRtrix3 git clone. For example, 
if you cloned the testing repo within your main MRtrix3 folder
```

Since the `install` byproducts are gone after the install block, I've been experimenting with this:

https://github.com/brainstorm/homebrew-science/commit/3356d245ee2f50bfa21

But for some reason, `qt5` does not compile successfully anymore (even when it does great during `install`).

dpo: It's a bit odd that tests are separate from the source repo. Isn't there a simpler test that could be run?

brainstorm: I know, I had some comments on that on the upstream developers @MRtrix3 and apparently the reason for separating those is repo size (43MB of test data):

https://github.com/MRtrix3/testing/issues/2

Could we just drop the testing part and let travisci do the testing instead of Homebrew?

dpo: That sounds too large for a Homebrew install. I guess it's ok to ship those separately, but it's too bad they require the build dir. I haven't looked at the tests, but would it be possible to write a short test that you compile, link and run, as in, e.g., [`armadillo.rb`](https://github.com/Homebrew/homebrew-science/blob/master/armadillo.rb#L36). That's the best kind of Homebrew test.

brainstorm: I hope it was that easy, but not really. The tests just run compiled tools and then compare the output(s), they don't have small (functional/unit) tests like that of `armadillo.rb`, just high level integration tests :_/ ... which are covered by Travis anyhow, so if a bug got introduced it would show there.

dpo: Sorry, how are the tests covered by Travis?

brainstorm: https://github.com/MRtrix3/mrtrix3/blob/master/.travis.yml#L19

dpo: Yes I see, but that doesn't test a user's local install. It seems the `testing` repository is meant for people doing development work. Your current `test` method performs the first steps of `install` all over again, which is overkill. The point of the `test` method isn't only to ensure that the software runs. Tests are also run when a dependent is updated, to see if anything broke in the toolchain. Currently, nothing depends on `mrtrix` obviously, but down the road, it might bite.

@tseemann You're more versed in bioinformatics software. Do you have any insights here?

brainstorm: That is neuro software... don't know guys, I would just remove the test method then. Not sure what/how to test this properly :-/

dpo:
Sorry for the delay. Is there a `test` method that could be added?

dpo:
@BrewTestBot test this please

dpo:
```
== /usr/local/Library/Taps/homebrew/homebrew-science/mrtrix.rb ==
C: 17: 10: Don't use parentheses around a method call.

1 file inspected, 1 offense detected
Error: 1 problem in 1 formula

==> audit problems
homebrew/science/mrtrix:
```

dpo:
The warning about parentheses can be ignored. But a `test` method is necessary.

brainstorm:
Would it be ok if it's empty?

mrtrix.rb
```diff
@@ -0,0 +1,20 @@
...
+  depends_on "gsl"
+  depends_on "qt5"
+
+  def install
+    system "./configure"
+    system "./build"
+
+    bin.install Dir["bin/*"]
+
+    so = (OS.mac?) ? "dylib" : "so"
```
dpo: `OS.mac? ? "dylib" : "so"`

dpo:
No, the test shouldn't be empty.

brainstorm:
@dpo, as it is now, the testsuite for mrtrix is quite heavy, that's why I hesitated to include it:

https://github.com/MRtrix3/testing/tree/master/data

What would be the best course of action here?

dpo:
It must be possible to either write a minimal program that links with the library, or to run one of the utilities on a sample image?!

dpo:
Will this be updated?

Issue #2970, opened 2015-11-04
Root6 MathMore support is missing


kratsg:
Should be very easy to add (two lines)

Add the following to dependencies
```ruby
  depends_on "gsl" => :recommended
```

Then add this line in with the `args`
```ruby
      #{config_opt("mathmore", "gsl")}
```

PR #3023, opened 2015-11-14
root6: add mathmore option with gsl dependency

Moisan:
Following #2970, this PR adds the `mathmore` option to the configure arguments which depends on the `gsl` formula.


dpo:

PR #3026, opened 2015-11-15
Bump graph-tool version to 2.12

count0:
This also changes the git repository to the default upstream URL.

### DIFF
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -1,10 +1,10 @@
 class GraphTool < Formula
   homepage "http://graph-tool.skewed.de/"
-  url "http://downloads.skewed.de/graph-tool/graph-tool-2.9.tar.bz2"
-  sha256 "d3df98adc8a7ea6202e270b62c05825a483fc08c51000721356971d76af1146e"
+  url "http://downloads.skewed.de/graph-tool/graph-tool-2.12.tar.bz2"
+  sha256 "ac5fdd65cdedb568302d302b453fe142b875f23e3500fe814a73c88db49993a9"
 
   head do
-    url "https://github.com/count0/graph-tool.git"
+    url "https://git.skewed.de/count0/graph-tool.git"
     depends_on "autoconf" => :build
     depends_on "automake" => :build
     depends_on "libtool" => :build
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -50,18 +50,18 @@ class GraphTool < Formula
     depends_on "pygobject3" => with_pythons
   end
 
-# We need a compiler with C++14 support.
-fails_with :llvm
+  # We need a compiler with C++14 support.
+  fails_with :llvm
 
-fails_with :clang do
-  build 3.5.2  # the highest build version for which compilation will fail
-  cause "graph-tool must be compiled in c++14 mode"
-end
+  fails_with :clang do
+    build 3.5.2  # the highest build version for which compilation will fail
+    cause "graph-tool must be compiled in c++14 mode"
+  end
 
-fails_with :gcc
-fails_with :gcc => "4.8" do
-  cause "graph-tool must be compiled in c++14 mode"
-end
+  fails_with :gcc
+  fails_with :gcc => "4.8" do
+    cause "graph-tool must be compiled in c++14 mode"
+  end
 
   def install
     ENV.cxx11
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -54,7 +54,7 @@ class GraphTool < Formula
   fails_with :llvm
 
   fails_with :clang do
-    build "3.5.2"  # the highest build version for which compilation will fail
+    build 352  # the highest build version for which compilation will fail
     cause "graph-tool must be compiled in c++14 mode"
   end
 
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -50,6 +50,19 @@ class GraphTool < Formula
     depends_on "pygobject3" => with_pythons
   end
 
+# We need a compiler with C++14 support.
+fails_with :llvm
+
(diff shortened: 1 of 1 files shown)


dpo:
There's a compilation issue on Mavericks:
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/tuple:180:73: error: reference to type 'graph_tool::GraphInterface' requires an initializer
             _NOEXCEPT_(is_nothrow_default_constructible<_Hp>::value) : value()
                                                                        ^
```

count0:
Graph-tool requires a C++14 compiler, for instance clang 3.7. I don't know how to force this in the formula.

dpo:
Now I remember we've been over this in #2897. The `fails_with` block still seems like the way to go at this point.

count0:
Well, I'm not sure how the syntax goes. Could you please point to the documentation?

count0:
@dpo Just a bump on this. A quick nudge in the right direction might help me close this. Surely you could provide an example of how to write the two lines of the `fails_with` block that will force the use of clang 3.7...

dpo:
Sorry for the delay. Something like this should do:
```
fails_with :llvm

fails_with :clang do
  build xyz  # insert the highest build version for which compilation will fail
  cause "graph-tool must be compiled in c++14 mode"
end

fails_with :gcc
fails_with :gcc => "4.8" do
  cause "graph-tool must be compiled in c++14 mode"
end
```

count0:
@dpo Thanks. I've modified the formula accordingly.

graph-tool.rb
```diff
@@ -50,6 +50,19 @@ class GraphTool < Formula
     depends_on "pygobject3" => with_pythons
   end
 
+  # We need a compiler with C++14 support.
+  fails_with :llvm
+
+  fails_with :clang do
+    build 352  # the highest build version for which compilation will fail
```
dpo: Should probably be 600.

dpo:
It seems ok on 10.10 and 10.11, but on 10.9 try adding `-std=gnu++11` to the c++ flags.

count0:
@dpo `-std=gnu++11` is definitely wrong, since the code requires C++14, i.e. `-std=gnu++14` (which is what the build uses). C++11 lacks lambdas with auto parameter types, which are used extensively. I cannot see the full context for the build failure in 10.9, so I don't see what is the root of the problem.

dpo:
https://travis-ci.org/Homebrew/homebrew-science/jobs/99304342#L5061

count0:
@dpo Well, it seems to be using the Xcode clang, which is probably too old. We need to force that a newer vanilla clang is installed. The minimum clang build should be increased, probably. Is it possible to see from that log which build was used?

dpo:
clang build: https://travis-ci.org/Homebrew/homebrew-science/jobs/99345873#L5511. I believe 600 is the latest on 10.9. We shouldn't depend on another clang, but if we crank the build number, Homebrew should revert to gcc on 10.9.

This is the error message I saw initially: https://travis-ci.org/Homebrew/homebrew-science/jobs/99345873#L5042

count0:
I'm not sure if using GCC is a good idea, but let's see.

dpo:
On 10.9:
```
/usr/local/include/boost/math/constants/constants.hpp:277:3: error: unable to find numeric literal operator 'operator""Q'
   BOOST_DEFINE_MATH_CONSTANT(half, 5.000000000000000000000000000000000000e-01, "5.00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e-01")
   ^
/usr/local/include/boost/math/constants/constants.hpp:277:3: note: use -std=gnu++11 or -fext-numeric-literals to enable more built-in suffixes
```
It seems to be using gcc5.

count0:
No clue. As you can see, this is a boost problem. I've never seen this error before. We should just try enabling "-fext-numeric-literals" and hope it goes away.

count0:
@dpo Ugh, it seems `-fext-numeric-literals` breaks the clang builds. How do I enable it conditionally only for GCC? Is there something similar to `fails_with` for that?

graph-tool.rb
```diff
@@ -64,6 +77,7 @@ def install
 
     config_args << "--disable-cairo" if build.without? "cairo"
     config_args << "--disable-sparsehash" if build.without? "google-sparsehash"
+    config_args << "CXXFLAGS=-fext-numeric-literals" # fix issue with boost and C++11/C++14
```
dpo: Should probably be something like
```
ENV.append "CXXFLAGS", "-fext-numeric-literals" unless ENV.compiler == :clang
```

count0:
Hopefully it will work now. Note that I removed "--disable-optimization" from config_args, since this completely turns off compiler optimizations, and nobody wants that.

dpo:
Let's see if it works but there's an audit warning
```
==> brew audit homebrew/science/graph-tool
==> FAILED
Error: 1 problem in 1 formula
homebrew/science/graph-tool:
 * 77: Trailing whitespace was found
```
and a test error
```
==> python test.py
dyld: lazy symbol binding failed: Symbol not found: __ZN5boost6python7objects23register_dynamic_id_auxENS0_9type_infoEPFSt4pairIPvS2_ES4_E
  Referenced from: /usr/local/lib/python2.7/site-packages/graph_tool/libgraph_tool_core.so
  Expected in: flat namespace
```

count0:
The test error is because boost python was compiled with clang, whereas graph-tool was compiled with gcc. The user needs to compile both with the same compiler.

dpo:
Thank you!

count0:
Thank you for your help!

PR #3029, opened 2015-11-15
Bump clblas to 2.8

GOFAI:


clblas.rb
```diff
@@ -1,24 +1,11 @@
...
   homepage "https://github.com/clMathLibraries/clBLAS"
-  url "https://github.com/clMathLibraries/clBLAS/archive/v2.2.tar.gz"
-  sha256 "0563e028485d240f8b6e6efcae61677049627b19e924dcb1ef8014065b6416df"
-
-  bottle do
-    cellar :any
-    sha256 "95e08bbfde36c49f8f4fae3f420980d329c67215f7198475f34a0bae22a05f4f" => :yosemite
-    sha256 "0315da2f6b54edfff65315709f2d16cc9c9514adafa2ea7400b6a759f16001c1" => :mavericks
-    sha256 "665bd59d4dc12933722003a7092eea19104ef5722aacfdc27e423c988df0f5ca" => :mountain_lion
-  end
```
dpo: Please leave the `bottle` section as it is. It will be automatically updated.

GOFAI:
It turns out that the build only worked for me because I'm lazy and haven't updated XCode--it's failing under XCode 7.1 because of the presence of pragma directives within macros in three files in clBLAS/src/library/blas/trtri.

dpo:
So what do we do here?

GOFAI:
I opened an issue on clBLAS, but it looks like the issue won't be fixed for the immediate future. The clBLAS formula should probably stay at an earlier version until then.

PR #3031, opened 2015-11-17
fastml 3.1 (bug fix)

andrewjpage:
Fastml bundles a standard library (getopt) and it causes a conflict on 10.11. This update renames the fastml version of getopt to fastml_getopt to avoid this conflict and allow for compilation. It has been tested on a clean install of 10.11 and 10.10.


dpo:
Please add a revision number.

andrewjpage:
Added a revision number to the formula

dpo:
@BrewTestBot test this please

PR #3033, opened 2015-11-17
update armadillo to 6.200.4

zeeMonkeez:
Update armadillo to latest version. b8448fb also fixes super dependency to version 4.3.


zeeMonkeez:
So the build fails because when a dependent package is tested, it is not rebuilt from source (and thus linked against an old version of the library). When an executable is called it cannot find the library.
Does that mean the automated build process is wrong? Shouldn't all packages be rebuilt from source?

PR #3036, opened 2015-11-19
bump Astrometry.net version to latest release, 0.64

dstndstn:
Just keeping up with the latest release.  Minor changes from 0.56 to 0.64.


tseemann:
Hey @dpo - did the test work for you? It failed for me:

```
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20151120-44527-ut25za.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-science/astrometry-net.rb -v
==> solve-field
ERROR: You didn't specify any files to process.
This program is part of the Astrometry.net suite.
For details, visit http://astrometry.net.
Git URL https://github.com/dstndstn/astrometry.net
Revision 0.64, date Thu_Oct_22_10:00:31_2015_-0400.

Usage:   solve-field [options]  [<image-file-1> <image-file-2> ...] [<xyls-file-1> <xyls-file-2> ...]
```

dstndstn:
It's not really a "test"... it just runs the main executable.  Apparently
it returns an error if you don't specify any arguments.  Changing it to
"solve-field -h" should at least make it return 0.

dpo:
There doesn't seem to be any difference with or without `-h`. Isn't there a test file that we could feed the driver?

PR #3041, opened 2015-11-21
breseq 0.28.1

tseemann:
breseq is a computational pipeline for finding mutations relative to a reference sequence in short-read DNA re-sequencing data for microbial sized genomes. breseq is a command line tool implemented in C++ and R. It will compile and function on a variety of Unix platforms, including MacOSX. It reports single-nucleotide mutations, point insertions and deletions, large deletions, and new junctions supported by mosaic reads (such as those produced by new mobile element insertions) in an annotated HTML format.


tseemann:
Hmm, it fails only on OS X 10.9  - any ideas?

```
Error: homebrew/science/breseq: failed
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/minitest/unit.rb:302:in `=~'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/minitest/unit.rb:302:in `assert_match'
/usr/local/Library/Taps/homebrew/homebrew-science/breseq.rb:31:in `block in <class:Breseq>'
/usr/local/Library/Homebrew/formula.rb:1284:in `block in run_test'
/usr/local/Library/Homebrew/formula.rb:1280:in `run_test'
/usr/local/Library/Homebrew/test.rb:28:in `block in <main>'
/usr/local/Library/Homebrew/test.rb:27:in `<main>'
```

tseemann:
Here we go again:
```
Error: 1 problem in 1 formula
homebrew/science/breseq:
```
It is published and most people don't get it via github.

ilovezfs:
Is there a programmatic way to check for a doi? If so, we can fix this with a pull request to modify the notability check to also include that as a valid way to fulfill the diagnostic.

tseemann:
@ilovezfs at the moment the `doi` and `tag` attributes are commented out, but in practice @sjackman encourages us to use them, and we do. they are useful for generating links to the publications etc.  eventually they will be official I think.

ilovezfs:
@tseemann we could just have audit grep the formula text in the mean time.

tseemann:
@ilovezfs i assume this logic is in OSX homebrew core?

sjackman:
Yep. It's at https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L602

PR #3043, opened 2015-11-21
Update glpk 4.57, fix test

ackinetics:
Updated to the latest glpk 4.57.  Fixed the test so that it does not assume that /usr/local/lib and /usr/local/include are in the search path for the compiler

### DIFF
--- a/glpk.rb
+++ b/glpk.rb
@@ -1,4 +1,5 @@
 class Glpk < Formula
+  desc "GLPK (GNU Linear Programming Kit) solves Linear Programming (LP) and Mixed Integer Programming (MIP) problems"
   homepage "https://www.gnu.org/software/glpk/"
   url "http://ftpmirror.gnu.org/glpk/glpk-4.57.tar.gz"
   mirror "https://ftp.gnu.org/gnu/glpk/glpk-4.57.tar.gz"
--- a/glpk.rb
+++ b/glpk.rb
@@ -1,8 +1,8 @@
 class Glpk < Formula
   homepage "https://www.gnu.org/software/glpk/"
-  url "http://ftpmirror.gnu.org/glpk/glpk-4.52.tar.gz"
-  mirror "https://ftp.gnu.org/gnu/glpk/glpk-4.52.tar.gz"
-  sha1 "44b30b0de777a0a07e00615ac6791af180ff4d2c"
+  url "http://ftpmirror.gnu.org/glpk/glpk-4.57.tar.gz"
+  mirror "https://ftp.gnu.org/gnu/glpk/glpk-4.57.tar.gz"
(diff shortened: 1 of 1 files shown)


dpo:
@ackinetics Thanks! @schoeps Is this version supported by `octave`? It was mentioned in #https://github.com/Homebrew/homebrew-science/pull/1250#issuecomment-59082065 that the GLPK API has changed.

dpo:
At least the `octave` test passes, and it seems only `shogun` needs to be bumped.

schoeps:
I think that is save. There should be no problem. However, I will do some tests.

glpk.rb
```diff
@@ -1,8 +1,8 @@
 class Glpk < Formula
```
dpo: Would you mind adding a `desc` field with a brief description of the software?

glpk.rb
```diff
@@ -34,7 +34,7 @@ def install
         return 0;
     }
     EOF
-    system ENV.cc, "test.c", "-lglpk", "-o", "test"
+    system ENV.cc, "test.c", "-L/usr/local/lib", "-I/usr/local/include", "-lglpk", "-o", "test"
```
dpo: Why is this necessary? Those locations should be searched automatically?!

ackinetics: Dominique,

added the desc.  

I put the full paths because my new El Capitan System did not have /usr/local/lib or /usr/local/include in the default path.  The test failed as written.  I can add those to the search path but out of the box it is not done for us.

Thanks




  CONFIDENTIAL

 

--------------------------------
 
Neil Singer, PhD
President & CTO

AC Kinetics, Inc.
20 Bedford Road
Armonk, NY 10504
www.ackinetics.com
 
(914) 816-1221
Fax: (914) 273-2882

dpo: On Mon, Nov 23, 2015 at 05:11:23PM -0800, ackinetics wrote:
>> @@ -34,7 +34,7 @@ def install
>>          return 0;
>>      }
>>      EOF
>> -    system ENV.cc, "test.c", "-lglpk", "-o", "test"
>> +    system ENV.cc, "test.c", "-L/usr/local/lib", "-I/usr/local/include", "-lglpk", "-o", "test"
>
>Dominique,
>
>added the desc.
>
>I put the full paths because my new El Capitan System did not have /usr/local/lib or /usr/local/include in the default path.  The test failed as written.  I can add those to the search path but out of the box it is not done for us.

Neil,

I think users normally have /usr/local/bin in their PATH and /usr/local/lib in
their LD_LIBRARY_PATH. There shouldn't be any need to specify the paths
explicitly here.

Dominique

dpo: Sorry, that isn't entirely right. It would be safe to say
```
-L#{lib} -I#{include}
```
in the `test` method. But outside of the formula, users will have added `$HOMEBREW_PREFIX/bin` to `PATH`, etc.

PR #3088, opened 2015-12-04
abyss-explorer: add a test

Moisan:
Add a simple test to the `abyss-explorer` formula.


dpo:

PR #3098, opened 2015-12-06
parmetis, superlu_dist: update patch shas.

### LINKED ISSUE #3060: bitbucket commit shasum changed (affects superlu_dist and parmetis)
It seems bitbucket changed the format for `raw` commit which affects patches (at least) in `parmetis` and `superlu_dist`. As a result those formulae fail due to sha mismatch.

Alternative could be to download those from PETSc bitbucket with patches already applied:
https://bitbucket.org/petsc/pkg-parmetis/get/v4.0.3p2.tar.bz2
https://bitbucket.org/petsc/pkg-superlu_dist/get/v4.2.tar.bz2

dpo:
Fixes #3060.


davydden:
@BrewTestBot test this please

dpo:
Need to rebottle `petsc` first.

PR #3100, opened 2015-12-07
update superlu_dist.rb

bknaepen:
see commit message


dpo:
It's already in #3098 and fails for the same reason. We have to wait for the MPI bug to get fixed.

PR #3103, opened 2015-12-08
superlu 5.1

ntkme:
This PR updates superlu to version 5.1


ntkme:
Nothing really failed except the `brew test` has timeout:

```
Error: homebrew/science/superlu: failed
execution expired
```

dpo:
It does hang at that test. I can reproduce locally. So there seems to be a problem. Do the tests finish for you?

dpo:
I tested `sitersol` on another matrix and it completed successfully.

PR #3104, opened 2015-12-08
augustus 3.2.1

ntkme:
This PR updates augustus to version 3.2.1.


augustus.rb
```diff
@@ -15,16 +16,26 @@ class Augustus < Formula
...
+    depends_on "gsl"
+    depends_on "lp_solve"
+    depends_on "suite-sparse"
   end
 
   def install
-    system "make"
-    rm_r %w[include mysql++ src]
+    inreplace "common.mk", "#ZIPINPUT = true", "ZIPINPUT = true" if build.with? "zlib"
+    inreplace "common.mk", "#COMPGENEPRED = true", "COMPGENEPRED = true" if build.with? "cgp"
```
dpo: Instead of using `inreplace`, is it possible to invoke `make` with `ZIPINPUT=true`, etc.?

dpo:
Much cleaner, thank you!

PR #2930, opened 2015-10-24
Trilinos 12.2.1 (fix hypre options)

davydden:

PR #3120, opened 2015-12-13
trilinos: update to 12.4.2; fix hypre options and mumps libraries order

davydden:
superseeds https://github.com/Homebrew/homebrew-science/pull/2930

### DIFF
--- a/trilinos.rb
+++ b/trilinos.rb
@@ -1,10 +1,9 @@
 class Trilinos < Formula
   desc "Algorithms for the solution of large-scale, complex multi-physics engineering and scientific problems"
   homepage "http://trilinos.sandia.gov"
-  url "https://trilinos.org/oldsite/download/files/trilinos-12.2.1-Source.tar.bz2"
-  sha256 "4a884fd5eef885815d25fc24c5a4b95e5e67b9eefaa01d61524eef92ae9319fd"
+  url "https://trilinos.org/oldsite/download/files/trilinos-12.4.2-Source.tar.bz2"
+  sha256 "78225d650ddcfc453b40cddb99b2fbb79998ff6359f9090154727b916812a58e"
   head "https://software.sandia.gov/trilinos/repositories/publicTrilinos", :using => :git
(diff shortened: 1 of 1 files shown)


dpo:
https://github.com/Homebrew/homebrew/pull/46461

trilinos.rb
```diff
@@ -58,7 +57,7 @@ class Trilinos < Formula
 
   # Experimental TPLs:
   depends_on "eigen"        => :recommended
-  depends_on "hypre"        => [:recommended] + mpidep + openblasdep # EpetraExt tests fail to compile
+  depends_on "hypre"        => [:recommended] + ((build.with? "mpi") ? [] : ["without-mpi"]) + openblasdep # EpetraExt tests fail to compile
```
dpo: Could you check whether this can be restored, now that the MPI issues have been fixed?

davydden: it's not related to mpi issues, AFAIK. The problem was that `hypre` does not have `--with-mpi`, it has `--without-mpi`.

dpo: Ah, you're right.

dpo:
@BrewTestBot test this please

dpo:
It times out. We need two hours of build time...

dpo:
@mikemcquaid Is there any chance we could request extra build time for specific formulae? This one is particularly tricky and there's no way it will fit in 60 minutes.

mikemcquaid:
@dpo I'll try and see if I can find a time when the CI isn't too busy. In general though: unfortunately not.

mikemcquaid:
@BrewTestBot test this please

mikemcquaid:
@dpo I've bumped this to 120m and scheduled a rebuild.

dpo:
@mikemcquaid Thanks! That should do it.

davydden:
i could be wrong, but it seems the build was not restarted, or maybe it's still in the queue...

dpo:
It seems to have worked.

PR #3122, opened 2015-12-13
calculix-ccx 2.9 (new formula)

costerwi:
Calculix ccx is a finite element solver with input file syntax similar to Abaqus.
The source code is available from the [main website](http://calculix.de/).


calculix-ccx.rb
```diff
@@ -0,0 +1,118 @@
...
+    # The spooles library is not currently maintained and so would not make a
+    # good brew candidate. Instead it will be static linked to ccx.
+    url "http://www.netlib.org/linalg/spooles/spooles.2.2.tgz"
+    sha256 "a84559a0e987a1e423055ef4fdf3035d55b65bbe4bf915efaa1a35bef7f8c5dd"
+  end
+
+  # Add <pthread.h> to Calculix.h
+  # Read arpack link options from pkg-config
+  # u_free must return a void pointer
+  patch :DATA
```
dpo: Was this submitted upstream?

calculix-ccx.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    # Extend library with multi-threading (MT) subroutines
+    system "make", "-C", "spooles/MT/src", "makeLib"
+
+    # Build Calculix ccx
+    args = [
+      "CC=#{ENV.cc}",
+      "FC=#{ENV.fc}",
+      "CFLAGS=-O2 -I../../spooles -DARCH=Linux -DSPOOLES -DARPACK -DMATRIXSTORAGE -DUSE_MT=1",
+      "FFLAGS=-O2 -fopenmp",
```
dpo: Is this a Fortran-only library? Keep in mind that clang, which is the default compiler on OSX, doesn't (yet) support OpenMP.

calculix-ccx.rb
costerwi: I'll suggest them to the author but these changes are already summarized in [compilation instructions for OSX](http://www.dhondt.de/INST_CCX_2_9_MAC_11_2015.pdf) on the website.

calculix-ccx.rb
costerwi: The openmp is only used in Fortran which seems to compile correctly.

costerwi:
Do you know why I get strange messages about Fortran when compiling? I do not set FC, FCFLAGS, or FFFLAGS. I have gfortran (gcc) installed by the brew dependency which seems to be detected initially but then reports it is unsupported:
```
$ brew install calculix-ccx
==> Installing calculix-ccx from homebrew/science
==> Building with an alternative Fortran compiler
This is unsupported.
Warning: No Fortran optimization information was provided.  You may want to consider
setting FCFLAGS and FFLAGS or pass the `--default-fortran-flags` option to
`brew install` if your compiler is compatible with GCC.
```

costerwi:
Sorry, it looks like the fortran messages are a common problem that has been reported [here](Homebrew/homebrew#31156) and elsewhere.

dpo:
Yes, no need to worry about those messages.

PR #3125, opened 2015-12-16
opengrm-thrax 1.2.0

stevenbedrick:
Updates the opengrm-thrax recipe to install the latest version of the Thrax library.


stevenbedrick:
Turns out that there's a sandbox issue- the test case is trying to write to `#{share}/thrax/grammars` which isn't one of the whitelisted directories.

dpo:
The solution is probably to `cp_r share/"thrax/grammars", testpath` and `cd` to that location before running the test.

stevenbedrick:
Aha! Yes! That totally did the trick. Thanks!

dpo:
Thank you!

PR #3131, opened 2015-12-18
ESS: 15.09-2 (new formula)

simonbyrne:
Add Emacs Speaks Statistics, an add-on package for emacs for interaction with R and other statistical packages.

### DIFF
--- a/ess.rb
+++ b/ess.rb
@@ -1,0 +1,38 @@
+class Ess < Formula
+  desc "Emacs Speaks Statistics"
+  homepage "http://ess.r-project.org/"
+  url "http://ess.r-project.org/downloads/ess/ess-15.09-2.tgz"
+  version "15.09-2"
+  sha256 "706c41237e1edf33a369902f503bb25254b2bbb750b9ed1adee244e875264afb"
+
+  depends_on :emacs => "23"
+
+  def install
+    system "make", "install", "PREFIX=#{prefix}", "EMACS=#{ENV["EMACS"]}",
+           "LISPDIR=#{elisp}"
+  end
+
+  def caveats; <<-EOS.undent
+      To load the package, add
+
+        (require 'ess-site)
+
+      to your '~/.emacs' file.
+    EOS
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


ess.rb
```diff
@@ -0,0 +1,38 @@
...
+  desc "Emacs Speaks Statistics"
+  homepage "http://ess.r-project.org/"
+  url "http://ess.r-project.org/downloads/ess/ess-15.09-2.tgz"
+  version "15.09-2"
+  sha256 "706c41237e1edf33a369902f503bb25254b2bbb750b9ed1adee244e875264afb"
+
+  depends_on :emacs => "23"
+
+  def install
+    system "make", "install", "PREFIX=#{prefix}", "EMACS=#{which "emacs"}"
```
dpo: The `:emacs` dependency sets `ENV["EMACS"]`. Is that not sufficient here?

ess.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def caveats
+    <<-EOS.undent
```
dpo: Most formulas combine lines 14 and 15. We should stick with the most common style.

ess.rb
simonbyrne: Ah, I didn't know that, I was basing it on the [auctex one](https://github.com/Homebrew/homebrew-tex/blob/95905f3106773e90955503f000fe27b3915dd450/Formula/auctex.rb#L23). Will fix it.

ess.rb
simonbyrne: will do.

ess.rb
simonbyrne: Actually, I don't know: by default the Makefile set `EMACS=emacs`, will the environmental variable override that? Or do I need to do something else?

dpo: I was thinking `EMACS=#{ENV["EMACS"]}`

simonbyrne:
@dpo Thanks, I've updated the formula.

dpo:
Great! Thank you!

PR #3142, opened 2015-12-22
armadillo: update to 6.400.3

zeeMonkeez:


dpo:
Thanks. Seems like `mlpack` will need to be rebottled.

PR #3145, opened 2015-12-24
vtk5: Add option to compile with system python

zeroyonichihachi:
Currently, when we install vtk5 with python bindings, it always installs Homebrew's python and uses it for the compilation. So we can't install vtk5 with the bindings for the system python.

This PR adds an option (`--with-system-python`) which enables to install vtk5 with a binding on the system python.


vtk5.rb
```diff
@@ -23,13 +23,16 @@ class Vtk5 < Formula
   option "with-qt-extern",  "Enable Qt4 extension via non-Homebrew external Qt4"
   option "with-tcl",        "Enable Tcl wrapping of VTK classes"
   option "without-legacy",  "Disable legacy APIs"
+  option "with-system-python", "Compile with system python bindings"
 
   depends_on "cmake" => :build
   depends_on :x11 => :optional
   depends_on "qt" => :optional
-  depends_on :python => :recommended
```
dpo: It should probably be
```
```
so as to use system Python when available.

zeroyonichihachi: Thanks for the comment, and that should be.

I'll close this PR and send again according to your advice.

tdsmith:
Does it make a difference? Does vtk actually link to the Python framework?

dpo:
It does:
```
$ otool -L /usr/local/Cellar/vtk5/5.10.1_2/lib/python2.7/site-packages/vtk/vtkCommonPython.so 
/usr/local/Cellar/vtk5/5.10.1_2/lib/python2.7/site-packages/vtk/vtkCommonPython.so:
	/usr/local/Cellar/vtk5/5.10.1_2/lib/vtk-5.10/libvtkCommonPythonD.5.10.dylib (compatibility version 5.10.0, current version 5.10.1)
	/usr/local/Frameworks/Python.framework/Versions/2.7/Python (compatibility version 2.7.0, current version 2.7.0)
	/usr/local/Cellar/vtk5/5.10.1_2/lib/vtk-5.10/libvtkPythonCore.5.10.dylib (compatibility version 5.10.0, current version 5.10.1)
	/usr/local/Cellar/vtk5/5.10.1_2/lib/vtk-5.10/libvtkCommon.5.10.dylib (compatibility version 5.10.0, current version 5.10.1)
	/usr/local/Cellar/vtk5/5.10.1_2/lib/vtk-5.10/libvtksys.5.10.dylib (compatibility version 5.10.0, current version 5.10.1)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
```

vtk5.rb
dpo: Thanks. Actually, I'd say
```
depends_on :python => :recommended if MacOS.version <= :snow_leopard
```

tdsmith:
Thanks, @dpo. It looks like the "right" thing to do here is to make that link unnecessary if we can (by e.g. passing `-undefined dynamic_lookup` instead of `-lpython` per my manifesto: http://blog.tim-smith.us/2015/09/python-extension-modules-os-x/)

PR #3171, opened 2016-01-04
Update visp.rb

fspindle:
Update to ViSP 3.0.0 release


visp.rb
```diff
@@ -14,19 +14,16 @@ class Visp < Formula
...
-  depends_on "libpng"   => :recommended
-  depends_on "zbar"     => :recommended
-  depends_on "ffmpeg"   => :optional
-
   option :cxx11
 
-  option "with-demos", "Build with demos"
-  option "with-examples", "Build with examples"
-  option "without-tests", "Build without tests"
-  option "with-tutorials", "Build with tutorials"
```
dpo: What is the rationale for removing these options?

visp.rb
```diff
@@ -37,17 +34,24 @@ def install
...
-    args << "-DUSE_FFMPEG="       + arg_switch("ffmpeg")
-    args << "-DUSE_LIBJPEG="      + arg_switch("jpeg")
-    args << "-DUSE_LIBPNG="       + arg_switch("libpng")
-    args << "-DUSE_OPENCV="       + arg_switch("opencv")
-    args << "-DUSE_ZBAR="         + arg_switch("zbar")
+    if build.cxx11?
+      args << "-DUSE_CPP11==ON"
+    else
+      args << "-DUSE_CPP11=OFF"
+    end
```
dpo: A one-liner would be more compact and more in line with `arg_switch`:
```
args << ("-DUSE_CPP11=" + (build.cxx11? "ON" : "OFF"))

File: visp.rb (hunk 1 above)
Comment (fspindle) [inline visp.rb]: Demos, examples, tests and tutorials take time to build and are anyway not installed. Enabling/disabling these 4 options doesn't change the content of the material that is installed, that's why I removed these options.

File: visp.rb (hunk 2 above)
Comment (fspindle) [inline visp.rb]: ok

Comment (dpo) [inline visp.rb]: Sorry, there's a typo (I didn't test; please test first):
```
args << ("-DUSE_CPP11=" + ((build.cxx11?) ? "ON" : "OFF"))
```

File: visp.rb (hunk 1 above)
Comment (dpo) [inline visp.rb]: Ok, thanks.

File: visp.rb [hunk 3]
```diff
@@ -37,17 +34,20 @@ def install
...
+      -DBUILD_TESTS=OFF
+      -DBUILD_TUTORIALS=OFF
     ]
 
-    args << "-DUSE_FFMPEG="       + arg_switch("ffmpeg")
-    args << "-DUSE_LIBJPEG="      + arg_switch("jpeg")
-    args << "-DUSE_LIBPNG="       + arg_switch("libpng")
-    args << "-DUSE_OPENCV="       + arg_switch("opencv")
-    args << "-DUSE_ZBAR="         + arg_switch("zbar")
+    args << "-DUSE_CPP11=ON" if build.cxx11?
```
fspindle: The syntax you suggest was not working. Since USE_CPP11=OFF is the default, I just turn ON when needed.

PR #3178, opened 2016-01-06
adol-c 2.6.0

dpo:
- install documentation
- implement test.

### DIFF
--- a/adol-c.rb
+++ b/adol-c.rb
@@ -1,7 +1,9 @@
 class AdolC < Formula
+  desc "Automatic Differentiation by Overloading in C/C++"
   homepage "https://projects.coin-or.org/ADOL-C"
-  url "http://www.coin-or.org/download/source/ADOL-C/ADOL-C-2.5.2.tgz"
-  sha256 "2fa514d9799989d6379738c2bcf75070d9834e4d227eb32a5b278840893b2af9"
+  url "http://www.coin-or.org/download/source/ADOL-C/ADOL-C-2.6.0.tgz"
+  sha256 "add322a59f4b038ed24a53cf848235c0a22bf27ac00a389e8e594b2cfb1bb2f0"
+  head "https://projects.coin-or.org/svn/ADOL-C/trunk/", :using => :svn
 
   bottle do
     revision 1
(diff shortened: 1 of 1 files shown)


dpo:
Why does the test fail to compile???
```
Failed executing: /usr/bin/clang++ -std=c++11 -stdlib=libc++ test.cpp -I/usr/local/opt/adol-c/include -o test -L/usr/local/Cellar/adol-c/2.6.0/lib -L/usr/local/lib -L/usr/local/opt/colpack/lib -ladolc -Wl,-rpath,/usr/local/Cellar/adol-c/2.6.0/lib -lboost_system -lColPack -Wl,-rpath,/usr/local/opt/colpack/lib
```

dpo:
@UniqMartin Sorry to bother, but I can't see why `brew` refuses to compile the test [here](https://github.com/Homebrew/homebrew-science/pull/3178/files#diff-5d2c080af71809791ad3484c82aae485R93). I'm wondering if you see something wrong. Rubocop doesn't.

UniqMartin:
With Clang being the C++ compiler, `ENV.cxx11` appends `-std=c++11` and `-stdlib=libc++` to the `CXX` environment variable. This environment variable is then returned by `ENV.cxx` as a string and reads `"/usr/bin/clang++ -std=c++11 -stdlib=libc++"`. Now `system` treats it as a unit and (since it's the first argument) as the file name of the binary to be executed. There is no such file, so it fails. This works:

```ruby
cxx_with_args = ENV.cxx.split + %W[
  test.cpp
  -I#{opt_include}
  -o test
  -L#{opt_lib} -ladolc
  -L#{Formula["colpack"].opt_lib} -lColPack
]
system *cxx_with_args
```

dpo:
Tricky! That works and makes a lot of sense. Many thanks!

adol-c.rb
```diff
@@ -39,5 +39,59 @@ def install
...
+        free(H);
+        cout << yp - 1 / (1.0 + n) << "\\n";
+        cout << errg << "\\n";
+        cout << errh << "\\n";
+        return 0;
+      }
+    EOS
+    ENV.cxx11
+    flags = "-I#{opt_include} -L#{opt_lib} -ladolc -L#{Formula["colpack"].opt_lib} -lColPack".split
+    system *(ENV.cxx.split), "test.cpp", "-o", "test", *flags
```
UniqMartin: Not sure what your policy with Ruby 1.8 compatibility is in this tap, but the core still tries to stay compatible. This line isn't, because Ruby 1.8 only allows a splat in the last argument. That's the reason I put all arguments into a single array in my proposed fix.

I would also recommend against constructing an array by splitting a string like you do for `flags`. Making it an array from the start makes things more robust (maybe some day paths will be allowed to have spaces) and with the `%W[]` syntax I suggested it is equally succinct.

UniqMartin:
> Tricky! That works and makes a lot of sense. Many thanks!

Indeed tricky! The error message certainly wasn't helpful as it didn't disclose where one argument end and where the next one begins. You're welcome!

adol-c.rb
dpo: You're right. I didn't realize the compatibility issue. Thanks.

PR #3194, opened 2016-01-10
ParaView 5.0.0 (update)

utkarshayachit:
Update to ParaView 5.0.0.

### DIFF
--- a/paraview.rb
+++ b/paraview.rb
@@ -1,8 +1,9 @@
 class Paraview < Formula
+  desc "ParaView: an open-source data analysis and visualization application"
   homepage "http://paraview.org"
-  url "http://www.paraview.org/paraview-downloads/download.php?submit=Download&version=v4.2&type=source&os=all&downloadFile=ParaView-v4.2.0-source.tar.gz"
-  sha256 "ac26cc5fe5ce82d27531727a01242353d40984826eaa580edea0791887a07b6b"
-  head "git://paraview.org/ParaView.git"
+  url "http://www.paraview.org/paraview-downloads/download.php?submit=Download&version=v5.0&type=source&os=all&downloadFile=ParaView-v5.0.0-source.tar.gz"
+  sha256 "b0ecfc8f590a696a4374752961abcf663acfc367ced8101ae4419cfbc6c60534"
+  head "https://gitlab.kitware.com/paraview/paraview.git"
 
   bottle do
     sha256 "f3c77c1007faebe1294e72f6e431b654916ccfaa632411080932e827e5efa2bf" => :yosemite
@@ -10,8 +11,9 @@ class Paraview < Formula
     sha256 "1a8b184f3cfa7adf51c1cb43a8cf5a6b7297cca7da3741afa0b2074e204a744a" => :mountain_lion
   end
 
-  depends_on "cmake" => :build
+  option "without-opengl2", "Use legacy OpenGL 1.1 rendering backend."
 
+  depends_on "cmake" => :build
   depends_on "boost" => :recommended
   depends_on "cgns" => :recommended
   depends_on "ffmpeg" => :recommended
@@ -25,10 +27,6 @@ class Paraview < Formula
   depends_on "libtiff"
   depends_on "fontconfig"
   depends_on "libpng"
-
-  # Temporary fix for a cast issue related to FreeType.
-  # See https://bugs.gentoo.org/show_bug.cgi?id=533444
-  patch :DATA
 
   def install
     args = std_cmake_args + %W[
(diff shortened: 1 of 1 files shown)


dpo:
`cgns` needs to be bumped first:
```
==> brew test --verbose homebrew/science/paraview
==> FAILED
Testing homebrew/science/paraview
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160110-65066-tjxp8m-0.sb /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-science/paraview.rb --verbose
==> /usr/local/Cellar/paraview/5.0.0/paraview.app/Contents/MacOS/paraview --version
dyld: Library not loaded: /usr/local/lib/libhdf5.9.dylib
  Referenced from: /usr/local/opt/cgns/lib/libcgns.3.2.dylib
  Reason: image not found
```

paraview.rb
```diff
@@ -10,8 +11,9 @@ class Paraview < Formula
     sha256 "1a8b184f3cfa7adf51c1cb43a8cf5a6b7297cca7da3741afa0b2074e204a744a" => :mountain_lion
   end
 
-  depends_on "cmake" => :build
```
dpo: You're using `cmake` on line 71.

utkarshayachit: `cmak`e depedency still exists, it just moved to line 16 as I added the new option for `without-opengl2`.

dpo: Apologies. I missed it. Thanks.

dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

utkarshayachit:
Please let me know if there's anything I can do to get this merged. If I am not mistaken, looks like the testbot keeps on failing due to timeouts. That's not too surprising. ParaView can take a long time to build on slower systems.

dpo:
@mikemcquaid This is another one where we need more than 60 minutes to build. Any chance to get such resources at some point?

mikemcquaid:
@dpo I'll do it when I get the chance. I think we'll need to work out a better long-term solution for this, though.

mikemcquaid:
CC @DomT4 as you often keep a closer look at the build queue than I do now.

DomT4:
https://github.com/Homebrew/homebrew-science/pull/3314#issuecomment-188310049

dpo:
This just won't build under the time constraints so I'd like to ask a question here. Is there any advantage to having this formula in Homebrew/Science given that it's available from Homebrew/Cask?

@utkarshayachit @davydden

utkarshayachit:
@dpo The one reason for having this formula in homebrew-science is that it will enable SDK installs in near future (currently it doesn't do that). That would be needed for folks trying to develop plugins for ParaView or custom applications or instrumenting simulation codes to use ParaView insitu components.

davydden:
I, personally, do not build paraview and rather use the homebrew-cask version. 

As far as i recall, building paraview could be advantageous when one wants to add some plugins which are not included automatically. So i would keep it.

dpo:
Ok. Thanks both for the feedback. That makes sense.

DomT4:
Don't rely on Travis here.

If you need bottles you'll only have to go through Jenkins regardless to get them. Just wait till 23:00 - 09:00 UK time (18:00 - 04:00 EST), check the CI queue is clear/clearish, set the Science timeout north of 4 hours and re-queue the build by calling `@BrewTestBot` with the magical phrase.

As long as you revert the timeout relatively quickly after the build completes it's fine to go higher temporarily.

dpo:
If I have authority to do that, I don't know how.

DomT4:
Fairly sure if you can access the bot you're allowed to play with the timeouts. Can you access the http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/configure page?

dpo:
Ok thanks.

DomT4:
@BrewTestBot test this please

DomT4:
This one failed to build on all three platforms, sadly. Actual build failure rather than timeout, though.

dpo:
Sorry about that. I'll check it out locally.

dpo:
The bug is upstream and related to recent changes in `ffmpeg`, similar to #3380.

utkarshayachit:
Thanks @dpo, I will push a fix to upstream and this PR to address this issue.

dpo:
It seems 5.1 is out now. @utkarshayachit do you want to push an update?

utkarshayachit:
@dpo Yes, I'll have an update soon. Should I close this PR and open a new one when that happens, or just update this one?

dpo:
Let's open a new one. I'll close this.

PR #3201, opened 2016-01-11
update nip2.rb to 8.2.0

jcupitt:
update nip2.rb to 8.2.0, the latest stable version


dpo:
@BrewTestBot test this please

PR #3205, opened 2016-01-12
pastix 5.2.2.22

UniqMartin:
Aside from updating the formula, this also applies a cosmetic fix and fixes the dynamic linking of the examples such that they actually work when used in conjunction with `brew test pastix`.

As a nice side effect, this will also fix the currently broken bottles, that are linked against some MPI library that no longer exists (it is probably from an older version of the `open-mpi` formula).


dpo:
:+1:

UniqMartin:
Are you okay with me pulling these kinds of updates (version bumps, minor fixes) without prior review? Or would you prefer if someone from the science team had a look first and gave a :+1: as you just did?

(I'm asking because I'm not sure what the customs around maintaining a tap are, and thus far I'm only a core maintainer.)

dpo:
I had a look and it looked good ;-). Thanks. But yes, please feel free to pull.

UniqMartin:
Thanks, will do in the future! And feel free to shout at me if I ever mess up when doing that. :smile_cat:

PR #3206, opened 2016-01-12
travis: various improvements

UniqMartin:
I picked the latest available images per OS X release as seen in the [Travis CI documentation](https://docs.travis-ci.com/user/languages/objective-c).


dpo:
There is still a number of differences with the [`.travis.yml`](https://github.com/Homebrew/homebrew/blob/master/.travis.yml) used in core. Do they matter?

UniqMartin:
I think there are some minor adjustments that could be carried over to keep the differences to core as small as necessary. If you want, I can do that and extend this PR a bit.

And then there's the slightly more intrusive change of testing with Ruby 1.8 on OS X 10.9. If you're not opposed to that, I would apply equivalent changes to this tap's `.travis.yml`. (Don't know if the science team cares about this. And I'm not the one to dictate how to run this tap. :smiley_cat:)

dpo:
> I think there are some minor adjustments that could be carried over to keep the differences to core as small as necessary. If you want, I can do that and extend this PR a bit.

I'm in favor of keeping things in sync if possible.

> And then there's the slightly more intrusive change of testing with Ruby 1.8 on OS X 10.9.

What would be the consequences/benefits of that?

UniqMartin:
> I'm in favor of keeping things in sync if possible.

I went ahead and added the missing stuff piece by piece. I tried to break down the changes into separate commits and motivated them where necessary, the primarily aim being to ensure that Homebrew is in a more defined state once the test bot starts building. What do you think?

> What would be the consequences/benefits of that?

The main benefit is that there's a reduced chance of code/formulae being added that use features from newer Rubies and thus would break builds for people with more outdated OS X installations. The core hasn't abandoned Ruby 1.8 yet and the `emacs` tap also recently added this part to their `.travis.yml`, so I figured you might want to do that, too.

dpo:
Many thanks!

> Step 2: Tap required taps (homebrew/python, homebrew/versions) as many
science formulae depend on them, symlink homebrew/science tap into
pre-installed Homebrew, and adjust environment accordingly.

Is it necessary to do this? Don't we want to be sure `brew` itself will tap what's necessary based on a formula's dependencies? Fresh users won't necessarily have those taps ready to go.

> The core hasn't abandoned Ruby 1.8 yet

Ok. So are we saying that as long as Mavericks is officially supported, all formulae should be Ruby 1.8 compliant?

UniqMartin:
> Is it necessary to do this? Don't we want to be sure brew itself will tap what's necessary based on a formula's dependencies? Fresh users won't necessarily have those taps ready to go.

Being a relatively new maintainer, I can't answer this with full confidence. I'll have to investigate that. In a way, things are still as they were, i.e. `homebrew/python` is tapped as before and `homebrew/versions` was already there as it came pre-installed with the Homebrew in the Travis image. But you ask good questions and I'll try to answer them soon!

> Ok. So are we saying that as long as Mavericks is officially supported, all formulae should be Ruby 1.8 compliant?

Maybe earlier, as we are currently working towards shipping a portable Ruby 2.0 with Homebrew that will work on OS X 10.5+. (There are still some building blocks missing before we can ship and use that.) Once this has happened, we're likely to stop caring about Ruby 1.8.

dpo:
> Maybe earlier, as we are currently working towards shipping a portable Ruby 2.0 with Homebrew that will work on OS X 10.5+.

Sounds excellent. I recall discussions mentioning Traveling Ruby.

Please feel free to pull this when you're ready!

dpo:
Good to go?

UniqMartin:
> Good to go?

It is good now, after rebasing to resolve a conflict. Sorry for the delay! I was too occupied with other stuff. Feel free to give a final review, if you wish, but I guess I'll just merge it later today.

UniqMartin:
>> Step 2: Tap required taps (homebrew/python, homebrew/versions) as many science formulae depend on them, symlink homebrew/science tap into pre-installed Homebrew, […]
>
> Is it necessary to do this? Don't we want to be sure `brew` itself will tap what's necessary based on a formula's dependencies? Fresh users won't necessarily have those taps ready to go.

You are right! The `test-bot` is certainly clever enough to do that, so there's no point to tap prematurely. Updated the relevant commit to reflect that. This saves some time if the taps are not needed and avoids any unwanted influences that may come from those taps. Thanks a lot for asking!

.travis.yml
```diff
@@ -1,36 +1,36 @@
...
-      git -C $(brew --repo) reset --hard origin/master;
-      git -C $(brew --repo) clean -qxdff;
-      brew tap homebrew/python;
-      brew update;
-      ln -s $PWD $(brew --repo)/Library/Taps/homebrew/homebrew-science;
-      cd $(brew --repo)/Library/Taps/homebrew/homebrew-science;
+      sudo chown -R "$USER" "$(brew --repo)";
+      git -C "$(brew --repo)" reset --hard origin/master;
+      git -C "$(brew --repo)" clean -qxdff;
+      brew update || brew update;
```
dpo: This is interesting ;)

.travis.yml
```diff
@@ -1,36 +1,36 @@
...
       os: osx
-      osx_image: xcode7
+      osx_image: xcode7.1
       rvm: system
-    - env: OSX=10.9
+    - env: OSX=10.9 HOMEBREW_RUBY=1.8.7
       os: osx
       osx_image: beta-xcode6.2
       rvm: system
+    - os: linux
```
dpo: Aren't we going to want to use LinuxBrew on Linux? It moved to another organization recently.

dpo:
Many thanks for your updates!

.travis.yml
UniqMartin: In the past, `brew update` would rarely (but sometimes still) fail. A second run would often fix that, as is also suggested in the [Troubleshooting](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Troubleshooting.md#check-for-common-issues) guide. This makes sure the bots are mostly immune to this problem and builds don't fail just because of that.

.travis.yml
UniqMartin: We are. Not much has changed so far and this only appears in the `diff` because I reorganized the build system matrix so that all test environments are on the same level. (I haven't used Linuxbrew, so I can't tell if and when this needs some adjustment.)

UniqMartin:
Sorry, I actually managed to break the Travis builds with this PR (they completed successfully, but they didn't build any formulae), but I fixed it now in ef845154682ffc4d615fe2a88cc9b558b5f127f0. Didn't realize that by removing all pre-installed taps I'm also removing the directory this repository is symlinked to …

PR #3215, opened 2016-01-14
updated to cmdstan 2.9.0

robertfeldt:
Changed from version 2.7.0 to 2.9.0. The dirs had changed and I use new stansummary instead of renaming print since from cmdstan 3.0 there will be no print command, only stansummary.

### DIFF
--- a/cmdstan.rb
+++ b/cmdstan.rb
@@ -10,19 +10,19 @@ class Cmdstan < Formula
 
   # tag "math"
 
-  url "https://github.com/stan-dev/cmdstan/releases/download/v2.7.0/cmdstan-2.7.0.tar.gz"
-  sha256 "8f8f119a88327a7cef9829cb593a1d726a65c6bc37cc85ef11f5442add31d9e1"
+  url "https://github.com/stan-dev/cmdstan/releases/download/v2.9.0/cmdstan-2.9.0.tar.gz"
+  sha256 "a9f2858caa5b55576da85ef31b4eae632c97837aa042514242a9aad7ada97121"
 
   depends_on "boost"
   depends_on "eigen"
 
   def install
     system "make", "build"
-    bin.install "bin/print" => "stansummary"
+    bin.install "bin/stansummary"
     bin.install "bin/stanc"
     doc.install "CONTRIBUTING.md", "LICENSE", "README.md", "examples"
-    include.install "stan/src/stan"
-    (include/"stan").install Dir["stan/lib/stan_math_*/stan/*"]
+    include.install "stan_2.9.0/src/stan"
+    (include/"stan").install Dir["stan_2.9.0/lib/stan_math_*/stan/*"]
   end
 
   test do
--- a/cmdstan.rb
+++ b/cmdstan.rb
@@ -19,19 +19,23 @@ class Cmdstan < Formula
   def install
     system "make", "build"
 
-    # symlink the two commands
+    # symlink the two commands. These are the only ones that should be symlinked to main bin dir.
     bin.install "bin/stansummary" 
     bin.install "bin/stanc"
 
     # But we need more from the bin dir...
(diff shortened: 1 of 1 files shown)


robertfeldt:
This actually does not seem to work. More files need to be copied over but I'm not sure which ones. Will debug in coming days but unsure of the solution.

dpo:
```
==> brew audit homebrew/science/cmdstan
==> FAILED
Error: 5 problems in 1 formula
homebrew/science/cmdstan:
 * 23: Trailing whitespace was found
 * 27: Trailing whitespace was found
 * 31: Trailing whitespace was found
 * Non-executables were installed to "/usr/local/Cellar/cmdstan/2.9.0/bin"
The offending files are:
  /usr/local/Cellar/cmdstan/2.9.0/bin/cmdstan
  /usr/local/Cellar/cmdstan/2.9.0/bin/libstanc.a
  /usr/local/Cellar/cmdstan/2.9.0/bin/stan
```
The whitespace issues are easy to fix. The library should be installed to `lib`. I'm not sure what the two other files are. Brew says they're not executables. Do you know what they are?

robertfeldt:
Sorry for the simple mistakes; I tried to fix them in latest commits. 

The non-executables copied to bin dir are needed by make script that each use of cmdstan runs in order to build a model-specific executable. Unfortunately the cmdstan team installs these files to the bin dir and I have not found a way to work around this. The only two commands that really should be symlinked to main homebrew bin dir are stanc and stansummary, but I'm not sure how to avoid symlinking other things that are installed to bin dir.

A lengthy discussion of how I worked with the developer of Stan.jl (which is based on cmdstan) to try and resolve these issues can be found here:

https://github.com/goedman/Stan.jl/issues/18

dpo:
I had a quick look. It might be sufficient to make simple changes to `makefile` before installing it.

robertfeldt:
Ok, I think I did all the necessary edits now. Currently I first use the old make file, then copy files built into bin to lib and then change the makefile. I will do another commit soon that first changes the makefiles so we build to right place in the initial build. This should be cleaner.

robertfeldt:
Ok, I cleaned it up so changes to make files happen first, before the build. Then we can build to lib instead of bin and install the whole lib dir.

````
feldt:/usr/local/Library/Taps/homebrew/homebrew-science$ brew audit --strict cmdstan.rb
==> brew style cmdstan

````

**Comment by robertfeldt** [comment]
Test also works fine on my machine now.

File: cmdstan.rb [hunk 1]
```diff
@@ -8,27 +12,64 @@ class Cmdstan < Formula
...
+    prefix.install "src", "makefile", "make", "stan_2.9.0", "examples"
   end
 
   test do
     system "#{bin}/stanc", "--version"
-    cp doc/"examples/bernoulli/bernoulli.stan", "."
+    cp prefix/"examples/bernoulli/bernoulli.stan", "."
     system "#{bin}/stanc", "bernoulli.stan"
-    system "make", "bernoulli_model.o", "CPPFLAGS=-I#{Formula["eigen"].include/"eigen3"}"
+    system "make", "bernoulli_model.o", "CPPFLAGS=-I#{Formula["eigen"].include/"eigen3"} -I#{prefix + "stan_2.9.0/src"} -I#{prefix + "stan_2.9.0/lib/stan_math_2.9.0"}"
```
dpo: Almost there. Small audit warning: https://travis-ci.org/Homebrew/homebrew-science/jobs/102999669#L2022
You can change this to `-I#{prefix}/stan_2.9.0/src ...`.

dpo:
You can prevent the audit errors by running `brew audit --strict cmdstan` after it's built and installed.

robertfeldt:
Fixed. Audit is fine now on my machine.

dpo:
@BrewTestBot test this please

dpo:
Thank you for your contribution!

robertfeldt:
Thanks for your help and guidance!

PR #3217, opened 2016-01-14
DES 3.12 (new formula).

lifepillar:
DES (Datalog Educational System) is an educational Deductive database with Datalog, Relational Algebra and SQL.

I have written my own shim scripts, because DES must be executed from inside its own directory.

### DIFF
--- a/des.rb
+++ b/des.rb
@@ -1,0 +1,53 @@
+class Des < Formula
+  desc "Datalog Educational System"
+  homepage "http://des.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/des/des/des3.12/DES3.12ACIDE0.17UnixesSWI.zip"
+  version "3.12"
+  sha256 "1cba61cfbe0d64e4ccf425691b650a352e43047bef4622262acd66b8eb3d62d5"
+
+  depends_on :java => "1.6+"
+  depends_on "homebrew/x11/swi-prolog"
+
+  def shim_script; <<-EOS.undent
+      #!/usr/bin/env bash
+      cd #{libexec}
+      swipl des.pl
+    EOS
+  end
+
+  def jar_script(target_jar); <<-EOS.undent
+      #!/bin/bash
+      cd #{libexec}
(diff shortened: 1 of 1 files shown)


des.rb
```diff
@@ -0,0 +1,48 @@
+class Des < Formula
+  desc "Datalog Educational System"
+  homepage "http://des.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/des/des/des3.12/DES3.12ACIDE0.17UnixesSWI.zip"
+  version "3.12"
+  sha256 "1cba61cfbe0d64e4ccf425691b650a352e43047bef4622262acd66b8eb3d62d5"
+
+  depends_on :java => "1.6+"
+  depends_on "swi-prolog"
```
dpo: Should be `homebrew/x11/swi-prolog`.

des.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def shim_script
+    <<-EOS.undent
```
dpo: Please merge lines 11 and 12, as in `caveats`.

des.rb
```diff
@@ -0,0 +1,48 @@
...
+      #!/usr/bin/env bash
+      cd #{libexec}
+      swipl des.pl
+    EOS
+  end
+
+  def jar_script(target_jar)
+    <<-EOS.undent
```
dpo: Same here.

des.rb
```diff
@@ -0,0 +1,48 @@
...
+        #{bin}/des
+
+    The executable to launch the GUI version of DES is
+        #{bin}/des-gui
+    EOS
+  end
+
+  test do
+    system "false"
+  end
```
dpo: This won't do. Is there a useful test to run here? For example, the test could run `#{bin}/des` to output the version number.

lifepillar: I have prepared a test. Is there a way to test it locally?

lifepillar: Never mind, I've read about `brew test`.

PR #3219, opened 2016-01-15
superlu: Matlab interface

dpo:


dpo:
@BrewTestBot test this please

PR #3221, opened 2016-01-15
spades.rb: compile with GCC for OpenMP support

bede:
Further to https://github.com/Homebrew/homebrew-science/pull/3218, a PR with support for multithreading in SPAdes 3.6.2

### DIFF
--- a/spades.rb
+++ b/spades.rb
@@ -1,18 +1,12 @@
 class Spades < Formula
   desc "SPAdes: de novo genome assembly"
   homepage "http://bioinf.spbau.ru/spades/"
-  bottle do
-    cellar :any
-    sha256 "9d69a98a442701f818df56325beeb1ddc8457da5932da172706becf18f6d3fed" => :yosemite
-    sha256 "89ff010f4542bb2f63144c642a125fdedee8367a14410bfd3b30c6b27caeb1c9" => :mavericks
-    sha256 "834d20f5a8b812cdb8c1e2b9c36a8baf35b5f6e5d7f3af910350713a3357ae5e" => :mountain_lion
-  end
 
   # tag "bioinformatics"
   # doi "10.1089/cmb.2012.0021"
 
-  url "http://spades.bioinf.spbau.ru/release3.6.0/SPAdes-3.6.0.tar.gz"
-  sha256 "6c04e55d068ad3e5b44c862cb031d41fcd433ab97e5dedadebcc5f785cde13a9"
+  url "http://spades.bioinf.spbau.ru/release3.6.2/SPAdes-3.6.2.tar.gz"
+  sha256 "20897e2707623ee1033d3c88fefc42771fa4bbaafaf0f95642991d83b361eb5f"
 
   depends_on "cmake" => :build
 
--- a/spades.rb
+++ b/spades.rb
@@ -1,6 +1,12 @@
 class Spades < Formula
   desc "SPAdes: de novo genome assembly"
   homepage "http://bioinf.spbau.ru/spades/"
+  bottle do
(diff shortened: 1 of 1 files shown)


spades.rb
```diff
@@ -1,21 +1,25 @@
 class Spades < Formula
   desc "SPAdes: de novo genome assembly"
   homepage "http://bioinf.spbau.ru/spades/"
-  bottle do
-    cellar :any
-    sha256 "9d69a98a442701f818df56325beeb1ddc8457da5932da172706becf18f6d3fed" => :yosemite
-    sha256 "89ff010f4542bb2f63144c642a125fdedee8367a14410bfd3b30c6b27caeb1c9" => :mavericks
-    sha256 "834d20f5a8b812cdb8c1e2b9c36a8baf35b5f6e5d7f3af910350713a3357ae5e" => :mountain_lion
-  end
```
dpo: Please don't alter the `bottle` section. The bot will update it automatically.

spades.rb
```diff
@@ -1,21 +1,25 @@
...
 
-  url "http://spades.bioinf.spbau.ru/release3.6.0/SPAdes-3.6.0.tar.gz"
-  sha256 "6c04e55d068ad3e5b44c862cb031d41fcd433ab97e5dedadebcc5f785cde13a9"
+  url "http://spades.bioinf.spbau.ru/release3.6.2/SPAdes-3.6.2.tar.gz"
+  sha256 "20897e2707623ee1033d3c88fefc42771fa4bbaafaf0f95642991d83b361eb5f"
 
   depends_on "cmake" => :build
 
+  fails_with :clang do
+    build 700
```
dpo: You may want to avoid specifying a build, to avoid using clang in the future too.

dpo:
Does this supersede #3218?

bede:
Thank you for your comments. Yes it does supersede it. I'm happy to revise
and PR.

bede:
@dpo Revised in 0e658ea197bc520991e50008e0ff3085796ab554. Please review after Travis completes.

dpo:
Is it that `spades` does not build at all with clang, or is it that it's nice to use gcc because it has OpenMP support?

Also http://spades.bioinf.spbau.ru/release3.6.2/manual.html#sec2.3 mentions g++ >= 4.7, not 4.3.

bede:
@dpo 4.3 built for me although I could be mistaken. Will check… Better to revise to 4.7 for consistency with docs?

Yes, I believe clang was never targeted, which is hardly surprising given the uselessness of a single threaded genome assembler!

dpo:
Ok. Because there are plans for OpenMP support in clang in the near future, you'd be better off using

```
needs :openmp

fails_with :gcc => "4.7" do
  cause "..."
end
```
and removing the `fails_with :clang` bit.

bede:
>  Because there are plans for OpenMP support in clang in the near future
This was my rationale behind initially specifying a `fails_with` clause, following examples in the cookbook / other GCC-only homebrew-science formulae. I'll make the changes you've recommended.

dpo:
I see. It makes sense. Over time, the `needs :openmp` will probably be a more robust way to achieve the same result.

bede:
Actually, @dpo, surely the following code will simply block GCC 4.7 releases? That's not what we want at all. I think what I suggested in the beginning was better.

```
fails_with :gcc => "4.7" do
  cause "..."
end
```
According to https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#compiler-selection

dpo:
Thanks for your contribution and patience!

PR #3222, opened 2016-01-15
root: change download url from ftp to https

veprbl:


dpo:
The test tries to write in `prefix`, which isn't permitted. Would you mind updating it so it runs in `testpath`?
```
==> brew test --verbose homebrew/science/root
==> FAILED
Testing homebrew/science/root
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160115-73184-181bkm0-0.sb /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-science/root.rb --verbose
==> make -C /usr/local/Cellar/root/5.34.34/test/ hsimple
clang++  -O2 -pipe -Wall -W -Woverloaded-virtual -pthread -stdlib=libc++ -m64 -I/usr/local/Cellar/root/5.34.34/include/root -c hsimple.cxx
error: unable to open output file 'hsimple.o': 'Error opening output file 'hsimple.o': Operation not permitted'
1 error generated.
```
Thanks!

root.rb
```diff
@@ -1,9 +1,9 @@
 class Root < Formula
-  desc "Root object oriented framework for large scale data analysis"
+  desc "Object oriented framework for large scale data analysis"
   homepage "http://root.cern.ch"
   version "5.34.34"
+  url "https://root.cern.ch/download/root_v#{version}.source.tar.gz"
   sha256 "8c1faf893ed3b279f3500368b3dcd2087352020a69d3055c4d36726e7f6acd58"
-  url "ftp://root.cern.ch/root/root_v#{version}.source.tar.gz"
   head "https://github.com/root-mirror/root.git", :branch => "v5-34-00-patches"
```
dpo: Now we just need a revision number here: `revision 1`.

veprbl: Right

veprbl:
Thank you!

PR #3213, opened 2016-01-14
openimageio: add --with-python3 and update to 1.6.9.

jaguarondi:
Add the `--with-python3` option to build and install the python3
bindings.

Update to 1.6.9 release.

PR #3225, opened 2016-01-16
field3d: bump for openimageio

jaguarondi:
Required to build openimageio, see PR #3213


jaguarondi:
@dpo : should I also bump hdf5 ?

dpo:
No need. The issue is with `openimageio`, but that will be rebottled by your other PR.

PR #3228, opened 2016-01-16
vips: 8.2.1 and jpeg-turbo support / nip2: 8.2

felixbuenemann:
vips:
* Update to newest upstream release 8.2.1
* Include bilinear interpolation rounding bugfix from upcoming 8.2.2
* Add support for compiling against libjpeg-turbo

nip2:
* Update to newest upstream release 8.2

The nip2 formula depends on vips and was rather outdated, so I updated it in the same commit.


felixbuenemann:
Not sure if the CI can properly handle multiple formulas in a single commit, just tell me if you want me to split it into to PRs.

felixbuenemann:
Will submit two PRs for vips and nip2.

PR #3229, opened 2016-01-17
root: reuse test from root6 formula

veprbl:
Test from root6.rb is more robust because it calls thisroot.sh to ensure that the right ROOT version is selected and environment is set up. It also checks that the C++ interpreter works.


dpo:
Would you mind doing this as part of #3222?  Sorry for the trouble. I should have been clearer.

PR #3236, opened 2016-01-19
poretools: update ggplot2 URL

dfornika:
ggplot2 v1.0.1 has been moved to the CRAN Archive.


dpo:

PR #3255, opened 2016-01-25
edirect: formula now installs current version 3.60

ozagordi:
The file url points to the most recent version (3.60 as of now), for which the shasum was updated.

### DIFF
--- a/edirect.rb
+++ b/edirect.rb
@@ -3,8 +3,8 @@ class Edirect < Formula
   # tag "bioinformatics"
 
   url "ftp://ftp.ncbi.nlm.nih.gov/entrez/entrezdirect/edirect.tar.gz"
-  sha256 "694673fa248677eac4dcb67ed9cd748e5b0ea80fa07f563603b39469055e6e6a"
-  version "2.00"
+  sha256 "2d108d780cef39bc4fdfd265d97ebaafe59ec937667061f54274f17a40a1a74f"
+  version "3.60"
 
   bottle do
     cellar :any


dpo:
```
==> brew audit homebrew/science/edirect
==> FAILED
Error: 1 problem in 1 formula
homebrew/science/edirect:
 * Non-executables were installed to "/usr/local/Cellar/edirect/3.60/bin"
The offending files are:
  /usr/local/Cellar/edirect/3.60/bin/xtract.go
```
Does that file have to be in `bin`?

ozagordi:
That file needs to be compiled with `go build xtract.go`, meaning that a go compiler is needed. I managed to run `brew install go`, then manually go to the directory and compile it, but I don't know how to include this in the formula.

tdsmith:
setup.sh does this:
```bash
cd "$DIR"
osname=`uname -s`
cputype=`uname -m`
case "$osname-$cputype" in
    Linux-x86_64 | Darwin-x86_64 )
        ./ftp-cp ftp.ncbi.nlm.nih.gov /entrez/entrezdirect xtract."$osname"
    ;;
esac
if [ -f xtract."$osname" ]
then
    chmod +x xtract."$osname"
else
    echo "Unable to download a prebuilt xtract executable; attempting to"
    echo "build one from xtract.go.  A Perl fallback is also available, and"
    echo "will be used if necessary, so please disregard any errors below."
    go build -o xtract."$osname" xtract.go
fi
```
which seems bad, if only because it's transferring binaries over an insecure connection without checksumming

tdsmith:
Oh, but we never run setup.sh.

Let's just delete it: `rm "xtract.go"`

PR #3263, opened 2016-01-27
Update formula for Cantera

speth:
New minor version, plus a fix for the Matlab toolbox.

### DIFF
--- a/cantera.rb
+++ b/cantera.rb
@@ -39,11 +39,27 @@ class Cantera < Formula
                   "f90_interface=n"]
 
     matlab_path = ARGV.value("with-matlab")
-    build_args << "matlab_path=" + matlab_path if matlab_path
+    if matlab_path
+      build_args << "matlab_path=" + matlab_path
+      # Matlab doesn't play nice with system Sundials installation
+      if build.head?
+        build_args << "system_sundials=n" # Cantera 2.3+
+      else
+        build_args << "use_sundials=n" # Cantera 2.2.x
+      end
+    end
+
     build_args << "python3_package=" + (build.with?("python3") ? "y" : "n")
 
     scons "build", *build_args
-    scons "test" if build.with? "check"
+    if build.with? "check"
+      if not matlab_path
+        scons "test"
+      else
+        # Matlab test stalls when run through Homebrew, so run other sub-tests explicitly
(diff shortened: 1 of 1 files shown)


cantera.rb
```diff
@@ -3,7 +3,6 @@ class Cantera < Formula
   url "https://github.com/Cantera/cantera/releases/download/v2.2.0/cantera-2.2.0.tar.gz"
   sha256 "306c218500eaabdf1e920601348d2b3acc1fb66b02eea842d98b3fbb41ebbc78"
   head "https://github.com/cantera/cantera.git"
-  revision 1
```
dpo: You'll want to increase the revision number to 2.

dpo:
For some reason the python tests cause a failure on 10.10 and 10.11:
```
*****************************
***    Testing Summary    ***
*****************************

Tests passed: 296
Up-to-date tests skipped: 0
Tests failed: 100
Failed tests:
    - python2 ***no results for entire test suite***

*****************************
scons: *** [test_problems/finish_tests] One or more tests failed.
scons: building terminated because of errors.
```

speth:
I'm looking at that now. Trying to figure out what's causing it -- I haven't been able to replicate the failure locally yet. I don't think there's anything in 2.2.0 -> 2.2.1 that should be able to cause this kind of failure, so I'm currently trying to get the test bot to run the tests for 2.2.0, but it's taking its sweet time to get around to them.

cantera.rb
speth: Okay then. I had removed the revision because the previous error message was "revision should be removed".

speth:
The previous test failure is now fixed, but I don't understand the failure to build on 10.10 (error building Sundials, which is a dependency).

dpo:
Looks all good now. Would you mind squashing your commits?

speth:
I can if you really want, but it feels wrong to me, at least as far as squashing 6f4711c and cb189b9 goes. The fix for the Matlab toolbox is independent of the new release.

dpo:
I see. It's ok to have two commits since the matlab fix wouldn't yield new bottles anyways. Thanks.

speth:
Ok, I can squash the Cython change into the update for the new release, since it doesn't have any effect on the outcome of the installation either way.

dpo:
@BrewTestBot test this please

PR #3264, opened 2016-01-27
nixio 1.1.0 (new formula)

gicmo:
The NIX project aims to develop standardized methods and models for storing neuroscience data together with their metadata in one common, open file format based. The default binary format current is HDF5. It started as an initiative of the Electrophysiology Task Force which is part of the INCF Datasharing Program. See the [homepage](http://www.g-node.org/nix) for details. We are also uploading to [NeuroDebian](http://neuro.debian.net) soon (and possibly Debian proper later).

There is a [GUI file viewer](https://github.com/bendalab/NixView) and an open source data recording tool, called [relacs](https://github.com/relacs/relacs), that use it already; besides its use in scientific analysis scripts in various labs.


nix.rb
```diff
@@ -0,0 +1,33 @@
...
+  url "https://github.com/G-Node/nix/archive/1.1.0.tar.gz"
+  sha256 "d607d96117621e5cc563002c4c161913d95db36eeb7c4e6f51afca5f6b788fcc"
+
+  head "https://github.com/G-Node/nix.git"
+
+  option "without-test", "Skip compile-time tests (not recommended)"
+
+  depends_on "cmake" => :build
+  depends_on "boost" => "c++11"
+  depends_on "homebrew/science/hdf5"
```
dpo: There's no need to fully qualify dependencies in this tap.

nix.rb
```diff
@@ -0,0 +1,33 @@
...
+  depends_on "cppunit"
+
+  needs :cxx11
+
+  def install
+    ENV.cxx11
+
+    cmake_args = std_cmake_args
+    cmake_args << "-DBUILD_TESTING=OFF"
```
dpo: Sorry for the silly question, but don't you want to set this to `ON` when building `--with-test`?

gicmo:
@dpo Thanks for the comments. Should all be fixed (indeed found another little bug, i.e. it was `*std_cmake_args` not `*cmake_args` in `system "cmake", ".", ...`).

tdsmith:
I don't feel good accepting a package called "nix" since that competes with the better-known package manager; is there something else we can call this?

nix.rb
```diff
@@ -0,0 +1,33 @@
+class Nix < Formula
+  desc "C++ library for the NIX scientific data format and model"
+  homepage "http://www.g-node.org/nix"
+  url "https://github.com/G-Node/nix/archive/1.1.0.tar.gz"
+  sha256 "d607d96117621e5cc563002c4c161913d95db36eeb7c4e6f51afca5f6b788fcc"
+
+  head "https://github.com/G-Node/nix.git"
+
+  option "without-test", "Skip compile-time tests (not recommended)"
```
tdsmith: You can remove this option and always run tests

nix.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  depends_on "cmake" => :build
+  depends_on "boost" => "c++11"
```
tdsmith: This isn't necessary on recent OS X releases.

```
if OS.mac? and MacOS.version < :mavericks
  depends_on "boost" => "c++11"
else
  depends_on "boost"
end
```

nix.rb
```diff
@@ -0,0 +1,33 @@
...
+    cmake_args << "-DBUILD_TESTING=OFF" if build.without? "test"
+
+    system "cmake", ".", *cmake_args
+    system "make"
+    system "make", "test" if build.with? "test"
+    system "make", "install"
+  end
+
+  test do
+    system bin/"nix-tool", "--help"
```
tdsmith: Can this do a real test that exercises the tool in some way? Downloading a smallish sample data set as a resource is fine.

nix.rb
gicmo: Fair enough

nix.rb
gicmo: Ha! I was wondering about that and was very close to doing the exact change above but then I looked at some other formulas and the boost formula itself and was left unsure. I am glad that is clarified. ;)

nix.rb
gicmo: Sounds like a good idea too.

gicmo:
About the package name: I understand (we have similar issues for our python bindings and PyPI). After quick discussion with some of the team I change the package name to `nixio`. Would that work?

tdsmith:
Sounds great; thanks very much.

tdsmith:
Thanks for your pull request! :tada: The new test is great, too.

Homebrew thrives because of contributions like this one!

PR #3265, opened 2016-01-27
libmatheval 1.1.11 (new formula)

khuston:
I made a formula for libmatheval, using the latest version on the GNU
webpage. GUILE is a dependency, but libmatheval has not been updated
after changes from GUILE 1.8 -> 2.0, so it still requires version 1.8.
The added formula includes homebrew/versions/guile18 as a dependency to
address this.


libmatheval.rb
```diff
@@ -0,0 +1,18 @@
+class Libmatheval < Formula
+  desc "Symbolic math parsing and evaluation"
+  homepage "https://www.gnu.org/software/libmatheval/"
+  url "http://ftpmirror.gnu.org/gnu/libmatheval/libmatheval-1.1.11.tar.gz"
+  sha256 "474852d6715ddc3b6969e28de5e1a5fbaff9e8ece6aebb9dc1cc63e9e88e89ab"
+  depends_on "homebrew/versions/guile18"
```
dpo: Please insert a blank line between the header and dependencies.

libmatheval.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "ls", "#{lib}/libmatheval.a"
+    system "ls", "#{lib}/libmatheval.dylib"
```
dpo: There must be a more meaningful test...

khuston: I don't really use libmatheval directly -- just as a dependency, but I wrote a simple test that uses the library and will push soon.

khuston:
It passes builds for OS X 10.10 and 10.11 but fails on 10.9 while installing a dependency, /homebrew/versions/guile18

dpo:
@BrewTestBot test this please

dpo:
The build completes in 6 minutes on 10.10 and 10.11, but fails after one hour on 10.9...

dpo:
@BrewTestBot test this please

dpo:
The problem is with a guile command never returning on 10.9 but I've absolutely no knowledge of it.

dpo:
Closing this. Please open a new pr if still interested and a resolution has come about.

PR #3266, opened 2016-01-29
sample 1.0.2 (new formula)

alexpreynolds:
This tool performs reservoir sampling (Vitter, "Random sampling with a reservoir"; cf. [http://dx.doi.org/10.1145/3147.3165](http://dx.doi.org/10.1145/3147.3165) and also: [http://en.wikipedia.org/wiki/Reservoir_sampling](http://en.wikipedia.org/wiki/Reservoir_sampling)) on very large text files that are delimited by newline characters (such as common genomic data formats like BED, SAM, VCF, etc.). Sampling can be done with or without replacement. The approach used in this application reduces the typical memory usage limitations in GNU `shuf` by instead storing a pool of byte offsets to the start of each line, instead of the line elements themselves, thus allowing considerably larger sample sizes.


sample.rb
```diff
@@ -0,0 +1,32 @@
+class Sample < Formula
+  homepage "https://github.com/alexpreynolds/sample"
```
dpo: Please add a `desc` field.

sample.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  head "https://github.com/alexpreynolds/sample.git"
+  version "1.0.2"
+
+  bottle do
+    cellar :any_skip_relocation
+    sha256 "4c5a2dfda0ea2658bc505ef5bdf47a199b810fe36f20a8fffe275395afc1cc80" => :el_capitan
+    sha256 "5faad7689e3ced8019f9f9833f629f581237922753fa8ff3785f8e23c5f48eab" => :yosemite
+    sha256 "18f3d8dbcd1bb202eb913c7238e5249dcf0ea98996306ecedbe090d2d56e5e91" => :mavericks
+  end
```
dpo: How are there bottles if this is a new formula?

dpo:
I'm not sure what's going on here:
```
==> make build
mkdir -p objects && /usr/bin/clang -Wall -Wextra -pedantic -std=c99 -D__STDC_CONSTANT_MACROS -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE=1 -O3 -c src/sample-library/mt19937.c -o objects/mt19937.o -iquote./include
ar rcs /Users/brew/Jenkins/science/version/mavericks/sample-library.a objects/mt19937.o
ar: /Users/brew/Jenkins/science/version/mavericks/sample-library.a: Operation not permitted
make: *** [sample-library] Error 1
```

sample.rb
```diff
@@ -0,0 +1,32 @@
...
+    ENV.O3
+    ENV.delete("CFLAGS")
+    ENV.delete("CXXFLAGS")
+    system "make", "build"
+    bin.install "sample"
+    doc.install %w[LICENSE README.md]
+  end
+
+  test do
+    system "make", "check"
```
dpo: This won't work for a couple of reasons. One is that (I think) the Makefiles won't be installed to the prefix (they shouldn't). Another is that brew won't allow any write operations to take place in the prefix after the install is complete.

tdsmith:
The bad paths are from use of `$(PWD)` vs `$(CURDIR)` in the Makefile, I think: https://www.gnu.org/software/make/manual/html_node/Recursion.html

@dpo Do you think this is notable enough to include?

sample.rb
alexpreynolds: Added.

sample.rb
alexpreynolds: I removed the block for bottles.

sample.rb
alexpreynolds: I'm not sure what's wrong, either. I don't run into any problems building the formula from source locally.

sample.rb
```diff
@@ -0,0 +1,26 @@
+class Sample < Formula
+  desc "This tool performs reservoir sampling (Vitter, 'Random sampling with a reservoir'; cf. http://dx.doi.org/10.1145/3147.3165 and also: http://en.wikipedia.org/wiki/Reservoir_sampling) with or without replacement on very large text files that are delimited by newline characters (such as common genomic data formats like BED, SAM, VCF, etc.) and scales to larger inputs than GNU shuf by sampling line offsets, instead of whole lines."
```
dpo: Sorry but that's really too long ;-)

alexpreynolds: Sorry about that. I shortened the description.

dpo:
@tdsmith I'm not sure. Random sampling is a noble pursuit but I can't tell how useful this one actually is.

sample.rb
dpo: Will this be updated? You can run `make check` in `install`. The best would be to add a meaningful test though, such as running the code on sample data.

alexpreynolds: I updated the project `makefile` and my repository's Travis CI tests pass. However, the Travis CI tests that are performed on the Homebrew side still do not. I'm looking at the Homebrew-side Travis CI logs and the following error comes up:

```
make: *** No rule to make target `check'.  Stop.
Error: homebrew/science/sample: failed
```

This is confusing to me, because the project `makefile` contains a target called `check` that runs three tests (that pass):

```
$ make check
mkdir -p objects && cc -Wall -Wextra -pedantic -std=c99 -D__STDC_CONSTANT_MACROS -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE=1 -O3 -c src/sample-library/mt19937.c -o objects/mt19937.o -iquote./include
ar rcs /Users/alexpreynolds/Developer/Github/sample/sample-library.a objects/mt19937.o
cc -Wall -Wextra -pedantic -std=c99 -D__STDC_CONSTANT_MACROS -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE=1 -O3 -c src/bin/sample.c -o objects/sample.o -iquote./include
cc -Wall -Wextra -pedantic -std=c99 -D__STDC_CONSTANT_MACROS -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE=1 -O3 objects/sample.o -o sample /Users/alexpreynolds/Developer/Github/sample/sample-library.a
/Users/alexpreynolds/Developer/Github/sample/sample README.md -d 123 | diff - /Users/alexpreynolds/Developer/Github/sample/test/README.md.seed123.txt > /dev/null || (echo "check: sample test failed on seed 123" && exit 1)
/Users/alexpreynolds/Developer/Github/sample/sample README.md -d 234 | diff - /Users/alexpreynolds/Developer/Github/sample/test/README.md.seed234.txt > /dev/null || (echo "check: sample test failed on seed 234" && exit 1)
/Users/alexpreynolds/Developer/Github/sample/sample README.md -d 987 | diff - /Users/alexpreynolds/Developer/Github/sample/test/README.md.seed987.txt > /dev/null || (echo "check: sample test failed on seed 987" && exit 1)
sample tests passed
```

Is there anything about the Homebrew-side log or what changes are made to the Travis CI build procedure that could suggest what the problem might be in more detail, which might help me make any necessary changes?

dpo: Remove line 11 if possible, and move `system "make", "check"` just after `system "make", "build"`. You can't execute `make check` in the `test` method; the `makefile` isn't available anymore at that point.

alexpreynolds: Thanks. I removed line 11 and replaced the `system "make", "check"` call with a more direct test. This passed a local audit check, but I'll see if this gets past the Homebrew/Travis CI checks.

alexpreynolds:
I added an API token via the [documentation](https://gist.github.com/christopheranderton/8644743), but the Travis CI test on the Homebrew side still fails.

iMichka:

Issue #3259, opened 2016-01-26
graph-tool: symbol binding failed


itcarroll:
Python package imports okay, by calling `graph_draw` gives a runtime error message:

```
dyld: lazy symbol binding failed: Symbol not found: __ZN10graph_tool6detail15action_dispatchIN5boost6lambda14lambda_functorINS3_19lambda_functor_baseINS3_12other_actionINS3_17assignment_actionEEENS2_6tuples5tupleINS4_INS3_8identityIRmEEEENS4_INS5_INS3_6actionILi2ENS3_15function_actionILi2EmEEEENSA_IKNS_15HardNumVerticesEKNS4_INS3_11placeholderILi1EEEEENS9_9null_typeESP_SP_SP_SP_SP_SP_SP_EEEEEESP_SP_SP_SP_SP_SP_SP_SP_EEEEEEN4mpl_5bool_ILb0EEEJNS0_15all_graph_viewsEEEC1ERKS10_
  Referenced from: /usr/local/lib/python3.5/site-packages/graph_tool/libgraph_tool_core.so
  Expected in: flat namespace

dyld: Symbol not found: __ZN10graph_tool6detail15action_dispatchIN5boost6lambda14lambda_functorINS3_19lambda_functor_baseINS3_12other_actionINS3_17assignment_actionEEENS2_6tuples5tupleINS4_INS3_8identityIRmEEEENS4_INS5_INS3_6actionILi2ENS3_15function_actionILi2EmEEEENSA_IKNS_15HardNumVerticesEKNS4_INS3_11placeholderILi1EEEEENS9_9null_typeESP_SP_SP_SP_SP_SP_SP_EEEEEESP_SP_SP_SP_SP_SP_SP_SP_EEEEEEN4mpl_5bool_ILb0EEEJNS0_15all_graph_viewsEEEC1ERKS10_
  Referenced from: /usr/local/lib/python3.5/site-packages/graph_tool/libgraph_tool_core.so
  Expected in: flat namespace
```

I uninstalled all the dependencies then built up from source with:

```
brew install --build-from-source boost
brew install --build-from-source boost-python --with-python3 --without-python
brew install --build-from-source graph-tool --HEAD --with-python3 --without-python
```

The graph-tool install issued the warning

```
Warning: homebrew/science/graph-tool dependency gcc was built with a different C++ standard
library (libstdc++ from clang). This may cause problems at runtime.
```

[...]

PR #3267, opened 2016-01-29
graph-tool: patch for boost 1.60

mistydemeo:
Refs #3259.


tdsmith:
:tada:

itcarroll:
got two identical hashes in the there...
betcha the first of the dupes should be
456247297df4a7db7ffa696bbd6cee3ab3a6ffa7e20dbd1a2058336efa33a782

mistydemeo:
Arg, good catch. Fixed and repushed.

PR #3271, opened 2016-01-30
atompaw 4.0.0.12 (new formula)

mtorrent:
Add formula for atompaw software, a small tool to produce atomic data for 1st-principle material science codes.


atompaw.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    system "./configure", *args
+    system "make"
+
+    if build.with? "test"
+      mkdir("example/F/test") unless File.directory?("example/F/test")
+      cd("example/F/test") do
+        system "../../../src/atompaw <../lda/F.input 2>&1 | tee make-test.log"
+        ohai `grep "</paw_dataset>" "F.LDA-PW-paw.xml" | tee make-test.log`.chomp
```
dpo: It would be great to install `F.input` to `pkgshare/"data"` (say) and to run this test in the `test` method.

atompaw.rb
```diff
@@ -0,0 +1,36 @@
...
+      args << "--with-libxc-incs=-I#{Formula["libxc"].opt_include}"
+      args << "--with-libxc-libs=-L#{Formula["libxc"].opt_lib} -lxc -lxcf90"
+    end
+
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  test do
+    atompaw_tests=%W[#{doc}/example/F/lda/F.input
```
dpo: Please use spaces around `=`.

PR #3281, opened 2016-02-03
geant 10.02

henryiii:
The GEANT4 formula is old and doesn't work with the brew version of CLHEP, so I'm proposing a change. If this is better done another way, feel free to try that instead, just trying to help.

This is an update for 10.02. I had to remove external CLHEP, since GEANT4 can't find it (on two different computers). It now uses internal CLHEP. If someone can get 10.02 working with external CLHEP, that would be great. Don't have a bottle, either. This works on El Cap. Added options for multithreading, and usolids (experimental).

Note that the current bottles are linked to an old version of CLHEP, so they don't work at all.


geant.rb
```diff
@@ -1,24 +1,20 @@
...
-  url "http://geant4.cern.ch/support/source/geant4.10.01.tar.gz"
-  sha256 "d37400e96423fedfbf8dbe1f49e2ef0367317c3893ad99f28eed06bf97e1feb7"
+  url "http://geant4.cern.ch/support/source/geant4.10.02.tar.gz"
+  sha256 "633ca2df88b03ba818c7eb09ba21d0667a94e342f7d6d6ff3c695d83583b8aa3"
 
-  bottle do
-    sha256 "1bed04e6dd00d665eab93c3625584790f1ad0a4c2a7d8f5fcd226c99833ebc08" => :yosemite
-    sha256 "c9ef2ae73e124f50645740808f5174c408c903b844c4a845b551313869a70649" => :mavericks
-    sha256 "0c36ec55ff66582e12e1c3e09606d0e4336d82f5a3ee7e956c80d33123569e48" => :mountain_lion
-  end
```
dpo: Please don't touch the `bottle` section. It will be updated automatically.

geant.rb
```diff
@@ -1,24 +1,20 @@
...
-
-  option "with-g3tog4", "Use G3toG4 Library"
-  option "with-gdml", "Use GDML"
-  option "with-notimeout", "Set notimeout in installing data"
+  option 'with-g3tog4', 'Use G3toG4 Library'
```
dpo: Please use double quotes everywhere. You can run `brew audit --strict` on your formula to fix audits before submitting.

geant.rb
```diff
@@ -1,24 +1,20 @@
...
+  option 'with-gdml', 'Use GDML'
+  option 'with-notimeout', 'Set notimeout in installing data'
+  option 'with-usolids', 'Use USolids (experimental)'
+  option 'without-multithreaded', 'Build without multithreading enabled'
```
dpo: Does this relate to OpenMP?

henryiii: Don't think it internally uses OpenMP, but I think I should make it an option rather than the default, since that's how it still works in cmake.

geant.rb
henryiii: Thanks, didn't know that. I'll fix it.

henryiii:
I've made multithreading default to off, since it is that way in Geant's build. I was thinking that the main reason to not have multithreading on was due to different systems, but there might be some other cases where one would not want it on. I was probably biased because I always develop in Geant with it on.

(The toolkit generates and handles tracks in separate threads with multithreading on. 95% sure it does not use OpenMP)

PR #3286, opened 2016-02-05
VTK: 7.0

jenshnielsen:
Upgrade VTK to 7.0 and add support for installing python3 binding. As far as I can tell there is no easy way to build both python 2 and 3 bindings so these are mutually exclusive at this stage. 

Im not sure is other formulas should be have a revision bump to build against this?

Brew audit complains that 

```
homebrew/science/vtk:
 * python modules have explicit framework links
```

But that is nothing new as far as I can see:

### DIFF
--- a/vtk.rb
+++ b/vtk.rb
@@ -31,6 +31,7 @@ class Vtk < Formula
   depends_on "qt5" => :optional
 
   depends_on :python => :recommended if MacOS.version <= :snow_leopard
+  depends_on :python3 => :optional
 
   depends_on "boost" => :recommended
   depends_on "fontconfig" => :recommended
@@ -48,6 +49,16 @@ class Vtk < Formula
     elsif build.with? "qt5"
       depends_on "sip"
       depends_on "pyqt5" => ["with-python", "without-python3"]
+    end
+  end
+
+  if build.with? "python3"
+    if build.with? "qt"
+      depends_on "sip"
+      depends_on "pyqt" => ["with-python3"]
+    elsif build.with? "qt5"
+      depends_on "sip"
+      depends_on "pyqt5" => ["with-python3"]
     end
   end
 
@@ -107,7 +118,7 @@ class Vtk < Formula
     ENV.cxx11 if build.cxx11?
 
     mkdir "build" do
-      if build.with? "python"
+      if build.with?("python") && build.without?("python3")
         args << "-DVTK_WRAP_PYTHON=ON"
         # CMake picks up the system"s python dylib, even if we have a brewed one.
         args << "-DPYTHON_LIBRARY='#{`python-config --prefix`.chomp}/lib/libpython2.7.dylib'"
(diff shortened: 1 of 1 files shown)


vtk.rb
```diff
@@ -49,6 +50,16 @@ class Vtk < Formula
     end
   end
 
+  if build.with? "python3"
+    if build.with? "qt"
+      depends_on "sip"
+      depends_on "pyqt" => ["with-python3"]
```
iMichka: So this means pyqt/pyqt5 will be built for python2 AND 3 ? Should we not use:
```depends_on "pyqt5" => ["with-python3", "without-python"]```

Not sure what is better

Another question: should the sip lines not be:
```depends_on "sip" => "with-python3"```

(Because it is defined as such in pyqt.rb and pyqt5.rb when using python3)

File: vtk.rb [hunk 2]
```diff
@@ -116,6 +127,22 @@ def install
...
+        args << "-DVTK_WRAP_PYTHON=ON"
+        args << "-DPYTHON_EXECUTABLE=/usr/local/bin/python3"
+        args << "-DPYTHON_INCLUDE_DIR='#{`python3-config --prefix`.chomp}/include/python3.5m'"
+        # CMake picks up the system"s python dylib, even if we have a brewed one.
+        args << "-DPYTHON_LIBRARY='#{`python3-config --prefix`.chomp}/lib/libpython3.5.dylib'"
+        # Set the prefix for the python bindings to the Cellar
+        args << "-DVTK_INSTALL_PYTHON_MODULE_DIR='#{lib}/python3.5/site-packages'"
+
+        if build.with? "qt"
+          args << "-DVTK_WRAP_PYTHON_SIP=ON"
```
iMichka: Does the qt5 case need to be covered here ?

iMichka:
> Brew audit complains that

>homebrew/science/vtk:
> * python modules have explicit framework links

>But that is nothing new as far as I can see:

This is a know problem, there was a discussion about this 2-3 days ago on the VTK mailing list. I gave them some instructions on how to fix this (we implemented this for ITK for example). I have no time to look a this for the moment, I hope somebody will have to write a patch for this soon (I'll keep an eye on this for the next VTK release though).

vtk.rb
```diff
@@ -49,6 +50,16 @@ class Vtk < Formula
     end
   end
 
+  if build.with? "python3"
+    if build.with? "qt"
+      depends_on "sip" => ["with-python3", "without-python"]
```
dpo: It seems like this line could be moved out of the inner `if`.

dpo: Oh, sorry, maybe not. Is `sip` a dependency only if `qt` or `qt5` is selected?

jenshnielsen: Yes sip is the python wrapper generator for pyqt4/5. Im not actually sure why we need to state it as an explicit dependency here is's already a dependency of pyqt4/5 but I just mirrored the python2 version

dpo:
Could you please rebase? I made a small change in `vtk.rb` to build against system Python when available.

jenshnielsen:
No problem, I have rebased

dpo:
Thank you!

PR #3288, opened 2016-02-06
insighttoolkit 4.9.0

iMichka:
VTK 7 is compatible with python 3


iMichka:
Please merge this only when VTK has been upgraded to VTK 7; see #3286

iMichka:
I update the formula:
- removed ```revision 1``. Brew audit told me to do this. No clue why, the documentations also does not explain what revision number a version bump should have. Maybe ```revision 0``` ?
- added the VTK dependency even no python bindings are used.

dpo:
We simply remove the `revision` when upgrading to a new version.

iMichka:
statismo was looking for ITK 4.8.1 dylibs. I updated the statismo revision to force a rebuild of statismo too.

iMichka:
Ok this is still failing due to statismo. Don't know if I should have touched that revision number.

Seems that the step ```brew uses --recursive homebrew/science/statismo``` is triggered, but ```brew uninstall``` has already been run, which means the insighttoolkit bottle has been deleted. It could be a bug in the build process ?

dpo:
@BrewTestBot test this please

iMichka:
I moved the statismo revision bump to a separate PR: #3344

dpo:
@BrewTestBot test this please

iMichka:
Thanks !

PR #3292, opened 2016-02-08
ncview 2.1.6

tomyun:


dpo:
```
==> brew test --verbose homebrew/science/ncview
==> FAILED
Testing homebrew/science/ncview
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160208-13851-1hi7ejj-0.sb /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-science/ncview.rb --verbose
==> /usr/local/Cellar/ncview/2.1.6/bin/ncview -c 2>&1
Error: homebrew/science/ncview: failed
<0> expected but was
<1>.
```

tomyun:
That's interesting. The same command works fine on my local machine running the latest El Capitan.

```
$ brew test --verbose homebrew/science/ncview
Testing homebrew/science/ncview
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160208-45822-1eh997h.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-science/ncview.rb --verbose
==> /usr/local/Cellar/ncview/2.1.6/bin/ncview -c 2>&1
```

Do you have any idea on this? Or I might remove `test` which I just added for this revision.

tseemann:
@tomyun  The command isn't failing, but it is returning exit code 1 instead of 0.

Change `shell_output("#{bin}/ncview -c 2>&1")` to `shell_output("#{bin}/ncview -c 2>&1", 1)`

tomyun:
@tseemann Actually it returns exit code 0 on my local machine. `shell_output("#{bin}/ncview -c 2>&1", 1)` fails like this:

```
$ brew test ncview
Testing homebrew/science/ncview
==> Using the sandbox
==> /usr/local/Cellar/ncview/2.1.6/bin/ncview -c 2>&1
Error: homebrew/science/ncview: failed
<1> expected but was
<0>.
```

tseemann:
@tomyun This happens to me too sometimes. On local machine, it's 0, but in @BrewTestBot it is 1. I have _no idea_ why this happens. Changing it will make the test pass ...

@sjackman - any ideas how this could happen?

sjackman:
I've only seen this once, and it took me ages to figure it out. I forget which formula it was. In that case, it was because the program worked when run at an interactive terminal, but failed when it was run on a server without a terminal. I was able to get it to fail on my local machine by running `foo </dev/null` instead of simply `foo`. It was very strange.

tomyun:
Strongly feel like it has something to do with headless x11, given that I also had a similar case with x11 and ncurses: Homebrew/homebrew-games#366.

sjackman:
If you have the time and inclination to sort out the failed test, go for it, otherwise I'm okay ignoring it and merging it if the test succeeds on your box. It would be nice if one other person can confirm the test works on their box as well.

sjackman:
And, yes, if it's an X11 app, I think that your intuition of headless X11 is probably correct.

tomyun:
@sjackman I don't think I can investigate the x11 issue more at this point. If you could just merge it, I'd happily take that.

sjackman:
@dpo @tseemann Any objections?

Off topic, how long does it take for the unpublished bottles to time out and be deleted from Bintray?

dpo:
No objection if we can get one more confirmation.

sjackman:
Works for me. When I run `brew test`, XQuartz starts, and `brew test` exits successfully.

PR #3296, opened 2016-02-09
openmodelica 1.9.3 (new formula)

RexFuzzle:
New formula for OpenModelica which a open source Modelica based modeling and simulation language.
This is a new formula for something which has been available for a while on macports. Unfortunately the stable version does not include the library files in the binary thus they cannot be installed on stable, but it does work on HEAD.

### DIFF
--- a/openmodelica.rb
+++ b/openmodelica.rb
@@ -1,0 +1,71 @@
+class Openmodelica < Formula
+  desc "Open-source modeling and simulation tool"
+  homepage "https://openmodelica.org/"
+  url "https://github.com/OpenModelica/OpenModelica/releases/download/v1.9.3/openmodelica_1.9.3.tar.xz"
+  sha256 "76d53076ae2681183be5f90574e4bd83203d455d1ddc31b93812c392859a667a"
+  head do
+    url "https://github.com/OpenModelica/OpenModelica.git"
+    option "with-library", "Build with OMLibraries"
+  end
+
+  depends_on "qt" =>:build
+  depends_on "autoconf" =>:build
+  depends_on "automake" =>:build
+  depends_on "cmake" =>:build
+  depends_on "gettext"
+  depends_on "gnu-sed" =>:build
+  depends_on "libtool" =>:build
+  depends_on "homebrew/science/lp_solve"
+  depends_on "openblas"
+  depends_on "pkg-config" =>:build
+  depends_on "readline" =>:build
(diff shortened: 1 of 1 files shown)


RexFuzzle:
So what do I do about this? 
El capitan fails on a time issue due to it taking too long- it takes 10 minutes on my system if all the dependencies are installed.
The other fails on sundials failing to build with the --without-mpi flag? Does this have to succeed?
Any help or pointers would be appreciated- seems so close.

RexFuzzle:
How do I get around this error? It is a very long build with lots of dependencies so it is going to take more than the 50min time limit- any advice?

openmodelica.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  depends_on "qt" =>:build
+  depends_on "autoconf" =>:build
+  depends_on "automake" =>:build
+  depends_on "cmake" =>:build
+  depends_on "gettext"
+  depends_on "gnu-sed" =>:build
+  depends_on "libtool" =>:build
+  depends_on "homebrew/science/lp_solve"
+  depends_on "openblas"
```
dpo: Doesn't this build against the Accelerate framework?

openmodelica.rb
```diff
@@ -0,0 +1,71 @@
...
+  depends_on "pkg-config" =>:build
+  depends_on "readline" =>:build
+  depends_on "xz" =>:build
+  conflicts_with "open-mpi", :because => "Causes issues in compiling."
```
dpo: Do you mean that it won't build if `open-mpi` is installed, or that it depends on some flavor of MPI, but not `open-mpi`?

openmodelica.rb
RexFuzzle: I'm not sure- I didn't think so, but don't know how to check. Any help would be much appreciated.

openmodelica.rb
RexFuzzle: So I think I can remove that conflict now as it was a problem when sundials was a dependency. With Sundials and open-mpi installed it would fail, but since sundials i no longer there, I haven't tried with without the conflicts with line.

openmodelica.rb
dpo: Will this be udpated? You can start by saying `depends_on "openblas" => :optional`. You might have to pass special flags during the build for the software to use OpenBLAS.

Also on line 18, there's no need to specify `homebrew/science`.

PR #3297, opened 2016-02-09
ecell4-4.0.0 : new formula

kozo2:
E-Cell4 is a multi-algorithm, multi-timescale biochemical simulation environment.
https://github.com/ecell/ecell4

### DIFF
--- a/ecell4.rb
+++ b/ecell4.rb
@@ -1,0 +1,67 @@
+class Ecell4 < Formula
+  desc "Multi algorithm-timescale bio-simulation environment"
+  homepage "https://github.com/ecell/ecell4"
+  # tag "systems biology"
+  # doi "10.1093/bioinformatics/15.1.72"
+  url "https://github.com/ecell/ecell4/archive/4.0.0.tar.gz"
+  sha256 "2cd3c82bc6e9666361f4d3bafa7272548ba2b52c179ad27a0fe35daf507bfd80"
+
+  head "https://github.com/ecell/ecell4.git"
+  option "with-python3", "Build python3 bindings"
+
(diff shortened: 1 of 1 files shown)


kozo2:
I don't know why cython call clang (not clang++) only on OSX=10.9.
https://travis-ci.org/Homebrew/homebrew-science/jobs/108005333#L4422

Do you have any idea?

ecell4.rb
```diff
@@ -0,0 +1,67 @@
...
+  url "https://github.com/ecell/ecell4/archive/4.0.0.tar.gz"
+  sha256 "2cd3c82bc6e9666361f4d3bafa7272548ba2b52c179ad27a0fe35daf507bfd80"
+
+  head "https://github.com/ecell/ecell4.git"
+  option "with-python3", "Build python3 bindings"
+
+  depends_on "cmake" => :build
+  depends_on "gsl" => :build
+  depends_on "homebrew/versions/boost155" => :build
+  depends_on "homebrew/science/hdf5"
```
dpo: There's no need to fully qualify dependencies that are in this tap.

dpo:
Normally, Python packages go in Homebrew/Python. But it seems that [this is available from `pip`](https://github.com/ecell/ecell4#mac), so I'm guessing it won't be acceptable.
@DomT4 ?

DomT4:
The general rule is that libraries aren't accepted in Homebrew, although `homebrew/python` may well make exceptions there, but applications (i.e. anything that generates an executable/binary/etc) is welcome.

kozo2:
OK, thank you for your comments.
Exactly this is available from pip.
I close this issue here.

PR #3298, opened 2016-02-09
cdsclient 3.81

glentner:
Version 3.80 is not available via ftp. So `brew install cdsclient` breaks with a 404 error. This updates the formula to the latest version.


glentner:
Most of the failures were from 'abibcode.awk' not having the appropriate executable permissions. I'm not sure what to do other than contact CDS and ask if they can patch this.

dpo:
Those aren't really failures, but rather something unusual. I'll launch a rebuild because the bottles failed to upload.

dpo:
@BrewTestBot test this please

dpo:
```
==> scamp -v
dyld: Library not loaded: /usr/local/lib/libplplotd.12.dylib
  Referenced from: /usr/local/bin/scamp
  Reason: image not found
```
It looks like `scamp` needs to be bumped.

PR #3301, opened 2016-02-11
rstudio-server v0.99.878 (update)

randy3k:
as title


randy3k:
> The job exceeded the maxmimum time limit for jobs, and has been terminated.

How should I fix this?

dpo:
@BrewTestBot test this please

PR #3308, opened 2016-02-13
xraylib: remove make check

tschoonj:
it fails on El Capitan for ruby, perl and python due to SIP...


dpo:
If the check fails, isn't that indicative of a problem with the software?

tschoonj:
Nope it's indicative of SIP :smile: 

The tests rely on the use of `DYLD_LIBRARY_PATH` in order to find libraries that were compiled. However, SIP doesn't allow this variable to be processed by executable residing in `/usr/bin`, including the system ruby, perl, python and others... Fortunately the installed software works fine since it relies on hardcoded paths to the libraries.

dpo:
Ok. Please bump the revision number.

tschoonj:
No need for that in this case I think. This only affects people who provided certain options when installing xraylib. There is no need for people with a successful installation to update.

tschoonj:
@dpo Could you please merge this one in? Thanks!

dpo:
@BrewTestBot test this please

PR #3313, opened 2016-02-15
vtk: use system python if available

dpo:


lantiga:
This PR broke our build on El Capitan (see our formula https://github.com/vmtk/vmtk/blob/master/distribution/homebrew/vmtk.rb). vmtk relies on vtk + python.
We have switched to brew so that we could rely on brew python, mainly as a workaround to SIP.

With this PR, if you happen to have brew python (which is what we want) and you install brew vtk (with python, as our formula does), when you `import python` from within the brew python interpreter, the interpreter segfaults (because vtk links against system python).

I don't think this is desirable, I would suggest to revert this behavior and assume users always go with brew python. It makes things more predictable, it avoids SIP issues and goes in the direction of not messing with system libraries, as Apple now recommends.

Removing `if MacOS.version <= :snow_leopard` on line 34 in vtk.rb forces vtk + python to build and everything works again.

dpo:
Sorry about that. Homebrew policy is to bottle against system Python on systems where one is available (i.e., >= 10.7). If Homebrew Python appears first on your `PATH`, you'll be able to build against it if you `brew install vtk --build-from-source`.

lantiga:
Thanks. But how do you get around the linking limitations imposed by SIP? E.g. if you install vtk with python bindings, the bindings are in a .dylib which ultimately links to the python dylib, which is disallowed by SIP.

dpo:
I am honestly not that familiar with SIP and I'll look into it. But I'm using VTK with Python bindings built against system Python with no problem.

lantiga:
Apart from the SIP issue, which I'm going to investigate too, this is what happened to us:
we had brew python installed and visible (`which python` gave `/usr/local/bin/python`) yet vtk built against system python. So doing `import vtk` from within the (brew) python intepreter led to a crash. 

My understanding is that cmake picked up system python even though brew python was first in the PATH. I'm now looking into that too.

lantiga:
I confirm it is the case:

Scenario 1:
- start with no brew python and no brew vtk
- `brew install vtk --with-python`
- `python -c "import vtk"`
=> Ok

Scenario 2:
- start with no brew python and no vtk
- `brew install python`
- `brew install vtk --with-python`
- `python -c "import vtk"` => ```Fatal Python error: PyThreadState_Get: no current thread. Abort trap: 6```

lantiga:
Ok, the issue is confirmed:

- start with no brew python and no vtk
- `brew install python`
- `which python` => `/usr/local/bin/python`
- `brew install vtk --with-python --build-from-source`
- `python -c "import vtk"` => `Fatal Python error: PyThreadState_Get: no current thread. Abort trap: 6`

So, if I'm not missing anything obvious, after the PR there is no way to use brew python and vtk.

dpo:
Thanks for investigating this. I'll have to try and reproduce the issue. Did you make sure that all `vtk` dependencies (e.g., `sip` and `pyqt`) were built against the same interpreter?

lantiga:
I don't rely on `sip` and `pyqt`, so my suggestion is to keep them off for now.

When I referred to SIP, it was https://en.wikipedia.org/wiki/System_Integrity_Protection. Sorry for the confusion.

dpo:
Ok I see. What does
```
otool -L /usr/local/lib/python2.7/site-packages/vtk/vtkCommonCorePython.so 
```
say?

lantiga:
```
/usr/local/lib/python2.7/site-packages/vtk/vtkCommonCorePython.so:
	/usr/local/Cellar/vtk/7.0.0/lib/libvtkCommonCorePython27D-7.0.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/Cellar/vtk/7.0.0/lib/libvtkWrappingPython27Core-7.0.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/Cellar/vtk/7.0.0/lib/libvtkCommonCore-7.0.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Python.framework/Versions/2.7/Python (compatibility version 2.7.0, current version 2.7.10)
	/usr/local/Cellar/vtk/7.0.0/lib/libvtksys-7.0.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
```

This is following:
- `brew install python`
- `which python => /usr/local/bin/python`
- `brew install vtk --with-python --build-from-source`

marcoesposito1988:
I can confirm the issue. After upgrading VTK, I cannot use it with the brewed python because of the PyThreadState_Get error mentioned above.

lantiga:
Should I open an issue on this? In my opinion the fact that vtk cannot work with brewed python, either bottled or built from source, should be fixed.

There are no workarounds except uninstalling brewed python, which is not a solution.

Thanks for any directions

dpo:
It's a bug and should be fixed. I'll look into it asap. Fel free to open an issue.

lantiga:
Done, thanks.

PR #3314, opened 2016-02-16
trilinos: use system python if available

dpo:


dpo:
@mikemcquaid Sorry to bother again with the question of long builds. Perhaps there could be a whitelist of formulas that are allowed to take up to 90 or 120 minutes or so? Another formula concerned is `paraview.rb` (see #3194).

dpo:
@mikemcquaid Just checking in.

mikemcquaid:
@dpo Sorry, this is on my list but I haven't got around to it yet. We may need to work out a better system to handle these cases.

mikemcquaid:
CC @DomT4 as you often keep a closer look at the build queue than I do now.

DomT4:
From 23:00 - 09:00 UK time (18:00 - 04:00 EST) the CI is quiet enough to get away with stacking up longer builds.

These sorts of situations make me tempted to push us quicker towards the direction where builds need to be manually called by a maintainer to activate, because at the moment on longer builds we're wasting an hour every time which could be used elsewhere.

We can't bump the timeout massively on a regular basis because it causes significant issues during the day with build stacking and if a build is failing there's no point letting it fail at 3 hours instead of 1, at least whilst other items are pending.

We also can't _really_ have a file in the repository which determines build times for each formulae because that opens Homebrew up to DoS where people could edit that file to increase the timeout and then edit the formula to build slowly/problematically and hold up the rest of the queue.

dpo:
Well the folks over at Travis generously bumped the build timeout for Homebrew/Science to *two hours*! I thought that would be more than enough to at least make sure that this formula builds before bothering with Jenkins, but it still times out! When I build locally, it takes between 60 and 90 minutes. Something on Travis makes the build quite a bit slower.

At this point, I'm thinking the bottles for this particular formula should just be built by hand somehow (how?). Is it possible for me to somehow fetch the three OSX images, build in there, and upload the bottles by hand?

DomT4:
Same comment as from the other thread: https://github.com/Homebrew/homebrew-science/pull/3194#issuecomment-188839287

DomT4:
@BrewTestBot test this please

trilinos.rb
```diff
@@ -21,9 +23,8 @@ class Trilinos < Formula
...
   depends_on :fortran       => :recommended
   depends_on :x11           => :recommended
-
-  depends_on :python        => :recommended
-  depends_on "homebrew/python/numpy"  if build.with? "python"
+  depends_on :python        => :recommended if MacOS.version <= :snow_leopard
+  depends_on "numpy"        => :python if build.with? "python"
   depends_on "swig"         => :build if build.with? "python"
 
   depends_on "cmake"        => :build
```
mikemcquaid: It seems you're also using this in the test which makes this not a build-time dependency.

dpo: Not in this formula (the test just runs a few executables) but in one of the dependencies, you're right (`dealii`). I'll have to fix that one. Thanks. For some reason, 10.11 didn't complain.

dpo:
Thanks for building!

DomT4:
No worries. Looks like this one is okay to merge when you're happy, once you've fixed the `dealii` test issue.

PR #3315, opened 2016-02-16
petsc: build against static mumps

dpo:


dpo:
Contingent on #3312.

dpo:
@BrewTestBot test this please

PR #3318, opened 2016-02-17
canu 1.0 (new formula)

sjackman:
Single molecule sequence assembler for genomes large and small


sjackman:
http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/4184/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_homebrew_science_canu___strict___online/
```
==> audit problems
homebrew/science/canu:
Length is calculated as homebrew/science/canu + desc. (currently 85)
```

PR #3319, opened 2016-02-17
OpenFST 1.5.1 (update)

stevenbedrick:

PR #3320, opened 2016-02-17
opengrm-thrax: bump revision

dpo:


stevenbedrick:
Hi, Dominique-

I have to confess I'm a touch confused about what to do here, Github-wise. Is the idea that once PR #3320 is complete and merged, we'll somehow cause PR #3319 to try and do the CI dance again, after which it can be merged?

Also, earlier this morning, I was looking into why the CI test failed, and also observed that the opengrm-thrax bottles were the root of the problem- does adding "revision 1" to the opengrm-thrax recipe actually solve the problem? I ask because, while investigating, I tried to build thrax 1.2.0 against OpenFST 1.5.1 (using `brew install --build-from-source opengrm-thrax`) and got a build error. Using thrax 1.2.1 instead of 1.2.0 (i.e., updating opengrm-thrax as well as openfst) worked just fine, though, so I am wondering if thrax 1.2.1 has a dependency against openfst 1.5.1. 

Is it acceptable homebrew etiquette to update both opengrm-thrax and openfst in one pull request?

Thanks much for your help and patience!

tdsmith:
> Is it acceptable homebrew etiquette to update both opengrm-thrax and openfst in one pull request?

That's preferred unless it takes too long to build them both together.

PR #3326, opened 2016-02-18
clBLAS and BLIS

jpauwels:
Updated BLIS to 0.1.8 and clBLAS to 2.6.

Regarding even newer versions of clBLAS: v2.8 fails for me while linking (but apparently there are [other issues too](https://github.com/Homebrew/homebrew-science/pull/3029)) and v2.10 can be patched to build, but then [segfaults](https://github.com/clMathLibraries/clBLAS/issues/215), verified on my system too. There are plenty of issues with OS X filed upstream, so it's well known.


dpo:
Does `arrayfire` work with CLBLAS 2.6?

Issue #1891, opened 2015-02-27
nfft: Installing bottle although brew install -cc=... would require an build from source


cmarquardt:
Hello,

this is probably more a question than an issue...

A while ago, the nfft formula was updated to optionally build a multithreaded version (#1767). This requires gcc, as clang doesn't support OpenMP. With

    brew install --cc=gcc-4.9 nfft

a multithreaded version of the library should be built along with the serial one; without the `--cc` option, i.e. if clang is the compiler, only the serial version of the compiler should be built.

I've had to redo my nfft installation and found that instead of building from source, an automatically build bottle is downloaded and installed. In principle, that's great - but there only is a bottle containing the serial version. When running `brew install`with the `--gcc=...` option, the same serial bottle is installed. The same is true for a `brew reinstall`, of course. In order to get the non-default argument `-cc=...` honoured, I have to issue an

    brew install --gcc=gcc-4.9 --build-from-source nfft

which works as expected.

My questions:

1. Is that behaviour (installing a bottle built with default settings even if non-default arguments are given to `brew (re)install`) intended? Shouldn't different bottles be selected based on the arguments given to `brew install`? Or a build from the sources be triggered as soon as non-standard arguments are given to `brew install`?
2. Alternatively, is there a way to force a build from source from within the formula if non-standard arguments are given to `brew (re)install`?

Thanks a lot,

   Christian.

dpo:
I thought a source build should be triggered if you supply non-default arguments, but @jacknagel will be able to confirm.

jacknagel:
We can't make "global" options like `--cc` trigger a source build because it would force installing all dependencies from source as well. Only `--interactive` or formula-specific options automatically trigger a source build.

cmarquardt:
Ok - it would be better then in the nfft formula to have a non-standard argument (say --multithreaded) which requires the gcc compiler?

How would one make sure in the formula that the gcc-x.y compiler (with a potentially unknown version x.y) is used for compilation, overwriting the default `clang`? I mean, is there a way to do the equivalent of --cc=gcc-x.y inside the formula?

Or alternatively, can one somehow trigger a build from source from within the formula?

Many thanks,

   Christian.

iml:
[...]

PR #3331, opened 2016-02-19
nest 2.10 (new formula)

tammoippen:
> NEST is a simulator for spiking neural network models that focuses on the dynamics, size and structure of neural systems rather than on the exact morphology of individual neurons.

I know there is already a PR #1876 for this software, but the PR is quite outdated. I based the formula on it, updated it to the latest release of NEST and tried to incorporate your suggestions.


nest.rb
```diff
@@ -0,0 +1,71 @@
...
+  depends_on "matplotlib" => :python
+  depends_on "cython" => :python
+
+  def install
+    args = ["--disable-debug",
+            "--disable-dependency-tracking",
+            "--prefix=#{prefix}",
+           ]
+
+    args << if ENV.compiler == :clang && build.with?(:mpi)
```
dpo: Should be `build.with?("mpi")`.

nest.rb
```diff
@@ -0,0 +1,71 @@
...
+              "--with-mpi"
+            else
+              "--without-mpi"
+            end
+
+    if ENV.compiler != :clang && build.with?(:mpi)
+      opoo "Cannot build nest with mpi due to bugs in configure. Use Clang+MPI."
+    end
```
dpo: This should probably be a `fails_with` block.

nest.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    if ENV.compiler == :clang
+      opoo "clang does not support OpenMP. Compile with gcc if this is not acceptable."
+      args << "--without-openmp"
+    end
```
dpo: Here I suggest using
```
option "with-openmp", "Build with OpenMP support"
needs :openmp if build.with? "openmp"
```
in the preamble. This will automatically pick an OpenMP-capable compiler.

dpo:
Thanks! It's great to revive that PR.

nest.rb
```diff
@@ -0,0 +1,71 @@
+class Nest < Formula
+  desc "nest::The NEural Simulation Tool"
```
dpo: Is this `nest::` necessary? I suggest just `NEural Simulation Tool`.

tammoippen:
Thank you for the review.

I simplify the formula:
* Building NEST with `clang` on OS X is not [stable](https://github.com/nest/nest-simulator/issues/74) => hence the `fails_with` block.
* Building NEST+GCC+MPI is difficult with Homebrew packages: the configure script is buggy, when only the actual compiler and not the MPI wrappers are used; and by default the wrappers are build with `clang`. The bugs are fixed for the next release.

dpo:
Does it work to set `ENV["CC"] = ENV["MPICC"]`, etc., before calling `configure` when building with MPI?

tammoippen:
Maybe (i am not able to test it right now), but since homebrew mpi is most likely built with `clang`, the same instability problems and OpenMP issues will be there, right?

dpo:
It could be. What I mean is something like [this](https://github.com/Homebrew/homebrew-science/blob/master/petsc.rb#L63).

tammoippen:
The environment variables `MPICC` and `MPICXX` are not set by default (at least not on my system after installing `open-mpi` or `mpich`). I would prefer to leave the MPI option for now. In the next release the issue will be solved and when I update the formula, I will add the MPI option.

dpo:
Those variables are set when in a Homebrew formula that `depends_on :mpi`.

tammoippen:
The formula needs `depends_on :mpi => [:cc, :cxx]` for these variables to be available. Thank you! Now installing with and without MPI/OpenMP is possible.

tammoippen:
@dpo Is there something else I should improve?

tammoippen:
Hello,
I really would like to see this PR merged. Can somebody please continue reviewing the formula? It is almost 3 weeks with no update. Is there something I can do? In the last update I added the line 

    depends_on "gcc" => :build

and I am not sure if this conforms with your guidelines.

dpo:
Sorry I'm currently impossibly busy. We don't want formulas to explicitly depend on `gcc` but you can use ` fails_with` block if you want to exclude specific compilers.

tammoippen:
In the last commit I added `args << "--with-mpi"`, so that NEST is actually compiled with MPI enabled. Further, the `test` command is improved:

* the ENV variable `exec_prefix` needs to be set for python tests
* MPI tests require the `~/.nestrc` file to specify the `/mpirun` command, since we are in a sandbox, tis file is not available and the formula creates one.

Some of our tests require access to the sources (otherwise they fail). Is it possible, to get a path to the sources? Or to copy them into the sandbox?

dpo:
@BrewTestBot test this please

nest.rb
```diff
@@ -0,0 +1,87 @@
...
+  option "without-openmp", "Build without OpenMP support."
+  needs :openmp if build.with? "openmp"
+
+  depends_on "gsl" => :recommended
+  depends_on :mpi => [:optional, :cc, :cxx]
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "numpy" => :python
+  depends_on "scipy" => :python
+  depends_on "matplotlib" => :python
+  depends_on "cython" => :python
```
dpo: Those should all say `if build.with? "python"`.

tammoippen: When I replace them like `depends_on "numpy" if build.with? "python"` I get errors:
```
==> Installing nest from homebrew/science
Error: No available formula with the name "cython" (dependency of homebrew/science/nest)
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

In [Python for Formula Authors](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md#python-module-dependencies) there is even the example for `numpy`. Or should they be loaded as [resources](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md#dependencies)? BTW: NEST uses `setup.py` scripts to install its bindings, but they are buried within its Makefiles.

dpo: Right, I think they should be
```
depends_on "numpy" => :python if build.with? "python"
```

dpo:
Also, currently, the Python option suggests that the default is `--without-python`, but `python` will be installed nevertheless if `MacOS.version <= :snow_leopard`. What you could do instead is remove the option and use
```
depends_on :python => :optional if MacOS.version <= :snow_leopard
```

nest.rb
```diff
@@ -0,0 +1,87 @@
...
+  needs :openmp if build.with? "openmp"
+
+  depends_on "gsl" => :recommended
+  depends_on :mpi => [:optional, :cc, :cxx]
+  depends_on :python => :optional if MacOS.version <= :snow_leopard
+  depends_on "numpy" => :python if build.with? "python"
+  depends_on "scipy" => :python if build.with? "python"
+  depends_on "matplotlib" => :python if build.with? "python"
+  depends_on "cython" => :python if build.with? "python"
+  depends_on "libtool"
```
ilovezfs: This is only a build-time dependency, right?

tammoippen:
2 of the tests in TravisCI `brew test --verbose homebrew/science/nest` fail, as we need access to the source code (see [above](https://github.com/Homebrew/homebrew-science/pull/3331#issuecomment-203977923)). Is there a way to copy the sources into the sandbox?

nest.rb
tammoippen: No. NEST actually uses the `libltdl` for loading user defined modules. See this (previous) TravisCI [output](https://travis-ci.org/Homebrew/homebrew-science/jobs/123263127) - here the tests fail, when NEST wants to load the library.

nest.rb
```diff
@@ -0,0 +1,87 @@
...
+  desc "The Neural Simulation Tool"
+  homepage "http://www.nest-simulator.org/"
+  url "https://github.com/nest/nest-simulator/releases/download/v2.10.0/nest-2.10.0.tar.gz"
+  sha256 "2b6fc562cd6362e812d94bb742562a5a685fb1c7e08403765dbe123d59b0996c"
+
+  head "https://github.com/nest/nest-simulator.git"
+
+  option "with-python", "Build Python bindings (PyNEST)."
+  option "without-openmp", "Build without OpenMP support."
+  needs :openmp if build.with? "openmp"
```
ilovezfs: You can remove
```
option "without-openmp", "Build without OpenMP support."
needs :openmp if build.with? "openmp"
```
and replace it with
```
depends_on :openmp => :recommended
```
You should see the automagic description in `brew info nest`.

tammoippen: Hi @ilovezfs ,

After replacing, I get:
```
(0)$ brew info nest
Error: Unsupported special dependency :openmp
Please report this bug:
/usr/local/Library/Homebrew/dependency_collector.rb:123:in `parse_symbol_spec'
...
```

Just before I did a `brew update`: `Updated Homebrew from a8ed06b to b70e434.`

ilovezfs: Ah, it's `:mpi`

tammoippen: This would be a nice extension, which was already mentioned [here](https://github.com/Homebrew/homebrew-science/issues/1891#issuecomment-77774828). So no update regarding the `:openmp`?

ilovezfs: Nope, all it does is verify the compiler has the mpi support, so you have it correct as it is.

ilovezfs: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/compilers.rb#L72

ilovezfs:
Two tests are failing:
http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/4538/version=el_capitan/console

If you install the source code, yes you can copy it from wherever it is in the prefix into the `testpath`. Examples:
https://github.com/Homebrew/homebrew-core/search?utf8=%E2%9C%93&q=cp_r

nest.rb
dpo: Then it should be `depends_on "libtool" => :run`.

tammoippen:
The tests should work fine now. But I am not sure about the python bindings at the moment. I have to investigate...

dpo:
@BrewTestBot test this please

dpo:
Thank you for your patience and efforts!

PR #3333, opened 2016-02-19
circlator 1.2.0

martinghunt:


circlator.rb
```diff
@@ -20,7 +20,6 @@ class Circlator < Formula
   depends_on "prodigal"
   depends_on "samtools"
   depends_on "spades"
-  depends_on "homebrew/python/numpy" => ["with-python3"]
   depends_on "homebrew/python/pymummer"
```
dpo: Does it have to be the `pymummer` from `homebrew/python` or could it be one installed with `pip`? It would be more general to just say `depends_on "pymummer" => :python`.

dpo: (in which case you would presumably also need `depends_on "numpy" => :python`).

sjackman:
@Homebrew/science When the package exists in `homebrew/python`, which do we prefer
`depends_on "homebrew/python/numpy"` or `depends_on "numpy" => :python`?

martinghunt:
Sorry, should have said this earlier: numpy is no longer a dependency for any of the python packages used by circlator. It used to be required by fastaq, but that is no longer the case. So we don't need depends_on numpy at all.

Re pymummer, I don't mind where it circlator gets it from, so whatever you prefer/is the convention. I only realised after making this PR that pymummer was in homebrew-python. I've put in a PR to get pymummer updated to the latest version (and to also remove the depends_on numpy).
https://github.com/Homebrew/homebrew-python/pull/285
... but that is currently failing ...

dpo:
@sjackman My understanding is that we prefer `depends_on "numpy" => :python` because the dependency will be satisfied whether the user installed `numpy` with `pip` or `brew`. See also https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md#python-module-dependencies.

sjackman:
Fine by me. I wasn't sure what the policy was.

martinghunt:
pymummer is now updated in homebrew-python. But Circlator is now up to version 1.2.0.

Do you want Circlator 1.2.0 as a separate new pull request and this one closed?

dpo:
That would make sense.

sjackman:
You can reuse this same pull request. That keeps the conversation all in one place. Just update your commit, `git push --force` it, and retitle this issue.

tseemann:
@martinghunt ping :)

martinghunt:
Tests passed, so hopefully that's it sorted now :)

circlator.rb
```diff
@@ -29,19 +28,20 @@ class Circlator < Formula
...
-    sha256 "3547d54b357702d659d244ed6493b6553eb2e72dcf79e88cdae2c595b933bf91"
+    url "https://pypi.python.org/packages/source/b/bio_assembly_refinement/bio_assembly_refinement-0.5.0.tar.gz"
+    sha256 "8d5e8d6c5ad23602015f8099842b607928e46c1393c9acb3beddba8d0351f263"
   end
 
   def install
     version = Language::Python.major_minor_version "python3"
     ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python#{version}/site-packages"
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python#{version}/site-packages"
+    ENV.prepend_create_path "PYTHONPATH", "#{HOMEBREW_PREFIX}/lib/python#{version}/site-packages"
```
sjackman: Change `"#{HOMEBREW_PREFIX}/lib/python#{version}/site-packages"`
to `HOMEBREW_PREFIX/"lib/python#{version}/site-packages"`

circlator.rb
```diff
@@ -55,6 +55,7 @@ def install
   end
 
   test do
-    assert_match "Available commands", shell_output("circlator -h 2>&1", 0)
+    assert_match "Available commands", shell_output("circlator 2>&1", 0)
```
sjackman: Change `shell_output("circlator 2>&1", 0)`
to `shell_output("#{bin}/circlator 2>&1", 0)`

circlator.rb
```diff
@@ -55,6 +55,7 @@ def install
...
+    assert_match "1.2.0", shell_output("circlator version 2>&1", 0)
```
sjackman: Ditto.

sjackman:
Looks good. Make those minor changes, and it's good to go.

ilovezfs:
Closing in favor of https://github.com/Homebrew/homebrew-science/pull/4050.

@martinghunt thanks for opening this! Maybe you can do the next upgrade.

PR #3339, opened 2016-02-19
vtk: cleanup unnecessary flags

iMichka:
IOKit was removed somewhere between vtk 6.1.0 and 6.2.0.
See commit 2eb00f97acd00fae495500c1c9ef2a730b14a183 in VTK

VTK_USE_TK was set to OFF because at one point the python bindings also built TK:
http://www.paraview.org/Bug/print_bug_page.php?bug_id=13318
This caused some troubles during the build, so we disabled it.
Now, VTK_USE_TK is only available (and set to ON by default) when using VTK_WRAP_TCL.

Cmake will warn about the VTK_USE_TK and IOKit flags being set but not used.

VTK_USE_CARBON was removed. The support for Carbon was dropped in VTK 7.
See commit da50d6c17e64c98f753a3091b617b78fe9fe8aa7 in VTK.

I also simplified the code for pyqt and sip in the python block.


dpo:
Seems `statismo` needs to be bumped, as in #3288. Should probably be done separately.

iMichka:
You mean, should I do a separate commit bumping it to revision 2, which you can then merge after this one and #3288 (so that we get no merge conflict) ?

dpo:
Yes, that would be great. Thank you!

PR #3357, opened 2016-02-23
Version bump for OpenFST 1.5.1, OpenGRM-Thrax 1.2.1, and OpenGRM-NGra…

stevenbedrick:
OK, I think *this* time I've got it- no stray mystery merge conflicts, and it's only affecting three files.

### DIFF
--- a/openfst.rb
+++ b/openfst.rb
@@ -1,7 +1,7 @@
 class Openfst < Formula
   homepage "http://www.openfst.org/"
-  url "http://openfst.org/twiki/pub/FST/FstDownload/openfst-1.5.0.tar.gz"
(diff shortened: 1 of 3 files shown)


stevenbedrick:
Commit 9b8abb3 included a bad SHA256 for OpenFST- they must've **just** changed the tarball, since I have on my laptop a version of the OpenFST 1.5.1 tarball from a few days ago with that hash. Whatever, let's see if the new hash does the trick.

stevenbedrick:
> `Failed to read test report file /Users/brew/Jenkins/science/version/mavericks/brew-test-bot.xml`


OK, what does **that** mean? Now the continuous integration is failing due to some sort of XML error, which I can only imagine has something to do with an internal problem within Jenkins? Any ideas?

dpo:
Thank you!

PR #3359, opened 2016-02-23
r: Add "-L" dirs to SHLIB_CXXLDFLAGS for gettext, etc

DabeDotCom:
While trying to install the 'RInside' package, I ran into the following:
```
prompt% echo 'install.packages("RInside")' | sudo R --no-save
...
clang++ -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/usr/local/lib -L/usr/local/Cellar/r/3.2.3/R.framework/Resources/lib -L/usr/local/opt/gettext/lib -L/usr/local/opt/readline/lib -L/usr/local/opt/openssl/lib -L/usr/local/lib -o RInside.so MemBuf.o RInside.o -F/usr/local/Cellar/r/3.2.3/R.framework/.. -framework R -lintl -Wl,-framework -Wl,CoreFoundation
clang++ -o libRInside.dylib MemBuf.o RInside.o -dynamiclib -Wl,-headerpad_max_install_names  -undefined dynamic_lookup -single_module -multiply_defined suppress -L/usr/local/lib   -F/usr/local/Cellar/r/3.2.3/R.framework/.. -framework R -lintl -Wl,-framework -Wl,CoreFoundation
ld: library not found for -lintl
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [libRInside] Error 1
ERROR: compilation failed for package ‘RInside’

prompt% clang++ --version
Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix
```

It would appear that "-L/usr/local/opt/gettext/lib" (et al) didn't make it into SHLIB_CXXLDFLAGS.

My one-line patch may not be the Most Correct way of doing it (I haven't tried it under different environments with different compilers, etc.) but it worked for me:

```
@@ -107,6 +107,7 @@ class R < Formula
     %w[gettext readline openssl].each do |f|
       ENV.append "CPPFLAGS", "-I#{Formula[f].opt_include}"
       ENV.append "LDFLAGS", "-L#{Formula[f].opt_lib}"
+      ENV.append "SHLIB_CXXLDFLAGS", "-L#{Formula[f].opt_lib}"
     end
```


DabeDotCom:
PS — Adding `depends_on :java` does silence the `brew audit` errors, but I didn't think that was appropriate for this patch (The continuous integration tests must be failing upstream as well...)

tdsmith:
I don't understand why the variables we set are passed into the first invocation you pasted but not the second! They're doing the same thing! (RInside.so looks misnamed, at any rate; it's a dylib, not a bundle. I guess something's making a distinction.) Anyway, if this works, I'll take it.

DabeDotCom:
The first looks like it creates a ".so" (presumably with the regular `LDFLAGS`, which _have_ the correct paths), whereas the second creates a ".dylib" using `SHLIB_CXXLDFLAGS`.

An alternate solution (which, arguably, _might_ be more appropriately applied upstream — though it's my understanding that the maintainers of R have poo-pooed supporting homebrew builds) would be to update "etc/Makeconf", thusly:

```
    inreplace r_home/"etc/Makeconf" do |s|
      s.gsub! /SHLIB_.*LDFLAGS =.*/, "\\0 $(LDFLAGS)"
    end
```

I tested the above in `post_install`, which has the advantage that it works with current bottles, as well.  I don't know if it's worth patching just the formula to preserve existing bottles, or if the Right Thing To Do is to go ahead and build new bottles from source, in which case I would actually use the following, which adds `-L/usr/local/lib` to the regular `LDFLAGS` (with no prefix), then appends _THAT_ to each of the other variables:

```
--- a/r.rb
+++ b/r.rb
@@ -140,5 +140,6 @@ class R < Formula
       # make Homebrew packages discoverable for R CMD INSTALL
       inreplace r_home/"etc/Makeconf" do |s|
         s.gsub! /CPPFLAGS =.*/, "\\0 -I#{HOMEBREW_PREFIX}/include"
-        s.gsub! /LDFLAGS =.*/, "\\0 -L#{HOMEBREW_PREFIX}/lib"
+        s.gsub! /^LDFLAGS =.*/, "\\0 -L#{HOMEBREW_PREFIX}/lib"
+        s.gsub! /.LDFLAGS =.*/, "\\0 $(LDFLAGS)"
       end
```

If you'd like that as a separate, single-commit PR [which, as a heads-up, will continue to fail testing until the "java_home" fix is made] let me know and I can close this one...

DabeDotCom:
NOTE: I cleaned this up and resubmitted as [#3386](https://github.com/Homebrew/homebrew-science/pull/3386)

PR #3361, opened 2016-02-24
Picard 2.1.0

bede:


tseemann:
Thanks @bede 
If you update it next time can you add a `desc` field? Not urgent, it's part of `brew audit --strict`.

bede:
Sure, thanks for the tip. Will check `--strict` output in the future.

PR #3372, opened 2016-02-25
mlpack: bump

fedorareis:
bump to version 2.0.1


mlpack.rb
```diff
@@ -1,9 +1,9 @@
 class Mlpack < Formula
   homepage "http://www.mlpack.org"
   # doi "arXiv:1210.6293"
-  url "http://www.mlpack.org/files/mlpack-1.0.12.tar.gz"
-  sha256 "f47abfc2ab75f1d7f4c73a3368c4428223f025cc6fbc6703735df6a2734a838c"
-  revision 2
+  url "http://www.mlpack.org/files/mlpack-2.0.1.tar.gz"
+  sha256 "87305f7003e060d3c93d60ce1365c4ec0fa7e827c356e857be316b0e54114f22"
+  revision 3
```
dpo: You can remove the `revision` line when you increase the version number.

dpo:
The test with `allknn` fails.

dpo:
[Here](https://github.com/Homebrew/homebrew-science/pull/3372/files#diff-2e6205c4e12cd1dafdb5284a684acf07R49) the executable should be `mlpack_allknn`.

fedorareis:
Yeah, I realized that was the issue on Friday. I fixed it locally but now the next test with test.cpp fails.

mlpack.rb
```diff
@@ -47,7 +46,7 @@ def install
 
   test do
     cd testpath do
```
dpo: Try inserting `ENV.cxx11` here and changing line 69 to
```
system *(ENV.cxx.split), "test.cpp",
```

dpo: @fedorareis 
In fact it would be better to do
```
cxx_with_flags = ENV.cxx.split + ["test.cpp", ... ]
system *cxx_with_flags
```

fedorareis:
Thank you @dpo for all of the help.

dpo:
Thank you!

PR #3377, opened 2016-02-26
Fix link to PETSc conf directory

mdave:
This PR fixes the installation of a symlink to the PETSc `conf` directory, which I believe changed location as of 3.6. Without this patch, programs which rely on the default `FindPETSc.cmake` module fail to configure with CMake, since the `conf` directory can't be found.


dpo:
It's true that line 150 is erroneous, but shouldn't it simply be removed? We have formulas here that depend on `petsc` and build correctly with `cmake` once we've set `PETSC_DIR` and `PETSC_ARCH`, e.g., `dealii`, `getdp` and `gmsh`.

mdave:
Actually I agree it's fine to just remove it. I was running into issues with the 'semi-official' version (the packages you mention use their own versions):

https://github.com/jedbrown/cmake-modules/blob/master/FindPETSc.cmake 

but this is because I was specifying `PETSC_ARCH=real` and `PETSC_DIR` to the opt prefix. Just setting `PETSC_DIR` works fine.

dpo:
OK thanks. Would you mind increasing the revision number, though?

mdave:
Done!

dpo:
Thank you!

PR #3381, opened 2016-02-28
scamp: bump for cdsclient

dpo:


dpo:
See #3298.

dpo:
@BrewTestBot test this please

PR #3384, opened 2016-02-29
SuiteSparse 4.5.1

syarra:
New version of SuiteSparse. Now builds shared library!!!

SuiteSparse is shipped with metis-5.1.0 and uses it by default. Setting some environment variables make possible the use of Homebrew's version of metis.

Wasn't tested with flag --with-matlab. No Matlab at home.

### DIFF
--- a/suite-sparse.rb
+++ b/suite-sparse.rb
@@ -49,7 +49,7 @@ class SuiteSparse < Formula
     # SuiteSparse is shipped with metis-5.1.0 but it can use Homebrew's version by
     # setting MY_METIS_LIB and MY_METIS_INC variables.
     make_args += ["MY_METIS_LIB=-L#{Formula["metis"].opt_lib} -lmetis",
-                  "MY_METIS_INC=#{Formula["metis"].opt_include}"] if build.with? "metis"
+                  "MY_METIS_INC=#{Formula["metis"].opt_include}"]
 
     # Demos failed to compile. Thus only building libraries.
     system "make", "library", *make_args
--- a/suite-sparse.rb
+++ b/suite-sparse.rb
@@ -26,7 +26,7 @@ class SuiteSparse < Formula
 
   def install
     cflags = "#{ENV.cflags}"
-    cflags += "-fopenmp" if build_with? "openmp"
+    cflags += "-fopenmp" if build.with? "openmp"
     cflags += " -I#{Formula["tbb"].opt_include}" if build.with? "tbb"
 
     make_args = ["CFLAGS=#{cflags}",
--- a/suite-sparse.rb
+++ b/suite-sparse.rb
@@ -14,9 +14,9 @@ class SuiteSparse < Formula
   option "with-matlab", "Install Matlab interfaces and tools"
   option "with-matlab-path=", "Path to Matlab executable (default: matlab)"
 
+  depends_on "metis"  # SuiteSparse must be compiled with metis 5.
+                      # It is shipped with metis-5.1.0
   depends_on "cmake" => :build
-  depends_on "metis" => :recommended  # SuiteSparse must be compiled with metis 5.
-                                      # It is shipped with metis-5.1.0
   depends_on "tbb" => :recommended
   depends_on "openblas" => :optional
 
--- a/suite-sparse.rb
+++ b/suite-sparse.rb
@@ -1,9 +1,8 @@
 class SuiteSparse < Formula
   desc "Suite of Sparse Matrix Software"
   homepage "http://faculty.cse.tamu.edu/davis/suitesparse.html"
-  url "http://faculty.cse.tamu.edu/davis/SuiteSparse/SuiteSparse-4.4.4.tar.gz"
-  sha256 "f2ae47e96f3f37b313c3dfafca59f13e6dbc1e9e54b35af591551919810fb6fd"
-  revision 2
+  url "http://faculty.cse.tamu.edu/davis/SuiteSparse/SuiteSparse-4.5.1.tar.gz"
+  sha256 "ac4524b9f69c4f8c2652d720b146c92a414c1943f86d46df49b4ff8377ae8752"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


suite-sparse.rb
```diff
@@ -15,22 +14,15 @@ class SuiteSparse < Formula
   option "with-matlab", "Install Matlab interfaces and tools"
   option "with-matlab-path=", "Path to Matlab executable (default: matlab)"
 
+  depends_on "cmake" => :build
+  depends_on "metis" => :recommended  # SuiteSparse must be compiled with metis 5.
```
dpo: If it *must* be compiled with `metis`, should this really be `:recommended` then? Or can `metis` support be disabled?

suite-sparse.rb
```diff
@@ -15,22 +14,15 @@ class SuiteSparse < Formula
...
-    ENV.deparallelize
-
-    # Switch to the Mac base config, per SuiteSparse README.txt
-    mv "SuiteSparse_config/SuiteSparse_config.mk",
-       "SuiteSparse_config/SuiteSparse_config_orig.mk"
-    mv "SuiteSparse_config/SuiteSparse_config_Mac.mk",
-       "SuiteSparse_config/SuiteSparse_config.mk"
-
     cflags = "#{ENV.cflags}"
     cflags += (ENV.compiler == :clang) ? "" : " -fopenmp"
```
dpo: I think we could transform this to
```
option "with-openmp", "Build with OpenMP support"
needs :openmp if build.with? "openmp"
```
which would be somewhat cleaner.

suite-sparse.rb
```diff
@@ -52,27 +44,16 @@ def install
...
-                  "METIS=-L#{Formula["metis4"].opt_lib} -lmetis"] if build.with? "metis4"
-    make_args << "CHOLMOD_CONFIG=-DNPARTITION" if build.without? "metis4"
+    # SuiteSparse is shipped with metis-5.1.0 but it can use Homebrew's version by
+    # setting MY_METIS_LIB and MY_METIS_INC variables.
+    make_args += ["MY_METIS_LIB=-L#{Formula["metis"].opt_lib} -lmetis",
+                  "MY_METIS_INC=#{Formula["metis"].opt_include}"] if build.with? "metis"
 
-    # Add some flags for linux
-    # -DNTIMER is needed to avoid undefined reference to SuiteSparse_time
-    make_args << "CF=-fPIC -O3 -fno-common -fexceptions -DNTIMER $(CFLAGS)" unless OS.mac?
```
dpo: What happens to this stuff?

dpo:
Thanks! That is excellent news!

suite-sparse.rb
syarra: It is not needed anymore. The new makefile takes care of it!

suite-sparse.rb
syarra: I found contradictory information on that!
- `SuiteSparse_config/SuiteSparse_config.mk` says `metis` support could be disabled by setting `CHOLMOD_CONFIG` to `-DNPARTITION`. (doesn't work!)
- `Makefile` in the root directory says you can either provide your own `metis-5.1.0` or `SuiteSparse` will build it for you.

For now I think we should make `metis` a non-optional dependency.

dpo: I agree. Let's make it mandatory.

suite-sparse.rb
dpo: For an example of this, see, e.g., `superlu.rb`.

dpo: @syarra

suite-sparse.rb
```diff
@@ -14,25 +13,20 @@ class SuiteSparse < Formula
...
-
-    # Switch to the Mac base config, per SuiteSparse README.txt
-    mv "SuiteSparse_config/SuiteSparse_config.mk",
-       "SuiteSparse_config/SuiteSparse_config_orig.mk"
-    mv "SuiteSparse_config/SuiteSparse_config_Mac.mk",
-       "SuiteSparse_config/SuiteSparse_config.mk"
-
     cflags = "#{ENV.cflags}"
-    cflags += (ENV.compiler == :clang) ? "" : " -fopenmp"
+    cflags += "-fopenmp" if build_with? "openmp"
```
dpo: Should be `build.with?`.

syarra:
What kind of tests do we add?

dpo:
It would be great to compile and run one of the demos. They're not installed by default but it would be easy to do:
```
pkgshare.install "Demo"
```
Then you could proceed as in, e.g., `plasma` or `superlu`.

schoeps:
Is anyone still working on this pull request? I wanted to try whether octave works with SuiteSparse 4.5.1  such that we can drop support of the old 4.2.1. However, I get the same error as on Travis

`clang: error: no such file or directory: '/private/tmp/suite-sparse20160402-49495-1fgxbia/SuiteSparse/lib/libmetis.dylib'`

suite-sparse.rb
```diff
@@ -14,25 +13,20 @@ class SuiteSparse < Formula
 
   option "with-matlab", "Install Matlab interfaces and tools"
   option "with-matlab-path=", "Path to Matlab executable (default: matlab)"
+  option "with-openmp", "Build with OpenMP support"
```
schoeps: add `option "without-metis", "Use Metis shipped with SuiteSparse (not recommended)"`

suite-sparse.rb
```diff
@@ -14,25 +13,20 @@ class SuiteSparse < Formula
...
 
+  depends_on "metis"  # SuiteSparse must be compiled with metis 5.
```
schoeps: change to `depends_on "metis" if build.with? "metis"`

schoeps:
Option metis was missing. See inline comments.

suite-sparse.rb
syarra: `metis` isn't an optional dependancy. SuiteSparse needs `metis` but fails to build with homebrew if it is build with the `metis` shipped with SuiteSparse.

schoeps:
I rather suggest to keep "with-metis" as an option for backward compatibility. Other packages (like octave, cf. line 171) read this option.

dpo:
I had a similar situatuon in `parmetis`. I'd much rather build against Homebrew metis if possible, or there might be conflicts.

suite-sparse.rb
schoeps: ...so it's intended to be optional but a proper patch is missing to make the shipped version work!?

schoeps:
@dpo: I agree but do you suggest to drop the "--with-metis" option?

suite-sparse.rb
syarra: Yes.

dpo:
What would be the advantage of building against the `metis` that ships with SuiteSparse? For now it doesn't build and we don't have as much control (unless the versions get way out of sync, which I doubt). What I'd really like though is to have a good test for `suite-sparse`.

schoeps:
@syarra since this pull request should be merged before octave, I made a proposal for a simple test: 
https://github.com/schoeps/homebrew-science/blob/suitesparse/suite-sparse.rb

dpo:
See #3515.

PR #3386, opened 2016-03-01
r: Add gettext (et al) to *ALL* LDFLAGS variables

DabeDotCom:
[Note: This is a cleaned-up version of [#3359](https://github.com/Homebrew/homebrew-science/pull/3359), which I've closed.  -dabe]

While trying to install the 'RInside' package, I ran into the following:
```
prompt% echo 'install.packages("RInside")' | sudo R --no-save
...
clang++ -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/usr/local/lib -L/usr/local/Cellar/r/3.2.3/R.framework/Resources/lib -L/usr/local/opt/gettext/lib -L/usr/local/opt/readline/lib -L/usr/local/opt/openssl/lib -L/usr/local/lib -o RInside.so MemBuf.o RInside.o -F/usr/local/Cellar/r/3.2.3/R.framework/.. -framework R -lintl -Wl,-framework -Wl,CoreFoundation
clang++ -o libRInside.dylib MemBuf.o RInside.o -dynamiclib -Wl,-headerpad_max_install_names  -undefined dynamic_lookup -single_module -multiply_defined suppress -L/usr/local/lib   -F/usr/local/Cellar/r/3.2.3/R.framework/.. -framework R -lintl -Wl,-framework -Wl,CoreFoundation
ld: library not found for -lintl
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [libRInside] Error 1
ERROR: compilation failed for package ‘RInside’

prompt% clang++ --version
Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix
```

It would appear that "-L/usr/local/opt/gettext/lib" (et al) didn't make it into SHLIB_CXXLDFLAGS.


rundel:
On a related note, there is also an issue with how the CPPFLAGS are handled in Makevar as well. The current formula appends `-I#{HOMEBREW_PREFIX}/include"` to every variant of the CPPFLAGS variable which is problematic. In particular by modifying `R_XTRA_CPPFLAGS` it can cause incompatibilities by giving `-I#{HOMEBREW_PREFIX}/include"` preference over local package includes.

Specifically, this causes an issue with the igraph package when suite-sparse is installed (see this [issue](https://github.com/igraph/rigraph/issues/135)) since R will attempt to use the homebrew headers for CHOLMOD instead of the packages version.

Fixing [this line](https://github.com/Homebrew/homebrew-science/blob/master/r.rb#L143) to `s.gsub! /^CPPFLAGS =.*/, "\\0 -I#{HOMEBREW_PREFIX}/include"` solves the problem.

DabeDotCom:
Man, it bums me out when a CI build fails because of another, completely unrelated dependency:

```
==> brew install rampart
==> Installing homebrew/science/rampart dependency: seqan
==> Downloading http://packages.seqan.de/seqan-library/seqan-library-2.1.0.tar.xz
Error: SHA256 mismatch
Expected: 97bbeffcb059d650e4bdf6ffcd3d585910a382525a1a5054941eb080df0c88d2
Actual: 986a55d028fb08e0befc538af927cbcfb880e7d4402f9b008d09f76abe63f86e
```

How does one ask the BrewBuildBot to try again?

tdsmith:
Can you add `revision 1` under the sha256 to bump the Homebrew package version? I'd like all users to get this update. If you can incorporate @rundel's change, that would be A+ too.

Don't worry about the CI failure; I'll go slap rampart around.

DabeDotCom:
@tdsmith: "Done."

(I wasn't sure if it was correct to bump the revision number when I didn't know the SHAs, yet.  Thanks for the tip!)

r.rb
```diff
@@ -12,7 +12,7 @@ class R < Formula
   sha256 "b93b7d878138279234160f007cb9b7f81b8a72c012a15566e9ec5395cfd9b6c1"
```
tdsmith: Oops; I wanted to bump the formula revision, not the bottle revision (which is automatic). Can you add `revision 1` up here on the class body? I'm pestering you because it needs to round-trip through our CI to generate the bottle.

DabeDotCom:
@tdsmith: No worries!  I'm always happy to help (and it helps me wrap my head around the whole process...)

tdsmith:
I'll take care of bumping littler. Thanks for your patience and your pull requests! :tada: Homebrew thrives because of contributions like this. :)

PR #3390, opened 2016-03-02
graph-tool: 2.18

count0:

### DIFF
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -56,7 +56,7 @@ class GraphTool < Formula
 
   fails_with :clang do
     cause "Older versions of clang have buggy c++14 support."
-    build 700
+    build 702
   end
 
   fails_with :gcc => "4.8" do
@@ -66,11 +66,12 @@ class GraphTool < Formula
     cause "We need GCC 5.0 or above for sufficient c++14 support"
   end
 
-  if MacOS.version == :mavericks && (Tab.for_name("boost").stdlib == "libcxx" || Tab.for_name("boost-python").stdlib == "libcxx")
-    odie "boost and boost-python must be built against libstdc++ on Mavericks. One way to achieve this, is to use GCC to compile both libraries."
-  end
-
   def install
+
+    if MacOS.version == :mavericks && (Tab.for_name("boost").stdlib == "libcxx" || Tab.for_name("boost-python").stdlib == "libcxx")
+      odie "boost and boost-python must be built against libstdc++ on Mavericks. One way to achieve this, is to use GCC to compile both libraries."
+    end
+
     system "./autogen.sh" if build.head?
 
     config_args = %W[
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -56,7 +56,7 @@ class GraphTool < Formula
 
   fails_with :clang do
     cause "Older versions of clang have buggy c++14 support."
-    build 703
+    build 702
   end
 
   fails_with :gcc => "4.8" do
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -43,6 +43,9 @@ class GraphTool < Formula
   if build.with? "cairo"
     depends_on "py2cairo" if build.with? "python"
     depends_on "py3cairo" if build.with? "python3"
+    depends_on "cairo" => :linked
+    depends_on "libsigc++" => :linked
+    depends_on "gmp" => :linked
   end
 
   if build.with? "gtk+3"
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -69,12 +69,18 @@ class GraphTool < Formula
     cause "We need GCC 5.0 or above for sufficient c++14 support"
   end
 
+  if MacOS.version == :mavericks
+    fails_with :gcc => "6" do
+      cause "GCC 6 fails with 'Internal compiler error' on Mavericks. You should install GCC 5 instead with 'brew tap homebrew/versions; brew install gcc5"
+    end
+  end
+
   def install
 
     if MacOS.version == :mavericks && (Tab.for_name("boost").stdlib == "libcxx" || Tab.for_name("boost-python").stdlib == "libcxx")
       odie "boost and boost-python must be built against libstdc++ on Mavericks. One way to achieve this, is to use GCC to compile both libraries."
     end
-
+  
     system "./autogen.sh" if build.head?
 
     config_args = %W[
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -22,6 +22,7 @@ class GraphTool < Formula
   option "without-numpy", "Use a numpy you've installed yourself instead of a Homebrew-packaged numpy"
   option "without-python", "Build without python2 support"
   option "without-scipy", "Use a scipy you've installed yourself instead of a Homebrew-packaged scipy"
+  option "with-openmp", "Enable parallel algorithms with OpenMP (requires GCC or clang >= 3.8)"
 
   cxx11 = MacOS.version < :mavericks ? ["c++11"] : []
   with_pythons = build.with?("python3") ? ["with-python3"] : []
@@ -75,6 +76,7 @@ class GraphTool < Formula
     ENV.append "CXXFLAGS", "-fext-numeric-literals" unless ENV.compiler == :clang
     config_args << "--disable-cairo" if build.without? "cairo"
     config_args << "--disable-sparsehash" if build.without? "google-sparsehash"
+    config_args << "--enable-openmp" if build.with? "openmp"
 
     Language::Python.each_python(build) do |python, version|
       config_args_x = ["PYTHON=#{python}"]
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -65,7 +65,7 @@ class GraphTool < Formula
   end
   
   fails_with :clang => "3.7" do
-    version 3.7.0
+    version "3.7.0"
     cause "We need clang version 3.7.1 or above, to avoid compiler bugs with undefined symbols"
   end
 
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -66,6 +66,10 @@ class GraphTool < Formula
     cause "We need GCC 5.0 or above for sufficient c++14 support"
   end
 
+  if MacOS.version == :mavericks && (Tab.for_name("boost").stdlib == "libcxx" || Tab.for_name("boost-python").stdlib == "libcxx")
+    odie "boost and boost-python must be built against libstdc++ on Mavericks. One way to achieve this, is to use GCC to compile both libraries."
+  end
+
   def install
     system "./autogen.sh" if build.head?
 
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -56,7 +56,7 @@ class GraphTool < Formula
 
   fails_with :clang do
     cause "Older versions of clang have buggy c++14 support."
-    build 702
+    build 703
   end
 
   fails_with :gcc => "4.8" do
--- a/graph-tool.rb
+++ b/graph-tool.rb
@@ -54,7 +54,9 @@ class GraphTool < Formula
   # We need a compiler with C++14 support.
   fails_with :llvm
 
-  fails_with :clang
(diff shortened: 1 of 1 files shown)


graph-tool.rb
```diff
@@ -30,12 +11,6 @@ class GraphTool < Formula
     depends_on "libtool" => :build
   end
 
-  bottle do
-    sha256 "1403a9699ab147da134b39d4cc3b64cd1a49be1d7d9b5941a842f87f5b9132b6" => :el_capitan
-    sha256 "c0259303a55befb147a40d60e58e283d2b9774d5a52b9d7f13b8eacc8e0d063f" => :yosemite
-    sha256 "7f2ecc68e838a493bcfe533b36c94d1e2b7dc90bd03ea0e1453f94bc1d80ef11" => :mavericks
-  end
-
```
dpo: Please don't alter the `bottle` section.

dpo:
```
Error: 3 problems in 1 formula
homebrew/science/graph-tool:
 * revision should be removed
 * 13: Trailing whitespace was found
 * 19: Trailing whitespace was found
```
Please build and audit locally before submitting.

count0:
@dpo can't really build locally, sorry.

count0:
@dpo One of the test failures (on 10.10) was due to a timeout. Would it be possible to restart it? The failure on 10.9 ("dyld: lazy symbol binding failed") seems like a compiler/linker bug, since the symbols in question need to be generated automatically by the compiler. Is there a mechanism to flag/exclude a given OS version?

dpo:
Doesn't look like an appropriate compiler was selected.

count0:
@dpo This is a bit of whack-a-mole... The only compilers that will work are clang >= 3.7.1 and gcc >=5, but here we have to rule out all others one by one. Is there a more straightforward way?

count0:
@dpo I'm quite stuck here. I think I might drop this merge request. In macports this works flawlessly, since it easier to select the right compiler. But here we have to flag all that **don't** work, but there is only two that will: the newest vanilla clang and gcc. I don't even know all other possible compiler versions and build numbers out there in homebrew...

dpo:
@count0 Should we simply force gcc using `fails_with :clang` without specifying a build number?

count0:
@dpo GCC should work, but it uses libstdc++ per default. Clang uses libc++. These two don't mix. Hence, if something else (e.g. boost) was compiled with clang/libc++ it will not work with graph-tool compiled with gcc/libstdc++. It could be that homebrew does things right and always selects the same C++ library independent of compiler, but I'm not sure.

dpo:
That works, but how about #3532? It seems to succeed on 10.9 with your previous exclusions.

count0:
@dpo I updated it to the current version too.

dpo:
@count0 Sorry for the back-and-forth but I just found [this](https://trac.macports.org/changeset/146449) linked from one of your issues. How about
```ruby
fails_with :clang do
  build 700.1.81
end
```
but it seems it might not be necessary with version 2.14?!

count0:
AFAIK nothing really changed in version 2.14 with regards to this problem... If the current code does not trigger the bug any longer, it would be by accident. However, I have not seen any evidence of this.

I had not seen this specific build mask from macports.  I will put it in the formula, and we will see how far we go.

dpo:
Sorry, should be `build 700`.

dpo:
ps: why depend on `homebrew/science/numpy` explicitly? Is there something specific about that `numpy`, or could you just say `depends_on "numpy" => :python`?

count0:
I can't see any particular reason (I'm not the original author of this formula). We do need the C API from numpy to be available, but it should be there in any complete numpy installation.

But let us wait and see if the current version works at all. If it does, we can make this other change.

dpo:
Locally, on 10.9, it builds but I get the same lazy symbol error.

count0:
Which version of clang got selected?

dpo:
gcc-5 because I only have clang 600 on this machine.

count0:
Weird, I've never seen this issue on GCC before. Maybe it is related to libc++, not clang itself?

What specific version of GCC 5 do you have?

dpo:
I have gcc 5.3.0.

dpo:
If `boost-python` is built with clang, it uses libc++.

count0:
@dpo OK, then this is most likely a libc++ issue. Can you verify which version do you have installed?

dpo:
I have libc++ 1.0. Is there a leaner version of `graph-tool` we could experiment with? It takes forever to build.

dpo:
I don't get the lazy symbol error when `boost`, `boost-python` and `graph-tool` are all compiled with gcc-5 on OSX 10.9.

count0:
Ok, so that was indeed a compiler issue. It must also work with vanilla clang >= 3.7.

dpo:
Ok, but I'm not sure we can force dependencies to also be built with a certain compiler.

count0:
Actually, since you used GCC for everything, it probably used libstdc++, right? Hence we cannot rule out that the problem is actually with libc++. It is possible that we can have only graph-tool compiled with GCC + libc++, provided libc++ is new enough. However, I'm not even sure homebrew has a separate formula for libc++; it could just come bundled with clang. If that is the case, we are bach at square one with the whole problem of selecting the right clang version...

count0:
So, from what I understand this really affects only 10.9, since clang 3.7 is not available there? Would a reasonable compromise be simply to display a warning for 10.9 that boost and boost_python must be compiled with the same compiler?

dpo:
There's no formula for libc++ as it ships with clang. But we can query how `boost` and `boost-python` were installed. For example:
```
$ brew irb
==> Interactive Homebrew Shell
Example commands available with: brew irb --examples
irb(main):004:0> Tab.for_name("boost").stdlib
=> "libcxx"
irb(main):005:0> Tab.for_name("boost").compiler
=> "clang"
irb(main):007:0> Tab.for_name("boost-python").stdlib
=> "libcxx"
irb(main):008:0> Tab.for_name("boost-python").compiler
=> "clang"
```
On OSX 10.9, the latest Xcode is version 6.2, so clang > 700 is not even available. We can simply fail in that case I guess. Something along the lines of
```ruby
fails_with :clang
  build 700
end

...

if MacOS.version == :mavericks && (Tab.for_name("boost").stdlib == "libcxx" || Tab.for_name("boost-python").stdlib == "libcxx")
  odie "boost and boost-python must be built against libstdc++ on Mavericks"
end
```
I'm not quite sure how the build bot will react though.

count0:
Well, let's try.

graph-tool.rb
```diff
@@ -66,6 +66,10 @@ class GraphTool < Formula
     cause "We need GCC 5.0 or above for sufficient c++14 support"
   end
 
+  if MacOS.version == :mavericks && (Tab.for_name("boost").stdlib == "libcxx" || Tab.for_name("boost-python").stdlib == "libcxx")
+    odie "boost and boost-python must be built against libstdc++ on Mavericks. One way to achieve this, is to use GCC to compile both libraries."
+  end
+
```
dpo: Sorry, this should be in `install`.

dpo:
It failed correctly on 10.9 but it still fails due to the lazy symbol thing on 10.10. Might need to request clang >= 703.

dpo:
@BrewTestBot test this please

dpo:
@count0 Sorry for the long delay. Do you want to update your PR to 2.16 and we'll try and wrap this up?

kikohs:
Hey, what's the status of this formula ? Could it be possible to have the latest graph-tool? Thanks

count0:
@dpo I've updated to version 2.16, but I'm still at loss on how to fix this. I simply can't hit the mark with the clang whitelisting/blacklisting that satisfies the test bot.

dpo:
OSX 10.9 is tricky because there's a requirement that `boost` and `boost-python` be built with gcc. But there's also a build error on 10.10:
```
In file included from /usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/tuple:39:0,
                 from /usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/functional:55,
                 from /usr/local/include/boost/config/no_tr1/functional.hpp:21,
                 from /usr/local/include/boost/function/detail/prologue.hpp:14,
                 from /usr/local/include/boost/function/function_template.hpp:13,
                 from /usr/local/include/boost/function/detail/maybe_include.hpp:13,
                 from /usr/local/include/boost/function/function0.hpp:11,
                 from /usr/local/include/boost/python/errors.hpp:13,
                 from /usr/local/include/boost/python/handle.hpp:11,
                 from /usr/local/include/boost/python/args_fwd.hpp:10,
                 from /usr/local/include/boost/python/args.hpp:10,
                 from /usr/local/include/boost/python.hpp:11,
                 from ../../../../src/graph/community_old/graph_blockmodel_covariates.cc:19:
/usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/array: In constructor 'constexpr std::array<std::pair<long unsigned int, long unsigned int>, 2ul>::array()':
/usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/array:90:12:   in constexpr expansion of 'std::pair<long unsigned int, long unsigned int>()'
/usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/array:90:12: internal compiler error: in cxx_eval_call_expression, at cp/constexpr.c:1449
     struct array
            ^~~~~
```

count0:
Internal compiler errors are, by definition, compiler bugs. Hence we need to see if clang will work for 10.10 instead, but I think we went that road before.

dpo:
2.16 built for me on 10.10 with clang but not with gcc-6.1.

count0:
@dpo and what clang build would that be?

dpo:
I have build 700.1.81.

count0:
@dpo We have a fail on mavericks only because of the libstdc++ and libc++ thing. Yosemite and  El Capitain seem to be working.

dpo:
Working on it.

dpo:
I forced compilation of `boost` and `boost-python` with libstdc++ but now I'm getting an internal compiler error on Mavericks:
```
/usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/array: In constructor 'constexpr std::array<std::pair<long unsigned int, long unsigned int>, 2ul>::      array()':
/usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/array:90:12:   in constexpr expansion of 'std::pair<long unsigned int, long unsigned int>()'
/usr/local/Cellar/gcc/6.1.0/include/c++/6.1.0/array:90:12: internal compiler error: in cxx_eval_call_expression, at cp/constexpr.c:1449
     struct array
            ^~~~~
```
It's not the first time I'm regretting the gcc 6.1 upgrade. I've encountered internal errors in gfortran too. I'm not sure why they don't show up on 10.10 and 10.11.

At this point, I think we just have to disable the bottle for Mavericks.

count0:
I agree with just disabling it for Mavericks. Should I just remove the line from the formula?

dpo:
Because there's no syntax to explicitly disable bottling for one architecture, let's add this to the formula:
```ruby
if MacOS.version == :mavericks
  fails_with :gcc => "6" do
    cause "Internal compiler error"
  end
end
```
which we'll hopefully be able to remove at the next gcc upgrade. `graph-tool` builds on 10.9 with gcc-5, so users will just have to `brew tap homebrew/versions`, `brew install gcc5` and use it to build `boost`, `boost-python` and `graph-tool`. Does that sound reasonable?

count0:
@dpo Yes, that is reasonable. I've made the change.

dpo:
It seems we're good to go. Would you mind squashing all your commits?

count0:
Can't you do this automatically when doing merge? i.e. https://help.github.com/articles/about-pull-request-merge-squashing/

Otherwise I have to close this pull request, and do another one.

dpo:
Nevermind, I'll do it. It doesn't happen automatically because there's more to just merging a Homebrew PR.

dpo:
@BrewTestBot test this please

count0:
This is awesome, thanks.

Issue #3385, opened 2016-02-29
masurca: Problems installing MaSuRCA and GLIBC using Linuxbrew


philippbayer:
As discussed on Twitter in this [thread](https://twitter.com/sjackman/status/704025671845040128)

Currently the installation of MaSuRCA fails with ([entire log here](https://gist.github.com/philippbayer/3c83b7131b3433e2c1e2))

```
cd /tmp/masurca20160229-8746-1mt8ybf/MaSuRCA-2.3.2b/CA/Linux-amd64/obj; \
          /gpfs1/homes/pbayer/.linuxbrew/bin/g++-5 -o /tmp/masurca20160229-8746-1mt8ybf/MaSuRCA-2.3.2b/CA/Linux-amd64/bin/fragmentDepth /tmp/masurca20160229-8746-1mt8ybf/MaSuRCA-2.3.2b/CA/Linux-amd64/obj/fragmentDepth.o /tmp/masurca20160229-8746-1mt8ybf/MaSuRCA-2.3.2b/CA/Linux-amd64/lib/libCA.a   -L/tmp/masurca20160229-8746-1mt8ybf/MaSuRCA-2.3.2b/CA/Linux-amd64/lib  -L/tmp/masurca20160229-8746-1mt8ybf/MaSuRCA-2.3.2b/CA/Linux-amd64/lib  -L/usr/lib64  -L/usr/X11R6/lib64 -L/gpfs1/homes/pbayer/.linuxbrew/lib -Wl,--dynamic-linker=/gpfs1/homes/pbayer/.linuxbrew/lib/ld.so -Wl,-rpath,/gpfs1/homes/pbayer/.linuxbrew/lib -pthread -lm -Wl,-O1 -pthread -lm -Wl,-O1
++++++++++++++++++++  markUniqueUnique ++++++++++++++++++++++
cd /tmp/masurca20160229-8746-1mt8ybf/MaSuRCA-2.3.2b/CA/Linux-amd64/obj; \
[...]

PR #3397, opened 2016-03-03
MaSuRCA: fix undefined reference to clock_gettime@GLIBC_2.17

philippbayer:
This fixes issue #3385 (on my system) and adds a description line to satisfy `brew audit`. It may not be necessary to restrict the fix to Linux only, not sure how Mac systems work in that regard.


sjackman:
https://travis-ci.org/Homebrew/homebrew-science/jobs/113302964#L876
```
############################## AS_OVL ##############################
make[1]: *** No rule to make target `jellyfish/jellyfish.hpp', needed by `AS_OVL_overlap.o'.  Stop.
make: *** [objs] Error 1
```
I'm not sure why this error should come up now and not before. It doesn't seem related to this patch. It might require `ENV.deparallelize` to fix. In any case, I'm going to merge this patch without updating bottles, as it doesn't affect Mac.

sjackman:
Merged! Thanks for your contribution, Philipp.

PR #3414, opened 2016-03-10
dynare: fix formula that does not work with new version of flex, building against flex 2.5.39 as resource

houtanb:
The update to `flex` broke this formula. As upstream won't update the formula until the new `flex` is in Debian stable, add `flex 2.5.39` as a resource to this formula and use it in the build process as per Homebrew/homebrew-versions#1114


dpo:
@BrewTestBot test this please

dpo:
Do you not want this merged any more?

houtanb:
@dpo we ended up fixing our code for the new `flex` version, so the `dynare.rb` build should be ok without this change. I will test it soon and make a new PR if that's not the case

PR #3448, opened 2016-03-16
minia 2.0.3 [WIP]

sjackman:
https://gist.github.com/sjackman/e5bde2df6f5d2d17c259#file-02-make-L754
```
[ 82%] Built target hdf5_tools
[ 82%] Building CXX object ext/gatb-core/src/CMakeFiles/gatbcore-static.dir/gatb/bank/impl/BankRandom.cpp.o
cd /tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/build/ext/gatb-core/src && /Users/sjackman/.homebrew/Library/ENV/4.3/clang++    -I/tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/build/ext/gatb-core/include -I/tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/src -I/tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/thirdparty  -DNDEBUG    -O3  -DINT128_FOUND  -DWITH_LAMBDA_EXPRESSION -std=c++0x    -DWITH_MPHF  -DDONT_USE_TR1  -DBOOST_NO_CXX11_RVALUE_REFERENCES=1 -Wno-invalid-offsetof -o CMakeFiles/gatbcore-static.dir/gatb/bank/impl/BankRandom.cpp.o -c /tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/src/gatb/bank/impl/BankRandom.cpp
In file included from /tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/src/gatb/bank/impl/BankConverterAlgorithm.cpp:20:
In file included from /tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/src/gatb/bank/impl/BankConverterAlgorithm.hpp:33:
In file included from /tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/src/gatb/tools/storage/impl/Storage.hpp:45:
In file included from /tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/src/gatb/tools/math/NativeInt8.hpp:33:
/tmp/minia20160316-71438-1fl6wf2/minia-2.0.3-Source/thirdparty/gatb-core/src/gatb/tools/misc/api/Abundance.hpp:34:10: fatal error: 'hdf5/hdf5.h' file not found
#include <hdf5/hdf5.h>
```


sjackman:
@rchikhi Any thoughts on this build error?

rchikhi:
Hmm the following recipe seems to have found the same bug and fixed it:
https://github.com/d235j/homebrew-science/blob/master/gatb.rb

Trick is to remove "-DCMAKE_BUILD_TYPE=Release" from cmake arguments.

sjackman:
Thanks, Rayan.

sjackman:
Travis fails because `latex` and `imagemagick` are missing.

sjackman:
@Homebrew/science Minia depends by default on LaTeX to build documentation. Is it possible to add a dependency on LaTeX, or is the preferred solution to disable documentation?

apjanke:
LaTeX is a giant dependency (like, >1GB and a 10 minute install on a beefy machine). It's possible (with the `:tex` requirement), but I think we prefer to disable TeX-dependent documentation by default and make `--with-tex` optional, even if it's on by default in upstream. See the [`gnuplot` formula in `homebrew/core`](https://github.com/Homebrew/homebrew-core/blob/a3c7a2118a6d6d60750c1919e1617a241ee81de5/Formula/gnuplot.rb) for an example.

(User-friendly distributions will include their TeX-generated doco as part of the source tarball, IMHO.)

apjanke:
(Apologies if that's not true for `homebrew/science` specifically. I realize I'm new here and may be speaking out of turn.)

dpo:
I believe you can say `depends_on :tex` and that will require the user to obtain LaTeX; Homebrew won't install it. So for instance, `dynare.rb` has `depends_on :tex => :build` and `pcb.rb` has `depends_on :tex if build.with? "doc"`. The latter may be appropriate for you (with doc disabled by default so the bot can build bottles)? It's easy to `brew cask install mactex`.

sjackman:
@rchikhi Hi, Rayan. Would you consider including the rendered documentation with `minia` to eliminate the LaTeX dependency of `minia`? Currently I'm using `-DSKIP_DOC=1` to skip building the documentation and avoid the LaTeX dependency.

sjackman:
Minia 2.0.3 is in Homebrew-science now.

sjackman:
A minor gaff. It should have used
```ruby
depends_on :tex => :optional
```
rather than
```ruby
  option "with-docs", "Install documentation. Requires LaTeX"
```

sjackman:
Well, may not actually, because building the documentation requires both `:tex` and `imagemagic`, so the one `--with-docs` option to pull in both dependencies is probably better.

ilovezfs:
It's usually best to use `depends_on :tex => [:build, :optional]` and `depends_on "imagemagic" => :build if build.with? "tex"`. See `camlp5.rb` in core.

sjackman:
Changed. Thanks for the tip, Joe.

rchikhi:
Hi Shaun,

Yes, good idea, we'll drop the latex dependency in the upcoming 3.0 releases.

Rayan

sjackman:
Thanks, Rayan.

rchikhi:
Thanks for putting minia on homebrew :)

sjackman:
No worries at all!

PR #3452, opened 2016-03-17
snpeff: updated to 4.2

heuermh:
Looking at previous revisions, I wonder why SnpSift was installed with different case (snpEff vs. SnpSift), e.g.
https://github.com/Homebrew/homebrew-science/blob/master/snpeff.rb#L29
https://github.com/Homebrew/homebrew-science/blob/master/snpeff.rb#L42

Should that also be fixed in this pull request?


tseemann:
@heuermh they are inconsistent with upper/lower case because that's how the author names them (at least in 4.1l it was), Has that changed?

heuermh:
@tseemann You're right!

Even though the tools are referred to as SnpEff and SnpSift on the web page, the jar files have different case
```
snpEff/snpEff.jar
snpEff/SnpSift.jar
```
I will leave it as is then.

Is there anything else holding back this pull request?

tseemann:
@BrewTestBot test this please

dpo:
@BrewTestBot test this please

heuermh:
Thanks!

PR #3455, opened 2016-03-20
nglib 5.3.1 (new formula)

ianrrees:
nglib is a tetrahedral meshing library, which we use in FreeCAD to generate meshes for finite element simulations.

### Have you:

### New formula: have you


ianrrees:
Changed the download URL to pass the audit AND resolve correctly!

FWIW - Hadn't caught this before submitting as the original download URL I used for development worked, but then the audit tool suggested changing http to https and using downloads.sourceforge.  Unfortunately, the rest of the URL is subtly different in that case, but homebrew's cache already had the right file with the same name...

ianrrees:
I suspect the CI failure is due to one of the patches having windows-style line endings, where the others have unix ones.  What's the preferred way to fix this?

ianrrees:
sed to the rescue!  I couldn't figure out a way to say "if Formula["opencascade"].version >= 6.8.1" to put around the sed replacements, but can't imagine there's much interest in nglib with old versions of opencascade.

nglib.rb
```diff
@@ -0,0 +1,113 @@
...
+      }
+      int main(int argc, char **argv) {
+          nglib::Ng_Init();
+          auto mesh(nglib::Ng_NewMesh());
+          nglib::Ng_DeleteMesh(mesh);
+          nglib::Ng_Exit();
+          return 0;
+      }
+    EOS
+    system ENV.cxx, "-Wall", "-o", "test", "test.cpp", "-std=c++11",
```
dpo: Does the library need to be compiled in c++11 mode?

ianrrees: No, the library is all C++98/03 AFAICT.

I only used C++11 for the test out of habit, could change the "auto" to "nglib::Ng_Mesh" if it's better to compile the test case without C++11 support?

dpo: It might be simpler to remove c++11 support. For a moment I wondered if the entire formula needed to depend on `:cxx11`.

ianrrees: @dpo - sorry if you've already gotten a notification for this, just wanted to mention that I changed the test to use default C++ standard.  Thanks!

dpo:
Thanks!

PR #3462, opened 2016-03-23
diamond formula compatible with linuxbrew and homebrew

schwager-hsph:
Deals with issue #1593 by building boost 1.57 locally as per README.rst instructions.

### DIFF
--- a/diamond.rb
+++ b/diamond.rb
@@ -7,6 +7,12 @@ class Diamond < Formula
   url "https://github.com/bbuchfink/diamond/archive/v0.7.9.tar.gz"
   sha256 "25dc43e41768f7a41c98b8b1dcf5aa2c51c0eaf62e71bff22ad01c97b663d341"
 
+  bottle do
+    sha256 "e0eb3edc6f875a6b8e37b2d369b27510ec714faf2e94ff414edb94fbd34a1141" => :yosemite
+    sha256 "da0750e96465902fbd7827dc2220c2cb298f71ae488d6175885eb2044e2066ad" => :mavericks
+    sha256 "459c5a98274de600b7a270e51a589c21bae1add6384fe76d4a4b5c3d988778cb" => :mountain_lion
+  end
+  
   depends_on "wget" => :build
 
   def install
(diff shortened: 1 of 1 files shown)


diamond.rb
```diff
@@ -13,13 +13,11 @@ class Diamond < Formula
...
 
-  depends_on "boost"
+  depends_on "wget" => :build
 
   def install
     Dir.chdir("src") do
-      inreplace "Makefile", "-Iboost/include", "-I#{Formula["boost"].include}"
-      inreplace "Makefile", "LIBS=-l", "LIBS=-L#{Formula["boost"].lib} -l"
-      inreplace "Makefile", "-lboost_thread", "-lboost_thread-mt"
+      system "./install-boost"
```
dpo: I don't think we want that. Have you reported any issue to Linuxbrew?

schwager-hsph: The install-boost script downloads and compiles boost for use only within the diamond source.
Essentially, the script downloads boost 1.57.0 and compiles with these commands:
```
./bootstrap.sh --with-libraries=timer,chrono,system,program_options,thread,iostreams --prefix=../boost
./b2 install
```
Here's the guts of the script: https://github.com/bbuchfink/diamond/blob/master/src/install-boost

I can re-implement the install-boost script as a resource if that's kosher?

schwager-hsph: Also, I didn't report an issue because I'd be reporting a dupe of #1909.
At present, the formula downloads the boost 1.60 bottle. Since the 1.60 bottle is compiled with clang and since I'm compiling diamond with gcc, when I try to compile diamond I get linker errors as described in #1593 and #1909.

schwager-hsph:
No word yet on whether you'd like me to leave it as-is, download and compile boost 1.57 as the install-boost script does, or something else.
What do I need to do to get this pushed through?

tomck:
@schwager-hsph There's a difference in compiling between the 0.7.x series and the 0.8.x series, as referenced in my final comment in Homebrew/homebrew-science#3365 - not sure how to resolve yet

ilovezfs:
You need to use CMake.

ilovezfs:
@schwager-hsph I'm closing this as "stale" because I've just pushed diamond 0.8.16 in https://github.com/Homebrew/homebrew-science/commit/6e8a1bcc406ef212d286f700e50c0b4eb34c5aff. Please let us know if you still have any issues compiling, etc.

PR #3464, opened 2016-03-23
gmsh update 2.12.0

schoeps:
update to version 2.12.0 and some clean ups according to `brew audit --strict --online gmsh`


schoeps:
I do not really understand the problem of travis. gmsh compilation was ok but then the old getdp (version 2.7) was loaded... Can someone e.g. @dpo  explain the issue to me? thanks.

dpo:
Because `getdp` depends on `gmsh`, Homebrew tries to run the `getdp` tests after updating the `gmsh` bottles to check ABI compatibility. The issue is that `getdp` links to the versioned `gmsh` library.

dpo:
I'll pull and rerun the `getdp` build.

PR #3466, opened 2016-03-24
parallel-netcdf 1.7.0 (new formula)

mkrapp:
Here's a new formula for `parallel-netcdf`, which works with and without MPI support.


parallel-netcdf.rb
```diff
@@ -0,0 +1,172 @@
+class ParallelNetcdf < Formula
+  desc "NetCDF library for parallel I/O"
+  homepage "https://trac.mcs.anl.gov/projects/parallel-netcdf"
+  url "http://cucis.ece.northwestern.edu/projects/PnetCDF/Release/parallel-netcdf-1.7.0.tar.gz"
+  sha256 "52f0d106c470a843c6176318141f74a21e6ece3f70ee8fe261c6b93e35f70a94"
+
+  deprecated_option "without-check" => "without-test"
```
dpo: You don't need this for new formulas.

parallel-netcdf.rb
```diff
@@ -0,0 +1,172 @@
...
+
+  option "without-cxx", "Don't compile C++ bindings"
+  option "without-fortran", "Don't compile Fortran bindings"
+  option "without-test", "Disable checks (not recommended)"
+  option "with-mpi", "Enable parallel compiling"
+
+  depends_on :mpi => [:optional, :cc, :cxx, :f90]
```
dpo: Line 12 is redundant with this. You only need line 14.

parallel-netcdf.rb
```diff
@@ -0,0 +1,172 @@
+class ParallelNetcdf < Formula
+  desc "NetCDF library for parallel I/O"
+  homepage "https://trac.mcs.anl.gov/projects/parallel-netcdf"
+  url "http://cucis.ece.northwestern.edu/projects/PnetCDF/Release/parallel-netcdf-1.7.0.tar.gz"
+  sha256 "52f0d106c470a843c6176318141f74a21e6ece3f70ee8fe261c6b93e35f70a94"
+
+  deprecated_option "without-check" => "without-test"
+
+  option "without-cxx", "Don't compile C++ bindings"
+  option "without-fortran", "Don't compile Fortran bindings"
```
dpo: Here you need `depends_on :fortran => :optional` instead.

parallel-netcdf.rb
```diff
@@ -0,0 +1,172 @@
...
+  depends_on :mpi => [:optional, :cc, :cxx, :f90]
+
+  def install
+    common_args = %W[
+      --prefix=#{prefix}
+      --mandir=#{man}]
+
+    args = common_args.clone
+    if build.with? "mpi"
+      args << "--enable-mpi"
```
dpo: What's the point of having both `args` and `common_args`?

parallel-netcdf.rb
```diff
@@ -0,0 +1,172 @@
...
+          // global attributes:
+                      :string = "Hello World\n",
+              "" ;
+          data:
+
+               v1 = 0, 1, 2, 3 ;
+
+               v2 = 0, 1, 2, 3 ;
+          }
+      */
```
dpo: Wondering if all those comments are necessary for the test. Isn't this file found elsewhere?

parallel-netcdf.rb
```diff
@@ -0,0 +1,172 @@
...
+          MPI_Finalize();
+          return 0;
+      }
+    EOS
+    system ENV.cc, "test.c", "-o", "test"
+    assert_match "version: 1.7", shell_output("./test")
+    if build.with? "mpi"
+      system "mpicc", "test_mpi.c", "-L#{lib}", "-I#{include}", "-lpnetcdf", "-o", "test"
+      system "mpiexec", "-n", "4", "./test", "output.nc"
+      assert File.exist?("output.nc")
```
dpo: Great test! Thanks! Don't you want to check the contents of `output.nc` though?

dpo:
Will this be updated?

mkrapp:
Sorry. I forgot about this. I can work on it next week.

jonchang:
Hi @mkrapp, it looks like your build failed in `make check`:

```
/usr/local/bin/mpicc  -g -O2   -o tst_nofill tst_nofill.o -L../common  /private/tmp/parallel-netcdf-20160905-8595-1gszs3u/parallel-netcdf-1.7.0/src/lib/libpnetcdf.a -ltestutils  -lm 
rm -f ./scratch.nc
rm -f ./testfile.nc
rm -f ./tooth-fairy.nc
./nc_test -c    -d .
./nc_test -d       .
opening a nonexistent file expects NC_ENOENT, but got NC_EFILE

	FAILURE at line 370 of test_write.c: file ./scratch.nc should not exist
	### 1 FAILURES TESTING test_ncmpi_abort! Stop ... ###

./nc_test: expects 0 failures ... *** TESTING C   ./nc_test for format CDF-1                         ------ rank 0: MPI error (MPI_File_open) : MPI_ERR_FILE: invalid file
fail with 1 mismatches
make[2]: *** [testing] Error 1
make[1]: *** [check-nc_test] Error 2
make: *** [check] Error 2
```

Audit warnings:

```
==> brew audit homebrew/science/parallel-netcdf --new-formula
Error: 7 problems in 1 formula
homebrew/science/parallel-netcdf:
  * 127: Trailing whitespace was found
  * 132: Trailing whitespace was found
  * 134: Trailing whitespace was found
```

parallel-netcdf.rb
```diff
@@ -0,0 +1,145 @@
+class ParallelNetcdf < Formula
+  desc "NetCDF library for parallel I/O"
+  homepage "https://trac.mcs.anl.gov/projects/parallel-netcdf"
+  url "http://cucis.ece.northwestern.edu/projects/PnetCDF/Release/parallel-netcdf-1.7.0.tar.gz"
+  sha256 "52f0d106c470a843c6176318141f74a21e6ece3f70ee8fe261c6b93e35f70a94"
+
+  option "without-cxx", "Don't compile C++ bindings"
```
jonchang: It doesn't look like you test for this option anywhere?

mkrapp:
The check is successful when building with `mpich` but apparently the
default MPI is`open-mpi`.

jonchang:
Does the library work with `open-mpi` or is it only compatible with `mpich`?

mkrapp:
I'm not sure. It's just one simple check that leads to the fail. If I omit that check, i.e., `--without-test`, the build with `open-mpi` succeeds.

From the `parallel-netcdf` [`INSTALL`](https://trac.mcs.anl.gov/projects/parallel-netcdf/browser/trunk/INSTALL):
```
...
PnetCDF can automatically detect the available MPI compilers and compile flags.
If alternate compilers or flags are desired, they can be specified by the
following environment variables and/or configure options.
...
```

parallel-netcdf.rb
mkrapp: Will remove that line.

jonchang:
Does the test program compile? That is, if you install --without-test, and then run 

brew test parallel-netcdf

If that's the case then please just disable the test by default and file an upstream bug about the make check error and add it as a comment. When upstream fixes the issue we can re-enable the check for everyone.

jonchang:
@BrewTestBot test this please

jonchang:
I *think* this formula works now but it's encountering weird transient errors on the builders. Stay tuned...

jonchang:
It looks like a race condition on the `make install` step. Could you wrap that step in a deparallelize block, like

```
ENV.deparallelize do
 ...
end
```

and re-push?

jonchang:
This is now merged. Thanks for being so patient while wrestling with the test bot!

PR #3467, opened 2016-03-24
ckon: bump revision for boost and fix test

ilovezfs:
ckon: bump revision for boost and fix test

- attempting to run the test would throw

  dyld: Symbol not found: __ZN5boost9re_detail12perl_matcherIN9__gnu_cxx17__normal_iteratorIPKcSsEESaINS_9sub_matchIS6_EEENS_12regex_traitsIcNS_16cpp_regex_traitsIcEEEEE14construct_initERKNS_11basic_regexIcSD_EENS_15regex_constants12_match_flagsE
    Referenced from: /usr/local/Cellar/ckon/0.7.1/bin/ckon
      Expected in: /usr/local/lib/libboost_regex-mt.dylib
       in /usr/local/Cellar/ckon/0.7.1/bin/ckon

  so ckon needs a new revision built against the latest boost.

- once rebuilt, the test was still slightly broken due to changes in the
  tested repository, so fix that, and peg the test to a specific commit
  revision of tschaume/BesCocktail


ilovezfs:
@DomT4 I noticed that ckon broke on Mavericks CI during your curl PR Homebrew/homebrew#50385 so this should fix that.

DomT4:
Some of those Boost symbol errors seem to be false positives on the older CIs, we used to see a ton of them on Mountain Lion before that CI was phased out and `revision` bumps didn't resolve the issue. Consequently didn't think much of it, but if this does resolve on Mavericks with a bump :+1:.

ilovezfs:
@DomT4 Yeah it was broken on Mountain Lion, too (old bottle hanging around). This fixes it on both Mavericks and Mountain Lion. Here's the Mavericks log:
https://gist.github.com/ilovezfs/1ecc66301f7dae0e14a4

DomT4:
:+1: Nice sleuthing. Guess we should ease off widely ignoring the symbol failures in this situation.

Issue #3461, opened 2016-03-22
Error installing metis


irisvanzelst:
Hi all,

I have an issue installing `metis` on OSX 10.11.2.

I am using the following `clang`:

```bash
$ clang -v
Apple LLVM version 7.3.0 (clang-703.0.29)
Target: x86_64-apple-darwin15.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

Steps to reproduce:

```bash
$ brew tap homebrew/science
$ brew install metis
==> Installing metis from homebrew/science
==> Downloading http://glaros.dtc.umn.edu/gkhome/fetch/sw/metis/metis-5.1.0.tar.gz
Already downloaded: /Library/Caches/Homebrew/metis-5.1.0.tar.gz
==> make config shared=1 prefix=/usr/local/Cellar/metis/5.1.0 openmp=0
==> make install
Last 15 lines from /Users/irisvanzelst/Library/Logs/Homebrew/metis/02.make:
[ 67%] Building C object libmetis/CMakeFiles/metis.dir/timing.c.o
cd /tmp/metis20160322-8216-15gx1te/metis-5.1.0/build/Darwin-x86_64/libmetis && /usr/local/Library/ENV/4.3/clang  -Dmetis_EXPORTS -I/tmp/metis20160322-8216-15gx1te/metis-5.1.0/GKlib -I/tmp/metis20160322-8216-15gx1te/metis-5.1.0/include -I/tmp/metis20160322-8216-15gx1te/metis-5.1.0/libmetis/.  -DLINUX -D_FILE_OFFSET_BITS=64 -DNDEBUG -DNDEBUG2 -DHAVE_EXECINFO_H -DHAVE_GETLINE -O3 -fPIC   -o CMakeFiles/metis.dir/timing.c.o   -c /tmp/metis20160322-8216-15gx1te/metis-5.1.0/libmetis/timing.c
[ 68%] Building C object libmetis/CMakeFiles/metis.dir/util.c.o
[...]

PR #3472, opened 2016-03-26
nektar 4.3.1: new formula

mdave:
This PR adds a Formula for the Nektar++ spectral/hp element framework (www.nektar.info). To the best of my knowledge, I've complied with all of the guidelines. A slight grey area is that we use CMake's ExternalProject to download and compile modified third party software. I've encapsulated this using resources, so that homebrew can perform the download and place the tarballs in the appropriate location, but let me know if you think there are major issues with this.
### Have you:
### New formula: have you

### DIFF
--- a/nektar.rb
+++ b/nektar.rb
@@ -1,0 +1,178 @@
+# Nektar++ will, by default, download and build its ThirdParty
+# dependencies. This download strategy has been added so that Homebrew can
+# download resources and check hashes independently.
+class NektarThirdPartyDownloadStrategy < CurlDownloadStrategy
+  def stage
+    cp cached_location, basename_without_params
+  end
+end
+
+class Nektar < Formula
+  desc "Nektar++ spectral/hp element framework"
+  homepage "https://www.nektar.info/"
+  url "http://www.nektar.info/downloads/file/nektar-4.3.3.tar.gz"
+  sha256 "3167d68154d0c0b2d2205456254d6812b958b0e1e6f649952f8b411dd6b3a9e9"
+  head "https://gitlab.nektar.info/nektar/nektar.git"
+
+  option "with-demos", "Compile Nektar++ demo executables"
+
+  depends_on "cmake"  => :run
+  depends_on "boost"
+  depends_on "tinyxml"
+  depends_on "homebrew/dupes/zlib"
+  depends_on "vtk"    => :recommended
+  depends_on :mpi     => :recommended
+  depends_on "arpack" => :recommended
+  depends_on "fftw"   => :recommended
+  depends_on "petsc"  => :recommended
+  depends_on "scotch" => :optional
+
+  # Modified version of the METIS library
+  resource "modmetis" do
+    url "https://www.nektar.info/thirdparty/modmetis-5.1.0_2.tar.bz2",
+      :using => NektarThirdPartyDownloadStrategy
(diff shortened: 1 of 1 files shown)


mdave:
Okay, I think I've gone as far as I can with testing this for now. Would appreciate any input/comments! Some notes from my side:

* The Jenkins build fails due to #3461.
* `cmake` is listed as `:run` since my test uses our CMake configuration file. If it's not there, Travis will uninstall `cmake` before running tests, causing failure.

nektar.rb
```diff
@@ -0,0 +1,176 @@
+class Nektar < Formula
+  # Nektar++ will, by default, download and build its ThirdParty
+  # dependencies. This download strategy has been added so that Homebrew can
+  # download resources and check hashes independently.
+  class NektarThirdPartyDownloadStrategy < CurlDownloadStrategy
```
dpo: This should be defined outside of the formula class.

nektar.rb
```diff
@@ -0,0 +1,176 @@
...
+  depends_on "fftw"   => :recommended
+  depends_on "petsc"  => :recommended
+  depends_on "scotch" => :optional
+
+  # Modified version of the METIS library
+  resource "modmetis" do
+    url "https://www.nektar.info/thirdparty/modmetis-5.1.0_2.tar.bz2",
+      :using => NektarThirdPartyDownloadStrategy
+    sha256 "c6496cb32c892b5ab1f78484040c1f8d53b48da232c7a7e085239ea5bd431392"
+  end
```
dpo: This library won't work with the regular METIS?

nektar.rb
mdave: Have now moved this.

nektar.rb
mdave: Unfortunately not -- one relatively important part of the code relies on an internal METIS data structure that isn't exposed.

dpo: Is there any chance to patch the `metis` formula, or better yet, to submit a patch upstream to expose the data structure in question?

mdave:
I've updated this to v4.3.1 to fix clang 7.3 issues with METIS compilation. I would prefer to use our modified version, at least in the short term. The patch is pretty hairy, so I am not sure I can get a patch that I would be happy submitting upstream or modifying in the formula, at least in the short term. If this is a deal breaker then please let me know.

dpo:
Does the formula install `libmetis`? If so, it should be stashed away so Homebrew doesn't put it in `/usr/local/lib` as it will conflict with the official `libmetis`.

mdave:
All of the libraries are stored in `/usr/local/lib/nektar++-VERSION` so as to separate our third-party dependencies from the main library.

iMichka:
Hi. There was no progress here since some time. It is partly understandable as this formula is pretty hairy, due to the metis question. If the change can not be done in metis, why not fork metis and do your own stuff, if you are not able to come to a compromise with the metis authors.

Please understand that we try to avoid exceptions and non-standard builds; because all this increases the maintenance burden. And we are an understaffed open-source project, so if we can avoid edge cases that would be great.

mdave:
No problem. As a developer on another understaffed open-source project I can certainly relate!

I am going to close this issue for now and resubmit later. We have a branch under development to remove the requirement on the modified Metis library over to Scotch, which I believe natively supports what we need through the API. We've also removed the requirement for the (now abandoned) Loki header library, so this should remove all of the 'hairy' components for this Formula.

Issue #1915, opened 2015-03-08
change root6 file to add option


leoredi:
Hi guys we should change the root6 file like this

"cmake-build" do system "cmake", "..", "-Dgnuinstall=ON", "-Dbuiltin_freetype=ON", "-Dall=ON", cmake_opt("python"), cmake_opt("ssl", "openssl"), cmake_opt("xrootd"), *std_cmake_args system "make", "install" end 

this will allow us to install root6 with all the possible dependences.

lgeiger:
> install root6 with all the possible dependencies.

this would be great.

dpo:
I'm not convinced that will lead to a clean install because some optional components are already available as separate Homebrew formulae. But if you feel strongly, please feel free to submit a pull request that we'll evaluate. Thanks.

PR #3477, opened 2016-03-27
Build ROOT 6 with CMake

alexpearce:
### Have you:

---

This PR changes the way ROOT is built, from the old `configure && make install` style to `cmake && make install`.

There are a couple of reasons for doing this. Firstly, to quote [the ROOT developers](https://root.cern.ch/building-root):

> The classic build with configure/make is is still available but it will not be evolving with the new features of ROOT.

Secondly, I've come across, and heard other people talk about, problems that only appear in the `configure` build (the current Homebrew method). I guess this is because the ROOT developers are focusing on the CMake build nows, so they don't notice changes that break functionality with `configure` builds and/or they have `configure`-related bug reports as a low priority.

Switching to a CMake-based build seems like The Right Thing.

We could also address #1915 in this PR by using `-Dall=ON`, rather than specifying a few particular flags [as is done now](https://github.com/Homebrew/homebrew-science/compare/master...alexpearce:root6-cmake?expand=1#diff-b316bc9bd3af2682b35a6b917a29b62dR54). I have no preference, the ROOT build this PR gives me is suitable for my needs.

### DIFF
--- a/root6.rb
+++ b/root6.rb
@@ -62,7 +62,7 @@ class Root6 < Formula
     # ROOT forbids running CMake in the root of the source directory,
     # so run in a subdirectory (there's already one called `build`)
     mkdir "build_dir" do
-      system "cmake", "..", *std_cmake_args, *args
+      system "cmake", "..", *(std_cmake_args + args)
       system "make", "install"
     end
 
--- a/root6.rb
+++ b/root6.rb
@@ -2,11 +2,12 @@ class Root6 < Formula
   # in order to update, simply change version number and update sha256
   version_number = "6.06.02"
   desc "Object oriented framework for large scale data analysis"
-  homepage "http://root.cern.ch"
+  homepage "https://root.cern.ch"
   url "https://root.cern.ch/download/root_v#{version_number}.source.tar.gz"
   mirror "https://fossies.org/linux/misc/root_v#{version_number}.source.tar.gz"
   version version_number
   sha256 "18a4ce42ee19e1a810d5351f74ec9550e6e422b13b5c58e0c3db740cdbc569d1"
+  revision 1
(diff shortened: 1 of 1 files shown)


alexpearce:
[Build 2023.4 fails](https://travis-ci.org/Homebrew/homebrew-science/jobs/118867541), which is a Linux build (I didn't know there were plans for Homebrew on Linux, cool! :beers:). I've only tried this formula on OS X, but fixing it should just be a case of adding an X11 dependency. Should we?

[Build 2023.3 fails](https://travis-ci.org/Homebrew/homebrew-science/jobs/118867540) on a syntax error:

```
/usr/local/Library/Taps/homebrew/homebrew-science/root6.rb:65: syntax error, unexpected tSTAR, expecting tAMPER
    system "cmake", "..", *std_cmake_args, *args
                                            ^
```

I don't have Ruby 1.8.7 locally and I couldn't find a REPL online to play with. Multiple splats might not be allowed in older Rubies. Could work around it by doing `*(std_cmake_args + args)`.

mdave:
Just FYI, I have similar issues with my splats for another build and have ruby 1.8.7 installed through rvm -- multiple splats would indeed seem to be the problem and your fix works fine for me:

```irb
1.8.7-p374 :003 > puts 'cmake', '..', *args, *args2
SyntaxError: compile error
(irb):3: syntax error, unexpected tSTAR, expecting tAMPER
puts 'cmake', '..', *args, *args2
                            ^
	from (irb):3
1.8.7-p374 :004 > puts 'cmake', '..', *(args+args2)
cmake
..
asd
asd2
asd3
asd4
asd5
 => nil
```

alexpearce:
Thanks for the confirmation @mdave. I've pushed a commit with the workaround.

PR #3483, opened 2016-03-29
metis: use gcc if clang is 7.3

ilovezfs:


metis.rb
```diff
@@ -14,6 +15,11 @@ class Metis < Formula
 
   depends_on "cmake" => :build
 
+  # Failure seems restricted to clang 7.3
+  # Check if fixed in next version of clang when it's released
+  # https://llvm.org/bugs/show_bug.cgi?id=27059
+  fails_with :clang if MacOS::Xcode.version == "7.3" || MacOS::CLT.version.to_f == 7.3
```
dpo: Many thanks for this! You should be able to exclude that version of clang using
```
fails_with :clang
  build 703
  cause "brief explanation here"
end
```

ilovezfs: Ah, thanks for that. And it actually needs to be `fails_with :clang do`, etc.

ilovezfs:
I'm wondering if
```
homebrew/science/cppad            homebrew/science/ipopt            homebrew/science/slepc          
homebrew/science/dealii           homebrew/science/mumps            homebrew/science/superlu_dist   
homebrew/science/getdp            homebrew/science/parmetis         homebrew/science/trilinos       
homebrew/science/gmsh             homebrew/science/petsc            homebrew/science/zoltan
```

are all going to need to be rebuilt with gcc as well. Also, Travis seems to be on 7.2 right now, so it's not really testing this PR.

dpo:
@UniqMartin is there a Xcode 7.3 image available for Travis?

ilovezfs:
OK, I figured out how to get metis to compile with clang 7.3, so closing this.

UniqMartin:
> @UniqMartin is there a Xcode 7.3 image available for Travis?

There is: `xcode7.3`. It existed since the first public betas and was now updated with the release build of Xcode 7.3. See [this blog post](https://blog.travis-ci.com/2016-03-23-xcode-image-updates) for details. (The Travis CI blog is a good way to stay up-to-date on these kinds of changes; it is also very low traffic.)

ilovezfs:
```
/* These are the jmp_buf for the graceful exit in case of severe errors.
   Multiple buffers are defined to allow for recursive invokation. */
#define MAX_JBUFS 128
__thread int gk_cur_jbufs=-1;
__thread jmp_buf gk_jbufs[MAX_JBUFS];
__thread jmp_buf gk_jbuf;

```
Not a problem.

UniqMartin:
Good; it seems to be safe in this case. Just wanted to leave a word of caution just in case.

PR #3488, opened 2016-03-30
exonerate: Add desc

sjackman:
Pairwise sequence alignment of DNA and proteins


tseemann:
@BrewTestBot test this please

PR #3494, opened 2016-04-01
Octave Update to 4.0.2

schoeps:
Since the updates of gmsh and getdp worked rather fine, I am brave enough to propose an octave.rb update....

Updates
- version bumb to 4.0.2
- openblas settings are checked
- bottle is disregarded if graphicsmagick was locally built
- plotting with gnuplot/qt is default


octave.rb
```diff
@@ -44,6 +35,11 @@ class Octave < Formula
     end
   end
 
+  # snapshots of Octave 4.1.0+
+  devel do
+    url "http://hydra.nixos.org/job/gnu/octave-default/tarball/latest/download/1/octave-4.1.0+.tar.gz"
```
dpo: You'll need a sha sum here.

octave.rb
```diff
@@ -166,6 +167,24 @@ def install
...
+    elsif !["arpack", "qrupdate", "suite-sparse421"].all? { |n| Tab.for_name(n).without? "openblas" }
+      odie "Octave is compiled --without-openblas but arpack, qrupdate and suite-sparse421 are."
+    end
+
+    # If GraphicsMagick was built from source, it is possible that it was
+    # done to change quantum depth.  If so, our Octave bottles are no good.
+    # https://github.com/Homebrew/homebrew-science/issues/2737
+    def pour_bottle?
+      Tab.for_formula("graphicsmagick").bottle?
+    end
```
dpo: Nice! I think `pour_bottle` should be outside of `install` though.

dpo:
Very brave indeed. Thanks!

octave.rb
schoeps: One gets the latest (successfully tested) snapshots from this url. It's rather like HEAD but with some more stability. I cannot give a sha sum here since it chances frequently. If the sha sum is a hard requirement, I would probably remove "--devel" (which is a pity, since I think these snapshots should be preferred over HEAD, however most users would probably care anyway).

octave.rb
schoeps: done.

octave.rb
dpo: Ah I see. `brew` might have a problem if the sha changes.

dpo:
```
dldfcn/__fltk_uigetfile__.cc:33:10: fatal error: 'FL/Fl.H' file not found
#include <FL/Fl.H>
         ^
1 error generated.
```

schoeps:
I have disabled FLTK by default since the regular FLTK formula is still incompatible with recent OS X.

dpo:
Seems like the docs should be disabled by default?
```
You don't have a working TeX binary (tex) installed anywhere in
your PATH, and texi2dvi cannot proceed without one.  If you want to use
this script, you'll need to install TeX (if you don't have it) or change
your PATH or TEX environment variable (if you do).  See the --help
output for more details.
```

schoeps:
1. Can you help me to understand why Jenkins fails but Travis works? (what is actually the difference?)

2. the creation of the manual is disabled (DOC_TARGETS=""). The only documentation that is must be created is the function help. Unfortunately, I cannot reproduce the problem on my local machine (see below). Any idea?

```
Making install in doc
Making install in doxyhtml
make[4]: Nothing to be done for `install-exec-am'.
make[4]: Nothing to be done for `install-data-am'.
Making install in interpreter
./config-images.sh ../..
../../doc/interpreter/images.mk is unchanged
./config-images.sh ../..
../../doc/interpreter/images.mk is unchanged
make[4]: Nothing to be done for `install-exec-am'.

```

schoeps:
my (local) build log: https://gist.github.com/schoeps/6701122a747795a9ff8cd5a3307d6c6d

dpo:
The Jenkins bot has nothing related to LaTeX installed on it at all. I'm just guessing but Travis might have certain LaTeX tools installed on their images, as you do on your local machine. What does the help feature rely on in Octave?

schoeps:
I had compiled octave on a virtual machine without latex. So, I rather suspect that somewhere "DOC_TARGETS=" was lost....

schoeps:
By the way, good news: octave can indeed be compiled with suitesparse 4.5.1, so suitesparse 4.2.1 can be removed. So we should probably wait with this pull request after #3384 is accepted.

dpo:
The doc issue might be the same as in https://github.com/Homebrew/homebrew-science/issues/3387. Perhaps there's a clue there?

dpo:
That would be great.

schoeps:
>  What does the help feature rely on in Octave?

only makeinfo is needed at compile time. Latex is only necessary for the pdf manual.

dpo:
There are just audit warnings on 10.11: https://travis-ci.org/Homebrew/homebrew-science/jobs/120328348#L1842

Does the LaTeX problem come from SuiteSparse?

schoeps:
> Does the LaTeX problem come from SuiteSparse?
No there was a left-over from the "--with-check" to "--with-test" migration :)

I will switch to suite-sparse if #3384 is accepted. Is anything missing there besides proper tests?

With both suite-sparse versions I see sometimes segfaults in the octave tests if Apple's BLAS (in contrast to openblas) is used. However, they do not occur when actually using octave. It don't know what's going on. Maybe @dpo or @mcg1969 have an idea?

dpo:
> I will switch to suite-sparse if #3384 is accepted. Is anything missing there besides proper tests?

No, just a test but it would be best if it consisted in compiling a sample program. Feel free to override #3384. 

> With both suite-sparse versions I see sometimes segfaults in the octave tests if Apple's BLAS (in contrast to openblas) is used. However, they do not occur when actually using octave. It don't know what's going on. Maybe @dpo or @mcg1969 have an idea?

Not sure. Did you try veclibfort?

schoeps:
new suite-sparse: #3503

dpo:
Want to try changing [this line](https://github.com/Homebrew/homebrew-science/pull/3494/files#diff-fee4cb7a8e3fa693dadd2ee9d45ed482R111) to `depends_on "suite-sparse"`?

schoeps:
@dpo: suite-sparse is working on my computer with openblas. there are still problems with apple's blas (segfaults during the tests). Interestingly this already happens with octave 4.0.0 if it is build from the hg-repository instead from the tarball. I have no clue what's going on... Maybe you have an idea? 

It can be tested with this formula (HEAD will build version 4.0.0 from hg)
https://gist.github.com/schoeps/d24d9b74fc2d487919e15b63285385b0

dpo:
@schoeps Just to make sure I understand, are the segfaults in the `suite-sparse` tests, or the `octave` tests? If the latter, with which `suite-sparse`? The new one, or `suite-sparse421`?

dpo:
With `HEAD` I get build errors on 10.9: https://gist.github.com/e068b0805b521423e4ba4ab47e548e56

combefis:
As asked by @dpo, here is a link to the patch file with some typos fixes that I made in another PR.

https://patch-diff.githubusercontent.com/raw/Homebrew/homebrew-science/pull/3552.patch

schoeps:
> Just to make sure I understand, are the segfaults in the suite-sparse tests, or the octave tests? If the latter, with which suite-sparse? The new one, or suite-sparse421?

Indeed, I was referring to the octave tests. They fail with segfaults on my machine when using Apple's blas. Can you please verify, that this command works in your Octave:
`single ([1+i 2+i 3+i]) * single ([ 4+i ; 5+i ; 6+i]) `

> ps: I just built the formula you link to with small changes to use the new suite-sparse and Apple's BLAS on OSX 10.9, and I'm not seeing any segfaults. I get: ...

I added the missing "-lsuitesparse" to the formula. I guess if the tests are fine on travis, we are good...

schoeps:
Does anyone know how I can access the files that Travis created?

schoeps:
@dpo: Jenkins sees the same segfaults that I mentioned.....

```
Integrated test scripts:

  liboctave/array/Array.cc-tst ................................ PASS     18/18  
  liboctave/array/CMatrix.cc-tst .............................. PASS      9/11  
                                                                  FAIL    2
  liboctave/array/CSparse.cc-tst .............................. PASS     10/10  
  liboctave/array/Sparse.cc-tst ............................... PASS    103/103 
  liboctave/array/dMatrix.cc-tst .............................. PASS     10/10  
  liboctave/array/dSparse.cc-tst .............................. PASS     12/12  
  liboctave/array/fCMatrix.cc-tst .............................panic: Segmentation fault: 11 -- stopping myself...
```

dpo:
On 10.9, I'm not seeing those segmentation faults at install time, but the command you gave
```
octave:1> single ([1+i 2+i 3+i]) * single ([ 4+i ; 5+i ; 6+i])
panic: Segmentation fault: 11 -- stopping myself...
attempting to save variables to 'octave-workspace'...
save to 'octave-workspace' complete
Segmentation fault: 11
```

dpo:
Ah, it's the good old buggy OSX BLAS problem. It's alright if you use `veclibfort`:
```octave
octave:1> single ([1+i 2+i 3+i]) * single ([ 4+i ; 5+i ; 6+i])
ans =  29 + 21i
```
Here are the changes I made (quickly; need to be smoothed out):
```diff
diff --git a/octave.rb b/octave.rb
index f9b0c28..9cf4b63 100644
--- a/octave.rb
+++ b/octave.rb
@@ -103,6 +103,7 @@ class Octave < Formula
   depends_on :java => "1.6"    if build.with? "java"
   depends_on "llvm"            if build.with? "jit"
   depends_on "openblas"        if build.with? "openblas"
+  depends_on "veclibfort"      unless build.with? "openblas"
   depends_on "pstoedit"        if build.with? "ghostscript"
   depends_on "qhull"           if build.with? "qhull"
   depends_on "qrupdate"        if build.with? "qrupdate"
@@ -170,7 +171,7 @@ class Octave < Formula
       args << "--without-umfpack"
     else
       # add -lsuitesparseconfig here if suite-sparse >4.5.1
-      ENV.append_to_cflags "-L#{Formula["suite-sparse"].opt_lib}"
+      ENV.append_to_cflags "-L#{Formula["suite-sparse"].opt_lib} -lsuitesparseconfig"
       if Tab.for_name("suite-sparse").with? "metis"
         ENV.append_to_cflags "-L#{Formula["metis"].opt_lib} -lmetis"
       end
@@ -187,6 +188,10 @@ class Octave < Formula
       odie "Octave is compiled --without-openblas but arpack, qrupdate and suite-sparse are."
     end
 
+    if build.without? "openblas"
+      args << "--with-blas=-L#{Formula["veclibfort"].opt_lib} -lveclibFort"
+    end
+
     system "./bootstrap" if build.head?
 
     # libtool needs to see -framework to handle dependencies better.
```

schoeps:
@dpo: can you help me interpreting the results?

dpo:
Ok so we have a few issues:
- I'm not sure why the test failed on 10.11. It worked on Travis, so it's probably not a big concern.
- It would be great to make the formula work on Linux. I think we should be good to go if we force `openblas` to be used there. I did the same in `suite-sparse`, lines [22](https://github.com/Homebrew/homebrew-science/blob/master/suite-sparse.rb#L22) and [41](https://github.com/Homebrew/homebrew-science/blob/master/suite-sparse.rb#L41).
- On 10.10, we have the LaTeX problem we've had before. Somehow `DOC_TARGETS=` is ignored or lost. I can try locally when I get a chance but it might not be right away.

schoeps:
the second item was easy. I also updated to 4.0.2.

however, I might still need help with 1 and 3.

octave.rb
```diff
@@ -163,7 +173,23 @@ def install
...
+        odie "Octave is compiled --with-openblas but arpack, qrupdate or suite-sparse are not."
+      else
+        args << "--with-blas=-L#{Formula["openblas"].opt_lib} -lopenblas"
+      end
+    else #without "openblas"
+      if ["arpack", "qrupdate", "suite-sparse"].any? { |n| Tab.for_name(n).with? "openblas" }
+        odie "Arpack, qrupdate or suite-sparse are compiled --with-openblas but Octave is not."
+      else
+       args << "--with-blas=-L#{Formula["veclibfort"].opt_lib} -lveclibFort"
+      end
```
dpo: On Linux, you'll want to link with `-lblas -llapack` if `openblas` isn't chosen.

dpo:
Seems to pass now! Do you want to squash your commits?

schoeps:
squashed.

dpo:
Done! Many thanks for the hard work and patience!

ted-piotrowski:
Just came by to say thank you for landing this PR. I use Octave for Coursera's Stanford Machine Learning class and there was a bug in 4.0.0 that prevented submission of Coursera assignments. We had to apply this patch for every assignment (https://drive.google.com/file/d/0B6lXyE7fgSlXZjlqQ3FIRExmTDA/view) and it was tedious. Now I've just run brew upgrade and submitted my assignment successfully without the patch. I am confident you have saved a lot of time and frustration for many current and future students. Well done!

dpo:
All thanks should go to @schoeps!

schoeps:
Thanks for letting me know. It's nice to hear appreciation. Typically one sees only bug reports :)

kittythecat:
Thank you very much schoeps and dpo! I was anxiously waiting for this update! :)
Only one issue: printing figures doesn't work "print -dpng test.png" (Octave 4.0.2 and Mac OS X 10.11). It says it misses gl2ps and fig2dev. It seems I do have gl2ps installed through homebrew, but it doesn't recognise it.

schoeps:
@kittythecat: if the following does not help, then please open a bug report for this issue. Did it worked with octave 4.0.1? Do you use Octave on the command line are the GUI? Which graphics_backend do you use?

Three possible solutions:
* you are most probably missing fig2dev. I think there is no official formula. For now 
```
./brew install homebrew/x11/imake
./brew install odousse/homebrew-xfig/transfig
```
@dpo are you aware of any reason why transfig is not in homebrew?
* change the graphics backend to qt:
```
graphics_toolkit ('qt');
x=0:10;plot(x,x.^2);
print -dpng test.png
```
* Export to pdf and convert afterwards.
```
graphics_toolkit ('gnuplot')
x=0:10;plot(x,x.^2);
print test.pdf
```

kittythecat:
@schoeps : Tried on both cli and gui.
With 'qt' and the formulas solved the fig2dev problem, but it still misses the gl2ps library.
However, with gnuplot (+aquaterm) it works, also directly to png (without the additional formulas).
Thanks for the help!

schoeps:
@dpo I like to come back to this question: are you aware of any reason why transfig is not in homebrew?

dpo:
@schoeps I believe it's available from https://github.com/mistydemeo/homebrew-xfig.

schoeps:
Yes and there is an update at  https://github.com/odousse/homebrew-xfig/transfig. However, it must be part of core or science such that we can make it a proper octave dependency, right?

dpo:
Technically, you could use
```
depends_on "mistydemeo/xfig/transfig"
```
The disadvantage is that it doesn't appear to be bottled. Perhaps it should be included here?
@mistydemeo

schoeps:
Ok, let's see what @mistydemeo says. I would be in favor of adding transfig to science or core.

PR #3503, opened 2016-04-04
SuiteSparse 4.5.1 (update)

schoeps:
Update of #3384. 
- compiles demos (and installs them for `test do`)
- switch `--with-metis` is default


suite-sparse.rb
```diff
@@ -14,26 +13,30 @@ class SuiteSparse < Formula
...
   option "with-matlab", "Install Matlab interfaces and tools"
   option "with-matlab-path=", "Path to Matlab executable (default: matlab)"
+  option "with-openmp", "Build with OpenMP support"
 
+  option "without-test", "Do not perform build-time tests (not recommended)"
+  option "without-metis", "Use shipped metis (not recommended) "
+
+  # SuiteSparse must be compiled with metis 5.
+  # It is shipped with metis-5.1.0
+  depends_on "metis" => :recommended
```
dpo: Wasn't there a problem when building against the METIS that ships with SuiteSparse?

suite-sparse.rb
```diff
@@ -14,26 +13,30 @@ class SuiteSparse < Formula
...
-    ENV.deparallelize
-
-    # Switch to the Mac base config, per SuiteSparse README.txt
-    mv "SuiteSparse_config/SuiteSparse_config.mk",
-       "SuiteSparse_config/SuiteSparse_config_orig.mk"
-    mv "SuiteSparse_config/SuiteSparse_config_Mac.mk",
-       "SuiteSparse_config/SuiteSparse_config.mk"
+  if build.without? "metis"
+    odie "Building with shipped metis is currently not supported."
+  end
```
dpo: Oh ok. Do we really want to do this? I'd think it's always best to build against Homebrew METIS to avoid risking a duplicate `libmetis`.

dpo:
Thanks!

suite-sparse.rb
schoeps: The original author also didn't like it; I did this because of backward compatibility. For example octave.rb used to check if `suite-sprase.rb` was compiled with `--with-metis4` to include `-lmetis` for linking. However, I have to admit that compatibility is broken anyway since metis 5 is used. 

For me it's the same. I can easily fix the problem with octave myself. What about other packages?

dpo:
@BrewTestBot test this please

suite-sparse.rb
```diff
@@ -107,4 +101,8 @@ def caveats
     end
     s
   end
+
+  test do
+    system "#{libexec}/amd_demo"
```
dpo: Sorry for being a bother but is there a chance to compile a demo as part of the test? That would ensure reliability when a dependency is updated.

dpo:
I wonder why there's a link error:
```
ld: library not found for -lsuitesparseconfig
```
Also, Travis isn't kicking in. Are you on an outdated branch by any chance?

suite-sparse.rb
schoeps: No problem, but how do I do that without repeating all the make_args?

schoeps:
> Also, Travis isn't kicking in. Are you on an outdated branch by any chance?

No?! ... not that I am aware of. Suitesparseconfig should also be already compiled before, as the log shows:
```
clang -L/usr/local/Cellar/suite-sparse/4.5.1/lib -dynamiclib -compatibility_version 4 -current_version 4.5.1 -shared -undefined dynamic_lookup SuiteSparse_config.o -o /usr/local/Cellar/suite-sparse/4.5.1/lib/libsuitesparseconfig.4.5.1.dylib -lm
( cd /usr/local/Cellar/suite-sparse/4.5.1/lib ; ln -sf libsuitesparseconfig.4.5.1.dylib libsuitesparseconfig.dylib )
( cd /usr/local/Cellar/suite-sparse/4.5.1/lib ; ln -sf libsuitesparseconfig.4.5.1.dylib libsuitesparseconfig.4.dylib )
```

My formula works on two different machines (10.10, 10.11). Any idea how to debug this?

suite-sparse.rb
```diff
@@ -107,4 +97,25 @@ def caveats
...
+      system ENV["CC"], "-o", "colamd_example", "-O", pkgshare/"colamd_example.c", "-L#{lib}", "-I#{include}", "-lsuitesparseconfig", "-lcolamd"
+      system "./colamd_example"
+      system ENV["CC"], "-o", "cs_demo1", "-O", pkgshare/"cs_demo1.c", "-L#{lib}", "-I#{include}", "-lsuitesparseconfig", "-lcxsparse"
+      system "./cs_demo1 < #{pkgshare}/t1"
+      system ENV["CC"], "-o", "klu_simple", "-O", pkgshare/"klu_simple.c", "-L#{lib}", "-I#{include}", "-lsuitesparseconfig", "-lklu"
+      system "./klu_simple"
+      system ENV["CC"], "-o", "umfpack_simple", "-O", pkgshare/"umfpack_simple.c", "-L#{lib}", "-I#{include}", "-lsuitesparseconfig", "-lumfpack"
+      system "./umfpack_simple"
+    end
+  end
```
dpo: Beautiful! Thanks so much!

dpo:
Hm. It works for me on 10.9. I wonder what's going on.

dpo:
Could you try adding `ENV.deparallelize` before `make` in case there's a race going on?

schoeps:
Still not working :(

suite-sparse.rb
```diff
@@ -14,28 +13,25 @@ class SuiteSparse < Formula
 
   option "with-matlab", "Install Matlab interfaces and tools"
   option "with-matlab-path=", "Path to Matlab executable (default: matlab)"
+  option "with-openmp", "Build with OpenMP support"
 
+  depends_on "cmake" => :build
```
dpo: This formula doesn't need `cmake`, only `metis` does.

dpo:
I think the problem is that when `libsuitesparseconfig.dylib` is created, it's created directly in `/usr/local/Cellar/suite-sparse/4.5.1/lib` (i.e., in the Homebrew location `lib`). It looks like all other `dylibs` will also be created there. Now it looks like `lib` isn't on the `LD_LIBRARY_PATH` on the bot. Could you try adding `lib` using something like
```
ENV.prepend "LD_LIBRARY_PATH", "#{lib}"
```

dpo:
In #3515 I resolved the problem by building with `env: std` but I suspect that's the cause of a build failure in `netcdf`.

schoeps:
Ok, let's close this request. It's superseeded by #3515.

PR #3504, opened 2016-04-05
BEDOPS 2.4.16

alexpreynolds:
- bedmap

  Added new `--echo-ref-row-id` option to report reference row ID elements.

- Starch C++ API

  Fixed a bug with extraction of archives made with `starch --gzip` (thanks to Brad Gulko for the bug report and Paul Verhoeven and Peter Weir for compile and testing assistance).


alexpreynolds:
Just wondering if I need to do anything additional to integrate this commit?

PR #3509, opened 2016-04-06
sailfish 0.9.2

sjackman:


sjackman:
Upstream issue: https://github.com/kingsfordgroup/sailfish/issues/92
and fix: https://github.com/kingsfordgroup/sailfish/issues/93

sjackman:
Merged upstream.

rob-p:
I agree with your suggestion, and I'll avoid moving the tag in the future.  However, I think there may be a problem with the old tag now (which is why I moved it).  There was a bug-fix update to RapMap (which is pulled in, in source form, as a dependency), that will cause an issue with the previous commit tagged as 0.9.2 (it creates a runtime logger with the same name, which results in an exception).  This wasn't an issue in older releases, because they pull in an older version of RapMap that wasn't doing this logging during index generation.

PR #3515, opened 2016-04-09
Suitesparse451

dpo:
### Have you:

### New formula: have you

- [ ] ~~Written a sensible test? (The best test is to compile and run a sample program.)~~

- [ ] ~~Removed the `revision` line, if any, when bumping the version number?~~
- [ ] ~~Added/bumped the `revision` number if the changes affect the precompiled bottles?~~


suite-sparse.rb
```diff
@@ -52,26 +51,24 @@ def install
...
-      (doc/pkg).install Dir["#{pkg}/Doc/*"]
+
+    # Install docs and demos
+    %w[AMD CAMD CCOLAMD CHOLMOD COLAMD CXSparse KLU LDL SPQR UMFPACK].each do |m|
+      pkgshare.install Dir["#{m}/demo/*.c",
+                           "#{m}/demo/*.h",
+                           "#{m}/demo/Matrix/*",
+                           "#{m}/Matrix/*"
+                          ]
+      (doc/m.to_s).install Dir["#{m}/Doc/*"]
```
schoeps: I realized that this can be done more elegantly since Suitesparse is already collecting the docs. Do outside of the loop:
`doc.install Dir["share/doc/suitesparse-*/*"]`

dpo: Thanks. I'll change it.

dpo:
Strangely enough, adding `env: std` fails on my machine; the openmp flag isn't processed correctly.

dpo:
Ok, this is cleaner, but `netcdf` has some issues.

schoeps:
but the 'netcdf' error should habe nothing to do with suitesparse?!

schoeps:
@dpo: I have realized that there is a problem with the path. See `otool -L /usr/local/opt/suite-sparse/lib/*.dylib`. You will find temporary directories as e.g., `/private/tmp/suite-sparse20160419-82821-1kvs00h/SuiteSparse/lib/libamd.2.4.4.dylib`. I guess one could fix it with `install_name_tool` but this seems quite troublesome.

dpo:
`SuiteSparse_config.mk` defines `SO_INSTALL_NAME = install_name_tool -id` but it's only used to install the METIS library, not the SuiteSparse libraries. That looks like a bug.

schoeps:
Are you aware of an easy fix? Looping over all libs afterwards is rather ugly.... (I need to correct this because I need to install suite-sparse in a non-standard location)

dpo:
It might be because the formula short-circuits `make install`, which caused other troubles.

sjackman:
Why isn't `make install` used?

dpo:
It's broken (at least for me) because we're using an external `metis`.

dpo:
But #3578 is green on both OSX and Linux now.

schoeps:
is the path correct? i.e. `otool -L ...`

dpo:
I got it. update coming up.

dpo:
@schoeps @sjackman Could you try #3578? Seems to be ok, though the bots haven't quite finished, and a quick check suggests that the lib paths are ok.

schoeps:
Brilliant, works for me also in non-standard locations. Thanks a lot.

dpo:
Great! I'll pull when the bottles are available.

dpo:
Alright. Let's try `octave` again.

PR #3530, opened 2016-04-12
mfem 3.1

goxberry:
### Have you:

### New formula: have you

### DIFF
--- a/mfem.rb
+++ b/mfem.rb
@@ -1,0 +1,85 @@
+class Mfem < Formula
+  desc "Free, lightweight, scalable C++ library for FEM."
+  homepage "http://www.mfem.org"
+  url "http://goo.gl/xrScXn"
+  version "3.1"
+  sha256 "841ea5cf58de6fae4de0f553b0e01ebaab9cd9c67fa821e8a715666ecf18fc57"
+
+  option "with-mpi", "Build with mpi support (implies --with-hypre --with-metis)"
+  option "with-suite-sparse", "Build with suite-sparse support (implies --with-metis)"
+
+  depends_on :mpi => [:cxx, :optional]
+  if build.with?("mpi")
+    depends_on "metis"
+    depends_on "hypre" => "with-mpi"
+  else
+    depends_on "hypre" => :optional
+    depends_on "metis" => :optional
+  end
+
+  depends_on "suite-sparse" => :optional
+  depends_on "openblas" => :optional
+
+  def install
+    make_args = ["PREFIX=#{prefix}"]
+
+    if build.with?("openblas")
+      lapack_lib = "-L#{Formula["openblas"].opt_lib} -lopenblas"
+    else
+      lapack_lib = "-llapack -lblas"
+    end
+    make_args += ["MFEM_USE_LAPACK=YES", "LAPACK_LIB=#{lapack_lib}"]
+
+    if build.with?("hypre") || build.with?("mpi")
+      make_args += ["HYPRE_DIR=#{Formula["hypre"].opt_prefix}",
+                    "HYPRE_OPT=-I#{Formula["hypre"].opt_include}",
+                    "HYPRE_LIB=-L#{Formula["hypre"].opt_lib} -lHYPRE"]
+    end
+
+    if build.with?("metis") || build.with?("mpi") || build.with?("suite-sparse")
+      metis_lib = "-L#{Formula["metis"].opt_lib} -lmetis"
+      make_args += ["MFEM_USE_METIS_5=YES",
+                    "METIS_DIR=#{Formula["metis"].opt_prefix}",
+                    "METIS_OPT=-I#{Formula["metis"].opt_include}",
+                    "METIS_LIB=#{metis_lib}"]
+    end
+
+    make_args += ["MFEM_USE_MPI=YES"] if build.with?("mpi")
+
+    if build.with?("suite-sparse")
+      ss_lib = "-L#{Formula["suite-sparse"].opt_lib} "
+      ss_lib += "-lumfpack -lcholmod -lcolamd -lamd -lcamd -lccolamd "
+      ss_lib += "-lsuitesparseconfig #{metis_lib} #{lapack_lib}"
+      make_args += ["MFEM_USE_SUITESPARSE=YES",
+                    "SUITESPARSE_DIR=#{Formula["suite-sparse"].opt_prefix}",
+                    "SUITESPARSE_OPT=-I#{Formula["suite-sparse"].opt_include}",
+                    "SUITESPARSE_LIB=#{ss_lib}"]
+    end
+
+    system "make", "config", *make_args
+    system "make", "all"
+    system "make", "install"
+
+    # Install some examples in pkgshare for users learning the library
+    (pkgshare/"examples").mkpath
+    (pkgshare/"examples").install Dir["examples/*.cpp"]
+    (pkgshare/"examples").install "examples/makefile"
+    (pkgshare/"data").mkpath
+    (pkgshare/"data").install Dir["data/*.mesh"]
+    (pkgshare/"data").install Dir["data/*.vtk"]
+  end
+
+  test do
+    cp_r "#{Formula["mfem"].opt_pkgshare}/examples", "./"
+    cd "examples" do
+      cp "#{Formula["mfem"].opt_pkgshare}/data/star.mesh", "./"
+      system "make", "all", "MFEM_DIR=#{Formula["mfem"].opt_prefix}", "CONFIG_MK=$(MFEM_DIR)/config.mk"
+      args = ["-m", "star.mesh", "--no-visualization"]
+      if Tab.for_name("mfem").with? "mpi"
+        system "mpirun", "-np", "4", "./ex1p", *args
+      else
+        system "./ex1", *args
+      end
+    end
+  end
+end


goxberry:
Placeholder PR until the MFEM devs create a release on GitHub. (See https://github.com/mfem/mfem/issues/53, which I filed earlier tonight.)

mfem.rb
```diff
@@ -0,0 +1,74 @@
+class Mfem < Formula
+  desc "Free, lightweight, scalable C++ library for FEM."
+  homepage "http://www.mfem.org"
+  url "https://github.com/goxberry/mfem/archive/3.1.tar.gz"
+  sha256 "b0d223ddc66a4f8ca01fd39d9ac9b7b235f60b2bfd8e100293ea14e9febc8682"
+
+  option "with-mpi", "Build with mpi support (implies --with-hypre --with-metis)"
+  option "with-suite-sparse", "Build with suite-sparse support (implies --with-metis)"
```
dpo: This is redundant with `depends_on "suite-sparse" => :optional`, which you could factor out of the `if` below.

goxberry: Ah, right, yes, I will make this change. Thanks!

goxberry:
@dpo My guess is that MFEM will fail the Travis-CI build checks even if the URL points to the canonical repository because the `brew audit --online` "notability" check will fail, since the [mfem](https://github.com/mfem/mfem/) repo doesn't have 50+ stars, 20+ watchers, and 20+ forks. It's not an openly developed package in the way that SUNDIALS (also at LLNL) is not an openly developed package, and it's a scientific package with a limited audience; both of these qualities depress the stars/watchers/forks figures.

goxberry:
I have no idea why the OS X build fails on Travis-CI, but the failing build for Linux can be fixed if I have some idea of how to satisfy a BLAS & LAPACK dependency for Linux, because the `suite-sparse` option requires `blas` and `lapack`. I looked at [suite-sparse in homebrew-science](https://github.com/Homebrew/homebrew-science/blob/master/suite-sparse.rb) and I see that, for linuxbrew, it does the same thing I currently do. Should I change the formula in some way to make it more linuxbrew friendly? (Require openblas, blis, or some BLAS+LAPACK replacement for Linux?)

For everything else in MFEM, BLAS & LAPACK are optional.

mfem.rb
```diff
@@ -0,0 +1,70 @@
...
+    depends_on "metis" => :optional
+  end
+
+  depends_on "suite-sparse" => :optional
+  depends_on "openblas" => :optional
+
+  def install
+    make_args = ["PREFIX=#{prefix}"]
+
+    lapack_lib = (build.with?("openblas") ? "-L#{Formula["openblas"].opt_lib}" : "")
```
dpo: Here, the relevant library is `libopenblas`.

goxberry: Thanks! Fixed this line; got the Homebrew conventions mixed up with those of spack.

mfem.rb
```diff
@@ -0,0 +1,73 @@
+class Mfem < Formula
+  desc "Free, lightweight, scalable C++ library for FEM."
+  homepage "http://www.mfem.org"
+  url "http://goo.gl/xrScXn"
+  version "3.1"
+  sha256 "841ea5cf58de6fae4de0f553b0e01ebaab9cd9c67fa821e8a715666ecf18fc57"
+
+  option "with-mpi", "Build with mpi support (implies --with-hypre --with-metis)"
+  option "with-suite-sparse", "Build with suite-sparse support (implies --with-metis)"
```
dpo: These options are redundant. Because `:mpi` and `suite-sparse` are `:optional`, there will automatically be options `with-mpi` and `with-suite-sparse`.

mfem.rb
```diff
@@ -0,0 +1,73 @@
...
+    make_args = ["PREFIX=#{prefix}"]
+
+    if build.with?("openblas")
+      lapack_lib = "-L#{Formula["openblas"].opt_lib} -lopenblas"
+    else
+      lapack_lib = "-llapack -lblas"
+    end
+    make_args += ["MFEM_USE_LAPACK=YES", "LAPACK_LIB=#{lapack_lib}"]
+
+    if build.with?("hypre") || build.with?("mpi")
```
dpo: Wouldn't `build.with? "hypre"` would be enough here?

mfem.rb
```diff
@@ -0,0 +1,73 @@
...
+      make_args += ["HYPRE_DIR=#{Formula["hypre"].opt_prefix}",
+                    "HYPRE_OPT=-I#{Formula["hypre"].opt_include}",
+                    "HYPRE_LIB=-L#{Formula["hypre"].opt_lib} -lHYPRE"]
+    end
+
+    if build.with?("metis") || build.with?("mpi") || build.with?("suite-sparse")
```
dpo: The `"mpi"` bit seems redundant.

mfem.rb
goxberry: If I omit `build.with?("mpi")` in this line, then `brew install mfem --with-mpi` omits the hypre options. The package still builds, but without passing those options for hypre, I don't know what hypre library it links to.

mfem.rb
goxberry: If I omit `build.with?("mpi")` on this line, then `brew install mfem --with-mpi` omits the METIS options for `make config`, and the build of the example problem in the `test do` block fails. Both this `or` statement and the one before it appear to be necessary for cases where `"hypre"` and `"metis"` are not specified as options, even though they are listed as dependencies. I don't know why this happens.

mfem.rb
dpo: Oh right. Sorry, it looks ok.

mfem.rb
dpo: Same thing. Because they're not optional/recommended. Sorry, my bad.

mfem.rb
```diff
@@ -0,0 +1,73 @@
...
+  version "3.1"
+  sha256 "841ea5cf58de6fae4de0f553b0e01ebaab9cd9c67fa821e8a715666ecf18fc57"
+
+  option "with-mpi", "Build with mpi support (implies --with-hypre --with-metis)"
+  option "with-suite-sparse", "Build with suite-sparse support (implies --with-metis)"
+
+  depends_on :mpi => [:cxx, :optional]
+  if build.with?("mpi")
+    depends_on "metis"
+    depends_on "hypre" => :mpi
```
dpo: What do you mean here? `"with-mpi`? That's the default.

mfem.rb
goxberry: I added these options intentionally to override the existing text that is printed with `brew info mfem`; if these lines are not added, then a user looking at `brew info` has to go to the source code to see that `brew install mfem --with-mpi` implies installing `--with-hypre` and `--with-metis`. If it's policy not to provide this documentation when the dependency DAG of a formula is complicated, I can remove these lines, but based on earlier experience with Homebrew having undocumented behavior (e.g., regarding MPI), I'd urge you to reconsider.

mfem.rb
goxberry: I think the goal of this line is to try to ensure that a user hasn't installed hypre via `brew install hypre --without-mpi`; I'm not sure if omitting `=> :mpi` (the hash map?) will achieve this goal.

In the MFEM install docs, the developers suggest copying the hypre directory into a specific location in the MFEM directory tree, so if MFEM is built with MPI compilers, hypre will be as well, so this seems like requiring this dependency is the safest course of action.

dpo: That's ok but it should be `"with-mpi` then.

mfem.rb
dpo: It's not policy, just a matter of preference. I prefer to omit them because the user can't use `--with-mpi` and `--without-metis` anyways. If you find undocumented behavior, that is or isn't a bug, please report it. If I remember well, the MPI issues (that lingered for a while) have been fixed. It's ok if you prefer to leave those options.

mfem.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    ENV.deparallelize
+    system "make", "config", *make_args
+    system "make", "all"
+    system "make", "install"
+  end
+
+  test do
+    args = ["-m", "data/star.mesh", "--no-visualization"]
+    if build.with?("mpi")
```
dpo: I don't think this works because you don't pass `--with` options when you test. You have to query what options were used at install time using `Tab.for_name("mfem").with? "mpi"`.

mfem.rb
goxberry: Thanks for the edit! Fixed in the version I'm about to push.

mfem.rb
goxberry: Thank you for the edit! This issue is also fixed in the version I'm about to push.

mfem.rb
```diff
@@ -0,0 +1,73 @@
...
+    ENV.deparallelize
+    system "make", "config", *make_args
+    system "make", "all"
+    system "make", "install"
+  end
+
+  test do
+    args = ["-m", "data/star.mesh", "--no-visualization"]
+    if Tab.for_name("mfem").with? "mpi"
+      system "mpirun", "-np", "4", "examples/ex1p", *args
```
dpo: Where's this located? In `share`? It should be indicated because currently, the executable can't be found.

goxberry: You're right; I didn't realize that `test do` would create its own temporary directory. I think I might have a fix for this, by copying the examples source and the accompanying makefile over to its own directory within the homebrew installation, but I also need to copy along with it some configuration information from the `system "make", "config", *make_args` step and then do a couple string replacements. I'll push up a replacement and notify you when I patch this in the next couple days.

goxberry:
@dpo Updated with a functioning `test do` block, although the test block is a little kludgey because it copies files that are now installed with MFEM.

dpo:
Thanks. We have a few audits now: https://travis-ci.org/Homebrew/homebrew-science/jobs/124903333#L243

mfem.rb
```diff
@@ -0,0 +1,89 @@
...
+    (prefix/"examples").install "examples/makefile"
+    (prefix/"data").mkpath
+    (prefix/"data").install Dir["data/*.mesh"]
+    (prefix/"data").install Dir["data/*.vtk"]
+
+  end
+
+  test do
+    f = Formulary.factory("mfem")
+    system "cp #{f.opt_prefix}/examples/makefile ./"
```
dpo: You can use the function `cp_r`.

goxberry: Sounds good. Had a personal emergency over the weekend. Will get to this
tonight. Thank you for all of your feedback and patience!

On Monday, April 25, 2016, Dominique <notifications@github.com> wrote:

> In mfem.rb
> <https://github.com/Homebrew/homebrew-science/pull/3530#discussion_r60990304>
> :
>
> > +
> > +    # Install some examples for users learning the library
> > +    (prefix/"config").mkpath
> > +    (prefix/"config").install "config/config.mk"
> > +    (prefix/"examples").mkpath
> > +    (prefix/"examples").install Dir["examples/*.cpp"]
> > +    (prefix/"examples").install "examples/makefile"
> > +    (prefix/"data").mkpath
> > +    (prefix/"data").install Dir["data/*.mesh"]
> > +    (prefix/"data").install Dir["data/*.vtk"]
> > +
> > +  end
> > +
> > +  test do
> > +    f = Formulary.factory("mfem")
> > +    system "cp #{f.opt_prefix}/examples/makefile ./"
>
> You can use the function cp_r.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/Homebrew/homebrew-science/pull/3530/files/2e4a4783268e2a7725e780f0e356d12f4e42589d#r60990304>
>


-- 
Geoffrey Oxberry, Ph.D., E.I.T.
goxberry@gmail.com

goxberry:
@dpo Pushed up changes in reply to your suggested edits. Let me know if there is anything else I can do to help this pull request along.

mfem.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    # Install some examples for users learning the library
+    (prefix/"config").mkpath
+    (prefix/"config").install "config/config.mk"
+    (prefix/"examples").mkpath
+    (prefix/"examples").install Dir["examples/*.cpp"]
+    (prefix/"examples").install "examples/makefile"
+    (prefix/"data").mkpath
+    (prefix/"data").install Dir["data/*.mesh"]
+    (prefix/"data").install Dir["data/*.vtk"]
```
dpo: Normally these would go into `pkgshare`, if you don't mind.

mfem.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  test do
+    f = Formulary.factory("mfem")
+    cp_r "#{f.opt_prefix}/examples", "./"
+    cd "examples" do
+      cp "#{f.opt_prefix}/data/star.mesh", "./"
+      system "make", "all", "MFEM_DIR=#{f.opt_prefix}", "MFEM_LIB_DIR=#{f.opt_prefix}/lib", "MFEM_INC_DIR=#{f.opt_prefix}/include"
+      args = ["-m", "star.mesh", "--no-visualization"]
+      if Tab.for_formula(f).with? "mpi"
```
dpo: And here you could use `Tab.for_name("mfem")`

dpo:
Many thanks. We're almost there. Just a few details to fix if you don't mind.

goxberry:
@dpo No problem! Added the changes you suggested; I had to hack the makefile for the MFEM examples a little bit with `inreplace`, but it's nothing drastic.

goxberry:
Figured out a way to simplify the formula very slightly; `config.mk` is copied over by the installation, so there's no need to copy it again. Still should pass the Travis-CI tests.

dpo:
Thanks. Are there single precision calculations in this library? The reason I ask is that the default BLAS/LAPACK on OSX should then be `veclibfort`, or else the Accelerate framework can cause segfaults or errors. See the most up-to-date `suite-sparse` formula for an example.

tzanio:
@dpo, there shouldn't be any single precision calculations in mfem proper.

goxberry:
@dpo There aren't any single precision calculations in mfem that I can see, having searched for some of the common LAPACK calls (e.g., sgemm, sgemv). @tzanio is the lead developer of MFEM, so he's the authority on what is in the library.

dpo:
Ok thank you.

dpo:
@BrewTestBot test this please

PR #3531, opened 2016-04-12
.travis.yml: Xcode 7.3, one OS X build.

mikemcquaid:
We don't need three OS X builds here (I mean really we don't need one) considering they are handled by Jenkins now.


dpo:
Wasn't the plan for Travis builds to take some load off of Jenkins?

mikemcquaid:
That was the plan but Travis doesn't work sufficiently reliably, fast enough or allow bottling so: we can't do that.

dpo:
Got it!

PR #3532, opened 2016-04-12
graph-tool 2.14

samifahed:


dpo:
There's already #3390 that we couldn't get to work right.
cc @count0

samifahed:
Closing this in favor of #3390 since now it's pointing to the 2.14 version

PR #3537, opened 2016-04-13
blast 2.3.0

ilovezfs:
- the boost 1.58 to 1.59 broke the formula unless you passed
  `--without-check`. So let's vendor in 1.58.
- remove the patches that upstream has merged
- deprecate the option `--without-check` and rename it `--without-test`
- have the test match against the version


ilovezfs:
@BrewTestBot test this please

ilovezfs:
@sjackman @DomT4 Jenkins CI is :green now (I don't expect the same out for Travis due to timeout while building Boost). :ship:?

sjackman:
Thanks for this work, Joe! I'm not super keen on bundling in `boost` 1.58. If it's only `make check` that's broken with current `boost`, I'd probably rather disable `make check`. @iml @tseemann Your thoughts?

ilovezfs:
@sjackman yeah let's just get rid of it.

ilovezfs:
@sjackman All :green_apple: again. Boost removed. Also, made the default dynamic libs, which I then discovered was broken. The symlink hack worked around the problem. Not sure how you feel about making dynamic the default.

sjackman:
I usually go with whatever the upstream default is, assuming that they have their reasons. I believe upstream defaults to static linking to speed up startup time, because people often run `blast` in a tight loop. I'm fine with dynamic linking personally.

Looks good to me anyway. Merge it! Thanks, Joe.

tseemann:
@sjackman Yes, I would just skip the `make check`. Bundling `boost` will just cause problems.

PR #3540, opened 2016-04-14
bowtie: fixes stdout for builds with clang on OS X

ilovezfs:
Upstream PR BenLangmead/bowtie#25 avoids implementation specific
`pubsetbuf`.


ilovezfs:
@mikemcquaid Do you prefer those trinity commits squashed together?

sjackman:
Can you split this PR into three separate PR, one for each formula?
If the first `trinity` commit fixes 2.0.6, and the second commit bumps the version to 2.2.0, then I prefer two commits.

ilovezfs:
@sjackman Sure. The purpose of having them together was to get the CI to actually test that it fixed the issue (trinity needed to at least build with gcc5 or nothing would be tested with respect to the original bug) but now that we know it all works, there's no reason to keep them together.


>If the first trinity commit fixes 2.0.6

Yep, it does.

sjackman:
Ah, thanks for the explanation.

ilovezfs:
@sjackman OK, they're all split now. I'd review them in this order: atram, bowtie, trinity206, trinity220

sjackman:
:+1: Thanks, Joe.

ilovezfs:
@BrewTestBot test this please

PR #3546, opened 2016-04-15
trinity: audit fix and re-test with patched bowtie

ilovezfs:
- audit fix for description
- make sure the trinity test-do works with fixed bowtie
- java req if not linux
- gcc < 5 restriction


trinity.rb
```diff
@@ -17,12 +19,16 @@ class Trinity < Formula
...
 
+  depends_on :java => "1.7+" unless OS.linux?
+
   needs :openmp
 
   fails_with :llvm do
     cause 'error: unrecognized command line option "-std=c++0x"'
   end
 
+  fails_with :gcc => "5"
```
sjackman: Can you include a `cause` with the compiler error message?

sjackman: Ah, I see this `fails_with` is removed in the next commit. Don't worry about it then.

sjackman:
:+1:

sjackman:
> java req if not linux

Why not Linux? Is the `:java` requirement not working on Linux?

ilovezfs:
@sjackman The commit history seemed to suggest that something regarding Linux was the reason `:java`, which was there previously, had been removed, so I was trying to respect that, but for all I know, it's actually fine.

PR #3547, opened 2016-04-15
trinity 2.2.0

ilovezfs:
- version bump to 2.2.0
- verify the test now works with patched bowtie
- unbundle deps we already have
- removes the < gcc 5 restriction

PR #3556, opened 2016-04-16
express: use boost dynamic libs

ilovezfs:
Fixes "libboost_thread-mt.a: error adding symbols: Bad value"

Also,
- bump revision
- audit fixes
- options for protobuf and gperftools


sjackman:
:+1: Thanks, Joe!

ilovezfs:
@sjackman Do you have any interest in making either/both of the new options (protobuf and gperftools) recommended instead of optional?

SInce Linuxbrew doesn't use superenv, the build fails if the user happens to already have protobuf installed on the machine. Building `--with-protobuf` worked around that forcing it to use Linuxbrew' protobuf instead. Since the build succeeds if there's no protobuf installed at all, the conservative thing to do would be to make it optional as it is here. However, since it enhances the functionality and avoids the potential gotcha, it might be ok to make it recommended despite the additional dependency so that that functionality it makes it into the bottles.

Similar situation with gperftools, but its presence didn't cause build failure. However, it also enhances the functionality of the software (speed apparently), so also might be justifiable as recommended so that that functionality makes it into the bottles.

sjackman:
Both `protobuf` and `gperftools` have no dependencies, so I'm fine with making them `:recommended`. I'm generally okay with making small-to-medium dependencies that add useful functionality `:recommended`, especially once bottling is more common on Linux.

> SInce Linuxbrew doesn't use superenv, the build fails if the user happens to already have protobuf installed on the machine.

`brew install gcc --with-glibc` is the default (a dependency of all bottles) on older machines with `glibc` < 2.19, and it removes `/usr/lib` and `/usr/include` from the default search path. I haven't done this for newer machines with `glibc` ≥ 2.19, where the host `glibc` and `gcc` are used.

sjackman:
Care to merge this PR and #3547? Thanks for the work, Joe.

ilovezfs:
@sjackman I'd been hoping for some help with https://github.com/Homebrew/homebrew-core/pull/383#issuecomment-211298660 but yes, let's go ahead without making gperftools recommended for now until that's sorted out.

sjackman:
Thanks, Joe!

ilovezfs:
@sjackman You're welcome. Any idea what's up with https://travis-ci.org/Homebrew/homebrew-science/jobs/125705454

sjackman:
My best guess is that Travis is using Hackintosh hardware that does not support an instruction supported by Core2Duo, which is the target bottle architecture. See https://github.com/Homebrew/legacy-homebrew/issues/49829

ilovezfs:
@sjackman that's a testable theory if we do something similar to this to check: https://github.com/Homebrew/homebrew-x11/pull/202/files#diff-78480132cbc2f9c290708f88bc7b6a6cR30

By the way, why is that Travis link seemingly disconnected from the PR? Or does is it trigger after merge as opposed to before or something?

sjackman:
It stops displaying the CI status once the PR is closed.

sjackman:
I get this error on my Linux system when compiling `express`:
```
[100%] Linking CXX executable express
cd /var/tmp/sjackman/express20160425-123339-8xxjia/express-1.5.1-src/src && /gsc/btl/linuxbrew/Cellar/cmake/3.5.1/bin/cmake -E cmake_link_script CMakeFiles/express.dir/link.txt --verbose=1
/gsc/btl/linuxbrew/bin/g++-5   -Wall -DNDEBUG   -L/gsc/btl/linuxbrew/lib -Wl,--dynamic-linker=/gsc/btl/linuxbrew/lib/ld.so -Wl,-rpath,/gsc/btl/linuxbrew/lib CMakeFiles/express.dir/main.cpp.o CMakeFiles/express.dir/fragments.cpp.o CMakeFiles/express.dir/mapparser.cpp.o CMakeFiles/express.dir/sequence.cpp.o CMakeFiles/express.dir/mismatchmodel.cpp.o CMakeFiles/express.dir/biascorrection.cpp.o CMakeFiles/express.dir/bundles.cpp.o CMakeFiles/express.dir/robertsfilter.cpp.o CMakeFiles/express.dir/markovmodel.cpp.o CMakeFiles/express.dir/targets.cpp.o CMakeFiles/express.dir/lengthdistribution.cpp.o CMakeFiles/express.dir/threadsafety.cpp.o CMakeFiles/express.dir/directiondetector.cpp.o CMakeFiles/express.dir/alignments.pb.cc.o CMakeFiles/express.dir/targets.pb.cc.o  -o express -rdynamic /gsc/btl/linuxbrew/lib64/libboost_thread-mt.so /gsc/btl/linuxbrew/lib64/libboost_system-mt.so /gsc/btl/linuxbrew/lib64/libboost_filesystem-mt.so /gsc/btl/linuxbrew/lib64/libboost_program_options-mt.so /gsc/btl/linuxbrew/lib64/libboost_date_time-mt.so /gsc/btl/linuxbrew/lib64/libboost_chrono-mt.so /gsc/btl/linuxbrew/lib64/libboost_atomic-mt.so /gsc/btl/linuxbrew/lib64/libz.so ../bamtools/lib/libbamtools.a -lpthread -Wl,-Bstatic -lprotobuf -Wl,-Bdynamic -Wl,-rpath,/gsc/btl/linuxbrew/lib64 
CMakeFiles/express.dir/main.cpp.o: In function `parse_options(int, char**)':
main.cpp:(.text+0x6e7): undefined reference to `boost::program_options::options_description::options_description(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int, unsigned int)'
```

I believe it `depends_on "boost" => "c++11"`. I have seen similar errors before with other formula. This problem is more general, and need an encompassing solution.

ilovezfs:
Hmm. Travis status was green when I merged and the report below appears to be still there but different one from the failed one I noticed only because it emailed me.

ilovezfs:
@sjackman Maybe boost bottles on Linux (or with GCC?) should be default?

PR #3557, opened 2016-04-16
harry 0.4.2 (new release)

rieck:
This release fixes the incorrect implementation of the bag distance.


harry.rb
```diff
@@ -1,7 +1,7 @@
 class Harry < Formula
   homepage "http://www.mlsec.org/harry"
```
dpo: Would you mind adding a `desc` field?

rieck:
Squashing is such a pain. I'll post a new request.

PR #3565, opened 2016-04-17
glpk: trigger CI

ilovezfs:
Let's make sure this passes now.


ilovezfs:
OK, CI now passes. Cf. https://github.com/Homebrew/homebrew-science/pull/3560

PR #3566, opened 2016-04-17
bamtools: needs el capitan bottle

ilovezfs:


ilovezfs:
Hmm, wonder if we can de-conflict the CI so this doesn't fail in perpetuity.

PR #3572, opened 2016-04-20
augustus 3.2.2

ntkme:
This PR updates augustus to 3.2.2:

- `boost` becomes a hard requirement.
- `ZIPINPUT=true` is on by default. In addition, because `boost` is now a hard requirement, there is no reason to turn this option off. 
- ~~The tar contains linux ".o" files, which requires `make clean` before `make`. I have already reported that to the developer. I know it is an upstream problem, but I'm submitting this PR anyways, since I found existing formula (swetest.rb) doing the same thing for the same reason.~~ Fixed in upstream.


dpo:
Thanks. Could you please add a comment in the formula indicating the reason for the `make clean`, and that that line should be removed in future upgrades?

ntkme:
It just got fixed in upstream. Unfortunately, the developer didn't bump the version number, so I have to remove fossies.org mirror due to the shasum mismatch.

nextgenusfs:
`etraining` needs to also be soft linked into the executable bin, i.e.
```
bin.install_symlink "../libexec/bin/etraining"
```

dpo:
@BrewTestBot test this please

PR #3584, opened 2016-04-25
g2o 2016-04-24 (new formula)

adrelino:
Added minimal g2o formula with tagged version 2016-04-24

### DIFF
--- a/g2o.rb
+++ b/g2o.rb
@@ -1,0 +1,24 @@
+class G2o < Formula
+  desc "A General Framework for Graph Optimization"
+  homepage "http://openslam.org/g2o.html"
+  url "https://github.com/RainerKuemmerle/g2o/archive/20160424_git.tar.gz"
+  version "2016-04-24"
+  sha256 "31abd5f4acf8407c18318b04f3ffc867c8ea7de89f18d51f9e92eb734d523b24"
+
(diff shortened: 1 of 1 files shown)


dpo:
Small audit:
```
==> audit problems
homebrew/science/g2o:
 * Description shouldn't start with an indefinite article (A)
```

g2o.rb
```diff
@@ -0,0 +1,24 @@
...
+  depends_on "cmake" => :build
+  depends_on "eigen"
+  depends_on "suite-sparse" => :recommended
+  depends_on "qt" => :optional
+  depends_on "libqglviewer" => :optional
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make", "install"
```
dpo: Are there unit tests that can be run at install time?

g2o.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+  end
+
+  test do
+    system "#{bin}/g2o", "-v"
```
dpo: Isn't there a better test than this? Ideally, a test should be a sample program that links with the library.

adrelino:
Thanks for closing this! I didn't have enough experience to properly write the install and test targets.

PR #3590, opened 2016-04-26
reapr 1.0.18

ilovezfs:
- bump version
- update homepage url
- switch tabix dependency over to htslib
- add smalt dependency in lieu of the new upstream vendored dependency
- modify test to run the upstream test.sh
- install manual


ilovezfs:
@sjackman mind reviewing this? In particular, requiring r might be a little heavy handed, but it seemed worth it so that the actual upstream test.sh could be run in the test block.

sjackman:
If it's only needed for `brew test`, perhaps make the `r` dependency `:recommended` rather than required. Perhaps there's an argument for `depends_on "r" => :test`.

sjackman:
Otherwise looks good.

PR #3594, opened 2016-04-26
express: trigger CI

ilovezfs:
hoping to cause the illegal instruction error


ilovezfs:
@sjackman any idea why this didn't engage Travis?

sjackman:
No, I'm not sure. It triggered BrewTestBot. Can you try `git commit --amend && git push --force`?

ilovezfs:
This time it did trigger the Travis CI. Still no sign of the illegal instruction error. I'm not sure where that failed run was even coming from, since Travis CI had also passed before the latest commit and bottles.

The two passed runs:
https://travis-ci.org/Homebrew/homebrew-science/builds/125698937
https://travis-ci.org/Homebrew/homebrew-science/builds/127975388

The failed run:
https://travis-ci.org/Homebrew/homebrew-science/jobs/125705454

If someone has access to core-only hardware, it would be interesting to see if this is reproducible there.

PR #3596, opened 2016-04-26
Update SeqAn to install CMake and pkg-config files

laserson:
### Have you:

- [ ] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?

### New formula: have you

The current Formula does not install the `FindSeqAn.cmake` file in the `share` directory.  This change installs the files to the correct locations.  cc @sjackman

### DIFF
--- a/seqan.rb
+++ b/seqan.rb
@@ -15,7 +15,7 @@ class Seqan < Formula
 
   def install
     include.install "include/seqan"
-    share.install Dir["share/*"]
+    pkgshare.install Dir["share/*"] unless build.head?
   end
 
   test do
--- a/seqan.rb
+++ b/seqan.rb
@@ -15,7 +15,7 @@ class Seqan < Formula
 
   def install
     include.install "include/seqan"
-    doc.install Dir["share/doc/seqan/*"] unless build.head?
+    share.install Dir["share/*"]
   end
 
   test do


seqan.rb
```diff
@@ -15,7 +15,7 @@ class Seqan < Formula
 
   def install
     include.install "include/seqan"
-    doc.install Dir["share/doc/seqan/*"] unless build.head?
+    share.install Dir["share/*"]
```
sjackman: The upstream git repository doesn't have a `share` directory. Does `brew install --HEAD seqan` work with this patch? If not, it should probably be `share.install Dir["share/*"] unless build.head?`

sjackman: It should be `pkgshare.install` I think rather than `share`.

laserson: Good point wrt the `share/` dir

seqan.rb
```diff
@@ -15,7 +15,7 @@ class Seqan < Formula
 
   def install
     include.install "include/seqan"
-    doc.install Dir["share/doc/seqan/*"] unless build.head?
+    pkgshare.install Dir["share/*"] unless build.head?
```
dpo: It would really be best to install the documentation separately in `doc` if that's possible.

sjackman: `doc` expands to `share/doc/seqan` and
`pkgshare` expands to `share/seqan`

ilovezfs:
Closing in favor of https://github.com/Homebrew/homebrew-science/pull/3897.

@laserson Thanks for opening this PR. I wouldn't have noticed the request from upstream for these changes otherwise.

PR #3606, opened 2016-04-29
ropebwt2: Add desc and tag bioinformatics

sjackman:


sjackman:
@BrewTestBot test this please

PR #3607, opened 2016-04-29
tabix: Add desc and tag bioinformatics

sjackman:


ilovezfs:
I was actually planning on just boneyarding this formula once https://github.com/Homebrew/homebrew-science/pull/3590 is merged.

sjackman:
I've merged this for posterity. Please go ahead with boneyarding it.

ilovezfs:
:+1:

PR #3612, opened 2016-04-29
new recipe: DSDP-5.8

sxprophet:


dsdp.rb
```diff
@@ -0,0 +1,77 @@
...
+    system "make", "dsdpapi", "LAPACKBLAS=" + lapacklib
+
+    # bin contains just test files
+    # bin.install Dir['bin/*']
+    lib.install Dir["lib/*"]
+    include.install Dir["include/*"]
+  end
+
+  test do
+    # tests alreaddy included in build phase
```
dpo: We still need a test here. Preferably a short program that links with the library.

sxprophet: Let's use an official example and data for test.

dsdp.rb
```diff
@@ -0,0 +1,86 @@
...
+    # bin.install Dir['bin/*']
+    lib.install Dir["lib/*"]
+    include.install Dir["include/*"]
+    # reuse official example for test
+    pkgshare.install "examples/maxcut.c"
+    pkgshare.install "bin/graph1"
+  end
+
+  test do
+    if build.with? "openblas"
```
dpo: This won't work. You need to use `Tab.for_name("dsdp").with? "openblas"`.

dsdp.rb
```diff
@@ -0,0 +1,86 @@
...
+    DATA
+  end
+
+  depends_on "openblas" => :optional
+
+  def install
+    if build.with? "openblas"
+      lapacklib = "-L#{Formula["openblas"].opt_prefix}/lib -lopenblas"
+    else
+      lapacklib = "-framework Accelerate"
```
dpo: Is there any chance this will use single precision or complex arithmetic? If so, you need to use `veclibfort` instead of the Accelerate framework.

sxprophet: DSDP does all computation with `double`. There is no complex number computation in semidefinite programming domain. It seems we don't need `vecLibFort`.

dsdp.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    ENV["DSDPROOT"] = Dir.pwd
+    system "make", "dsdpapi", "LAPACKBLAS=" + lapacklib
```
dpo: Here it would be best to pass `CC` to `make` for consistency. In `test`, you use `ENV.cc`.

dsdp.rb
```diff
@@ -0,0 +1,86 @@
...
+    pkgshare.install "bin/graph1"
+  end
+
+  test do
+    if Tab.for_name("dsdp").with? "openblas"
+      lapacklib = ["-L#{Formula["openblas"].opt_prefix}/lib", "-lopenblas"]
+    else
+      lapacklib = ["-framework", "Accelerate"]
+    end
+    system ENV.cc, "-o", "maxcut", "-I#{include}", "#{pkgshare}/maxcut.c", "-L#{lib}", "-ldsdp", *lapacklib
```
dpo: `opt_include` and `opt_pkgshare`.

Issue #3150, opened 2015-12-27
vtk5 shouldn't link directly to python


tdsmith:
Per https://github.com/Homebrew/homebrew-science/pull/3145#issuecomment-167444931, an importable vtk5 module links directly to a Python framework, [which is bad](http://blog.tim-smith.us/2015/09/python-extension-modules-os-x/); we should remove that linkage if we can.

I've self-assigned to indicate that I intend to follow through with this but anyone should feel free to lend a hand.

iMichka:
VTK5 is gone since #3616. Nobody complained about this, at least not to my knowledge. The python linking problem has been fixed here #3811, for VTK 7.1.0.

There is nothing more to be done here; closing.

PR #3616, opened 2016-04-30
vtk5: remove formula

iMichka:
### Have you:

- [ ] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors? **Does not apply here**
- [ ] Built your formula locally prior to submission with `brew install <formula>`? **Does not apply here**

VTK 5.10.0 was released on Oct 17, 2012
VTK 6.0.0 was released on Jun 21, 2013

The vtk5 formula was moved from homebrew-versions to
homebrew-science, see commit a98415fa71f3ed2375fe60be27bca2d93aa0f9ea
There is no explanation in the commit message for this; but
the guess is that as there were big API changes between
vtk5 and vtk6, and it was probably good to keep
a vtk5 formula here.

Since then, vtk7 has been released. Most people seem to
have done the transition. Looking at the download statistics
on bintray for the vtk5 bottles, the number of downloads
has been very low for the last year.

If somebody really needs vtk5, they can still maintain this
formula in a separate tap.


iMichka:
I am ready to add the formula to homebrew-versions if needed. But I think that if somebody really needs it he can: tap it, or open it on homebred-versions himself and try to argument why this is still really needed.

Also, merging this PR will automatically close #3150, as this will no more be needed.

PR #3630, opened 2016-05-05
r 3.3.0

erichare:
### Have you:

### New formula: have you


tdsmith:
The formulas look great, thanks!

This looks like it'll be a bit of a project to pull; test-bot isn't building R first so everything else is dragging.

Other issues that want to be addressed:
* nonpareil tries to write directly to R's public site-library, which it should not be doing.
* poretools has an invalid checksum -- but actually no longer has R dependencies, so we should stop trying to install them.
* quake installs .R files to `bin`; is that right?

Thanks for your help; I don't think there's anything else you need to do. I'll use your commits and split it into a couple of PRs once we have the dependencies sorted out and pull one and then build the other; it'll be broken for a few minutes but hopefully nobody will notice.

erichare:
Sounds great. Thank you very much!

tdsmith:
Just to set expectations, my advancement exam is Thursday (!) and it's not likely that I'll work on this before then, if I know what's good for me

rkrug:
I would suggest to switch to https downloads instead of the http.

rkrug:
Would it be possible to pull the update of R and the other working recipies and to leave the recipies for poretools and nonpareil as is for the moment?

yihui:
Yeah I think it is a good idea to download from https.

tdsmith:
I'll do the https change; looks like both mirrors support it (good work, R-core!)

tdsmith:
Also I found myself with some breathing room tonight so I'll spend some time on this; NB that you'll have to reinstall all your packages

tdsmith:
Landed; thanks all!

yihui:
Excellent. Good luck to your exam!

PR #3663, opened 2016-05-16
rstudio-server 0.99.902

randy3k:
### Have you:


Eric-Guo:
boost150 removed from Homebrew, so should using 155 instead.

randy3k:
Oh, why boost150 was removed...I remember there were problems in using boost155.
More importantly, RStudio is still using boost150.

randy3k:
Before the issue of boost150 is resolved, the formula for rstudio-server will be temperately hosted at https://github.com/randy3k/homebrew-rstudio-server

randy3k:
it turns out that rstudio-server could link against the most updated version of boost with a little patch.

ilovezfs:
@randy3k That's good news. Have you reported this upstream?

randy3k:
The issue was caused by the old [friend](https://github.com/Homebrew/legacy-homebrew/issues/30512), the c library `libc++` and `libstdc++`. 

RStudio assumes that the boost library was linked against `libstdc++`, so it forces to use `libstdc++` when compiling itself. And unfortunately the homebrew boost was linked against `libc++`, so I have to patch the CMakefile to make sure `libc++` is used indeed.

ilovezfs:
It would be good to have a comment in the formula with an upstream bug report link nudging upstream to fix that assumption.

randy3k:
Actually, upstream doesn't solely assume, they provide a [installation script](https://github.com/rstudio/rstudio/blob/master/dependencies/common/install-boost#L27) of boost (which linked against stdlibc++).

But anyway, I will inform them that there is no problem to link against `libc++` (at least on Mac OS X 10.10).

I believe one of the reasons for them to keep using `stdlibc++` is to retain support of old platforms.

randy3k:
any chance to review and merge it?

Eric-Guo:
I fetch & installing 902 of rstudio-server in this PR, my network is slow, but will give feedback within 1 day.

tdsmith:
Where did this shell script come from? We should ask RStudio to ship it instead of embedding it in the formula if it's useful.

Can you link to a place where RStudio have commented on the libc++/libstdc++ issue?

Thanks!

randy3k:
The script file is from [RStudio](https://github.com/randy3k/rstudio/blob/master/src/cpp/server/extras/admin/rstudio-server.in) repo, however they only render it for linux, (RStudio Server does not officially support Mac.) 

[This](https://github.com/rstudio/rstudio/commit/a07274ed4ef1b9eae3bfe672f3de1ff25b5b0856) is the relevant commit for the  libc++/libstdc++ issue.

Eric-Guo:
@randy3k Thanks your effort to maintain rstudio-server for brew, I have another question, maybe not relative to this PR, but I guess will affect user after merge.

When i running studio-server when start, always got below warning.

```text
During startup - Warning messages:
1: Setting LC_CTYPE failed, using "C" 
2: Setting LC_COLLATE failed, using "C" 
3: Setting LC_TIME failed, using "C" 
4: Setting LC_MESSAGES failed, using "C" 
5: Setting LC_MONETARY failed, using "C" 
```

I try to add `LC_ALL=en_US.UTF-8` to `/usr/local/Cellar/rstudio-server/0.99.902/bin/rserver` following [RStudio Support](https://support.rstudio.com/hc/en-us/community/posts/200662477-System-locale-on-Mac-os-X-set-by-default-to-an-UTF-8) or running `defaults write org.R-project.R force.LANG en_US.UTF-8` in [MacOS FAQ](https://cran.r-project.org/bin/macosx/RMacOSX-FAQ.html#Internationalization-of-the-R_002eapp), or not work.

Any comments?

tdsmith:
Seconded; thanks for your work, Randy!

> This is the relevant commit for the libc++/libstdc++ issue.

Thanks for the link! Can you post a comment on their support forum, link to that, and add it as a Ruby comment to the relevant line in the formula?

> The script file is from RStudio repo, however they only render it for linux,

Can we change that logic and request that upstream makes the same change instead of embedding the file in the formula?

randy3k:
@Eric-Guo 
Does it happen for RStudio desktop or R terminal console?
How about setting both of these?
```
export LANG=en_US.UTF-8
export LC_ALL=en_US.UTF-8
```

randy3k:
> Thanks for the link! Can you post a comment on their support forum, link to that, and add it as a Ruby comment to the relevant line in the formula?

It is easy.

> Can we change that logic and request that upstream makes the same change instead of embedding the file in the formula?

I can try.

randy3k:
I just pushed a PR to RStudio on the admin script https://github.com/rstudio/rstudio/pull/712
However, I think we should merge this for now instead of waiting RStudio to merge the PR and release a stable release.

Once the next stable release of RStudio is out, I will update the formula to reflect the update.

PS: the failure is not related to RStudio, but the upstream R.

randy3k:
@tdsmith 
My PR to RStudio has been merged so I have updated the formula to use the scripts on the RStudio git repo.

Eric-Guo:
@randy3k 

It's in rstudio-server browser, but your hit works! it need to set the LANG as well.

I suggest you add below two line to `rserver` startup script, so it won't show warning at after installation.

```bash
export LANG=en_US.UTF-8
export LC_ALL=en_US.UTF-8
```

rstudio-server.rb
```diff
@@ -109,51 +123,106 @@ def install
...
-              <array>
-                      <string>#{opt_prefix}/rstudio/bin/rserver</string>
-              </array>
-              <key>RunAtLoad</key>
-              <true/>
-      </dict>
-      </plist>
+    (bin/"rserver").write <<-EOS.undent
+      #!/usr/bin/env bash
+      export PATH=#{opt_prefix}/rstudio/bin:$PATH
```
Eric-Guo:
```
export LANG=en_US.UTF-8
export LC_ALL=en_US.UTF-8
```

Add above line will save user like me searching in google

randy3k: there may be users using other locale?

randy3k: Alternatively, I think you could specific 

```
LANG=en_US.UTF-8
```

in `$HOME/.Renviron`

tdsmith:
Looking good! Homebrew has a mechanism for writing plists and generating auto-help text; you can grep homebrew-core for `def plist` if you want to try to use that. Otherwise I'll do that and pull soon.

randy3k:
For now, the launchd plist is downloaded and rendered from the RStudio repo since 0.99.902 does not have it. In the future, the plist and the admin script will be installed automatically when building RStudio Server (becoz of my PR to RStudio). I think there is no need to use the `def plist` option since it will not be needed anyway in the future.

rstudio-server.rb
randy3k: It turns out that RStudio assume UTF-8, this probably is a good idea.

randy3k:
I feel good now. Let me know if there are other concerns. If not, I think it is ready to merge.

randy3k:
@tdsmith any chance to review it?

randy3k:
Perhaps I am being annoying, but it has been another two weeks, could somebody review this?

tdsmith:
@BrewTestBot test this please

tdsmith:
I'd like the caveats to be shorter but not as much as I'd like to land this; really sorry for the delay. Thanks again for all your work here.

tdsmith:
(Er, thanks in the first place, probably!)

PR #3673, opened 2016-05-17
asndisc: new formula

vivekkrish:
### Have you:

### New formula: have you


tdsmith:
Binary-only and unversioned; deferring to @sjackman's judgement

sjackman:
It would be better to compile and install this binary from source code. I expected the source code for asndisc to be included in `ncbi-c++-toolkit` (`ncbi_cxx--12_0_0.tar.gz`), but I don't see it in there. I've contacted upstream to ask them which tarball contains `asndisc`.

sjackman:
The upstream tarball is ftp://ftp.ncbi.nlm.nih.gov/toolbox/ncbi_tools/old/20120620/ncbi.tar.gz
These tools, `asdisc` and `asnval`, should be built from source. Take care not to install an executables that might conflict with `ncbi-c++-toolkit`, `blast`, `rmblast` or `sratoolkit`, also from NCBI.

vivekkrish:
Hi Shaun,

I created the 3 formulas for asndisc, asnval and asn2gb similar to the
already existing tbl2asn formula.

Locally, I've tried compiling from source, but it is a rather convoluted
process (due to the inherent complexity of the ncbi-toolbox codebase), also
one of the reasons why NCBI provides us with pre-built binaries for Mac and
Linux.

This also applies to all other tools in the toolkit, available for download
here: ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/.

Thank you.

>

vivekkrish:
All binaries provided by NCBI are already versioned, the information is
tracked here:
ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/sqn2agp/DOCUMENTATION/VERSIONS

>

sjackman:
I created the formula for `ncbi-c++-toolkit`, which was rather difficult, so believe me you have my understanding that it's not straight forward to build from source. It should all the same be built from source. Looking at the formula for `ncbi-c++-toolkit` may help you. Also the Debian package build rules: http://anonscm.debian.org/cgit/debian-med/ncbi-tools6.git/tree/debian/rules#n93

vivekkrish:
Ok. Sounds good! Will work on a new formula.

BTW, if we are building from source, should we have one "all encompassing"
formula for all the concerters withing the NCBI Toolbox, or should it be
one per converter?

Appreciate your input!

Thank you.

>

sjackman:
One all encompassing formula named `ncbi-tools` after the Debian package. Thanks, Vivek!

PR #3682, opened 2016-05-18
nest: enable building with HEAD

tammoippen:
### Have you:

- [X] Removed the `revision` line, if any, when bumping the version number? (There is none.)
- [X] Added/bumped the `revision` number if the changes affect the precompiled bottles? (Should not affect the bottle.)

In HEAD we changed the build system to `cmake`, hence the building needs to be changed in that case. Also add missing Python dependancy `nose` for testing and have the C/CXX flags enabled found by configure/cmake.

Many issues discovered creating this formula are fixed in the current master of NEST - this makes installing HEAD much simpler than 2.10.0 . In #3331 I mentioned that there is a problem with the [python bindings](https://github.com/Homebrew/homebrew-science/pull/3331#issuecomment-210463734) - the issue is fixed in HEAD, but still remaining in nest 2.10.0. So when using the Python bindings, please install from HEAD.


nest.rb
```diff
@@ -23,18 +23,46 @@ class Nest < Formula
...
+  depends_on "automake" => :build unless build.head?
+  depends_on "cmake" => :build if build.head?
 
   fails_with :clang do
     cause <<-EOS.undent
       Building NEST with clang is not stable. See https://github.com/nest/nest-simulator/issues/74 .
     EOS
   end
 
+  env :std
```
dpo: What's the issue with superenv?

nest.rb
```diff
@@ -23,18 +23,46 @@ class Nest < Formula
...
+      system "make", "install"
+    end
+  end
+
   def install
+    ENV.delete("CFLAGS")
+    ENV.delete("CXXFLAGS")
+
+    if build.head?
+      install_head
```
dpo: I think it would make more sense to inline the function since the configure mechanism will be phased out in future releases.

nest.rb
tammoippen: When I inspected `$HOME/Library/Logs/Homebrew/nest`, I saw that some flags were replaced, e.g. `-O2` was replaced by `-Os`. After changing to `env :std` and `ENV.delete("CFLAGS")` and `ENV.delete("CXXFLAGS")` the flags selected by `configure` were used.

nest.rb
tammoippen: Done.

nest.rb
dpo: Ok, thanks.

PR #3700, opened 2016-05-27
Octave: fix hdf5 library

schoeps:
### Have you:

### What's new?
- ensure that the correct hdf5 library is used


schoeps:
I would like to make `fltk` recommended (instead optional) but then I would have to specify `--devel` since fltk-1.3.3 does not work on recent Mac OS X. Objections?

dpo:
The trouble is that `fltk --devel` isn't bottled. That'll prevent `octave` from being bottled I believe.

schoeps:
Even worse, I realized that you cannot ask for `--devel` similarly as for other options... so, I will wait until the next official fltk release :(

dpo:
Good point. Thanks!

PR #3712, opened 2016-06-04
rstudio-server: bump to boost155 as boost150 removed

Eric-Guo:


Eric-Guo:
Duplicated to #3663, prefer merge that

Issue #3370, opened 2016-02-25
PCL 1.7.2 with VTK 7


ubuntuslave:
I'm having runtime issues when running *PCL 1.7.2* and *VTK 7.0*. The installation succeeds, but the libvtk dependencies looked for are still for VTK 6.2.

Installing PCL 1.7.2 `--without-vtk` lets PCL run at least for now. Has anyone know how to home-brew *vtk6* instead of  *vtk7* (the default nowadays)

dpo:
`pcl` probably just needs to be bumped. Feel free to submit a PR for that. I might not get to it right away. In the meantime, you'll get VTK7 support if you `brew install pcl --build-from-source`.

dpo:
There's a build error unfortunately:
```
[ 22%] Building CXX object visualization/CMakeFiles/pcl_visualization.dir/src/histogram_visualizer.cpp.o
[...]

Issue #3347, opened 2016-02-21
PCL, Error 2 for pcl_filters


dmitrievk:
I use `brew install pcl --with-example` to build pcl 1.7.2 with vtk 7.0 on el capitan. I got the following error:
cd /tmp/pcl20160221-37183-u6tl42/pcl-pcl-1.7.2/macbuild/filters && /usr/local/Cellar/cmake/3.4.3/bin/cmake -E cmake_symlink_library ../lib/libpcl_filters.1.7.2.dylib ../lib/libpcl_filters.1.7.dylib ../lib/libpcl_filters.dylib
[ 23%] Built target pcl_filters
make: *** [all] Error 2

Here are the build logs:
[https://gist.github.com/anonymous/302b50543d251421b80d](url)

earthflat:
Same error for me on Yosemite!

dmitrievk:
Works with --HEAD though

earthflat:
I need to use recent stable version 1.7.2 of PCL with regards to some dependencies. Could somebody fix or help with the problem of VTK 7.0 with stable PCL 1.7.2?

drhalftone:
I get the error with "brew install pcl --without-apps --without-tools --without-vtk --without-qt"

dpo:
`pcl` was updated to 1.8.0 which should build against vtk7. Please feel free to open a new issue if problems persist.

PR #3725, opened 2016-06-08
[PCL] Fixes VTK 7 support for 1.7.2

fran6co:
This fixes 2 issues:

  * ```brew install pcl``` isn't using the bottle because VTK is built without qt support by default
  * 1.7.2 doesn't support VTK 7 (https://github.com/Homebrew/homebrew-science/pull/3512, https://github.com/Homebrew/homebrew-science/issues/3370, https://github.com/Homebrew/homebrew-science/issues/3347)

Merges:

  * https://github.com/PointCloudLibrary/pcl/pull/1500
  * https://github.com/PointCloudLibrary/pcl/pull/1534


pcl.rb
```diff
@@ -31,19 +31,30 @@ def message
...
+  revision 1
 
   stable do
+    url "https://github.com/PointCloudLibrary/pcl/archive/pcl-1.7.2.tar.gz"
+    sha256 "479f84f2c658a6319b78271111251b4c2d6cf07643421b66bbc351d9bed0ae93"
     patch do
       url "https://gist.githubusercontent.com/fran6co/a6e1e44b1b43b2d150cd/raw/0c4aeb301ed523c81cd57c63b0a9804d49af9848/boost.patch"
       sha256 "5409b0899f65d918248a8fdfb820478cc0b191c50339e16692a911fab76c3f43"
     end
     # Fixes PCL for VTK 6.2.0
```
dpo: Is this useful? We don't support VTK 6.

fran6co: Yes, it's actually a patch for VTK >= 6.2.0

dpo: There's already a patch below for VTK 7, which is the only one in homebrew/science. I'm just checking whether both patches are needed for VTK 7.

fran6co: The patch below builds on the previous patch. This is going to go all away when 1.8.0 gets released. 1.7.2 is fairly old already.

dpo: Ok, thank you for clarifying.

PR #3753, opened 2016-06-16
sailfish 0.10.1

rob-p:
### Have you:

### New formula: have you

### DIFF
--- a/sailfish.rb
+++ b/sailfish.rb
@@ -3,8 +3,8 @@ class Sailfish < Formula
   homepage "http://www.cs.cmu.edu/~ckingsf/software/sailfish"
   # doi "10.1038/nbt.2862"
   # tag "bioinformatics"
-  url "https://github.com/kingsfordgroup/sailfish/archive/v0.9.2.tar.gz"
-  sha256 "4872c5241864ae12a3aa5e519cf6d2d38d0e510bfcfd420c8d05249f7ee83f82"
+  url "https://github.com/kingsfordgroup/sailfish/archive/v0.10.1.tar.gz"
+  sha256 "a0d6d944382f2e07ffbfd0371132588e2f22bb846ecfc3d3435ff3d81b30d6c6"
 
   bottle do
     sha256 "cc870770c98a55fafb75ceb80ec4eec18be196c165060eabf1366a87359b8fa9" => :el_capitan


rob-p:
I'm about to push a change that can adversely effect source builds of the current package — an update to a dependency that is pulled and built from source.  This pull request will guarantee stability, and that the homebrew package will not be affected by the upstream changes.

sjackman:
There can be no spaces in the checked box `[x]`. I usually submit the PR without typing `x`s (with empty boxes) and then use the GitHub GUI to check the boxes.

rob-p:
dohh! Thanks, Shaun.  I'm new to these fancy boxes etc. ;P.

sjackman:
No worries. Thanks for your contribution!

sjackman:
Did you notice that Linuxbrew has precompiled binary bottles now? I noticed that your package is not `cellar :any`, which means that the bottle can only be used on systems that install Linuxbrew in `/home/linuxbrew/.linuxbrew` and on Macs in `/usr/local`. After it builds binaries, it looks for the string `/home/linuxbrew` in the binary executables, and if it finds it, it marks the package as non-relocatable, meaning it thinks that it will break if it's installed in, for example, `/home/sjackman` instead of `/home/linuxbrew`. Do you know wether `sailfish` is relocatable?

rob-p:
Ahh, I didn't know this.  Sailfish should work so long as it's libraries are in `../lib` relative to the binary location.  I know this is the case on Linux, but I aways run into library problems on OSX.  What's the easiest way to test this?

sjackman:
```sh
brew install --build-bottle sailfish
brew bottle -v sailfish
```
and check the report for where it's finding the hardcoded paths in the executables, then modify the code to remove hard-coded paths and use relative paths instead.

rob-p:
I'll give this a go tonight and see what I come up with.  If it's not relocatable, I'll try to address this in the next update.

rob-p:
ok it's not relocatable b/c I get this output:

```
arning: String '/usr/local/Cellar' still exists in these files:
/usr/local/Cellar/sailfish/0.10.1/bin/sailfish
 --> match '/usr/local/Cellar/sailfish/0.10.1/lib' at offset 0xa34

./sailfish-0.10.1.el_capitan.bottle.1.tar.gz
  bottle do
    revision 1
    sha256 "836bbc795005365ec61579debd7c2b26aa7396b3184ad8e77cb553e7626919e0" => :el_capitan
  end
```

So, the thing I have to figure out now is how to tell CMake, to tell `install_name_tool` (I guess) to do the right thing . . . sigh.

sjackman:
Homebrew automatically changes references of `/usr/local/Cellar` in `install_name` to `@@HOMEBREW_CELLAR@@` (a dummy variable) using `install_name_tool`. See [`relocate_install_names `](https://github.com/Linuxbrew/brew/blob/master/Library/Homebrew/keg_relocate.rb#L31). I don't believe that it changes the `RPATH`. Perhaps that's where the reference is.

sjackman:
Yep, there is an RPATH entry:
```
❯❯❯ otool -l sailfish |grep -A2 RPATH
          cmd LC_RPATH
      cmdsize 56
         path /usr/local/Cellar/sailfish/0.10.1/lib (offset 12)
```
This issue could be improved in Homebrew.

sjackman:
I opened an issue in Homebrew/brew to discuss https://github.com/Homebrew/brew/issues/371

sjackman:
It'll probably get fixed in Homebrew eventually, but it turns out there's a few tricky details that don't apply to `sailfish` involving `dlopen`. To get it fixed quickly, your original plan of fixing the `cmake` script to delete the `RPATH` on Mac is best.

rob-p:
Ok.  I must admit, I'm still somewhat confused about how dynamic library search works on OSX.  If I delete the RPATH, how do I ensure that the executable knows where to look for the libraries relative to the binary?

sjackman:
Mac uses absolute paths to the executables dependent libraries. `RPATH` is only used if explicitly requested. `sailfish` doesn't use `RPATH` to locate its libraries, but uses absolute paths.
e.g.
`install_name = /usr/lib/libncurses.5.4.dylib`
is equivalent to
`install_name = @rpath/libncurses.5.4.dylib`, `RPATH=/usr/lib`

```
❯❯❯ otool -L /usr/local/bin/sailfish
/usr/local/bin/sailfish:
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	/usr/local/opt/tbb/lib/libtbb.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/opt/tbb/lib/libtbbmalloc.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
```

Issue #3401, opened 2016-03-07
vtk crashes when brewed python is installed


lantiga:
### Indicate troubleshooting steps taken
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Checked that a similar issue hasn't already been filed?
- [x] Looked up the software bug tracker if reporting a build error?
- [n/a] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?
### Bug reports / build errors:

After PR #3313 (see also discussion there), which leads vtk to link to system python when it is available, makes it impossible to have vtk coexist with brewed python without segfaulting. This is true both even if one forces building from source.

Scenario 1 (everything works with system python and no brewed python):
- start with no brew python and no brew vtk
- `which python` => `/usr/bin/python`
- `brew install vtk --with-python`
- `mkdir -p /Users/lantiga/Library/Python/2.7/lib/python/site-packages`
  `echo 'import site; site.addsitedir("/usr/local/lib/python2.7/site-packages")' >> /Users/lantiga/Library/Python/2.7/lib/python/site-packages/homebrew.pth`
- `python -c "import vtk"`
  => Ok

Scenario 2 (brew gets the bottled version, which links against system python, even when brewed python is installed, leading to a segfault):
- start with no brew python and no vtk
- `brew install python`
- `which python` => `/usr/local/bin/python`
- `brew install vtk --with-python`
- `python -c "import vtk"` => `Fatal Python error: PyThreadState_Get: no current thread. Abort trap: 6`

Scenario 3 (building vtk from source doesn't solve the problem, as cmake picks up system python even when brewed python is available and comes first in PATH):
- start with no brew python and no vtk
- `brew install python`
- `which python` => `/usr/local/bin/python`
- `brew install vtk --with-python --build-from-source`
- `python -c "import vtk"` => `Fatal Python error: PyThreadState_Get: no current thread. Abort trap: 6`

As an aside: - `/usr/bin/python -c "import vtk"` => Ok

To confirm, in Scenario 3:
`otool -L /usr/local/lib/python2.7/site-packages/vtk/vtkCommonCorePython.so`
gives

```
/usr/local/lib/python2.7/site-packages/vtk/vtkCommonCorePython.so:
[...]

PR #3755, opened 2016-06-16
Siconos 4.0.0 (new formula)

radarsat1:
### Have you:
### New formula: have you

The only caveat is that I get one message from `brew audit` regarding the linking of Siconos' Python modules, suggesting to add `-undefined dynamic_lookup`.  However, this requires an upstream change, and can be fixed in the next release of the library.  In the meantime, the tests pass for me.

### DIFF
--- a/siconos.rb
+++ b/siconos.rb
@@ -7,7 +7,7 @@ class Siconos < Formula
   head "https://github.com/siconos/siconos.git"
 
   depends_on "cmake" => :build
-  depends_on :fortran => :build
+  depends_on :fortran
   depends_on "swig" => :build
   depends_on "bullet"
   depends_on "boost"
--- a/siconos.rb
+++ b/siconos.rb
@@ -68,10 +68,7 @@ class Siconos < Formula
       new NewtonImpactFrictionNSL(0.8, 0., 0.0, 3));
       return 0; }
     EOS
-    system "pwd"
-    system "ls","-a"
     system "#{bin}/siconos", (testpath/"test.cpp")
-    system "ls","-a"
     # a mechanicsIO script
     (testpath/"test.py").write <<-EOS.undent
       from siconos.mechanics.contact_detection.tools import Contactor
--- a/siconos.rb
+++ b/siconos.rb
@@ -53,14 +53,14 @@ class Siconos < Formula
     # executable
     system "#{bin}/siconos", "-h"
     # python modules
-    system "python", "-c", "import siconos"
-    system "python", "-c", "import siconos.kernel"
-    system "python", "-c", "import siconos.kernel; siconos.kernel.SiconosVector()"
-    system "python", "-c", "import siconos.mechanics"
-    system "python", "-c", "import siconos.mechanics.joints"
-    system "python", "-c", "import siconos.mechanics.contact_detection.bullet"
-    system "python", "-c", "import siconos.control"
-    system "python", "-c", "import siconos.io"
+    system "python", "-B", "-c", "import siconos"
+    system "python", "-B", "-c", "import siconos.kernel"
+    system "python", "-B", "-c", "import siconos.kernel; siconos.kernel.SiconosVector()"
+    system "python", "-B", "-c", "import siconos.mechanics"
+    system "python", "-B", "-c", "import siconos.mechanics.joints"
+    system "python", "-B", "-c", "import siconos.mechanics.contact_detection.bullet"
+    system "python", "-B", "-c", "import siconos.control"
+    system "python", "-B", "-c", "import siconos.io"
     # a C++ user program
     (testpath/"test.cpp").write <<-EOS.undent
     #include <SiconosKernel.hpp>
@@ -82,6 +82,6 @@ class Siconos < Formula
       with Hdf5(mode='r+') as io:
         io.run(t0=0,T=1,h=1)
     EOS
-    system "siconos", "test.py"
+    system "siconos", "-B", (testpath/"test.py")
   end
 end
--- a/siconos.rb
+++ b/siconos.rb
@@ -6,7 +6,7 @@ class Siconos < Formula
   sha256 "8b8b6b11d225e8d7f8fc2d068aa5e24701edaf364ce334939f8175df7d1868ae"
   head "https://github.com/siconos/siconos.git"
 
-  depends_on "cmake" => :build
+  depends_on "cmake"
   depends_on :fortran
   depends_on "swig" => :build
   depends_on "bullet"
--- a/siconos.rb
+++ b/siconos.rb
@@ -11,7 +11,8 @@ class Siconos < Formula
   depends_on "swig" => :build
   depends_on "bullet"
   depends_on "boost"
-  depends_on "numpy" => :python
+  depends_on "numpy"
+  depends_on "scipy"
   depends_on "hdf5"
   depends_on "vtk"
 
@@ -36,23 +37,40 @@ class Siconos < Formula
     sha256 "6de44d8c482128efc12334641347a9c3e5098d807dd3c69e867fa8f84ec2a3f1"
   end
 
+  # six needed by h5py
+  resource "six" do
+    url "https://files.pythonhosted.org/packages/4e/aa/73683ca0c4237891e33562e3f55bcaab972869959b97b397637519d92035/six-1.4.1.tar.gz"
+    sha256 "f045afd6dffb755cc0411acb7ce9acc4de0e71261d4b5f91de2e68d9aa5f8367"
+  end
+
   def install
+    def pyver
+      Language::Python.major_minor_version "python"
+    end
+    pyinst = libexec/"vendor/lib/python#{pyver}/site-packages"
+
     # Install Python dependencies first
-    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
-    %w[cython h5py lxml].each do |r|
+    ENV.prepend_create_path "PYTHONPATH", pyinst
+    %w[cython six h5py lxml].each do |r|
       resource(r).stage do
         system "python", *Language::Python.setup_install_args(libexec/"vendor")
       end
     end
 
     # Install Siconos
-    ENV.prepend_create_path "PYTHONPATH", libexec+"lib/python2.7/site-packages"
+    ENV.prepend_create_path "PYTHONPATH", pyinst
     system "cmake", ".", "-DIN_SOURCE_BUILD=ON", "-DWITH_BULLET=ON", *std_cmake_args
     system "make", "install"
 
     # Install executables and set Python path
     bin.install Dir[libexec/"bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+
+    # Make Python module dependencies available
+    dest_path = lib/"python#{pyver}/site-packages"
+    dest_path.mkpath
+    ENV.prepend_create_path "PYTHONPATH", pyinst
+    (dest_path/"homebrew-siconos.pth").write "#{pyinst}\n"
   end
 
   test do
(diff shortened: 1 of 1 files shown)


radarsat1:
It appears the test failed but I don't understand why.  It passes on my machine.  The error seems to be:

    siconos
    test.cpp
    /usr/local/Cellar/siconos/4.0.0/libexec/bin/siconos: /private/tmp/siconos-test-20160616-37052-1l2us48/test build configuration failed, [Errno 2] No such file or directory

then there are a bunch of,

    ==> Sandbox log
    Jun 16 22:24:38 kernel[0]: Sandbox: Python(37061) deny(1) file-write-create /usr/local/lib/python2.7/site-packages/siconos/__init__.pyc
    Jun 16 22:24:38 kernel[0]: Sandbox: Python(37062) deny(1) file-write-create /usr/local/lib/python2.7/site-packages/siconos/__init__.pyc

What does this mean?  I have little experience with Travis.  Is it not finding the `test.cpp` file?  Once Siconos is installed, the `siconos` executable needs to be executed from the same directory as `test.cpp`.  It is created in the formula with,

       # a C++ user program
        (testpath/"test.cpp").write <<-EOS.undent
        #include <SiconosKernel.hpp>
        int main() { SP::NonSmoothLaw nslaw(
          new NewtonImpactFrictionNSL(0.8, 0., 0.0, 3));
          return 0; }
        EOS
        system "siconos", "test.cpp"

Is this incorrect somehow?

siconos.rb
```diff
@@ -0,0 +1,87 @@
...
+  head "https://github.com/siconos/siconos.git"
+
+  depends_on "cmake" => :build
+  depends_on :fortran => :build
+  depends_on "swig" => :build
+  depends_on "bullet"
+  depends_on "boost"
+  depends_on "homebrew/python/numpy"
+  depends_on "homebrew/science/hdf5"
+  depends_on "homebrew/science/vtk"
```
dpo: There's no need to fully qualify the `homebrew/science` formulas.

dpo:
From the BrewBot logs, it seems `cython` should also be installed as a resource on Mavericks. I'm not sure why it's not missing on the other platforms.

The `test.cpp` test also passes on my machine, but it seems the `siconos` command creates a temporary directory where the test is "installed". Perhaps that directory is removed too early? I'm not exactly sure what happens when you run the `siconos` command.

The Python errors occur because Python tries to create `.pyc` files, which Homebrew doesn't allow on the brew bot. I'm not seeing those errors locally. They can be ignored.

siconos.rb
radarsat1: So just `depends_on "vtk"` for example?

radarsat1:
Indeed, it creates a `.siconos` directory in which the source of the simulation is built by CMake and then executed.  It shouldn't be writing *outside* this folder however, and it's within the temporary test directory, so I'm surprised this causes any problem.

On the other hand, this is only for the C++ interface.  We actually use mostly the Python interface these days, so I could remove this test if it is really problematic.  Although I'd like to understand it better.

Note that the `siconos` script simply executes the Python scripts, normally we can equivalently just write `python test.py`, as long as `import siconos` et al. works that should be fine.  However, it seems that with homebrew, since `hdf5` and `lxml` are installed local to the Siconos package, I *must* use the `siconos` to indirectly execute `test.py`, so that it has a chance to fudge `PYTHONPATH` before running the script.  This is okay, but a bit weird.  Is there a better way to installed the Python package dependencies?  I am following the Homebrew and Python guide, and some examples, in which I found how to use `resource` etc.  If this is the right way then it's no problem, we just have to instruct users to use the `siconos` script as a front-end.

siconos.rb
```diff
@@ -0,0 +1,90 @@
+# coding: utf-8
+class Siconos < Formula
+  desc "Modeling and simulation of nonsmooth dynamical systems"
+  homepage "http://siconos.gforge.inria.fr"
+  url "https://github.com/siconos/siconos/archive/4.0.0.tar.gz"
+  sha256 "8b8b6b11d225e8d7f8fc2d068aa5e24701edaf364ce334939f8175df7d1868ae"
+  head "https://github.com/siconos/siconos.git"
+
+  depends_on "cmake" => :build
+  depends_on :fortran => :build
```
dpo: This shouldn't be `:build`. It's needed in the tests.

radarsat1:
The test is now passing, but Travis fails due to a GitHub API rate limit.

tdsmith:
If this is a Python library which is intended to be imported, the application pattern you're using will not work, since it intentionally hides the installed packages from sys.path. This project may find a happier home in homebrew-python.

radarsat1:
The two supported interfaces are as a C++ library and as a Python module.  However, we provide the "siconos" executable as a front-end interface, which seems to work fine with this package.  I'll take a look at how formulas are organized in homebrew-python and see if it's a better fit, but it's really a scientific package, and we hope that making it available both as an executable and as a library too will give it more usage.  But you're right that ideally people would be able to write Python applications that "import siconos" without issues, so I was surprised to see that this is hard to do with Homebrew.  (h5py and lxml are pretty standard modules after all).

tdsmith:
Well, let me take a step back: if these are bindings to a C/++ library that are hard to install with pip, this is fine. In that event, the Python bindings should be installed to `prefix`. The Python module dependencies (lxml, h5py) should be installed to `libexec/vendor` and you should drop a .pth file into `lib/python2.7/site-packages` that contains the line `#{libexec}/vendor/lib/python2.7/site-packages`.

The goal is that the package should still be installable without conflicts even if lxml and h5py have been installed from another source, and that users should not have to take any action to preinstall those modules before installing this formula.

We're not exactly a Python distribution :) An overriding goal is to avoid repackaging individual modules from pypi in order to use them as dependencies.

radarsat1:
Yes, that's just about what I was trying to do, however I understood that homebrew packages should not trigger "pip" so I went with the local modules installation.  Actually as it is, users can just "brew install siconos" and they get something working fine with the "siconos" executable.  It is only if they want to use it as a library that they need to install h5py and lxml themselves, otherwise it seems to happily use the modules that are local to the formula.  However, I'll give it a shot with that .pth file and see if it makes things easier/better somehow.  I'm afraid of breaking users installations if they already have those packages installed, do pip-installed packages take precedence over the .pth file? I'm not very familiar with .pth files.

Absolutely, I don't think of homebrew as managing Python packages, just want to make it easy to install and use Siconos from the intended interface(s).  So far I am happy with it, except Travis seems to have some problem accessing GitHub during the test.

tdsmith:
Awesome 👍 

>  do pip-installed packages take precedence over the .pth file?

In general, yes, because the site-packages directory is added to sys.path before the directory is scanned for .pth files. Paths discovered in .pth files are subsequently appended to sys.path. .pth files are documented (v succinctly) at https://docs.python.org/2/library/site.html.

siconos.rb
```diff
@@ -0,0 +1,87 @@
+# coding: utf-8
+class Siconos < Formula
+  desc "Modeling and simulation of nonsmooth dynamical systems"
+  homepage "http://siconos.gforge.inria.fr"
+  url "https://github.com/siconos/siconos/archive/4.0.0.tar.gz"
+  sha256 "8b8b6b11d225e8d7f8fc2d068aa5e24701edaf364ce334939f8175df7d1868ae"
+  head "https://github.com/siconos/siconos.git"
+
+  depends_on "cmake"
```
dpo: Should be `depends_on "cmake" => :run` if the `siconos` command calls `cmake`.

dpo:
The 10.9 build fails because it's missing Cython. You could simple add it as a resource for 10.9.

radarsat1:
I'm not sure.. Cython is not needed by Siconos, it uses swig.  However, it has a dependency on numpy, perhaps that is where Cython needs to be added?  Anyways I can add it, I see an example in the cantera.rb formula for example, but I don't think it's a direct dependence.  Is there a way to see a 10.9 build log?

radarsat1:
Hm, unless Cython is used by the h5py module... I'll check

radarsat1:
New formula works for me from a scratch install of homebrew on El Capitan.  However, it seems it is affected by #3401 when using brewed python.

siconos.rb
```diff
@@ -0,0 +1,111 @@
...
+  head "https://github.com/siconos/siconos.git"
+
+  depends_on "cmake" => :run
+  depends_on :fortran
+  depends_on "swig" => :build
+  depends_on "bullet"
+  depends_on "boost"
+  depends_on "numpy"
+  depends_on "scipy"
+  depends_on "hdf5"
```
dpo: `numpy` and `scipy` should be `=> :python`.

radarsat1:
I thought so too, but if I add `=> :python`, I get the following error:

```
$ emacs siconos.rb
MacBook-Pro-de-Melaine:homebrew-science sinclairs$ brew install ./siconos.rb
siconos: Unsatisfied dependency: numpy
Homebrew does not provide special Python dependencies; install with:
  `pip install numpy`
siconos: Unsatisfied dependency: scipy
Homebrew does not provide special Python dependencies; install with:
  `pip install scipy`
Error: Unsatisfied requirements failed this build.
```

Leaving them as-is seems to work fine as long as homebrew-python is tapped. So which is correct?

radarsat1:
Seems okay now, but how do I get past this?

```
26.00s==> brew audit homebrew/science/siconos --strict --online

==> Installing or updating 'rubocop' gem

Successfully installed rainbow-2.1.0

Successfully installed ast-2.3.0

Successfully installed parser-2.3.1.2

Successfully installed powerpack-0.1.1

Successfully installed ruby-progressbar-1.8.1

Successfully installed unicode-display_width-1.1.0

Successfully installed rubocop-0.40.0

7 gems installed

Error: GitHub API Error: API rate limit exceeded for 208.78.110.199. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

Try again in 46 minutes 52 seconds, or create a personal access token:

  https://github.com/settings/tokens/new?scopes=&description=Homebrew

and then set the token as: export HOMEBREW_GITHUB_API_TOKEN="your_new_token"

==> FAILED
```

Do I need to add this token to *my* github account?  I thought maybe it was a temporary problem with the Hombrew account, but it persists, and I can see that some other audits are passing.  Thanks in advance.

dpo:
@BrewTestBot test this please

iMichka:
@radarsat1 Seems work has stalled here. Would you like to continue to work on this or should we abandon here?

radarsat1:
@iMichka I am still interested in getting this accepted, but I am afraid I simply don't know how to continue since I never got an answer regarding why Travis failed.

That said, we are actually releasing Siconos 4.1.0 next week some time, so I will be updating this package and triggering a new, hopefully successful attempt.

In the meantime I have been able to use this script in my own fork to successfully install Siconos from scratch.

iMichka:
Some things have changed so I hope the travis build will work better now. I'm fine if we wait until the next version. I can help you then to fix the different problems.

radarsat1:
Absolutely, thanks for the answer.

iMichka:

radarsat1:
Hm, that's a shame.  Is there a discussion about this decision somewhere?  I searched but haven't found much.

iMichka:
We had a discussion between maintainers. It is sort of due to the high effort it needs to keep things up and running here. I think I will maybe open an issue to discuss this further.

Not everything has been defined for the moment but the idea is that homebrew-science will loose buildbot support for OS X around end of December, so it will sort of disappear, at least under it's current form.

Here is the change done for the "new formula" policy: #6348

We have also a list of formulae that we will migrate from homebrew-science to homebrew-core: #6331 
All these will be maintained.

The question is still open to what happens with all the other formulae. I'll open an issue for that.

iMichka:
You will be able to follow the discussion here: #6365

radarsat1:
Okay, thanks very much for the info!

PR #3763, opened 2016-06-18
nextflow: formula tweaks

ilovezfs:
- use `pipe_ouput` in the test
- set bottle :unneeded


ilovezfs:
As far as I can tell, the old test's last line did nothing except echo :)

PR #3767, opened 2016-06-20
Formula for Triqs 1.3.2

parcollet:
TRIQS (Toolbox for Research on Interacting Quantum Systems) is a scientific project providing a set of C++ and Python libraries to develop new tools for the study of interacting quantum systems.

The goal of this toolkit is to provide high level, efficient and simple to use libraries in C++ and Python, and to promote the use of modern programming techniques.

TRIQS is free software distributed under the GPL license.


triqs.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  depends_on "cmake" => :build
+  depends_on :mpi => :cxx
+
+  depends_on "boost"
+  depends_on "hdf5"
+  depends_on "cmake" => :build
+  depends_on "fftw"
+  depends_on "gmp"
+  depends_on "python"
```
dpo: `depends_on :python if MacOS.version <= :snow_leopard`

triqs.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def install
+    ENV.cxx11
+    args = std_cmake_args
+    args.delete "-DCMAKE_BUILD_TYPE=None"
+    args << "-DCMAKE_BUILD_TYPE=Release"
+
+    args << ("-DBuild_Tests=" + ((build.with? "test") ? "ON" : "OFF"))
+
+    args << ["-DPYTHON_INTERPRETER=/usr/local/bin/python", "-DALLOW_COMPILATION_IN_SOURCES=ON", "-DCMAKE_INSTALL_PREFIX=/usr/local"]
```
dpo: The Python interpreter should just be `python` and should default to system Python, not Homebrew Python.

Does this install anything else than Python bindings?

triqs.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    mkdir "tmp" do
+      args << ".."
+      system "pip", "install", "--upgrade", "h5py"
+      system "pip", "install", "--upgrade", "mako"
+      system "pip", "install", "--upgrade", "mpi4py"
```
dpo: I don't think this should be part of a formula. It's up to the user to upgrade.

triqs.rb
```diff
@@ -0,0 +1,56 @@
...
+      system "cmake", *args
+      system "make", "-j8"
```
dpo: `-j8` is not necessary. The build is parallelized by default.

triqs.rb
parcollet: Yes, it installs:
- a C++ compiled library
- C++ headers
- python modules
- (later) documentation files

It is a library in C++, with its Python interface

parcollet:
It fails with a strange error : 

Any idea ?

-----------------------------------------------------

```
Error: GitHub API Error: API rate limit exceeded for 208.78.110.199. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

Try again in 59 minutes 59 seconds, or create a personal access token:

  https://github.com/settings/tokens/new?scopes=&description=Homebrew

and then set the token as: export HOMEBREW_GITHUB_API_TOKEN="your_new_token"

travis_time:end:3488939f,start=1466472913000000000,finish=1466472957000000000,duration=44000000000

travis_fold:end:brew.audit.6

[1;39m==>[4;31m FAILED[0m



travis_time:end:02a2797e:start=1466472845000000000,finish=1466472957000000000,duration=112000000000
[0K

[31;1mThe command "if [ "${TRAVIS_OS_NAME}" = "osx" ]; then brew test-bot --tap=homebrew/science; else docker run -w /home/linuxbrew/.linuxbrew/Library/Taps/homebrew/homebrew-science --env-file /tmp/travis.env -t linuxbrew brew test-bot --tap=homebrew/science; fi" exited with 1.[0m



Done. Your build exited with 1.
```

parcollet:
Same error this morning...

parcollet:
@dpo : 
Any idea about this mysterious GitHub API error ?

dpo:
It's not an error due to your formula but rather a Travis glitch. I'll restart the build and we'll see.

parcollet:
Same error. It did the computation again, but the problem is still here.

dpo:
I installed `h5py` with `pip` but then CMake complains about HDF5 versions:
```
-- hdf5 version: 1.8.16
CMake Error at CMakeLists.txt:311 (message):
  The HDF5 version of the h5py (1.8.13) is different from the hdf5 library
  found for C++ (1.8.16).  You MUST have the same version.
```

parcollet:
Was it the origin of the issue ?

yes, it is the usual issue with hdf5 & brew.
brew installs one version of hdf5, 
pip installs h5py precompiled with another version, 
we forbid this, it can easily lead to erratic crashes.
It is a defect in brew not to install a coherent hdf5 & h5py.

One solution is to force pip to recompile from source, then it is fine
pip install h5py --no-binary :all: 
How do I do this in the Formula ?

The real solution would be to make a brew formula for h5py, 
make it depends on hdf5 and ensure that versions of the hdf5 lib are the same in both.
That would solve it for other projects too.

tdsmith:
Take a look at https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md#bindings; the `=> :python` dependencies should always be avoided in part because they lead to problems like this.

Because the recommended configuration for libraries does not override packages already in site-packages, you could consider depending on a `HdfVersionMatchesRequirement` that ensures that, if h5py already exists, its version matches the installed version of hdf5. You can grep for `Requirement` to see some examples of this construct.

parcollet:
Thanks.
Ok, I can add such a requirement, but how will this solve the current issue with the formula ?
The bottle will just fail to compile, no ?

dpo:
@parcollet An alternative is to install `mpi4py` and `h5py` as resources, e.g.:
```ruby
resource "mpi4py" do
  url "https://bitbucket.org/mpi4py/mpi4py/downloads/mpi4py-2.0.0.tar.gz"
  sha256 "6543a05851a7aa1e6d165e673d422ba24e45c41e4221f0993fe1e5924a00cb81"
end

def install
  # ...
  ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
    resource("mpi4py").stage do
      system "python", *Language::Python.setup_install_args(libexec/"vendor")
    end
  end
  # ...
```

dpo:
Just checking if this will be updated.

parcollet:
Yes it will, we really need the brew bottle.
What will the last suggestion do exactly ? 
Force h5py installation from source, i.e. recompilation ?

dpo:
Yes they will be installed from source.

parcollet:
ok, thanks.
1- I have discovered the python tap, it does most python dependencies.
2- the formula works on my machine, audit is ok, but when I push, 
I still get the same GITHUB error as before.

I have no clue how to fix this.


==> brew audit homebrew/science/triqs --new-formula
==> Installing or updating 'rubocop' gem
Successfully installed rainbow-2.1.0
...
Successfully installed rubocop-0.43.0
7 gems installed
Error: GitHub API Error: API rate limit exceeded for 208.78.110.199. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Try again in 2 minutes 55 seconds, or create a personal access token:
  https://github.com/settings/tokens/new?scopes=&description=Homebrew
and then set the token as: export HOMEBREW_GITHUB_API_TOKEN="your_new_token"

parcollet:
@dpo 
Still fails because of the installation of h5py, which does not compiles properly.
I do not understand why brew recompiles it, and does not simply use the bottle.

iMichka:
@BrewTestBot test this please

jonchang:
The homebrew/python tap is deprecated so these lines will need to be changed:

```ruby
  depends_on "homebrew/python/numpy"
  depends_on "homebrew/python/scipy"
  depends_on "homebrew/python/mpi4py"
  depends_on "homebrew/python/h5py"
```

The `numpy` and `scipy` dependencies should just be `depends_on "numpy"` and similar. `mpi4py` and `h5py` will need to be vendored as resources.

jonchang:
@parcollet if you're still interested in getting this merged, feel free to ask for help with any of these things.

iMichka:
Obviously nobody is currently working on this. I'm closing this to to try to reduce the huge amount of pull request that we currently have. Nothing is lost, if somebody wants to pick this up and work on it, a new pull request can be submitted.

PR #3781, opened 2016-06-25
arpack 3.3.0

ilovezfs:
- always run the tests
- style changes

### DIFF
--- a/arpack.rb
+++ b/arpack.rb
@@ -1,19 +1,17 @@
 class Arpack < Formula
   desc "Routines to solve large scale eigenvalue problems"
   homepage "https://github.com/opencollab/arpack-ng"
-  url "https://github.com/opencollab/arpack-ng/archive/3.2.0.tar.gz"
-  sha256 "ce6de85d8de6ae3a741fb9d6169c194ff1b2ffdab289f7af8e41d71bb7818cbb"
-  revision 1
+  url "https://github.com/opencollab/arpack-ng/archive/3.3.0.tar.gz"
+  sha256 "ad59811e7d79d50b8ba19fd908f92a3683d883597b2c7759fdcc38f6311fe5b3"
   head "https://github.com/opencollab/arpack-ng.git"
 
   bottle do
-    sha256 "40acf9a57730a926f68aa2cc54760adc1844d0431b212f1678c8d73afe7b8e77" => :el_capitan
-    sha256 "08a74247fa5caedfb62607debe6202ee66bf06cc7933b23825b1c641c0955675" => :yosemite
-    sha256 "7c44096b2ab077cf76dc4a4908b07819489fa5164e47dd3b91b5762b288cc779" => :mavericks
+    sha256 "f0d5c5dde75b198b126a31a26b784e6622560ee5cbae7d3bf722ebff25c26f4c" => :el_capitan
   end
 
-  # TODO: enable with 3.3.0
-  # option "without-check", "skip tests (not recommended)"
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
 
   depends_on :fortran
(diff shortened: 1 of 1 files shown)


dpo:
@BrewTestBot test this please

dpo:
Seems ready to go.

ilovezfs:
@dpo, yes although I suspect @sjackman may want the style of https://github.com/Homebrew/homebrew-science/pull/3781/files#diff-694b5d3d75fbe231fa066e8800056934R63, and following, tweaked.

sjackman:
It's fine by me.

ilovezfs:
Oh you're just saying that! Don't be shy!

arpack.rb
```diff
@@ -23,37 +21,56 @@ class Arpack < Formula
...
+      %w[
+        pcndrv1
+        pdndrv1
+        pdndrv3
+        pdsdrv1
+        psndrv1
+        psndrv3
+        pssdrv1
+        pzndrv1
+      ].each { |slv| system "mpirun", "-np", "4", slv }
```
sjackman: I'd be inclined to put the brackets on a single line. Other than that, it looks good to me.
```ruby
%w[pcndrv1 pdndrv1 pdndrv3 pdsdrv1 psndrv1 psndrv3 pssdrv1 pzndrv1].each do |slv|
  system "mpirun", "-np", "4", slv
end
```

ilovezfs: Yeah, I agree, but is there a reasonable way to reconcile that with the rule about not exceeding the 80th column?

sjackman: I think you can break the bracket line in two with a double indentation for the second line. Also 80 is a rule of thumb, but I think it's flexible.

ilovezfs: by double indentation do you mean two spaces or four

sjackman: As above: 4 spaces. (I updated my comment.)

ilovezfs: That makes sense in terms of trying to keep it visually separated from what's below.

But it looks like this occurs in two cases in homebrew/core, each with a three space indentation:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/emscripten.rb#L76-L77
https://github.com/Homebrew/homebrew-core/blob/master/Formula/android-sdk.rb#L39-L41

It turns out this is actually enforced by `brew audit --strict`, (i.e., by `brew style`):
```
== arpack.rb ==
C: 63: 12: Align the elements of an array literal if they span more than one line.
```

sjackman: Oh, right. I'm confusing my C style with my Ruby style. 3 spaces is fine, as is the following. Use whichever you prefer.
```ruby
%w[
    pcndrv1 pdndrv1 pdndrv3 pdsdrv1
    psndrv1 psndrv3 pssdrv1 pzndrv1].each do |slv|
  system "mpirun", "-np", "4", slv
end
```

dpo:
Version 3.4.0 is out now.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@dpo 3.4.0 looks good to go as well. Full test-bot run: https://gist.github.com/ilovezfs/4c43bccba89a327604ac11231d6682d6

I'd like to ship 3.3.0 for completeness first though.

dpo:
Thanks. I'm all for completeness but quite a bit needs to be rebottled on arpack updates:
```
homebrew/science/armadillo        homebrew/science/getdp            homebrew/science/slepc
homebrew/science/calculix-ccx     homebrew/science/octave
homebrew/science/dealii           homebrew/science/shogun
```

ilovezfs:
Hold off on the rebottling of those then. We'll ship this and then include revision bumps for those in the next one.

dpo:
Sounds like a plan.

ilovezfs:
@dpo shipped!

PR #3782, opened 2016-06-26
samtools 1.3.1

ilovezfs:


ilovezfs:
@sjackman please code review when you have a chance. By the way, the test was somewhat hilariously broken since it would pass if the command was not installed due to matching the program name in "command not found ..."

ilovezfs:
@sjackman any idea why Jenkins is whining about numpy?

samtools.rb
```diff
@@ -17,34 +16,36 @@ class Samtools < Formula
...
+      ace2sam
+      maq2sam-long
+      maq2sam-short
+      md5fa
+      md5sum-lite
+      wgsim
+      *.lua
+      *.pl
+      *.py
+    ].each { |f| bin.install Dir["misc/#{f}"] }
```
sjackman: Stylistically, I'd prefer
```ruby
cd "misc" bin.install %w[…]
```

samtools.rb
```diff
@@ -17,34 +16,36 @@ class Samtools < Formula
...
-    man1.install %w[samtools.1]
-    (share+"samtools").install %w[examples]
-    (include+"bam").install Dir["*.h"]
+    (include/"bam").install Dir["*.h"]
+    man1.install "samtools.1"
+    pkgshare.install "examples"
   end
 
   test do
-    system "samtools 2>&1 |grep -q samtools"
```
sjackman: Thanks for catching this bug. It has made itself into a number of bioinformatics packages and have been slowly caught and purged.

samtools.rb
```diff
@@ -17,34 +16,36 @@ class Samtools < Formula
     sha256 "d0e3e6f7ffba8ed4005ff14155727292b939a645c8c2376191aae0f24cf20eb3" => :x86_64_linux
   end
 
-  option "with-dwgsim", "Build with Whole Genome Simulation"
-  option "without-curses", "Skip use of libcurses, for platforms without it, or different curses naming"
-
-  depends_on "homebrew/dupes/ncurses" unless OS.mac?
   depends_on "htslib"
-  depends_on "dwgsim" => :optional
+  depends_on "homebrew/dupes/ncurses" => :recommended unless OS.mac?
```
sjackman: If you like, I'm okay making this dependency required. You've already done the work though, so either way I'm fine.

sjackman:
Two small comments. Looks good to me. Thanks, Joe!

samtools.rb
ilovezfs: Do you mean this?
```
    cd "misc" do
      bin.install %w[ace2sam maq2sam-long maq2sam-short md5fa md5sum-lite wgsim]
      bin.install Dir["{*.lua,*.pl,*.py}"]
    end
```

ilovezfs:
@sjackman PR refreshed.

ilovezfs:
woo hoo :green_apple:

samtools.rb
ilovezfs: And the solution reveals itself:

```
bin.install Dir["{samtools,misc/*}"].select { |f| File.executable?(f) }
```

PR #3784, opened 2016-06-26
dealii: revision for petsc

ilovezfs:
- PETSc library version number has changed so we need a revision bump
- PETSc 3.7.x changed the function signature of PetscOptionsSetValue()

### DIFF
--- a/dealii.rb
+++ b/dealii.rb
@@ -3,7 +3,7 @@ class Dealii < Formula
   homepage "http://www.dealii.org"
   url "https://github.com/dealii/dealii/releases/download/v8.3.0/dealii-8.3.0.tar.gz"
   sha256 "4ddf72632eb501e1c814e299f32fc04fd680d6fda9daff58be4209e400e41779"
-  revision 2
+  revision 3
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


dpo:
See #3730.

ilovezfs:
I saw it.

dpo:
It won't build because the problem comes from PETSc.

ilovezfs:
Right.

ilovezfs:
@dpo it looks like the entire patch actually amounts to this: https://gist.githubusercontent.com/ilovezfs/13192d7577f5abc89c1be3a3b3eb11f5/raw/8b79fa1172ce99aea6cd2728f8a9c0690d868c95/gistfile1.txt

So I think we don't need to wait for this upstream which is moving at a glacial rate and recommending to its users not to use brew since we don't have versions support for the old petsc. Did I mention I'm annoyed (with them not you)?

ilovezfs:
@dpo PR refreshed. Should be fixed now. Only needed this:
```
inreplace "source/lac/petsc_precondition.cc",
  "PetscOptionsSetValue(\"",
  "PetscOptionsSetValue(NULL,\""
```

dpo:
Great if it works. Maybe upgrade to the newer version then?

ilovezfs:
@dpo Let's ship fixed 8.3 bottles, and then you can rebase your PR :)

dealii.rb
```diff
@@ -94,7 +102,7 @@ def install
       log_name = "make-test.log"
       system "make test 2>&1 | tee #{log_name}"
       ohai `grep "tests passed" "#{log_name}"`.chomp
-      prefix.install "#{log_name}"
+      prefix.install log_name
```
dpo: In my other PR, I removed this log file stuff altogether. The log will be in `~/Library/Homebrew/logs` anyways.

ilovezfs: Yeah this is just to shut audit up. I'm sure whatever you have is better.

dpo:
Sounds good.

ilovezfs:
@dpo shipped.

PR #3785, opened 2016-06-26
Update root.rb to fix a python bug

cjohnson-phys:
### Have you:

- [ X] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [ X] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [ X] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [ X] Built your formula locally prior to submission with `brew install <formula>`?

### New formula: have you

- [ X] Not altered the `bottle` section?
- [ X] Checked that the tests still pass?

There was an issue when using pyROOT when root tries to link to python that is not Homebrew python (I think...)
A related issue is discussed a bit here: https://root.cern.ch/phpBB3/viewtopic.php?t=17714

### DIFF
--- a/root.rb
+++ b/root.rb
@@ -20,7 +20,7 @@ class Root < Formula
   depends_on "fftw" => :optional
   depends_on "qt" => [:optional, "with-qt3support"]
   depends_on :x11 => :optional
-  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on :python
 
   def install
     # brew audit doesn't like non-executables in bin


root.rb
```diff
@@ -20,7 +20,7 @@ class Root < Formula
   depends_on "fftw" => :optional
   depends_on "qt" => [:optional, "with-qt3support"]
   depends_on :x11 => :optional
-  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on :python
```
dpo: We don't want this. Root should build correctly against system Python.

cjohnson-phys: Then pyROOT stays broken if you use homebrew python. Why is this requirement needed now?

dpo: It's not new: https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md#python-declarations

`root` might have to be rebuilt from source if you want to use Homebrew Python. In that case, first `brew install python` and then `brew install -s root`. If Homebrew Python appears first on your PATH (it should), the bindings should be built against it.

cjohnson-phys: Thanks for the information.

tdsmith:
This may be a (py)root bug; importable modules should not link against an interpreter. http://blog.tim-smith.us/2015/09/python-extension-modules-os-x/

PR #3787, opened 2016-06-27
dgtal 0.9.2

ilovezfs:
- add an option to run the test suite, and always do so during CI
- install the examples
- add a test

### DIFF
--- a/dgtal.rb
+++ b/dgtal.rb
@@ -1,10 +1,10 @@
 class Dgtal < Formula
-  desc "Digital Geometry Tools and Algorithms Library (DGtal) is a toolkit to perform topology and geometry processing on (grid-based) digital data."
+  desc "Digital Geometry Tools and Algorithms"
   homepage "http://dgtal.org"
-  url "http://liris.cnrs.fr/dgtal/releases/DGtal-0.9.1-Source.tar.gz"
-  sha256 "9969eee23f0542d94f1a3f39b55665010baada65496223d671998f2c3eefb007"
+  url "http://dgtal.org/releases/DGtal-0.9.2-Source.tar.gz"
+  mirror "http://liris.cnrs.fr/dgtal/releases/DGtal-0.9.2-Source.tar.gz"
+  sha256 "ba044b7c353a8550dc740c1d80e3caf59a76d4332d956599359027f3e8e20ed9"
   head "https://github.com/DGtal-team/DGtal.git"
-  revision 2
 
   bottle do
     cellar :any
@@ -12,6 +12,9 @@ class Dgtal < Formula
     sha256 "e56a577077e8db005ca89e71e7dd2dee7db8aa839f66864a47cdb385dc40de58" => :yosemite
     sha256 "6990e203f40c8061d689b7ef70486df5a4fe21e2ff49441365981e16c4e6aa69" => :mavericks
   end
+
+  option "without-test", "Skip build-time tests"
+  option "without-examples", "Don't build the examples"
 
   depends_on "cmake" => :build
   boost_args = []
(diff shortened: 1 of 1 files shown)


ilovezfs:

dgtal.rb
```diff
@@ -45,7 +47,29 @@ def install
...
+        typedef double Ring;
+        MPolynomial<3,Ring,std::allocator<Ring> > P;
+        MPolynomialReader<3,Ring> reader;
+        string str = "xyz^3-4yz";
+        reader.read( P, str.begin(), str.end() );
+        std::cout << P << std::endl;
+        return 0;
+      }
+    EOS
+    system ENV.cxx, "poly.cpp", "-o", "poly", "-lDGtal"
```
sjackman: Add the option `-L#{lib}`.

sjackman:
One comment for non-`/usr/local` installations.

Do the tests take a very long time, and that's why they're not on by default?

ilovezfs:
@sjackman 
```
iMac-TMP:local joe$ time brew install -s dgtal
==> Installing dgtal from homebrew/science
==> Downloading http://dgtal.org/releases/DGtal-0.9.2-Source.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/dgtal-0.9.2.tar.gz
==> cmake .. -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREF
==> make
==> make install
🍺  /usr/local/Cellar/dgtal/0.9.2: 983 files, 282.5M, built in 15 seconds

real	0m15.567s
user	0m30.893s
sys	0m3.928s
iMac-TMP:local joe$ brew uninstall dgtal
Uninstalling /usr/local/Cellar/dgtal/0.9.2... (983 files, 282.5M)
iMac-TMP:local joe$ time brew install -s --with-test dgtal
==> Installing dgtal from homebrew/science
==> Downloading http://dgtal.org/releases/DGtal-0.9.2-Source.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/dgtal-0.9.2.tar.gz
==> cmake .. -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREF
==> make
==> make test
==> make install
🍺  /usr/local/Cellar/dgtal/0.9.2: 983 files, 282.5M, built in 2 minutes 46 seconds

real	2m47.300s
user	15m57.046s
sys	1m7.838s
```

ilovezfs:
And if we wanted to be especially thorough during CI, there's also the option to build the examples.

sjackman:
If it's only a matter of 3 minutes, I'd be inclined to enable the tests by default.

ilovezfs:
They are on by default for CI. I'm not sure the users need to sit through 3 minutes of testing if it's actually done in 15 seconds, though.

sjackman:
Most users install bottles, and it's `cellar :any`. In Homebrew-Science we typically enable tests by default.

ilovezfs:
Okie dokie. And what about building the examples?

ilovezfs:
The examples add an additional minute.

sjackman:
I have no real opinion. I'd suggest going with whatever the upstream default is.

ilovezfs:
@sjackman PR refreshed.

ilovezfs:
@sjackman Ug, how tedious ... looks like there is one failed test with gcc5 and gcc6, but not with clang and not with gcc4. Does that ring any bells for you in terms of cause?

Specifically, the failing test is tests/base/testClone2.cpp. Fails in testCloneCases() where it returns
```
  return nbok == nb;
```
But printing the values,
```
nbok 0x4
nb 0x25
```

sjackman:
`nbok` is the number of passed tests and `nm` is the number of tests, so 4 of 37 tests passed and 33 of 37 failed. See http://liris.cnrs.fr/dgtal/doc/0.9.2/testClone2_8cpp_source.html You'll have to contact upstream to sort that one out.

dcoeurjo:
Hi everyone. 
I'm one of the main devs of this library. I'll have a look to this issue.

ilovezfs:
@dcoeurjo thanks!

dcoeurjo:
@ilovezfs just a newbie question: why the bot is testing build with gcc? I thought that hombrew  formula only have to build  for clang mac compilers.

(issue reported in the upstream).

ilovezfs:
@dcoeurjo the build bot does indeed use clang with macOS as you thought.

The gcc6 and gcc5 failures were my own observation locally using `gcc` (https://github.com/Homebrew/homebrew-core/blob/master/Formula/gcc.rb) and `homebrew/versions/gcc5` (i.e., https://github.com/Homebrew/homebrew-versions/blob/master/gcc5.rb).

Also, everything in homebrew/science gets built and tested for Linux by TravisCI and that uses gcc as well.

ilovezfs:
@dcoeurjo by the way, don't be misled by that Yosemite failure

```
==> brew uses homebrew/science/dgtal
==> FAILED
Error: /usr/local/Library/Taps/homebrew/homebrew-core/Formula/gcc.rb:34: syntax error, unexpected <<
<<<<<<< Updated upstream
  ^
/usr/local/Library/Taps/homebrew/homebrew-core/Formula/gcc.rb:38: syntax error, unexpected ===, expecting keyword_end
=======
   ^
/usr/local/Library/Taps/homebrew/homebrew-core/Formula/gcc.rb:40: syntax error, unexpected >>, expecting keyword_end
>>>>>>> Stashed changes
```

That's happening everywhere right now and isn't the fault of dgtal.

ilovezfs:
@DomT4 this was just now ... I'm assuming it's broken again?

DomT4:
Good lord. Fixed, again, for now 👀.

ilovezfs:
@DomT4 makes me want to disconnect the Jenkins boxes from The Internets and never upgrade lol

ilovezfs:
@dcoeurjo I've added
```
  # GCC 4 works, and according to upstream issue, GCC <= 5.2.1 may also be fine
  ["5", "6"].each do |n|
    fails_with :gcc => n do
      cause "testClone2 fails: https://github.com/DGtal-team/DGtal/issues/1203"
    end
  end
```
to the formula. Since it builds OK with clang, this can probably :ship:

dcoeurjo:
@ilovezfs good idea. It solves the problem waiting for the fix in the upstream

ilovezfs:
@dcoeurjo Yup, and we know of at least two compilers that will work fine on Linux.

dcoeurjo:
@ilovezfs agree... So far, our travis build bot only checks dgtal build with  gcc4-8 on linux. Maybe we should consider more recent compilers as well

ilovezfs:
@dcoeurjo I'd recommend doing at least gcc-6.2.0 as well for good coverage.

ilovezfs:
@dcoeurjo :boat:'d! By the way, my apologies for unceremoniously closing your PR. I didn't realize you were upstream!

dcoeurjo:
@ilovezfs no problemo ;).  I forgot to double-check that no pending PR already existed...my bad

Issue #3793, opened 2016-06-29
edirect: Error: SHA256 mismatch


cmdcolin:
I have a short error with installing NCBI edirect

```
$ brew install edirect
==> Installing edirect from homebrew/science
==> Downloading https://homebrew.bintray.com/bottles-science/edirect-3.60.mavericks.bottle.tar.gz

curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "edirect"
Download failed: https://homebrew.bintray.com/bottles-science/edirect-3.60.mavericks.bottle.tar.gz
Warning: Bottle installation failed: building from source.
==> Downloading ftp://ftp.ncbi.nlm.nih.gov/entrez/entrezdirect/edirect.tar.gz
######################################################################## 100.0%
Error: SHA256 mismatch
Expected: 2d108d780cef39bc4fdfd265d97ebaafe59ec937667061f54274f17a40a1a74f
Actual: cb394e73b8370c66b22c48f8725ff5516a41daad7c9f6780c9728cbf8eb69030
Archive: /Users/biocmd/Library/Caches/Homebrew/edirect-3.60.tar.gz
To retry an incomplete download, remove the file above.
```

I am not sure why the bottle download failed, but also note the shasum error. I noticed that when downloading the ftp source manually reported a different version from 3.60 ftp://ftp.ncbi.nlm.nih.gov/entrez/entrezdirect/edirect.tar.gz and is now 4.50, so this ftp url might not point to a stable version

It might be worth to point to a specific version in ftp://ftp.ncbi.nlm.nih.gov/entrez/entrezdirect/versions/

ilovezfs:
The bottles for 3.60 were removed from bintray at some point. I'm not sure why, but they're quite gone.

The checksum doesn't match because that URL is generic not version specific, so whenever it gets update by upstream, the formula will currently break. I'll fix this now.

cmdcolin:
Super, thanks!

ilovezfs:
@cmarquardt https://github.com/Homebrew/homebrew-science/pull/3794

You can `brew pull https://github.com/Homebrew/homebrew-science/pull/3794` if you'd like to try it.

cmdcolin:
@ilovezfs I think this has issues. I ran this and then re-ran brew install edirect and it seemed to work with the newer installation (used go, etc) but I got this error running an esearch

```
esearch -db pubmed -query "opsin gene conversion"
[...]

PR #3794, opened 2016-06-29
edirect 2016-06-29

ilovezfs:
- use versioned URLs for the stable spec
- add a head spec using the URL pointing of the "current" version
- fix audit issue about non-binaries in bin

### DIFF
--- a/edirect.rb
+++ b/edirect.rb
@@ -1,10 +1,12 @@
 class Edirect < Formula
+  desc "Advanced method for accessing the NCBI's databases"
   homepage "http://www.ncbi.nlm.nih.gov/books/NBK179288/"
-  # tag "bioinformatics"
+  url "ftp://ftp.ncbi.nlm.nih.gov/entrez/entrezdirect/versions/2016-06-28/edirect.tar.gz"
+  version "2016-06-28"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@sjackman @DomT4 Although this PR specifically addresses #3793, I'm very reluctant to merge this in its current form because the formula seems to violate our rules about installing software after-the-fact, and what's worse, it does so without any sort of checksums for what it's installing, and seems to target random directories having nothing to do with the prefix. As it currently exists this thing belongs in the boneyard IMO unless it gets a major re-rewrite, or unless I'm missing something and the offending setup script is totally superfluous and everything is functional even if that script is not installed.

sjackman:
> and the offending setup script is totally superfluous and everything is functional even if that script is not installed

Yes, the setup script appears to be unnecessary. It works for me without running the setup script.

ilovezfs:
superseded by https://github.com/Homebrew/homebrew-science/pull/4065

sjackman:
Woops. I forgot about this PR.

sjackman:
I would like to use versioned URLs.

ilovezfs:
@sjackman hehe no problem... I should have shipped by now. Hoping for a :+1: from @cmdcolin that it's working for him.

sjackman:
Fixed by f645664841df1c66d6af1ce6cbd02db4b3bf605e.

PR #3796, opened 2016-06-30
Updating MFEM package with v3.2 info.

tzanio:
Hello,

I am not an expert on Homebrew, so I probably did not complete all your PR requirements. I hope this could be mitigated by the fact that the change is quite small -- it just bumps up the version to 3.2 with updated link and sha256.

Cheers,
Tzanio

### DIFF
--- a/mfem.rb
+++ b/mfem.rb
@@ -1,9 +1,9 @@
 class Mfem < Formula
   desc "Free, lightweight, scalable C++ library for FEM."
   homepage "http://www.mfem.org"
-  url "http://goo.gl/xrScXn"
-  version "3.1"
-  sha256 "841ea5cf58de6fae4de0f553b0e01ebaab9cd9c67fa821e8a715666ecf18fc57"
+  url "http://goo.gl/Y9T75B"
+  version "3.2"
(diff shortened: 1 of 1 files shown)


dpo:
Thanks @tzanio. Your changes look good. Something might have changed in MFEM since 3.1 because the test doesn't compile:
```
make all MFEM_DIR=/usr/local/opt/mfem CONFIG_MK=$(MFEM_DIR)/config.mk
makefile:44: /usr/local/opt/mfem/config/test.mk: No such file or directory
make: *** No rule to make target `/usr/local/opt/mfem/config/test.mk'.  Stop.
```
Any idea how it should be built?

tzanio:
Thanks @dpo,

This is strange -- `config/test.mk` is indeed a new file in 3.2, but I haven't seen this issue on any of the platforms we've tested. The intended build sequence is:

1. Configuration, e.g. `make config MFEM_USE_MPI=NO MFEM_DEBUG=YES`
2. Build: `make all` (or just `make` if you don't want the examples and miniapps)
3. Install: `make install`

We now also have the targets `make check` for a quick build check and `make test` for an exhaustive set of tests.

Does this help?
Tzanio

PR #3822, opened 2016-07-10
geant4: 10.02.p02

davidchall:
- updated to 10.02.p02
- renamed geant formula to geant4 (this is the actual name of the package, i.e. major version number is 10).
- added description
- made multi-threading on by default (most users will want this, and it is possible to build single-threaded applications with multi-threaded Geant4)


tdsmith:
Can you add a `test` block that exercises the Geant4 libraries with a simple program?

davidchall:
I've added a test as requested. The recommended way to build Geant4 applications is with CMake, so I have changed this from a `:build` to a `:run` dependency. Otherwise, people could install from a bottle and be unable to run the test.

davidchall:
Due to the comments in #3825, I have updated this formula to use Qt5 instead of Qt4. I checked that this works correctly using example B1 with Qt5 visualisation.

Please let me know if you need me to do anything else before this can be merged.

tdsmith:
lgtm; thanks very much!

PR #3825, opened 2016-07-12
mathgl (2.3.5.1)

### LINKED ISSUE #2820: MathGL fails to build with QT5
When building MathGL with `--with-qt=5`, MathGL fails to install with build error.

Gist log can be found at https://gist.github.com/anonymous/9fd2b27611c778fc7a83

Using Xcode 7 on Yosemite 10.10.5

pthariensflame:
Update MathGL to 2.3.5.1; may be sufficient to close #2820.

---

### Have you:


UniqMartin:
@dpo (I guess) You can't just move around the methods as done in ad5c286fdbfab60cd68f2ee5da23544ba9d008e1 as this completely breaks the formula. It cannot be loaded by Homebrew anymore. Those methods have to be already defined when they are first used, i.e. *before* the first `depends_on` line makes use of them.

UniqMartin:
Fixed the breakage caused by ad5c286fdbfab60cd68f2ee5da23544ba9d008e1 by reverting the offending changes in ef35225542872283bf282460dcf4783c12c750c3.

dpo:
Very sorry about that. I was away and couldn't fix right away. Thanks for fixing!!!

UniqMartin:
No worries! After all it should have only affected users that wanted to install this formula (and developers who would have seen error messages for various operations that read all formulae).

Best mid-term solution would probably to go away from the `--with-qt=` option with argument and adopt the pattern used in core (`--with-qt` and/or `-with-qt5`, e.g. `depends_on "qt" => :optional` and `depends_on "qt5" => :optional`) as this also works much nicer with the dependency resolution logic than a trailing modifier `if`. If that move is made, I'd suggest to drop the `qt` dependency completely, as we're slowly (sadly very slowly) but steadily dropping support for `qt` and it will need to go away sooner than later (not to mention that it doesn't and probably will never work on macOS 10.12).

pthariensflame:
@UniqMartin Out of curiosity, since I hadn't heard about this, why will Qt “probably” never work on Sierra?

UniqMartin:
@pthariensflame I'm not sure if I created any confusion with my comment, but when I'm talking about `qt` I'm meaning the Homebrew formula of that name, namely Qt 4.8.7. That one is already past the end of its life and the only people who are still getting support for it are paying Qt customers. Everyone else should have migrated to the Qt 5 series a long time ago. Programs compiled with Qt 4 will probably continue to work on macOS Sierra, but it seems like Qt 4 will no longer compile on that system. In any case, the legacy Qt 4 series is not receiving any bug fixes anymore and the only reason we haven't eliminated it yet is because too many formulae rely exclusively on it and we haven't had the manpower to clean that up and migrate as many formulae as possible to Qt 5 (and boneyard the remaining so Qt 4 can eventually be removed).

pthariensflame:
Ah, I see!  Thanks for the clarification.  :)

PR #3834, opened 2016-07-17
Gmsh: version 2.13.1

schoeps:
### Have you:

### Comments
- this release fixes many issues, e.g. the ones in https://github.com/Homebrew/homebrew-science/pull/3585
- `brew audit --strict` asks for `=> :linked`. It's messy because audit does not take care of conditional dependencies. So is this now required?


schoeps:
I am still not sure how ":linked" is supposed to work but I did it since travis complains otherwise.

ilovezfs:
@schoeps it's libraries the software is once built actually ending up linked to regardless of whether you specifically added those libraries as explicit dependencies in the formula.

schoeps:
@ilovezfs how do I handle conditionals? I need to link to y only if switch z was chosen...

ilovezfs:
@schoeps the CI will only build the default configuration, so to pass audit you don't have to worry about those. However, technically what you *should* do is build it locally with each of the different options and then run brew audit, which will spit out the linked set for that option. Then write a block containing only the ones that were not already in the linked set for the default configuration that you already added to the formula, like this:
```
if build.with? "foo"
  depends_on "this" => :linked
  depends_on "that" => :linked
end
```

ilovezfs:
@schoeps for more info on this it might help to read through

schoeps:
@ilovezfs this is a lot of manual effort and computational time that could be fully automatized (e.g. octave.rb has around 20 options). I have seen that you already suggested in https://github.com/Homebrew/brew/pull/514 to let the CI do it... I am not sure if contributors will be happy about this additional burden.

ilovezfs:
@schoeps the `=> :linked` for the default configuration (i.e., the one that ends up in the binaries we distribute) is the set that we're mainly concerned about. For the non-default options the `=> :linked`, at this point at least, are primarily pedantry and won't be enforced.

schoeps:
The test fails because the old getdp bottle is used. Can one enforce it's recompilation? Or pull this request anyway :)

dpo:
@BrewTestBot test this please

dpo:
@schoeps would you mind adding the updated `getdp` to this PR again and bump its revision number? It seems it's linked to the versioned library of `gmsh`.

schoeps:
done, let's see what oracle says :)

getdp.rb
```diff
@@ -10,6 +10,7 @@ class Getdp < Formula
   url "http://www.geuz.org/getdp/src/getdp-2.9.0-source.tgz"
   sha256 "08487f3f5a41012d06db0ec97206b883961c0e7853f47f8502f6d1ef80ef67c9"
   head "https://geuz.org/svn/getdp/trunk", :using => GetdpSvnStrategy
+  version 2
```
dpo: `revision`

dpo:
Thanks!

PR #3836, opened 2016-07-17
getdp: version 2.9.0

schoeps:
### Have you:

- [ ] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors? See below

### Comments
- merge after gmsh
- patches for petsc are not necessary anymore 
- `brew audit --strict` asks for `=> :linked`. It's messy because audit does not take care of conditional dependencies. So is this now required?


schoeps:
I am still not sure how ":linked" is supposed to work but I did it since travis complains otherwise.

getdp.rb
```diff
@@ -32,19 +31,16 @@ class Getdp < Formula
...
-    sha256 "1950275890bf8ba1ad23ec80ff012921711fa0f3b5ee16ceac8be27ac356deea"
-  end
+  depends_on "fftw" => :linked
+  depends_on "homebrew/science/hwloc" => :linked
+  depends_on "homebrew/science/netcdf" => :linked
+  depends_on "homebrew/science/parmetis" => :linked
+  depends_on "homebrew/science/scalapack" => :linked
+  depends_on "homebrew/science/suite-sparse" => :linked
+  depends_on "homebrew/science/sundials" => :linked
+  depends_on "homebrew/science/superlu_dist" => :linked
```
dpo: Did the bot complain that these dependencies were not fully qualified?

schoeps: brew audit suggested this. I made a new commit to see if we get away without the full path.

ilovezfs: @schoeps See https://github.com/Homebrew/brew/issues/540.

ilovezfs: @dpo https://github.com/Homebrew/brew/issues/540.

schoeps: @ilovezfs Thanks, I was not aware of this thread.... however, what is the conclusion? :)

ilovezfs: The conclusion is people need to complain on that thread if they want it changed ;)

ilovezfs: But to be clear, the fully qualified path is not required at all. It's just the bot being goofy. Fully qualified works. Not fully qualified works. The only time it has to be fully qualified is if it's a dependency in another tap other than homebrew/core (e.g., homebrew/dupes, homebrew/python, etc.). So within science, there's no reason to qualify other science formulae or any formulae in core.

schoeps: @ilovezfs I am still not convinced that it is good idea that contributors add this. I think the bot should do it automatically as it is doing it with bottles. 

Most local installations of contributors are probably not using only standard bottles (i.e., switches have been chosen) and this makes this process messy. I might have to find out which dependency of a dependency has nonstandard options such that some other formula is not linked too...

ilovezfs: One problem with just automating it is that the bot is only telling us the truth about what is linked, but it's not telling us that it's actually correct that it should be linked.

For example, take a look at this harfbuzz https://github.com/Homebrew/homebrew-core/pull/3184. The bot said
```
  * Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "cairo" => :linked
      depends_on "gettext" => :linked
```

However, you'll notice that in the actual changes I never added `depends_on "cairo" => :linked`. The reason I didn't add that is the preexisting formula already had cairo listed a dependency, but one that's supposedly optional https://github.com/ilovezfs/homebrew-core/blob/3a15d08dfa50d23b0e2f7d4341dabbdabb2a820c/Formula/harfbuzz.rb#L30.
```
depends_on "cairo" => :optional
```

But if it's optional, then it shouldn't be in the default configuration, and the default configuration is what the bot is building, so why was cairo ending up linked? The reason is that another one of the dependencies of harfbuzz is gobject-introspection, which has an unconditional dependency on cairo. So cairo happens to be installed during the build of harfbuzz, and the harfbuzz build system notices cairo is installed and decides to opportunistically link to cairo and enable it. But this is expressly not what the formula intended if cairo is only supposed to be optional.

So how can that be fixed?
```
    if build.with? "cairo"
      args << "--with-cairo=yes"
    else
      args << "--with-cairo=no"
    end
```
It turns out the build system has an explicit option `--with-cairo=no` which will prevent it from enabling that feature and linking to cairo even if cairo happens to be installed. So adding that, the harfbuzz build system no longer linked to cairo, and the bot audit error went away.

So ideally contributors and maintainers are paying attention to the specifics of these audit errors and not just blindly pasting in the `=> :linked` that bot spits out. But is that actually happening? :)

ilovezfs: Now of course you could argue that cairo should just be always enabled for harfbuzz if it's going to be installed anyway, but that would require further discussion, whereas this linked issue was unambiguously a bug if we stipulate that cairo is supposed to be optional.

schoeps: I thought that audit was indeed looking to what harfbuzz was actually linked to (e.g. using otool -L)? But you say it's looking what was necessary to install harfbuzz? This is unfortunate because then you basically loose the beauty of the package manager that every package just care about its dependencies?

ilovezfs: >I thought that audit was indeed looking to what harfbuzz was actually linked

Yes that's exactly what it's doing

>But you say it's looking what was necessary to install harfbuzz?

No. All it does is check whether everything it's linked to is declared in the formula as a dependency.

What I'm saying is that we can't (or at least "shouldn't") automate adding the `=> :linked` because sometimes the right thing to do is change something else in the formula to prevent the linkage from occurring.

PR #3837, opened 2016-07-17
tophat: fixes for python and script paths

DarwinAwardWinner:
### Have you:

This adds patches to the tophat recipe to make all the python scripts
look for "python2" instead of just "python", since otherwise the tophat
scripts only run under python 2, and will fail if "python" points to
python 3. In addition, the "tophat2" script that just acts as an alias
to "tophat" has been modified to use `readlink -f` instead of just
`readlink`, so it now works when symlinked.

### DIFF
--- a/tophat.rb
+++ b/tophat.rb
@@ -3,9 +3,9 @@ class Tophat < Formula
   homepage "http://ccb.jhu.edu/software/tophat"
   url "http://ccb.jhu.edu/software/tophat/downloads/tophat-2.1.1.tar.gz"
   sha256 "37840b96f3219630082b15642c47f5ef95d14f6ee99c06a369b08b3d05684da5"
+  revision 2
   # doi "10.1093/bioinformatics/btp120"
   # tag "bioinformatics"
-  revision 1
 
   bottle do
     cellar :any
@@ -18,6 +18,7 @@ class Tophat < Formula
   depends_on "boost"
   depends_on "bowtie2"
   depends_on "bowtie" => :optional
+  depends_on :python if MacOS.version <= :snow_leopard
 
   patch :p0, :DATA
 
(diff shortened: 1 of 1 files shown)


dpo:
There's an issue with the test:
```
==> brew test homebrew/science/tophat --verbose
==> FAILED
Testing homebrew/science/tophat
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160717-74368-8buq8k-0.sb /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-science/tophat.rb --verbose
==> /usr/local/Cellar/tophat/2.1.1_1/bin/tophat -r 20 /usr/local/Cellar/tophat/2.1.1_1/share/test_data/test_ref /usr/local/Cellar/tophat/2.1.1_1/share/test_data/reads_1.fq /usr/local/Cellar/tophat/2.1.1_1/share/test_data/reads_2.fq
env: python2: No such file or directory
```

DarwinAwardWinner:
It appears to be because `python2` doesn't doesn't exist as a symlink to the Python 2 installation. I'm not sure why this is the case in the CI server, since this symlink does exist on my own machine. Does linuxbrew require `which python` to lead to Python 2? If so, that's a pretty serious limitation.

dpo:
It seems like the formula is missing a `python` dependency, but it looks like the `python2` link only exists if the user brews `python`. On 10.9 and older, we don't force installation of `python` but rely on system Python by default:
```
```
Then the test should call `system "python"` instead of `python2`. If `python` was brewed, then Homebrew Python will appear on the `PATH` before system Python.

tdsmith:
There is no `/usr/bin/python2` on OS X. You can use `python2.7` though.

dpo:
@DarwinAwardWinner

dpo:
Could you rebase your branch? The diff seems to be against some old version of tophat. Thank you.

DarwinAwardWinner:
Ok, rebased.

tophat.rb
```diff
@@ -3,6 +3,7 @@ class Tophat < Formula
   homepage "http://ccb.jhu.edu/software/tophat"
   url "http://ccb.jhu.edu/software/tophat/downloads/tophat-2.1.1.tar.gz"
   sha256 "37840b96f3219630082b15642c47f5ef95d14f6ee99c06a369b08b3d05684da5"
+  revision 1
```
dpo: There are two `revision` lines now. That's what was bothering me. Could you bump the revision number below instead of adding this line? Thanks.

dpo:
Many thanks for your contribution and your patience!

@sjackman 🐧

PR #3840, opened 2016-07-18
kat 2.1.0: Disable documentation

sjackman:
Disable documentation to remove Python dependency on `sphinx-doc` and build bottles.


sjackman:
```
  * Non-executables were installed to "/usr/local/Cellar/kat/2.1.0/bin"
    The offending files are:
      /usr/local/Cellar/kat/2.1.0/bin/kat_plot_colormaps.py
```

sjackman:
@maplesond Disabling documentation means we can build bottles again. Sound good?

maplesond:
Fine by me.  You just need to take out `make man` and `make html` from the formula.  

The kat_plot_colormaps.py is actually just some shared code that the other python scripts require, we left it in bin just for our convenience.  Does this cause a problem your end?

sjackman:
It's better for us if it passes all audit checks. It makes it easier to quickly confirm that all checks have passed. The fix is to move the above file to `$prefix/share/kat/`. The workaround/hack is to add `#!/usr/bin/env python` to the first line of the script, which will silence the audit warning. I'm fine with either.

sjackman:
Merged and added bottles for Mac. Still doesn't build on Linuxbrew yet due to the `boost` issue.

PR #3856, opened 2016-07-24
concorde (new formula)

Moisan:
### Have you:

### New formula: have you


concorde.rb
```diff
@@ -0,0 +1,44 @@
...
+  url "http://www.math.uwaterloo.ca/tsp/concorde/downloads/codes/src/co031219.tgz"
+  sha256 "c3650a59c8d57e0a00e81c1288b994a99c5aa03e5d96a314834c2d8f9505c724"
+
+  option "with-cplex=", "Compile with CPLEX bindings, you must install CPLEX beforehand and pass the CPLEX installation directory. For example, with version 12.6.3 Community, the default installation directory is /Users/username/Applications/IBM/ILOG/CPLEX_Studio_Community1263/cplex"
+  depends_on "qsopt" if build.without? "cplex="
+
+  def install
+    args = %W[--prefix=#{prefix}
+              --disable-dependency-tracking
+              --host=x86_64]
```
dpo: Is Concorde also available for 32bit architectures? If so you can switch using `MacOS.prefer_64_bit?`.

concorde.rb
```diff
@@ -0,0 +1,44 @@
...
+    if build.with? "cplex="
+      cplex_path = ARGV.value("with-cplex")
+      mkdir "cplex_files"
+      cd "cplex_files" do
+        ln_s Dir["#{cplex_path}/include/ilcplex/*.h"], "."
+        ln_s "#{cplex_path}/lib/x86-64_osx/static_pic/libcplex.a", "."
+      end
+      args << "--with-cplex=#{buildpath}/cplex_files"
+    else
+      args << "--with-qsopt=#{Formula["qsopt"].lib}"
```
dpo: Let's use `opt_lib` so we're not constrained by a specific version of `qsopt`.

concorde.rb
Moisan: Actually the host argument doesn't matter. The configure script is supposed to be able to guess this it but was failing to do so on my machine. Giving any string as argument makes the build succeed.

However I supposed CPLEX was installed with 64bit architecture a few lines bellow. I'll make the switch for this one.

PR #3857, opened 2016-07-24
arrayfire 3.3.2

ilovezfs:
vendor glew1 in since glew2 dropped glewmx support

### DIFF
--- a/arrayfire.rb
+++ b/arrayfire.rb
@@ -4,8 +4,8 @@ class Arrayfire < Formula
 
   stable do
     url "https://github.com/arrayfire/arrayfire.git",
-      :tag => "v3.2.2",
-      :revision => "7507b61b3f89c99d23900ce4b66d3f8aeef4e609"
+      :tag => "v3.3.2",
+      :revision => "f65dd9798f8efeea4d55efe34ba62f4fc3ae7ca0"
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 any idea if the analytics data supports taking this one out back?

tdsmith:
_i_ use arrayfire so let's keep it :D

ilovezfs:
@tdsmith *is* the reverse dependency.

tdsmith:
in any case package popularity is a classic long-tail distribution; _most_ packages in the homebrew ecosystem have only a handful of users, so i don't know what a useful cutoff would be

ilovezfs:
>most packages in the homebrew ecosystem have only a handful of users

doesn't this imply most packages are dispensable? :)

ilovezfs:
:green_apple: for Jenkins.

tdsmith:
Maybe! But I expect many users will have a mixture of popular and unpopular packages installed; having unpopular packages provides a lot of value. Just... diffusely.

dpo:
I also use arrayfire.

ilovezfs:
two revdeps!

ilovezfs:
@tdsmith :ship:?

PR #3865, opened 2016-07-25
openimageio 1.6.16

JCount:
### Have you:

add support for optionally building HEAD with openjpeg-2.1.1 ;
vendor openjpeg-1.5.2 for all other builds ;
see https://github.com/Homebrew/homebrew-core/pull/3234


openimageio.rb
```diff
@@ -69,6 +74,11 @@ class Openimageio < Formula
         :revision => "9a70c65c7a29a50114a8208d61c87ba4fedd018e"
   end
 
+  resource "openjpeg-1.5.2" do
+    url "https://mirrors.ocf.berkeley.edu/debian/pool/main/o/openjpeg/openjpeg_1.5.2.orig.tar.gz"
+    sha256 "aef498a293b4e75fa1ca8e367c3f32ed08e028d3557b069bf8584d0c1346026d"
+  end
+
```
dpo: There's already a formula for `openjpeg`.

ilovezfs: @dpo Yes, @JCount is working on that: https://github.com/Homebrew/homebrew-core/pull/3234

dpo: So `openimageio` requires `openjpeg` 1.5.2 and not 2.1.1?

ilovezfs: @dpo It needs a new tag: https://github.com/OpenImageIO/oiio/issues/1449

I'd highly recommend pestering upstream for one!

DomT4: I asked upstream if there's a timeline on the next release yet.

JCount:
What I did to make the vendor itself work, and then get `FindOpenJpeg.cmake` to recognize its existence, is less than ideal. Any suggestions on how to clean it up before, and if, this PR is merged are more than welcome.

DomT4:
Per upstream:
> I'm planning to cut releases on Aug 1, and it will build fine against OpenJpeg 2.1. Is that soon enough, or do I need to try to do it sooner? 

I think that's reasonable enough.

JCount:
Agreed.

JCount:
There are some issues with this that have been there for a while, that I am not sure how to best solve. Building with the arg `EMBEDPLUGINS=1` causes behavior that might be unwanted; namely, it appears to automatically enable all the `Findfoo.cmake` modules, and the enable the `cmake` flags for building the plugins into itself. What I specifically observed was it adding `-DUSE_FFMPEG` to the args, and then explicitly and deeply linking against it.

JCount:
I actually currently have a few more changes in the works, aimed at allowing this to accomplish more than merely being the version bump needed to let the `openjpeg` switch move forward.

openimageio.rb
```diff
@@ -13,6 +13,7 @@ class Openimageio < Formula
   end
 
   option "with-test", "Dowload 95MB of test images and verify Oiio (~2 min)"
```
JCount: Currently the `--with-test` breaks the build, as it does in precisely the same way with the `1.6.15` formula.
```
84% tests passed, 14 tests failed out of 86

Total Test time (real) =  77.88 sec

The following tests FAILED:
	 10 - oiiotool-subimage (Failed)
	 11 - oiiotool-text (Failed)
	 12 - dither (Failed)
	 20 - python-typedesc (Failed)
	 21 - python-imagespec (Failed)
	 22 - python-roi (Failed)
	 23 - python-deep (Failed)
	 24 - python-imageinput (Failed)
	 25 - python-imageoutput (Failed)
	 26 - python-imagebuf (Failed)
	 27 - python-imagebufalgo (Failed)
	 32 - texture-derivs (Failed)
	 53 - texture-icwrite (Failed)
	 61 - openexr-v2 (Failed)
Errors while running CTest
make: *** [Makefile:366: test] Error 8
...
Error: homebrew/science/openimageio 1.6.16 did not build
```

DomT4: Unless you know how to fix it, I'd be tempted to just remove the option. Evidently either nobody is building `--with-test`, nobody cares enough to report the failure or whatever is causing the tests to fail isn't something people are seeing during run.

Up to @Homebrew/science ultimately, but would be nice to get this PR wrapped fairly soon & this is a preexisting issue.

openimageio.rb
```diff
@@ -132,9 +142,15 @@ def install
...
-    args << "USE_PYTHON3=1" if build.with? :python3
+    if build.with? "qt"
+      args << "USE_OPENGL=ON"
+      args << "USE_QT=ON"
+    else
+      args << "USE_OPENGL=OFF"
+      args << "USE_QT=OFF"
+    end
+
+    args << "USE_PYTHON3=ON" if build.with? :python3
```
DomT4: Should be `"python3"`.

DomT4:
> EDIT: I removed the hardcoded includes for both openjpeg and freetype, primarily because the one for openjpeg was breaking the build due to the currently bottled version.

Due to the way CI works we'd need to merge the core PR before Science can find the new versions you're telling it to look for. That's probably the issue.

JCount:
I could determine the above from the error output, but given that `upstream` has been working on its `cmake` build system, I felt that trying to see if it could dynamically find the include paths was worth exploring.

openimageio.rb
JCount: I will wait to fix this, so that I can just roll it into however the `--with-test` failure is going to be handled, preventing yet another automatically triggered rebuild, unless you feel I should do it immediately.

DomT4:
I see. Yeah, `cmake` theoretically should be able to hunt things down pretty well.

Given there's a release with the necessary `openjpeg` support available now I think we're in a good position to merge the core PR once CI is happy there, which will enable this PR to find the updated libraries & paint a more accurate picture of where things are in terms of readiness here.

DomT4:
@BrewTestBot test this please

JCount:
Rebased to remove `revision 1` from formula.

DomT4:
Alright, let's merge this in. Nobody from Science has replied on the test situation for a couple of days now, and it's no more broken than it was before, so let's get the immediate fix for the `openjpeg` change in & the rest can be hammered out later.

Merged in https://github.com/Homebrew/homebrew-science/commit/0b70e2494dbb4447e6c1bc3a861aca66591cd27c. Thank you for all your work around the `openjpeg` update @JCount; you've been great 🙇.

PR #3875, opened 2016-07-28
Added description for fastqc

smoe:
### Have you:

Actually, the current version is not passing the audit since that description is missing.

Had my own flavour of that formula.

While at it, please also have a look at my implementation that starts from the source code:

```
class Fastqc < Formula
  desc "quality control tool for high throughput sequence data"
  homepage "http://www.bioinformatics.babraham.ac.uk/projects/fastqc/"
  url "http://www.bioinformatics.babraham.ac.uk/projects/fastqc/fastqc_v0.11.5_source.zip"
  version "v0.11.5"
  sha256 "ae3056fabb59a2b8fcbd0e23edbf57b71ad68dbc8d407615dd7bf7cce0a439a3"

  # depends_on "cmake" => :build
  depends_on "ant" # if your formula requires any X11/XQuartz components
  depends_on "jdk"
  depends_on "perl"
  depends_on :x11 => :optional

  def install
    system "pwd"
    system "ant"
    system "mkdir","-p","#{prefix}/share"
    system "mkdir","-p","#{prefix}/bin"
    system "cp","-ar","bin","#{prefix}/share/fastqc"
    system "ln","-s","#{prefix}/share/fastqc/fastqc","#{prefix}/bin/fastqc"
    system "chmod","+x","#{prefix}/bin/fastqc"
  end

  test do
    system "fastqc","-v"
  end
end
```

Your implementation is very similar to mine, but you use the provided binaries instead of compiling.  It is Java, hence a bit pointless, I admit, but anyway, it would be nice.

Can one possibly distingiush between dependencies of the build and dependencies of the readily drinkable bottle?

An update to 0.11.5 would be nice.

Many thanks

Steffen


sjackman:
Thanks for your contribution, Steffen!

> Your implementation is very similar to mine, but you use the provided binaries instead of compiling. It is Java, hence a bit pointless, I admit, but anyway, it would be nice.

I'd prefer to use the provided binaries. @Homebrew/science ?

> Can one possibly distingiush between dependencies of the build and dependencies of the readily drinkable bottle?

Yes.
```ruby
depends_on :java => :build
```

> An update to 0.11.5 would be nice.

Please submit a pull request.

fastqc.rb
```diff
@@ -1,4 +1,5 @@
 class Fastqc < Formula
+  desc "quality control tool for high throughput sequence data"
```
sjackman: Please change `quality` to `Quality`

smoe:
To avoid too many patches for trivial bits, I have created a new quest at for the same thing on
https://github.com/Homebrew/homebrew-science/pull/3888

apjanke:
> I'd prefer to use the provided binaries. It's easier and more repeatable. The biggest argument for building from source is when it's necessary to patch the source.
@Homebrew/science ?

In homebrew/core, we've been accepting binary (JAR) distributions for Java programs, unless patching is needed. With Java, the binaries are cross-platform and (effectively) not optimized, and optional features and libraries are usually activated at run time from a single common JAR, instead of conditionally compiled in as with C programs. So there's not much reason to compile from source for Java packages.

sjackman:
@smoe Rather than open a new pull request, use `git commit --amend fastqc.rb` to modify your commit, and `git push --force` to update the pull request on GitHub.

PR #3888, opened 2016-07-28
fastqc: Added description

smoe:
As requested - capitalisation of first word in title.


sjackman:
```
❯❯❯ brew audit --strict fastqc
homebrew/science/fastqc:
  * `desc` (line 3) should be put before `homepage` (line 2)
Error: 1 problem in 1 formula
```

sjackman:
Merged. Thanks for your contribution, Steffen!

PR #3892, opened 2016-07-29
diamond 0.8.16

ilovezfs:
use CMake


ilovezfs:

sjackman:
👍 Thanks again, Joe.

PR #3913, opened 2016-07-30
clfft 2.12.1

JCount:
### Have you:

update to version 2.12.1


dpo:
Thanks. There seems to be an issue with the test (nevermind the compiler warning):
```
==> brew test homebrew/science/clfft --verbose
==> FAILED
Testing homebrew/science/clfft
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160730-44095-1yw64v9-0.sb /Users/brewadmin/Homebrew/Cellar/ruby187/1.8.7-p374_1/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-science/clfft.rb --verbose
==> /usr/bin/clang++ /usr/local/Cellar/clfft/2.12.1/share/clfft/examples/fft1d.c -I/usr/local/opt/clfft/include -L/usr/local/opt/clfft/lib -lclFFT -framework OpenCL -o fft1d
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated
/usr/local/Cellar/clfft/2.12.1/share/clfft/examples/fft1d.c:74:77: warning: format specifies type 'unsigned int' but the argument has type 'unsigned long' [-Wformat]
    printf("\nPerforming fft on an one dimensional array of size N = %u\n", (unsigned long)N);
                                                                     ~~     ^~~~~~~~~~~~~~~~
                                                                     %lu
1 warning generated.
Error: homebrew/science/clfft: failed
<"Platform found: Apple\nDevice found on the above platform: Intel(R) Core(TM) i7-3720QM CPU @ 2.60GHz\n\nPerforming fft on an one dimensional array of size N = 16\n(0.000000, 0.000000) (1.000000, 3.000000) (2.000000, 6.000000) (3.000000, 9.000000) (4.000000, 12.000000) (5.000000, 15.000000) (6.000000, 18.000000) (7.000000, 21.000000) (8.000000, 24.000000) (9.000000, 27.000000) (10.000000, 30.000000) (11.000000, 33.000000) (12.000000, 36.000000) (13.000000, 39.000000) (14.000000, 42.000000) (15.000000, 45.000000) \n\nfft result: \n(0.000000, 0.000000) (1.000000, 3.000000) (2.000000, 6.000000) (3.000000, 9.000000) (4.000000, 12.000000) (5.000000, 15.000000) (6.000000, 18.000000) (7.000000, 21.000000) (8.000000, 24.000000) (9.000000, 27.000000) (10.000000, 30.000000) (11.000000, 33.000000) (12.000000, 36.000000) (13.000000, 39.000000) (14.000000, 42.000000) (15.000000, 45.000000) \n"> expected to be =~
</\(120\.000000,\ 360\.000000\)/>.
```

JCount:
hmmm. I must have forgotten to enable `--verbose` when I ran the test, even though I normally do.

JCount:
Ok, so the issue seems to actually (simply) be an upstream typo in the provided example `fft1d.c` file that has been being used as the test. I intend to let upstream know about how `%u` was used instead of `%lu`. 

Note, though this makes it so that this is not currently an example file to use as a test, I think @dpo would agree with me that it is not indicative of an issue with the software itself.

Manually repairing the file that the test currently utilizes seems to resolve everything besides the deprecation warning. Unfortunately this issue is present within all of the examples currently. Therefore, at least temporarily, it probably will need to be necessary to create an entirely new test format, or put a hold on this and wait until upstream is able to fix this. (Adding a patch to fix the example seems like a bad idea)

This all seems to originate from a patch to correct an accidental use of `%ul` instead of the valid `%lu`; unfortunately the fix replaced the typo with `%u`, which is itself a typo that insidiously exists as a compiler type mismatch warning.

The relevant commit seems to be https://github.com/clMathLibraries/clFFT/commit/238cb52ac3c12a150377b3ef7f1feb705269b05d

dpo:
Thanks. The warning isn't what worries me though, it's this part of the output on the bot:
```
Expected /\(120\.000000,\ 360\.000000\)/ to match "Platform found: Apple\nDevice found on the above platform: Intel(R) Xeon(R) CPU           X5570  @ 2.93GHz\n\nPerforming fft on an one dimensional array of size N = 16\n(0.000000, 0.000000) (1.000000, 3.000000) (2.000000, 6.000000) (3.000000, 9.000000) (4.000000, 12.000000) (5.000000, 15.000000) (6.000000, 18.000000) (7.000000, 21.000000) (8.000000, 24.000000) (9.000000, 27.000000) (10.000000, 30.000000) (11.000000, 33.000000) (12.000000, 36.000000) (13.000000, 39.000000) (14.000000, 42.000000) (15.000000, 45.000000) \n\nfft result: \n(0.000000, 0.000000) (1.000000, 3.000000) (2.000000, 6.000000) (3.000000, 9.000000) (4.000000, 12.000000) (5.000000, 15.000000) (6.000000, 18.000000) (7.000000, 21.000000) (8.000000, 24.000000) (9.000000, 27.000000) (10.000000, 30.000000) (11.000000, 33.000000) (12.000000, 36.000000) (13.000000, 39.000000) (14.000000, 42.000000) (15.000000, 45.000000) \n".
```
see https://travis-ci.org/Homebrew/homebrew-science/jobs/148591613#L421.

I can't reproduce that error locally (though I get the warning). It might be due to the version of Ruby used on the bot though.

JCount:
@dpo Oh, ok. I am in the process of creating an upstream PR to fix the example files. I agree with you, that the output that concerned you is not really an issue, just a slight bug in the way the test continued to execute after the bot had triggered the `fail` logic. 

What seems to be happening is that the error thrown by `brew test`, is in fact also being passed to, and caught by, (probably indirectly) the native error handling code present in the example file. I think the travis-ci and BrewTestBot are both simply running into the same mismatch error, but due to their setup, it seems to seems to end up triggering the error handling logic in the example files in an unanticipated way, leading to the spewing forth of a garbled mess comprised of the example file's error logic printf()'s.

I hope that makes sense, it did in my head. I don't think this will, but it is somewhat the way I ended up putting it together in my own head:  I would term it an external error-triggered error in the error handling logic.

This does actually have a point. If all of the circles that my head went in to reach that conclusion produced anything at all correct, then there could potentially be a minor issue with how separately return values from `brew test --verbose` are passed up the call stack when they originate from somewhere within the logic of the Ruby code of the `test` module itself (in this case the warning), versus when module itself catches and handles (fatal) error output from the actual test it executes. (It might have to do with some sort of "cross-contamination" if the sandboxing mechanism has some flaw or minor imperfection.)

Ok, that was wild hypothesizing, but somewhere in it there could be a fragment of reality, but I cannot analyze any of it at all objectively.

PR #3916, opened 2016-07-31
shark 3.1.0

ilovezfs:
use boost 1.60.0


ilovezfs:
@BrewTestBot test this please

ilovezfs:
needs to be reported upstream but pulling as-is

PR #3918, opened 2016-07-31
gap 4.8.4

sjackman:


gap.rb
```diff
@@ -86,16 +79,11 @@ def install
 
     if build.with? "InstPackages"
       ohai "Trying to automatically build included packages"
-
-      resource("script_that_builds_included_packages").stage do
-        chmod "u+x", "InstPackages.sh"
-        (libexec/"pkg").install "InstPackages.sh"
-      end
-
+      chmod "u+x", libexec/"bin/BuildPackages.sh"
```
alex-konovalov: this file has already executable permissions in the GAP distribution, so this line is redundant (though it doesn't make any harm)

alex-konovalov:
Thank you @sjackman! the changes look good to me - it's essential that you have noted the new BuildPackages.sh script introduced in GAP 4.8.4. Perghaps the Travis build fails only because of not receiving any output for 10 minutes?

sjackman:
Merged!

PR #3931, opened 2016-08-02
gtsam: depend on boost155

ilovezfs:
primarily intended to trigger CI

### DIFF
--- a/gtsam.rb
+++ b/gtsam.rb
@@ -1,7 +1,9 @@
 class Gtsam < Formula
+  desc "Library of C++ classes that implement SAM in robotics and vision"
   homepage "https://collab.cc.gatech.edu/borg/gtsam/"
   url "https://research.cc.gatech.edu/borg/sites/edu.borg/files/downloads/gtsam-3.2.1.tgz"
   sha256 "1e9217c11d92e6838e2d0bec3a7dd0d36d2131acdf2e50264f6dc225d8ce1a97"
+  revision 1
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
@sjackman this is the same failure I'm seeing locally.

ilovezfs:
Just confirmed it builds fine with boost155, and that found that it also fails with boost159.

ilovezfs:
Hopefully, there can be an El Capitan boost155 bottle to fix this https://github.com/Homebrew/homebrew-versions/pull/1330.

gtsam.rb
```diff
@@ -1,23 +1,27 @@
...
 
   bottle do
     sha256 "5e1040d1a55f395b038e5ee5647377a2b8182f2af0536f8e8c028c81cc7f57d9" => :yosemite
     sha256 "b62f92c94589f4794cfe67076213aeae26234a464aa87727e2312da32c277eb4" => :mavericks
     sha256 "4c9a630f008d37f1ecd19cd1ce87912206f2eed24caf0c7b6404b5c08eba74cf" => :mountain_lion
   end
 
+  option "with-test", "Run unit tests at build-time"
+
+  deprecated_option "without-tests" => "without-test"
```
sjackman: I think this line should be 
```ruby
  deprecated_option "with-tests" => "with-test"
```

ilovezfs: fixed.

PR #3934, opened 2016-08-02
gap: Rename option InstPackages to BuildPackages

alex-konovalov:
Following #3918 by @sjackman, I suggest to rename the `InstPackages` option to `BuildPackages`.

### DIFF
--- a/gap.rb
+++ b/gap.rb
@@ -17,8 +17,10 @@ class Gap < Formula
   #   with Homebrew), and a number of GAP packages under `pkg/`, some of
   #   which need to be built.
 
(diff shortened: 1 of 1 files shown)


sjackman:
How about `--with-packages`?

sjackman:
Also add
```ruby
deprecated_option "with-InstPackages" => "with-packages"
```

alex-konovalov:
@sjackman good suggestions, thanks! I've just force-pushed the new version.

sjackman:
Merged. Thanks for your contribution, Alex.

PR #3940, opened 2016-08-03
abyss: bump for boost 1.60.0

tdsmith:


DomT4:
You sent me scrambling to check we hadn't broken the CI seeing so many pending builds on Science suddenly 😄. Thought it'd borked on doing the same one over and over again for a minute.

tdsmith:
boooooooost ¯\\\_(ツ)\_/¯

DomT4:
Won't miss these mass bumps once we handle major updates a little more gently, that's for sure. Break stuff & hope we can fix it becomes less acceptable as we creep towards 1.0 🙈.

PR #3959, opened 2016-08-03
dgtal: bump for boost 1.60.0

tdsmith:


tdsmith:
needs to be revision 2 since revision 1 bottles are stale but too old to delete

PR #3964, opened 2016-08-03
express: bump for boost 1.61.0

tdsmith:


tdsmith:
@sjackman What should I do about the Linux bottle here?

sjackman:
I haven't yet merged Homebrew/core, including boost. I'm doing that today. So, the Linuxbrew bottles will all be built against the old version of boost. I'll merge Homebrew/core today, then trigger rebuilds of the affected packages. Go ahead and merge the PR deleting the Linuxbrew bottle, and tag me in a comment reminding me to rebuild and merge the Linuxbrew bottle.

PR #3981, opened 2016-08-03
orocos-kdl: bump for boost 1.60.0

tdsmith:


tdsmith:
Verified it continues to build; no need to rebottle for header-only use.

PR #3986, opened 2016-08-03
quake: bump for boost 1.60.0

tdsmith:


tdsmith:
quake is known broken

PR #3993, opened 2016-08-03
trilinos: bump for boost 1.60.0

tdsmith:


tdsmith:
Ruby error in formula:

```
Error: undefined local variable or method `s' for #<Formulary::FormulaNamespacebc1388db77d99fa87f42d96420e57de5::Trilinos:0x007ffad1421480>
Please report this bug:
/usr/local/Library/Taps/homebrew/homebrew-science/trilinos.rb:290:in `block in install'
/usr/local/Library/Homebrew/extend/fileutils.rb:95:in `block in mkdir'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/fileutils.rb:125:in `chdir'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/fileutils.rb:125:in `cd'
/usr/local/Library/Homebrew/extend/fileutils.rb:94:in `mkdir'
/usr/local/Library/Taps/homebrew/homebrew-science/trilinos.rb:276:in `install'
```

dpo:
[Line 289](https://github.com/Homebrew/homebrew-science/blob/master/trilinos.rb#L289) should be replaced with
```
inreplace "#{lib}/cmake/Trilinos/TrilinosConfig.cmake" do |s|
  s.gsub! "PyTrilinos;", "" if s.include? "COMPONENTS_LIST"
end
```
I introduced the bug following a audit warning that wanted a one-liner `inreplace`.

dpo:
Seems to be the audit failure that's buggy. The 3-liner defines a variable that's used in a test.

ilovezfs:
You can use a regex.

I think this will work:
```
inreplace "#{lib}/cmake/Trilinos/TrilinosConfig.cmake", /((?=.*?COMPONENTS_LIST)[^\r\n]*)(PyTrilinos;)/, "\\1"
```

ilovezfs:
CC @UniqMartin for thoughts on the audit test's merit or lack thereof.

dpo:
Yes we could use a regex though I find that less readable.

ilovezfs:
>I find that less readable.

Indeed. I'd recommend using a `patch do` instead and putting it in https://github.com/Homebrew/formula-patches.

dpo:
I didn't know about homebrew/formula-patches. Is that for all taps together? There'll be a lot of stuff in there.

ilovezfs:
Yes it's for all taps. Whether that makes any sense is left as an exercise for the reader.

dpo:
I'll submit a PR soon to upgrade Trilinos. It appears as though that patch may no longer be necessary.

UniqMartin:
I think the true issue here is that the substitution via `inreplace` is somewhat bogus. It should probably be more specific and not conditional on the contents of the file itself. To me it feels like the replacement should instead be wrapped by `if build.with?("python")` or something similar. (But I haven't investigated this piece of software and its build system to be absolutely certain.)

> CC @UniqMartin for thoughts on the audit test's merit or lack thereof.

I think the audit check is still generally useful, though it fails to recognize this non-standard use. We might be able to improve the audit check for cases like this one, but I think our energy is better invested into other things and avoiding such uses entirely.

PR #4017, opened 2016-08-05
itensor 2.0.10 (new formula)

kyungminlee:


itensor.rb
```diff
@@ -0,0 +1,160 @@
+class Itensor < Formula
+  desc "C++ library for implementing tensor product wavefunction calculations"
+  homepage "http://itensor.org/"
+  url "https://github.com/ITensor/ITensor/archive/v2.0.10.tar.gz"
+  sha256 "3b5e829362ecfc6984227baa0cfbd5ef4aec45334bfb8cdda3cc7aa88c109ddb"
+  head "https://github.com/ITensor/ITensor.git"
+
+  option "with-openblas", "Use OpenBLAS instead of Apple's Accelerate Framework"
+
+  depends_on "homebrew/science/openblas" => (OS.mac? ? :optional : :recommended)
```
dpo: There's no need to fully qualify `openblas`, and the `option` above is redundant with the dependency, and can be removed.

dpo:
Thanks. The formula looks good. It would be useful to contribute a complete Makefile to Itensor.

kyungminlee:
Commit updated following PR comments.

PR #4020, opened 2016-08-06
remove formula suite-sparse421

dpo:


raffienficiaud:
We depend on this one, why is this removed?

dpo:
Nothing else in homebrew/science depends on it and it's been superseded by `suite-sparse`. If you're maintaining a tap, feel free to grab the old formula and add it to your tap.

raffienficiaud:
That is ok for homebrew-science, but not for the users. Are those versions compatible? We experienced a change in the API between those versions.

Maybe I misunderstood something about homebrew-science. Isn't it a science oriented set of package for homebrew, in which case it also has the contract of maintaining some compatibility?

Thanks,

dpo:
There was an API change after 4.2.1. Because we normally don't maintain formulae for different versions of a library in this tap (only the latest), and because no other homebrew/science formula depends on `suite-sparse421`, it was removed. If you still rely on that old version of SuiteSparse, the best is to include the formula in a tap. Alternatively, it could be moved to homebrew/versions or to homebrew/boneyard. Feel free to submit a PR there.

raffienficiaud:
Thanks, I will check there then.

raffienficiaud:
Sorry to bother again. It appears that

``brew install https://raw.githubusercontent.com/Homebrew/homebrew-science/89c1ed1083f197931f51aced73666c202a2a5f05/suite-sparse421.rb
``
just works. Apart from being difficult to handle (we have to find the right revision to install), do you see any other shortcomings in using this approach?

dpo:
There's no particular shortcoming other than the fact that it's cumbersome. If you feel others might also need that formula, we should just add it to homebrew/versions.

PR #4031, opened 2016-08-10
calculix-ccx: fix patch

dpo:


calculix-ccx.rb
```diff
@@ -117,7 +118,7 @@ def install
 @@ -41,5 +41,5 @@ void *u_free(void* ptr,const char *file,const int line, const char* ptr_name){
    if(log_realloc==1) {
        printf("FREEING of variable %s, file %s, line=%d: oldaddress= %ld\n",ptr_name,file,line,(long int)ptr);
-   }
+   }
```
dpo: @MikeMcQuaid When I use `brew pull --bottle`, trailing whitespace is removed. In this case, it's a problem because the patch fails to apply without the whitespace. Any idea where that behavior comes from?

MikeMcQuaid: Use `brew pull --clean --bottle`

dpo: Great! Thanks!

dpo: Hm. For some reason, that didn't close the PR.

PR #4046, opened 2016-08-12
libbuddy 2.4 (new formula)

mkhl:
This library is used by `maude`, which I’m submitting an (updated) source build separately.

PR #4047, opened 2016-08-12
maude 2.7.1

mkhl:
The current version of maude can (finally) be built from source using clang.

This change includes
* build instructions
* a script that sets `MAUDE_LIB` automatically
* a more extensive test
* a formula for the required `BuDDy` library, also submitted separately as #4046


dpo:
Is there a test that could be added to `libbuddy`? If not and if no other formula uses it, perhaps it should be a resource inside `maude`?!
```
==> brew audit homebrew/science/libbuddy --new-formula
==> FAILED
Error: 1 problem in 1 formula
homebrew/science/libbuddy:
```

mkhl:
It’s a C library, we could probably write a simple program using it and check that it compiles.
Would you prefer that?

dpo:
Yes, that would be the best. Thanks!

dpo:
Please remove `libbuddy` from here and update #4046.

mkhl:
done

dpo:
@BrewTestBot test this please

mkhl:
@BrewTestBot test this please

dpo:
Great, thank you!

PR #3789, opened 2016-06-28
circlator: revision bump required by audit

ilovezfs:
- audit doesn't care that the formula was "unrevisioned" in 734f0823;
  subsequent commits must have a revision at least as great as the
  revoked revision or audit will fail

- add GCC dependency to satisfy numpy's :fortran requirement

- fix audit issue regarding starting with an indefinite article

PR #4050, opened 2016-08-13
circlator 1.3.0

ilovezfs:
use new virtualenv feature

### DIFF
--- a/circlator.rb
+++ b/circlator.rb
@@ -1,10 +1,10 @@
 class Circlator < Formula
-  desc "Tool to circularize genome assemblies"
-  homepage "https://github.com/sanger-pathogens/circlator"
-  url "https://github.com/sanger-pathogens/circlator/archive/v0.16.0.tar.gz"
-  sha256 "c5331e1a687dacedd134356177671228e592c772df3306d6e037e55bd62d84df"
-  revision 1
+  include Language::Python::Virtualenv
 
+  desc "Tool to circularize genome assemblies"
+  homepage "https://sanger-pathogens.github.io/circlator/"
+  url "https://github.com/sanger-pathogens/circlator/archive/v1.3.0.tar.gz"
+  sha256 "0d5219ffa7b2c93fbd8277e6b39ac8e3926cb71b64a06c3f66e3f98a0b5be39d"
   head "https://github.com/sanger-pathogens/circlator.git"
 
   bottle do
@@ -17,48 +17,51 @@ class Circlator < Formula
 
   # tag "bioinformatics"
 
-  depends_on "zlib" unless OS.mac?
-  depends_on :fortran
   depends_on :python3
   depends_on "bwa"
+  depends_on "mummer"
   depends_on "prodigal"
   depends_on "samtools"
   depends_on "spades"
-  depends_on "homebrew/python/numpy" => ["with-python3"]
-  depends_on "homebrew/python/pymummer"
+  depends_on "zlib" unless OS.mac?
 
-  resource "pysam" do
-    url "https://pypi.python.org/packages/source/p/pysam/pysam-0.8.3.tar.gz"
-    sha256 "343e91a1882278455ef9a5f3c9fc4921c37964341785bf22432381d18e6d115e"
+  resource "et_xmlfile" do
+    url "https://files.pythonhosted.org/packages/22/28/a99c42aea746e18382ad9fb36f64c1c1f04216f41797f2f0fa567da11388/et_xmlfile-1.0.1.tar.gz"
+    sha256 "614d9722d572f6246302c4491846d2c393c199cfa4edc9af593437691683335b"
+  end
+
+  resource "jdcal" do
+    url "https://files.pythonhosted.org/packages/37/36/3199cfb80fcbf4e4df3a43647733d4f429862c6c97aeadd757613b9e6830/jdcal-1.2.tar.gz"
+    sha256 "5ebedb58b95ebabd30f56abef65139c6f69ec1687cf1d2f3a7c503f9a2cdfa4d"
+  end
+
+  resource "openpyxl" do
+    url "https://files.pythonhosted.org/packages/25/17/0d6096f9b2cb3e0ebb5b2fb60de016606f6dd599bec94ed524e84a1e2da8/openpyxl-2.3.5.tar.gz"
+    sha256 "4307578b94a708e1519295c333c51477ac51a06f01e81b2697cc301c286a4762"
   end
 
   resource "pyfastaq" do
-    url "https://pypi.python.org/packages/source/p/pyfastaq/pyfastaq-3.6.0.tar.gz"
-    sha256 "3a052466f89db1e1ac6cfb475117fdf8466ed319b62aabf9531e38f569b51588"
+    url "https://files.pythonhosted.org/packages/2a/46/6ece19838a79489556c97092e832bafeb46e7b28c52418a6c5a7568da999/pyfastaq-3.13.0.tar.gz"
+    sha256 "79bfe342e053d51efbc7a901489c62e996566b4baf0f33cde1caff3a387756af"
   end
 
-  resource "bio_assembly_refinement" do
-    url "https://pypi.python.org/packages/source/b/bio_assembly_refinement/bio_assembly_refinement-0.3.3.tar.gz"
-    sha256 "3547d54b357702d659d244ed6493b6553eb2e72dcf79e88cdae2c595b933bf91"
+  resource "pymummer" do
+    url "https://files.pythonhosted.org/packages/8d/28/e289144751e0b33685f00a54eb5947b989348d4472a83b64a8de0b0e3f2e/pymummer-0.7.1.tar.gz"
+    sha256 "7aab311c60fcb9fc5a2bce658e949d80f4801e73107eb2e835f46caed02cfedf"
+  end
+
+  resource "pysam" do
+    url "https://files.pythonhosted.org/packages/27/89/bf8c44d0bfe9d0cadab062893806994c168c9f490f67370fc56d6e8ba224/pysam-0.8.4.tar.gz"
+    sha256 "30cf23931edf8a426678811f234bca4a83a53438028b323f2ef55792562d9dea"
   end
 
   def install
-    version = Language::Python.major_minor_version "python3"
-    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python#{version}/site-packages"
-    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python#{version}/site-packages"
-
-    %w[pysam pyfastaq bio_assembly_refinement].each do |r|
-      resource(r).stage do
-        system "python3", *Language::Python.setup_install_args(libexec/"vendor")
-      end
-    end
-
-    system "python3", *Language::Python.setup_install_args(libexec)
-    bin.install Dir[libexec/"bin/*"]
-    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+    virtualenv_create(libexec, "python3")
+    virtualenv_install_with_resources
   end
 
   test do
-    assert_match "Available commands", shell_output("circlator -h 2>&1", 0)
+    output = shell_output("#{bin}/circlator test outdir")
+    assert_match "Finished run on test data OK", output
   end
 end


martinghunt:
Thanks for this. Is it possible to specify a spades version? If so, spades 3.7.1 gave me slightly better results with circlator on testing out all versions 3.6.2-3.9.0.

ilovezfs:
@martinghunt yeah, if you believe it gets better results, and given that upstream also specifies a 3.7.1 dependency specifically, at least in the short run, I'd be onboard with vendoring spades 3.7.1 into the formula, adding it to homebrew/versions, or adding it to homebrew/science as a separate formula. The problem with multiple formulae of different versions is that they can have conflicting paths when linked. In this case, currently the only reverse dependency is circlator, so that's not really a big concern.

In terms of vendoring versus a separate formula (regardless of which tap it goes in), the question is whether the spades binaries need to be exposed directly in /usr/local/bin or if that's not important.
```
/usr/local/Cellar/spades/3.9.0/bin/bwa-spades
/usr/local/Cellar/spades/3.9.0/bin/corrector
/usr/local/Cellar/spades/3.9.0/bin/dipspades
/usr/local/Cellar/spades/3.9.0/bin/dipspades.py
/usr/local/Cellar/spades/3.9.0/bin/hammer
/usr/local/Cellar/spades/3.9.0/bin/ionhammer
/usr/local/Cellar/spades/3.9.0/bin/metaspades.py
/usr/local/Cellar/spades/3.9.0/bin/plasmidspades.py
/usr/local/Cellar/spades/3.9.0/bin/rnaspades.py
/usr/local/Cellar/spades/3.9.0/bin/scaffold_correction
/usr/local/Cellar/spades/3.9.0/bin/spades
/usr/local/Cellar/spades/3.9.0/bin/spades.py
/usr/local/Cellar/spades/3.9.0/bin/spades_init.py
/usr/local/Cellar/spades/3.9.0/bin/truspades.py
```

ilovezfs:
@martinghunt https://github.com/Homebrew/homebrew-science/pull/4051.

So you should be able to test it, as follows:
```
brew pull https://github.com/Homebrew/homebrew-science/pull/4050
brew pull https://github.com/Homebrew/homebrew-science/pull/4051
perl -p -i -e 's/"spades"/"spades37"/' /usr/local/Library/Taps/homebrew/homebrew-science/circlator.rb
brew uninstall -f spades circlator
brew install circlator
```
Let me know how that goes :)

ilovezfs:
Make sure to do `brew install -s circlator` to avoid
```
==> Downloading https://homebrew.bintray.com/bottles-science/circlator-1.3.0.el_capitan.bottle.1.tar.

curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "circlator"
Download failed: https://homebrew.bintray.com/bottles-science/circlator-1.3.0.el_capitan.bottle.1.tar.gz
```

circlator.rb
```diff
@@ -1,10 +1,10 @@
...
-  desc "Tool to circularize genome assemblies"
-  homepage "https://github.com/sanger-pathogens/circlator"
-  url "https://github.com/sanger-pathogens/circlator/archive/v0.16.0.tar.gz"
-  sha256 "c5331e1a687dacedd134356177671228e592c772df3306d6e037e55bd62d84df"
-  revision 1
+  include Language::Python::Virtualenv
 
+  desc "Tool to circularize genome assemblies"
+  homepage "https://sanger-pathogens.github.io/circlator/"
+  url "https://files.pythonhosted.org/packages/fe/6b/6d57405c1f3279f0e382c22044c5a19271dbc4e531e730a33899b3791874/circlator-1.3.0.tar.gz"
```
sjackman: Why not use the GitHub url?

sjackman: The GitHub url and the pythonhosted url have different SHA256. How do they differ?
https://github.com/sanger-pathogens/circlator/archive/v1.3.0.tar.gz

GitHub: 0d5219ffa7b2c93fbd8277e6b39ac8e3926cb71b64a06c3f66e3f98a0b5be39d
pythonhosted: fc5cf0bfba57128ab0914d9ed735a3a9288d602edef35201703238cf83009b4b

ilovezfs: We see this for most python formulae in core. In this case, the difference turns out to be primarily be the tests directory.

https://gist.github.com/ilovezfs/ceb86d8dc2ac4c8116829271a3442967

ilovezfs:
@sjackman @tdsmith looks like it's a safe bet to say this is nice and broken:
```
iMac-TMP:homebrew-science joe$ circlator progcheck
  File "/usr/local/bin/circlator", line 57, in <module>
    exec('import circlator.tasks.' + task)
  File "/usr/local/Cellar/circlator/1.3.0/libexec/lib/python3.5/site-packages/circlator/__init__.py", line 26, in <module>
    from circlator import *
  File "/usr/local/Cellar/circlator/1.3.0/libexec/lib/python3.5/site-packages/circlator/bamfilter.py", line 2, in <module>
    import pysam
  File "/usr/local/Cellar/circlator/1.3.0/libexec/lib/python3.5/site-packages/pysam/__init__.py", line 6, in <module>
    from pysam.cutils import *
ImportError: dlopen(/usr/local/Cellar/circlator/1.3.0/libexec/lib/python3.5/site-packages/pysam/cutils.cpython-35m-darwin.so, 2): Library not loaded: @rpath/pysam/libchtslib.so
  Referenced from: /usr/local/Cellar/circlator/1.3.0/libexec/lib/python3.5/site-packages/pysam/cutils.cpython-35m-darwin.so
  Reason: image not found
```

sjackman:
> Thanks for this. Is it possible to specify a spades version? If so, spades 3.7.1 gave me slightly better results with circlator on testing out all versions 3.6.2-3.9.0.

I'm not in favour of vendoring SPAdes 3.7.1 into Circlator. Please contact the upstream SPAdes authors with your experience that 3.7.1 gives better results.

circlator.rb
sjackman: I'd prefer to use the GitHub URL. Please open an issue with upstream regarding the difference in tarballs.

ilovezfs:
@sjackman upstream circlator does specifically say to use spades 3.7.1: https://github.com/sanger-pathogens/circlator/blob/master/install_dependencies.sh#L11

ilovezfs:
Interesting... with the old method body that I'd written for 1.2.0 (https://github.com/Homebrew/homebrew-science/pull/3789#issuecomment-229115405) it works:
https://gist.github.com/ilovezfs/16198603a4a604e3ac714cd2d2238d29

So it might be one of the deps needs to be capped.

sjackman:
Is that a strict dependency (doesn't sound like it) or simply the last version of SPAdes that upstream tested?

sjackman:
> So it might be one of the deps needs to be capped.

What does capped mean? Do you mean pinned to a particular version?

ilovezfs:
Yeah, pinned to a lower version than pip pulled out of the sky.

ilovezfs:
likely it's pysam needs to be 0.8.3 but I'll poke it and find out

ilovezfs:
@sjackman it's not stated, but 3.7.1 was specifically chosen 3 days ago: https://github.com/sanger-pathogens/circlator/commit/a10eb05eb5d9f781423ab326de5e523a89ee25b8

ilovezfs:
@sjackman It's actually in the code hehe:
```
WARNING: SPAdes version 3.9.0 is being used. It will work, but better results are usually obtained from Circlator using SPAdes version 3.7.1. Although 3.7.1 is not the latest version, we recommend it for Circlator.
```

sjackman:
I'm not happy with this situation, but I'm a pragmatic person. Is there an open issue with the upstream developers of SPAdes to discuss this isssue?

ilovezfs:
@sjackman it looks like spades has a support email address not a bug system:

spades.support at bioinf.spbau.ru
http://bioinf.spbau.ru/spades

The justification for 3.7.1 is here:
https://github.com/sanger-pathogens/circlator/pull/70
"recommend spades 3.7.1 (worked the best on NCTC samples from Circlator paper)."

I do wonder if there's some other set of samples where 3.9.0 would work better...

sjackman:
The GitHub repository for SPAdes is https://github.com/ablab/spades

sjackman:
I've created an upstream issue to discuss: https://github.com/sanger-pathogens/circlator/issues/72

ilovezfs:
@sjackman Yup, I :+1:'d it. Thank you for doing that.

circlator.rb
```diff
@@ -17,48 +17,56 @@ class Circlator < Formula
...
+    virtualenv_install_with_resources
 
-    %w[pysam pyfastaq bio_assembly_refinement].each do |r|
-      resource(r).stage do
-        system "python3", *Language::Python.setup_install_args(libexec/"vendor")
-      end
+    if OS.mac?
+      site_packages = libexec/"lib/python3.5/site-packages"
+      libchtslib = site_packages/"pysam/libchtslib.cpython-35m-darwin.so"
+      (site_packages/"pysam").install_symlink libchtslib => "libchtslib.so"
```
ilovezfs: @sjackman @tdsmith This fixes the breakage with the current pysam (pysam-0.9.1.4). The alternative would be to pin pysam to 0.8.4. I'm not sure whose fault this problem is at this point.

```
	@rpath/pysam/libchtslib.so (compatibility version 0.0.0, current version 0.0.0)
```
But we've installed `libchtslib.cpython-35m-darwin.so` not `libchtslib.so`.

```
iMac-TMP:~ joe$ otool -L /usr/local/Cellar/circlator/1.3.0/libexec/lib/python3.5/site-packages/pysam/cutils.cpython-35m-darwin.so
/usr/local/Cellar/circlator/1.3.0/libexec/lib/python3.5/site-packages/pysam/cutils.cpython-35m-darwin.so:
	/usr/local/opt/circlator/libexec/lib/python3.5/site-packages/pysam/cutils.so (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	/usr/lib/libcurl.4.dylib (compatibility version 7.0.0, current version 8.0.0)
	@rpath/pysam/libchtslib.so (compatibility version 0.0.0, current version 0.0.0)
```

sjackman: > I'm not sure whose fault this problem is at this point.

Ask upstream if pysam 0.9.14 is supported?

sjackman: I get the following error when building this formula on Linux with pysam-0.9.1.4:
```
ValueError: dictionary update sequence element #0 has length 3; 2 is required
```

sjackman: I'm circlator doesn't work with pysam 0.9.14, I'm fine with pinning it to 0.8.4.

ilovezfs: Seems likely not: https://github.com/sanger-pathogens/circlator/blob/master/setup.py#L23

But the bug is most likely pysam's: https://github.com/pysam-developers/pysam/releases/tag/v0.9.1.4
since they seem to be dancing around htslib issues currently.

Let's see if 0.8.4 breaks on Linux too.

martinghunt:
Pysam 0.9.1.4 should be ok (but whatever works). The circlator travis build used 0.9.1.4 and passed ok.

circlator.rb
ilovezfs: Pretty sure it's entirely deliberate given how those tests require the assumptions fulfilled by https://github.com/sanger-pathogens/circlator/blob/master/install_dependencies.sh, which isn't what the pip-based install uses at all.

ilovezfs:
@martinghunt hmm so the pysam 0.9.1.4 problem may hae something to do with how homebrew does the install.

@martinghunt can you shed any light on the difference between your pypi tarball and the GitHub source archive tarball?

ilovezfs:
@sjackman well that's some progress: now Linux dies with `circlator.external_progs.Error: Found spades but couldn't get version.` in the test.

sjackman:
Martin (upstream author @martinghunt) says that using SPAdes 0.9.0 is fine.
https://github.com/sanger-pathogens/circlator/issues/72#issuecomment-239632498
> If it's a pain, go with 3.9.0.

ilovezfs:
@sjackman yeah, that's a relief. So do you think the Linux fail is problem with the Linux spade's bottle? Seems weird OS X one wouldn't have the same trouble with the version retrieval.

sjackman:
I get the following on Linux. What do you get?
```
$ spades.py --version
SPAdes v3.9.0
```

ilovezfs:
The same
```
iMac-TMP:~ joe$ spades.py --version
SPAdes v3.9.0
```

sjackman:
The regex used by circlator is `^SPAdes genome assembler v.?([0-9][0-9\.]+`, which doesn't agree with `SPAdes v3.9.0`.
https://github.com/sanger-pathogens/circlator/blob/master/circlator/external_progs.py#L22

ilovezfs:
The command it passes is '' not '-v', and then it does match.

ilovezfs:
What's output when you run
```
$ spades.py ''
```

sjackman:
```
$ spades.py
SPAdes genome assembler v3.9.0
```

ilovezfs:
Yeah, so that should match ...

ilovezfs:
Does the test fail for you locally?

ilovezfs:
@sjackman yikes
```
==> brew test homebrew/science/circlator --verbose

Testing homebrew/science/circlator

==> which spades.py

.

==> spades.py

.

Last 150 lines from /home/linuxbrew/.cache/Homebrew/Logs/circlator/02.spades.py:

2016-08-13 18:57:00 +0000

spades.py

/usr/bin/env: python: No such file or directory

Error: homebrew/science/circlator: failed

Failed executing: spades.py 

/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1508:in `block in system'

/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1445:in `open'

/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1445:in `system'

/home/linuxbrew/.linuxbrew/Library/Taps/homebrew/homebrew-science/circlator.rb:61:in `block in <class:Circlator>'

/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1350:in `block in run_test'

/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:14:in `block in mktemp'

/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:75:in `block in run'

/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:75:in `chdir'

/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:75:in `run'

/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:13:in `mktemp'

/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1344:in `run_test'

/home/linuxbrew/.linuxbrew/Library/Homebrew/test.rb:33:in `block in <main>'

/usr/lib/ruby/2.0.0/timeout.rb:66:in `timeout'

/home/linuxbrew/.linuxbrew/Library/Homebrew/test.rb:32:in `<main>'
```

sjackman:
Fixable with `depends_on :python` ?

tdsmith:
If distutils isn't rewriting the script's shebang on installation, we can propose using the newer setuptools entry_point feature, but I thought it should have!

ilovezfs:
Adding `:python` to spades was sufficient to fix this too. Thanks for coming up with the fix @sjackman.

ilovezfs:
@sjackman shipped if you want to #ProduceThePenguinProduct now

@martinghunt 1.3.0 is :ship:'d :)

sjackman:
Thanks, Joe! I've merged the bottle for Linuxbrew.

martinghunt:
Thanks all, you are awesome :)

ilovezfs:
@sjackman you're welcome :)

@martinghunt you're welcome! thanks for being such an engaged upstream! Is it your belief now that SPAdes 3.9.0 is at least as good as 3.7.2 after all? If so, would a minor patch release removing that warning be warranted so people don't go on wild goose chases for the older version?

martinghunt:
Put it this way: internally for our assembly pipeline, I'm sticking with 3.7.1. But I agree that the pragmatic solution for homebrew is to just use the latest spades.

Are you suggesting a patch just for homebrew? Or asking that I remove the warning completely? I'd like the warning in for anyone who installed via pip or github. Circlator can also find spades via an environment variable, so it's just a case of getting 3.7.1 and doing: export CIRCLATOR_SPADES=/foo/bar/bin/spades.py. I'm open to suggestions, as obviously those warnings don't look good when installed via homebrew.

sjackman:
We could either disable the warning, or add an explanation to `caveats`.

PR #4051, opened 2016-08-13
spades37 3.7.1 (new formula)

ilovezfs:
circlator, at least as of version 1.3.0, gives better results with
spades 3.7.1.

See https://github.com/Homebrew/homebrew-science/pull/4050.

### DIFF
--- a/spades37.rb
+++ b/spades37.rb
@@ -1,0 +1,34 @@
+class Spades37 < Formula
+  desc "SPAdes: de novo genome assembly"
+  homepage "http://bioinf.spbau.ru/spades/"
+  # tag "bioinformatics"
+  # doi "10.1089/cmb.2012.0021"
+  url "http://spades.bioinf.spbau.ru/release3.7.1/SPAdes-3.7.1.tar.gz"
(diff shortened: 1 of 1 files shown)


sjackman:
I would prefer not to accept this formula and instead work on the core Brew functionality to install arbitrary older versions of bottles. See https://github.com/Homebrew/brew-evolution/pull/6

ilovezfs:
Yeah, that's pure pie in the sky to me at this point, but I'm happy if you're happy ;)

sjackman:
Closing unless there's overwhelming support for this PR from other @Homebrew/science

sjackman:
I spoke and acted too soon. See https://github.com/Homebrew/homebrew-science/pull/4050#issuecomment-239625864

ilovezfs:
@sjackman based on https://github.com/sanger-pathogens/circlator/issues/72#issuecomment-239632498 it looks like your initial instinct to close this was correct.

sjackman:
Martin (upstream author @martinghunt) says that using SPAdes 0.9.0 is fine.
https://github.com/sanger-pathogens/circlator/issues/72#issuecomment-239632498
> If it's a pain, go with 3.9.0.

PR #4054, opened 2016-08-13
nanopolish 0.5.0 [WIP; do not merge]

sjackman:
wget is used to download eigen. I would prefer that nanopolish use
the eigen dependency installed by brew.


sjackman:
There's two issues with this PR.
+ It downloads and compiles `eigen` rather than using the brewed version.
+ It fails to build with GCC 6, which is reported upstream: https://github.com/jts/nanopolish/issues/57

ilovezfs:
@sjackman both issues seem to be resolved now in what I just shipped: https://github.com/Homebrew/homebrew-science/pull/4270

sjackman:
Superseded by #4270.

PR #4056, opened 2016-08-14
spades: remove python patch

ilovezfs:
- depend on gcc
- depend on :python if Linux

### DIFF
--- a/spades.rb
+++ b/spades.rb
@@ -15,6 +15,8 @@ class Spades < Formula
   end
 
   depends_on "cmake" => :build
+  depends_on "gcc"
+  depends_on :python if OS.linux?
 
   needs :openmp
 
@@ -22,16 +24,10 @@ class Spades < Formula
     cause "Compiling SPAdes requires GCC >= 4.7 for OpenMP 3.1 support"
   end
 
-
   def install
     mkdir "src/build" do
       system "cmake", "..", *std_cmake_args
       system "make", "install"
-    end
-
-    # Fix audit error "Non-executables were installed to bin"
-    inreplace bin/"spades_init.py" do |s|
-      s.sub! /^/, "#!/usr/bin/env python\n"
     end
   end
 


ilovezfs:
CC @tdsmith RE: https://github.com/Homebrew/homebrew-science/pull/4050#issuecomment-239671538 What change do you want here for spades?

tdsmith:
Probably none?

ilovezfs:
to get the test to pass for spades on Linux, it looks like the needed change was what @sjackman had suggested, namely adding the `depends_on :python`. The questions are a) whether that's an appropriate change for this formula b) if so, whether it should be `depends_on :python3` instead and c) whether this will be sufficient to also get the circlator test to pass on Linux or if an additional change is needed over there.

If you're cool with this as is, I guess it can be merged and then we can retest circlator without additional changes to see what happens then on Travis for Linux (presumably after @sjackman also updates the Linux bottle for spades), unless anyone has a better idea for how to test this interplay without committing to master as experimentation :)

sjackman:
:+1: Seems good to me.

tdsmith:
On Sun, Aug 14, 2016, 08:34 ilovezfs <notifications@github.com> wrote:

> to get the test to pass for spades on Linux, it looks like the needed
> change was what @sjackman <https://github.com/sjackman> had suggested,
> namely adding the depends_on :python. The questions are a) whether that's
> an appropriate change for this formula b) if so, whether it should be depends_on
> :python3 instead and c) whether this will be sufficient to also get the
> circlator test to pass on Linux or if an additional change is needed over
> there.
>
There's probably a more "correct" solution but if it works, :yolo:

>

sjackman:
```ruby
depends_on :python if OS.linux?
```

SPAdes depend on Python. Why isn't this simply
```ruby
depends_on :python
```

ilovezfs:
At least on macOS, if you're bottling as opposed to just building with --build-from-source, `depends_on :python` forces use of Homebrew python rather than allowing system python, which is always available and doesn't need an explicit `depends_on :python`.

ilovezfs:
@sjackman :ship:'d if you want to synthesize the Linux bottle in your magical laboratory now.

sjackman:
Oh. Interesting. So is `depends_on :python` only used when the formula includes a Python module that compiles C source to a native library?

tdsmith:
On OS X we try to use it only when we're building something that embeds an interpreter. Modules built against any (narrow Unicode) 2.7.x build are ABI-compatible with any other (narrow Unicode) 2.7.x build, so we tend to build against system python for those.

sjackman:
So the 44 formula in Homebrew-science that `depends_on :python` should probably be
```sh
depends_on :python => :recommended unless OS.mac?
```

tdsmith:
I should say "modules built _correctly_ against any 2.7.x python" because [people get it wrong](http://blog.tim-smith.us/2015/09/python-extension-modules-os-x/) so I wouldn't expect the transition to be seamless.

But, yes, that's the "best" outcome insofar as it doesn't insist on installing Homebrew's python for people on OS X. I don't think -science gets too many complaints about that, though, so maybe a few "extra" Homebrew python installs are not worth worrying about. Most of the "too many Pythons" problems are actually related to the linking issue I describe in the link above. Depending on `:python` is I think _semantically_ correct even though I don't like its interaction with bottles.

Anyway OS X system python kinda sucks insofar as its native TLS support depends on an antique OpenSSL that is becoming inappropriate on the modern Internet and I'm hearing rumors that it might disappear after Sierra so it's probably not worth a mass conversion since the tide might swing back.

ilovezfs:
It's probably also worth mentioning that this is why you'll sometimes see those weird

```
option "without-python", "Build without python support"
```
in homebrew/core that are not accompanied by any corresponding

```
depends_on :python => :recommended
```
(other than potentially an inert `depends_on :python => :recommended if MacOS.version <= :snow_leopard`)

For example,
https://github.com/Homebrew/homebrew-core/blob/master/Formula/protobuf.rb#L46-L47

ilovezfs:
>I'm hearing rumors that it might disappear after Sierra

because that won't break anything ... :see_no_evil:

PR #4058, opened 2016-08-15
Update salmon to v0.7.0

rob-p:
### Have you:

### New formula: have you

cc @sjackman

### DIFF
--- a/salmon.rb
+++ b/salmon.rb
@@ -3,16 +3,10 @@ class Salmon < Formula
   homepage "https://github.com/COMBINE-lab/salmon"
   # tag "bioinformatics"
 
-  url "https://github.com/COMBINE-lab/salmon/archive/v0.6.0.tar.gz"
-  sha256 "2a015c0f95b745fbed575d8610aea7e09cb1af55ca2c68e10ab15826fba263b1"
-  revision 2
+  url "https://github.com/COMBINE-lab/salmon/archive/v0.7.0.tar.gz"
+  sha256 "b0c30941814760095d6d302f7c3c74be0fc098c2851ff0fd58dd171ed1f8b4d8"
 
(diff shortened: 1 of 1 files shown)


rob-p:
There seems to be a problem with (and only with Mavericks), though I feel like that is Mavericks' fault (has to do with non-standard compliant behavior about `max_align_t`, I think.

sjackman:
The code that fails is a hack for `gcc`, so it could be disabled for `clang.`
```
/tmp/salmon-20160815-56369-11v446m/salmon-0.7.0/include/concurrentqueue.h:241:10: error: no type named 'max_align_t' in the global namespace; did you mean 'std::max_align_t'?
        typedef ::max_align_t max_align_t;      // GCC forgot to add it to std:: for a while
                ^~~~~~~~~~~~~
                std::max_align_t
```
https://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/5494/version=mavericks/testReport/junit/brew-test-bot/mavericks/install_homebrew_science_salmon/

I'm happy merging this as is. We can fix Mavericks later if you're inclined.

sjackman:
The preferred git commit log message is simply `salmon 0.7.0` to make search the git history for particular versions of software easier.

rob-p:
I think that the problem is that clang (on mavericks only) defines the GCC symbol, and so falls into the path that it shouldn't.  The issue is that this is coming from the [concurrent queue](https://github.com/cameron314/concurrentqueue) library, which isn't maintained by me.

rob-p:
Also can I "re-write" the history or some other git magic to change the commit message?

sjackman:
You can fix up the commit log message with `git commit --amend`
I'll do that when I pull the PR, right now.

sjackman:
Merged without a bottle for Mavericks (see above).

PR #4059, opened 2016-08-15
pocl: disable bottles

ilovezfs:
The latest pocl (0.13) fails to build with the current llvm formula
(3.8.1) as well as with the prior (3.8.0):

```
error "isfinite is not a macro"
```

pocl 0.12 is not compatible with llvm 3.8*, but it does build with our
prior major llvm version, which was llvm (3.6.2), but this version is no
longer available in core and the version in homebrew/dupes differs.
Also, despite the successful build, there is still a test failure in
`make check`:

```
 16: Kernel functions length, distance, and normalize FAILED
(testsuite.at:174)
```

pocl 0.11 also builds with llvm 3.6.2, and `make check` succeeds, but
this may simply be because the test that failed for 0.12 appears to be
new.

### DIFF
--- a/pocl.rb
+++ b/pocl.rb
@@ -3,10 +3,7 @@ class Pocl < Formula
   url "https://downloads.sourceforge.net/project/pocl/pocl-0.10.tar.gz"
   sha256 "e9c38f774a77e61f66d850b705a5ba42d49356c40e75733db4c4811e091e5088"
 
-  bottle do
-    sha256 "b79e64fe8bfeb42cc7a931fb67f6f02976f6d30798f4af928a32ee35eaa527d2" => :yosemite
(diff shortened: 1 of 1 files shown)


ilovezfs:
Build issue reported upstream here: https://github.com/pocl/pocl/issues/365.

ilovezfs:
Above issue is fixed in HEAD (tested at a62d357d94000fb23a306e1093e0417a9e4d3df1), however `make check` had 84/107  tests fail:
https://gist.github.com/ilovezfs/e44fbbfdf1c4a02433c5289b88c248f3

gist-logs:
https://gist.github.com/edab77acb8e3292ba64fa97b06d6c5e3

PR #4060, opened 2016-08-15
gmt: bump gshhg version to 2.3.5, University of Hawaii and mirrors no…

LuminousRubyist:
### Have you:

- [x] Added revision number due to updated dependency
- [x] Wrote a test, there was none previously


LuminousRubyist:
Travis Cl build fails with:
```
Error: No available formula with the name "homebrew/x11/gv" 
Please tap it and then try again: brew tap homebrew/x11
```

First, I will go ahead and admit my inexperience with Homebrew in general. The `gmt.rb` file does not have a `test do` block. I've gone ahead and written a simple one along the lines of 
```ruby
test do
  assert_match "5.2.1", shell_output("#{bin}/gmt --version",0)
end
```
Which passes on my local machine. However, I'm not sure if this would be ideal, and it doesn't have much to do with incrementing the gshhg version number, so I'm keeping it local.

I'm not sure if the issue is with some operating systems having the GV software installed by default while others do not, or if the gv dependency should be made explicit in the formula or removed entirely. Any help answers would be welcome.

dpo:
I'm not sure why Travis is having an issue. The X11 dependency looks fine and the build bot is happy. Would you mind bumping the revision number of `gmt`?

It would be great to add a test. The one you suggest will break on upgrades. Is it possible to run an executable on some sample data?

ilovezfs:
@dpo I think the problem is just that the .travis.yml needs to tap the x11 repo too, like it does with the python tap

dpo:
@ilovezfs Ah, I guess I'm always expecting `depends_on :x11` to know to tap the x11 repo. Any reason why it isn't the case?

ilovezfs:
@dpo the main reason I think is just that in homebrew/core non-optional, cross-tap dependencies are verboten. I added manual tapping at the beginning of each run of homebrew/python, homebrew/x11, and homebrew/versions to the Jenkins setup for science so that things would actually get tested as opposed to just give up like you're seeing on Travis. The corresponding change should be made for Travis. We may want to do homebrew/dupes too for the sake of astral, plink2, pymol, samtools-0.1, samtools, vcftools, and vsearch.

dpo:
Ok. See #4068.

LuminousRubyist:
I went ahead and wrote some test code and bumped `gmt`'s revision number as @dpo recommended. Travis still breaks on the cross-tap x11 issue. Waiting on #4068.

dpo:
Thanks! I just need to rebuild the bottles.

dpo:
@BrewTestBot test this please

PR #4062, opened 2016-08-15
oce 0.17.2

ilovezfs:


ilovezfs:
@sjackman the solution to the dealii problem I believe is going to be to use this, as the upstream documentation says oce is what's expected, and the latest version of opencascade is missing some headers that dealii expects to find.

PR #4063, opened 2016-08-15
dealii: switch to oce

ilovezfs:


ilovezfs:
@sjackman @dpo @davydden any objections/concerns with this change?

dealii.rb
```diff
@@ -86,7 +87,7 @@ def install
     args << "-DMETIS_DIR=#{Formula["metis"].opt_prefix}" if build.with? "metis"
     args << "-DMUPARSER_DIR=#{Formula["muparser"].opt_prefix}" if build.with? "muparser"
     args << "-DNETCDF_DIR=#{Formula["netcdf"].opt_prefix}" if build.with? "netcdf"
-    args << "-DOPENCASCADE_DIR=#{Formula["opencascade"].opt_prefix}" if build.with? "opencascade"
```
davydden: should stay `-DOPENCASCADE_DIR` with prefix of `oce`. Otherwise LGTM.

ilovezfs: @davydden FYI the reason for the change is that the opencascade 7.0.0 release is missing some headers dealii expects to be there, and similar such issues with some symbols renames, etc. Is dealii compatibility with opencascade 7.0.0 on the horizon, or is that not planned currently?

davydden: i was not aware of this, i created an issue to further track this https://github.com/dealii/dealii/issues/2983

ilovezfs: Thanks!

By the way, not sure if you'd noticed my string replacement patch in the formula (lines 49-54)
```
    # PETSc 3.7.x added a parameter to PetscOptionsSetValue()
    # https://bitbucket.org/petsc/petsc/src/5d547b27bccc01eacb9fc0eef6ae71e85dce2b0c/src/sys/objects/options.c?at=master&fileviewer=file-view-default#options.c-1078
    # See upstream PR: https://github.com/dealii/dealii/pull/2327
    inreplace "source/lac/petsc_precondition.cc",
      "PetscOptionsSetValue(\"",
      "PetscOptionsSetValue(NULL,\""
```

I just noticed that https://github.com/dealii/dealii/pull/2327 was superseded by https://github.com/dealii/dealii/pull/2754. Any chance of getting a new release tag, so that we can be inline with your chosen solution? Alternatively, I can attempt to apply https://github.com/dealii/dealii/commit/0ee2a57ce6679ef94894f7efa464b4d5b808b42b as a patch in the interim, or just leave the formula alone with the string replacement until the next release.

davydden: I saw that fix. I suppose in the homebrew context the easiest solution is keep it as is until 8.5 release (later this year). PETSc is updated in Homebrew to 3.7 and thereby there is no need to support compatibility with older PETSc's.

ilovezfs: Sounds like a plan. You may be interested in tracking https://github.com/Homebrew/brew/issues/620 by the way, as I know you'd expressed concern about Homebrew's version support more generally.

ilovezfs:
Closed via https://github.com/Homebrew/homebrew-science/commit/f765e094663fab20b622c06324e441a373d5827a

PR #4068, opened 2016-08-16
travis: tap python, x11 and dupes

dpo:


ilovezfs:
I'd do homebrew/versions too, but that LGTM other than that.

.travis.yml
```diff
@@ -31,9 +31,10 @@ before_install:
       docker run
         -v $(pwd):/home/linuxbrew/homebrew-science
         linuxbrew/linuxbrew
-        /bin/sh -c 'brew tap homebrew/python && brew update && cp -rf ~/homebrew-science $(brew --repo)/Library/Taps/homebrew/homebrew-science';
+        /bin/sh -c 'brew update && mkdir -p $(brew --repo)/Library/Taps/homebrew && cp -rf ~/homebrew-science $(brew --repo)/Library/Taps/homebrew/homebrew-science';
```
ilovezfs: I'm not familiar enough with docker to say whether it's valid to move the tapping out of here since presumably it's trying to be done before the `docker commit` below?

ilovezfs:
@dpo I think I wouldn't worry about duplication, and just put it below ` brew update || brew update;` for OS X (i.e., under line 23), and inline it where python was for Linux (i.e., as part of line 34).

dpo:
@sjackman Is linuxbrew [ignoring `HOMEBREW_DEVELOPER=1`](https://travis-ci.org/Homebrew/homebrew-science/jobs/152618359#L331)? Perhaps it should be `"1"`?

ilovezfs:
@dpo homebrew/versions too I think:
```
Josephs-MacBook-Pro:homebrew-science joe$ ag '\"homebrew/versions'|grep -v optional
arrayfire.rb:34:  depends_on "homebrew/versions/glfw3"
cufflinks.rb:15:    depends_on "homebrew/versions/boost155"
gtsam.rb:20:  depends_on "homebrew/versions/boost155"
nco.rb:27:  depends_on "homebrew/versions/antlr2"
nexusformat.rb:22:  depends_on "homebrew/versions/hdf4" => :recommended
pocl.rb:16:    depends_on "homebrew/versions/llvm34"
shark.rb:14:  depends_on "homebrew/versions/boost160"
```

dpo:
@ilovezfs Sorry, done now.

ilovezfs:
cool. :shipit: and let's re-run the one that was failing to see if it worked out as expected.

sjackman:
I think it's a bug that the required taps aren't autotapped, but this solution is pragmatic, and it's what I've done in other repos.

ilovezfs:
>I think it's a bug that the required taps aren't autotapped

If it's a bug, it's a very intentional bug. I'm not sure what to call those but we have enough of them around here that they probably deserve their own term of art. :see_no_evil:

PR #4076, opened 2016-08-20
rna-star 2.5.2b

ilovezfs:
- require `cxx11` as prescribed by the release notes
- install `STARlong` also
- update names of the macOS make targets
- install the `extras` directory in `pkgshare`
- allow `prefix.install_metafiles` to handle the standard docs
- add `STARlong` to the test block

### DIFF
--- a/rna-star.rb
+++ b/rna-star.rb
@@ -4,9 +4,9 @@ class RnaStar < Formula
   # doi "10.1093/bioinformatics/bts635"
   # tag "bioinformatics"
 
-  url "https://github.com/alexdobin/STAR/archive/STAR_2.4.1d.tar.gz"
-  sha256 "1c895fc4111798d62a0497c5746f3ae25fdd3f41a0afc4becd90959de183cd5b"
-
+  url "https://github.com/alexdobin/STAR/archive/2.5.2b.tar.gz"
+  version "2.5.2b"
+  sha256 "f88b992740807ab10f2ac3b83781bf56951617f210001fab523f6480d0b546d9"
   head "https://github.com/alexdobin/STAR.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


ilovezfs:

also, it works to build against the htslib in homebrew/science using symlinks if that's preferable to the vendored one

rna-star.rb
```diff
@@ -20,16 +20,23 @@ class RnaStar < Formula
...
+  needs :cxx11
+
   def install
+    ENV.cxx11
     cd "source" do
-      system "make", "htslib"
-      system "make", OS.mac? ? "STARforMac" : "STAR"
-      bin.install "STAR"
+      progs = %w[STAR STARlong]
+      targets = OS.mac? ? %w[STARforMacStatic STARlongForMacStatic] : progs
```
sjackman: Is there an option for dynamic linking rather than static linking? I usually prefer dynamic, given the choice.

rna-star.rb
```diff
@@ -20,16 +20,23 @@ class RnaStar < Formula
...
+      system "make", *targets
+      bin.install progs
     end
-    doc.install "CHANGES", "LICENSE", "README", "RELEASEnotes", "doc/STARmanual.pdf"
+    pkgshare.install "extras"
+    doc.install (buildpath/"doc").children
+    mv "RELEASEnotes.md", "NOTES.md"
```
sjackman: Does this line affect  the installation? It looks like it just renames a file in the build directory?

sjackman:
A couple of questions. Looks good otherwise.

rna-star.rb
ilovezfs: @sjackman the Linux target is shared by default, but it appears they killed off the macOS shared option: https://github.com/alexdobin/STAR/blob/master/source/Makefile

rna-star.rb
ilovezfs: Yes, this causes it to be picked up automatically by `prefix.install_metafiles`: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/metafiles.rb#L3-L10

rna-star.rb
sjackman: `LDFLAGS_Mac_static` adds only `-static-libgcc`, so it seems that it's not actually a static executable, just statically linked to `libgcc`.

rna-star.rb
sjackman: I don't see `prefix.install_metafiles` in the formula. Is it run automatically now? That's pretty cool.

rna-star.rb
ilovezfs: I think it is since it's packing in their libhts.a too.

rna-star.rb
ilovezfs: Yeah, that's why you see me sometimes remove the `doc.install ...` stuff that's in science since it shouldn't be there if it's one of the magic metafiles:

https://github.com/Homebrew/brew/blob/bf40dd64dff1fc86b0ecbde33dc9e3be3ef3fccb/Library/Homebrew/build.rb#L137-L139

tseemann:
What i find weird is why all the bugs listed at https://github.com/alexdobin/STAR/releases only warranted going from `2.5.2a` to `2.5.2b` !!!

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@sjackman :penguin:

sjackman:
> `@BrewTestBot test this please`
I need to get this command to trigger a rebuild of the Linuxbrew bottle as well.

PR #4103, opened 2016-08-24
arpack 3.4.0 with cmake

dpo:


dpo:
Actually, maybe the dependents don't need to be bumped. They appear to be linked against `libarpack.2.dylib`.

ilovezfs:
@dpo That would be pleasant! BTW it looks like a CMake-based build is now available if you feel like converting it over.

dpo:
I'll have a look.

dpo:
Switching to the CMake build requires bumping the dependents.

dpo:
The pkgconfig info file seems to be missing when installing with CMake: https://github.com/opencollab/arpack-ng/issues/44. That breaks `calculix-cxx` and `shogun`.

PR #4136, opened 2016-08-25
sundials: revision for open-mpi

ilovezfs:


ilovezfs:
@BrewTestBot test this please

dpo:
I just noticed you don't need to fill out a form to download sundials any more! So now we can update. I DO see a PR in my future!

ilovezfs:
@dpo notice I switched this over to the new version...but still build mysteriously fails on that example though I'm not seeing that locally...

ilovezfs:
oh there's a 2.6 series too? oy...

ilovezfs:
Not sure what's up with cvsHessian_ASA_FSA but it's the criminal.

PR #4121, opened 2016-08-25
getdp: revision for open-mpi

ilovezfs:
and pass `with-mpi` to the `arpack` and `hdf5` deps if building `getdp` `with-mpi`

PR #4137, opened 2016-08-25
superlu_dist: revision for open-mpi

ilovezfs:


DomT4:
Killed the build here because it completed within 20 minutes on Mavericks, 24 on El Cap but was still trucking at 1hr32m for Yosemite 😖.

ilovezfs:
@DomT4 and this time it was fine. :see_no_evil:

DomT4:
Ho hum, maybe Yosemite's Clang decided to briefly go on strike last night 😄.

ilovezfs:
that makes about as much sense as any other explanation I think

DomT4:
Most CI troubleshooting ends in a mildly resigned shrug 🙈.

ilovezfs:
@DomT4 bonus points if you can identify the cause of this failure https://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/5625/version=el_capitan/console without cheating (i.e., without looking at the change I made to the PR subsequently).

DomT4:
Oh the old:
```
==> Installing homebrew/science/getdp dependency: arpack
/usr/bin/sandbox-exec -f /tmp/homebrew20160826-23459-1kly7q6.sb nice /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/build.rb /usr/local/Library/Taps/homebrew/homebrew-science/arpack.rb --verbose --with-mpi
Error: /usr/local/opt/autoconf not present or broken
Please reinstall autoconf. Sorry :(
```
bug, I presume? https://github.com/Homebrew/legacy-homebrew/issues/42312

ilovezfs:
@DomT4 nope! https://github.com/Homebrew/homebrew-science/pull/4121

getdp had
```
depends_on :mpi => [:cc, :cxx, :f90, :recommended]
 depends_on "arpack" => :recommended
```

and arpack had
```
depends_on :mpi => [:optional, :f77]
```

So getdp asks for plain vanilla arpack in the formula, but when the rubber meets the road it's actually asking for arpack with-mpi, which means it has to build arpack from source, and that needs autoconf. So it ends up attempting a source build without having installed the build-time deps.

See also
https://github.com/Homebrew/homebrew-core/issues/3711#issuecomment-238617652
https://github.com/Homebrew/homebrew-core/issues/3711#issuecomment-239663737

for similar but not identical situations. #:bug:

dpo:
Good catch. That's why we have tons of ugly constructs, e.g., [in `trilinos.rb`](https://github.com/Homebrew/homebrew-science/blob/master/trilinos.rb#L44) to propagate BLAS and MPI settings, and attempts like https://github.com/Homebrew/homebrew-science/pull/2761. I'm not sure what to solution to this problem is but it deserves some thought.

PR #4140, opened 2016-08-26
fplll 5.0.1 (new formula)

tlepoint:
### Have you:

### New formula: have you

Add formula for fplll (lattice algorithms using floating-point arithmetic)


fplll.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      int main()
+      {
+        fplll::ZZ_mat<mpz_t> A;
+        A.resize(50, 50 + 1); // construct 50 × (50+1) integer relations matrix
+        A.gen_intrel(1000);   // generate 1000-bit coefficients
+        return fplll::lll_reduction(A, LLL_DEF_DELTA, LLL_DEF_ETA, LM_PROVED, FT_MPFR, 0, LLL_DEFAULT);
+      }
+    EOS
+    system ENV.cxx, "test.cpp", "-std=c++11", "-lgmp", "-lmpfr", "-lfplll", "-o", "test"
```
dpo: For better abstraction, you can call `ENV.cxx11` in `test`. You might have to do something [like this](https://github.com/Homebrew/homebrew-science/blob/master/adol-c.rb#L89) to compile your example code.

ilovezfs: I don't think that works.

dpo: It works in `adol-c`:
```
$ brew test -v adol-c
==> /usr/bin/clang++ -std=c++11 -stdlib=libc++ test.cpp -I/usr/local/opt/adol-c/include -o test -L/usr/local/opt/adol-c/lib -ladolc -L/usr/local/opt/colpack/lib -lColPack
```

ilovezfs: It didn't want to behave for me in https://github.com/Homebrew/homebrew-core/pull/3499

dpo: I learned about that trick over at https://github.com/Homebrew/homebrew-science/pull/3178#issuecomment-170177606. Don't forget the `split` thing.

ilovezfs: I was just trying to add it to the top of the test.
```
test do
  ENV.cxx11
  blah blah
```
etc.

Peculiar approach. I think that's mostly not used at all in core but maybe Martin wants them all changed :see_no_evil:

dpo: It's also used in `cppad`, `mlpack` and `sdsl-lite`.

ilovezfs: @dpo https://gist.github.com/ilovezfs/fbc6512f6f06ef3df11c84bd85f908f3

I see a PR in your future :)

dpo: @ilovezfs :) I know what you mean but look at https://github.com/Homebrew/brew/blob/498e81c9fcd3bfef4a14eb0fddce51af892b1e9d/Library/Homebrew/extend/ENV/std.rb#L183. If I install a formula with a specific compiler (say, gcc instead of clang), I want to run the test with the same compiler options. In this case, they're not quite the same for clang and gcc. I see it as the same thing as calling `ENV.fortran` in a test that requires gfortran. There was a related discussion recently about selecting an OpenMP-capable compiler in `test`, which Homebrew currently doesn't support.

ilovezfs: Mainly it just needs to work via test-bot in the CI environment to fulfill its purpose.

dpo: I find it also good if the user is able to run `test` correctly, even with non-default compiler, etc.

ilovezfs: Yup, but I'll bet if you look at the analytics "no one" ever actually does that except us lol

dpo: You're probably right. It's good to have extra tricks up our sleeves! :)

ilovezfs: I agree. It would be nice if there was a DSL for this instead of the crazy splitting thing.

dpo: Yes. I would also love `ENV.openmp` but that's another story.

PR #4148, opened 2016-08-27
open-mpi: migrate from core

ilovezfs:
Goes together with https://github.com/Homebrew/homebrew-core/pull/4274 and https://github.com/Homebrew/brew/pull/828.

### DIFF
--- a/bsponmpi.rb
+++ b/bsponmpi.rb
@@ -1,0 +1,17 @@
+class Bsponmpi < Formula
+  desc "Implements the BSPlib standard on top of MPI"
+  homepage "https://sourceforge.net/projects/bsponmpi/"
+  url "https://downloads.sourceforge.net/project/bsponmpi/bsponmpi/0.3/bsponmpi-0.3.tar.gz"
+  sha256 "bc90ca22155be9ff65aca4e964d8cd0bef5f0facef0a42bc1db8b9f822c92a90"
+
+  depends_on "scons" => :build
+  depends_on :mpi => [:cc, :cxx]
+
+  def install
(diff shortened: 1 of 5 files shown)


dpo:
What's the rationale? What about `mpich` and the `:mpi` dependency?

ilovezfs:
@dpo recheck the PR ;)

dpo:
Thanks. Is core happy for dependencies of core formulae to live in another tap? Any chance we could have real `:`-style requirements in science?

ilovezfs:
@dpo as far as I know the "default_formula" mechanism won't break just because the formula is in a tap, but obviously that will need to be tested.

ilovezfs:
@dpo see https://github.com/Homebrew/brew/pull/828.

ilovezfs:
@dpo the rationale is that the revisions for
```
abinit
abyss
alpscore
bless
dealii
dida
elemental
exabayes
garli
getdp
gmsh
hypre
lammps
lis
meme
mumps
p4est
parmetis
pastix
petsc
ray
scalapack
scotch
scotch5
slepc
sundials
superlu_dist
trilinos
zoltan
```
all should have been part of the single open-mpi update PR (just as we do for llvm, gcc, boost, etc. in core), but weren't solely because open-mpi is in another tap, thereby breaking all of them in one swoop, and leaving everyone in the dark about whether they're going to build or not and leaving science to scramble to fix all the broken bottles.

dpo:
At the very least, `zoltan` should have been marked as conflicting with `trilinos`.

PR #4150, opened 2016-08-27
jsbsim 1.0.rc2

slnovak:
This PR adds a formula for [JSBSim](http://jsbsim.sourceforge.net/).


tseemann:
@slnovak Thanks Stefan - it compiled cleanly on Linux too!

PR #4156, opened 2016-08-28
cfitsio: migrate from core

zmwangx:
Justification in https://github.com/Homebrew/homebrew-core/pull/4313 so I'm not going to repeat myself.


ilovezfs:
@zmwangx bottle blocks should be removed when the formula is initially imported from elsewhere because the links they point to are no longer valid (url goes by tap). Then bot does the bottle-add commit.

zmwangx:
Fixed.

PR #4157, opened 2016-08-29
arpack 3.4.0

dpo:


ilovezfs:
@dpo gave on cmake hehe?

dpo:
I wanted to ship 3.4.0 before the issues with CMake are resolved. See https://github.com/Homebrew/homebrew-science/pull/4103#issuecomment-243011773.

ilovezfs:
Ah, I see. That's weird it doesn't install the pkg-config files. Sounds like a :bug:

PR #3489, opened 2016-03-30
Dlib 19.1 (new formula)

tseemann:
Dlib is a modern C++ toolkit containing machine learning algorithms and tools for creating complex software in C++ to solve real world problems. It is open source software and licensed under the Boost Software License. The introduction contains everything you need to know to get started using the library. However, if after consulting the documentation, you have any questions, comments, or complaints feel free to post in the forums.

PR #4212, opened 2016-09-01
dlib: remove openblas dep, add test

dpo:


dpo:
This really fixes #3489.

PR #4225, opened 2016-09-06
beast2 2.4.3 (new formula)

### LINKED ISSUE #4141: Any chance of including BEAST-2 in homebrew-science & updating the version?
I've installed [`BEAST-2`](https://github.com/jonchang/homebrew-biology/blob/master/beast2.rb) via the [`homebrew-biology`](https://github.com/jonchang/homebrew-biology) cask that @jonchang set up.  But while attempting to install other add-ons (the BEAUTi add-ons menu is actually blank... any suggestions?), I noticed the formula installs version 2.1.1.  But the latest version [seems to be](http://beast2.org/) 2.4.3.  I have 2 questions:
1. Would it be possible to update the version of `BEAST-2` installed via Homebrew?
2. Is there any particular reason that `BEAST-1` is part of the `homebrew-science` collection, but not `BEAST-2`?

Thanks in advance for any feedback.

jonchang:
### Have you:
### New formula: have you

Imported from jonchang/biology

Fixes #4141.


beast2.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def caveats; <<-EOS.undent
+    This install coexists with BEASTv1 as all scripts are suffixed with '-2':
+
+        beast-2 -help
+  EOS
+  end
+
+  test do
+    cp opt_prefix/"examples/testCalibration.xml", testpath
```
ilovezfs: usually a directory of "examples" would be installed in `pkgshare` so that it would then be accessible, once the formula is linked, at `/usr/local/share/beast2/examples`. there may be some other directories you'd want to similarly relocate to pkgshare or doc, though I haven't looked

beast2.rb
```diff
@@ -0,0 +1,49 @@
...
+    system "ant", "linux"
+
+    cd "release/Linux/beast" do
+      # Move jars to libexec and point wrapper scripts
+      # to the new locations
+      mv "lib", "libexec"
+      inreplace Dir["bin/*"], "$BEAST/lib", "$BEAST/libexec"
+
+      Dir["bin/*"].each do |f|
+        mv f, f + "-2"
```
ilovezfs: can the build script be modified with a string replacement or patch so that this isn't needed?

jonchang: These wrapper scripts are shipped in the tarball, not generated by the build script, so we have to change it this way.

beast2.rb
jonchang: Done

beast2.rb
ilovezfs: I think I'd probably do something like this: https://github.com/Homebrew/homebrew-core/blob/master/Formula/coreutils.rb#L52-L59

```
libexec.install "bin"
(libexec/"bin").children.each { |f| bin.install_symlink f => "#{f}-2" }
```

ilovezfs: you can do `each_child` instead of `children.each` actually

jonchang: Done

ilovezfs:
```
iMac-TMP:2.4.3 joe$ find . -maxdepth 1 -type d
bin
images
libexec
share
templates
```
Probably want images and templates in `pkgshare` too, no?

Otherwise this looks :100: 

Are there any formulae you know of off the top of your head that should depend on this either always (required) or sometimes (recommended or optional)?

jonchang:
The templates are used internally by the `beauti-2` script. But while testing this out I noticed that the `beauti-2` app doesn't work if you don't run it from the prefix directory since it can't find its templates. It's hardcoded to look in the current directory for `templates/Standard.xml`. I'm going to see if I can write a patch or file an upstream bug or something to fix this.

ilovezfs:
Hurray!

PR #3832, opened 2016-07-16
hdf5 1.10.0 patch1

mannau:
Pull Request to update hdf5 library to current 1.10.0 patch1

PR #3332, opened 2016-02-19
hdf5: Added option to build without high-level library (recommended for threadsafe build)

piermorel:

PR #4230, opened 2016-09-07
hdf5 1.8.17

tschoonj:

### DIFF
--- a/hdf5.rb
+++ b/hdf5.rb
@@ -1,9 +1,8 @@
 class Hdf5 < Formula
   desc "File format designed to store large amounts of data"
   homepage "http://www.hdfgroup.org/HDF5"
-  url "https://www.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.16/src/hdf5-1.8.16.tar.bz2"
-  sha256 "13aaae5ba10b70749ee1718816a4b4bfead897c2fcb72c24176e759aec4598c6"
-  revision 1
+  url "https://www.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.17/src/hdf5-1.8.17.tar.bz2"
+  sha256 "fc35dd8fd8d398de6b525b27cc111c21fc79795ad6db1b1f12cb15ed1ee8486a"
 
   bottle do
     sha256 "3ef70c3ba3d08e2e14f055d9fb3af6368a8cb3d9a825634fb30b83c66d648b1b" => :el_capitan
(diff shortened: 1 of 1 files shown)


tschoonj:
Looks like a lot of dependencies will need a revision bump.

@ilovezfs Is there a clever `brew` command to do that? :smile:

ilovezfs:
@tschoonj not yet. I'd like to have a `brew revise-formula-pr` that takes a `"--for=#[some other formula}"` argument for the commit message but that doesn't exist yet.

dpo:
Please consider whether this subsumes #3332 and #3832.

ilovezfs:
is 1.10 non-stable release?

tschoonj:
Don't know about #3332 (seems harmless)

Ideally 1.10.0-patch1 one would come here, as it is now officially designated as the stable release. That being said, I have serious doubts about the stability of this release after participating in a webinar by the HDF5 developers a couple of months ago when they released 1.10.0.

To the best of my knowledge, no API was broken in 1.10.0, so all dependent packages should still compile against it.

Perhaps a compromise could be to create a 1.8.17 Formula in Homebrew-Versions, while keeping 1.10.0 here?

ilovezfs:
>I have serious doubts about the stability of this release after participating in a webinar by the HDF5 developers

@tschoonj I tend to trust your judgment :) Maybe it should be the reverse for a while with a hdf5-110 in versions until it gets a TomSchoonjanSealOfApproval? CC @DomT4 @dpo @sjackman for thoughts.

dpo:
We've had two versions of a library in Science before (e.g, SuiteSparse, SuperLU), so I wouldn't be opposed to having `hdf5-110` here with an option to build dependents against it (but that's a lot of formulae to modify).

tschoonj:
It's probably best to make 1.10.0-patch1 the default: the stability issues I was referring to would likely impact those formulas that make use of the SWMR (single write - multiple read) technology that was added. Formulas that need SWMR will have a hard dependency on 1.10.0.

ilovezfs:
@tschoonj thanks for the PR! I shipped it, and also slipped in an audit fix (I hope you don't mind I didn't do this in a separate commit) for the non-library file libhdf5.settings installed in `lib` and corrected the conditional `build.with? "check"` to use the new option `build.with? "test"` since otherwise it won't work. I'll now generate the bottles with the Homebrew Testing job on Jenkins since the location of that file changed. Then I'll open separate PRs for each of those revision bumps. Then we'll do 1.10-patch1 in the next round after that.

ilovezfs:
Bottles shipped in https://github.com/Homebrew/homebrew-science/commit/0e2d0a9153b81ffe8445c6e79d6bfd53555932c7

PR #4229, opened 2016-09-07
xraylib: fix PYTHONPATH issue

tschoonj:
When building simultaneously the python2 and python3 bindings, the python3 bindings would end up in the wrong folder due to problems with PYTHONPATH being set to /usr/local/lib/python2.7/site-packages by Homebrew. 

Solution was to unset the variable in the Formula.

PR #4231, opened 2016-09-07
xrmc: rebuild against gcc6

tschoonj:
Should fix test failure in #4229

### DIFF
--- a/xrmc.rb
+++ b/xrmc.rb
@@ -12,12 +12,21 @@ class Xrmc < Formula
     sha256 "cc9fd9634165a26fcadfc8a7ec9632fea2122c5458db368f6bc111fe4e6ccaea" => :mountain_lion
   end
 
-  option "with-test", "Run build-time tests (may take a long time)"
+  option "without-test", "Don't run build-time tests (may take a long time)"
 
-  depends_on "xraylib"
-  depends_on "pkg-config" => :build
   needs :openmp
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "pkg-config" => :build
   depends_on "xmi-msim" => :optional
+
+  if build.with? "xmi-msim"
+    depends_on "xraylib" => "with-fortran"
+  else
+    depends_on "xraylib"
+  end
 
   fails_with :llvm do
     cause <<-EOS.undent
@@ -27,23 +36,33 @@ class Xrmc < Formula
   end
 
   def install
+    inreplace Dir.glob("{examples,test}/*/Makefile.am"),
+      "$(datadir)/examples/xrmc/", "$(datadir)/examples/"
+
     args = %W[
       --disable-dependency-tracking
       --disable-silent-rules
       --prefix=#{prefix}
       --enable-openmp
+      --docdir=#{doc}
+      --datarootdir=#{pkgshare}
     ]
 
-    args << ((build.with? "xmi-msim") ? "--enable-xmi-msim" : "--disable-xmi-msim")
+    if build.with? "xmi-msim"
+      args << "--enable-xmi-msim"
+    else
+      args << "--disable-xmi-msim"
+    end
 
+    system "autoreconf", "-fiv"
     system "./configure", *args
     system "make"
-    system "make", "check" if build.with? "check"
+    system "make", "check" if build.with? "test"
     system "make", "install"
   end
 
   test do
-    cp Dir.glob("#{share}/examples/xrmc/cylind_cell/*"), "."
-    system "#{bin}/xrmc", "input.dat"
+    cp_r (pkgshare/"examples/cylind_cell").children, testpath
+    system bin/"xrmc", "input.dat"
   end
 end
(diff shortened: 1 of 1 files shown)


tschoonj:
Not sure why this failed. Could you try restarting the build? Thanks!

ilovezfs:
build restarted

ilovezfs:
@tschoonj it's reproducible locally but it's fixed if I build from HEAD. Also there's a formula bug here since it should `depends_on "xraylib" => "with-fortran"` whenever it's `build.with? "xmi-msim"`.

ilovezfs:
@tschoonj it's also fixed if I re-run autotools so something's amiss with the generated files.

ilovezfs:
@tschoonj can you please apply this diff to your PR?
```
diff --git a/xrmc.rb b/xrmc.rb
index ce6929b..27cbab1 100644
--- a/xrmc.rb
+++ b/xrmc.rb
@@ -12,13 +12,22 @@ class Xrmc < Formula
     sha256 "cc9fd9634165a26fcadfc8a7ec9632fea2122c5458db368f6bc111fe4e6ccaea" => :mountain_lion
   end
 
-  option "with-test", "Run build-time tests (may take a long time)"
+  option "without-test", "Don't run build-time tests (may take a long time)"
 
-  depends_on "xraylib"
-  depends_on "pkg-config" => :build
   needs :openmp
+
+  depends_on "autoconf" => :build
+  depends_on "libtool" => :build
   depends_on "xmi-msim" => :optional
 
+  if build.with? "xmi-msim"
+    depends_on "xraylib" => "with-fortran"
+  else
+    depends_on "xraylib"
+  end
+
   fails_with :llvm do
     cause <<-EOS.undent
     llvm-gcc's OpenMP does not support the collapse statement,
@@ -27,23 +36,33 @@ class Xrmc < Formula
   end
 
   def install
+    inreplace Dir.glob("{examples,test}/*/Makefile.am"),
+      "$(datadir)/examples/xrmc/", "$(datadir)/examples/"
+
     args = %W[
       --disable-dependency-tracking
       --disable-silent-rules
       --prefix=#{prefix}
       --enable-openmp
+      --docdir=#{doc}
+      --datarootdir=#{pkgshare}
     ]
 
-    args << ((build.with? "xmi-msim") ? "--enable-xmi-msim" : "--disable-xmi-msim")
+    if build.with? "xmi-msim"
+      args << "--enable-xmi-msim"
+    else
+      args << "--disable-xmi-msim"
+    end
 
+    system "autoreconf", "-fiv"
     system "./configure", *args
     system "make"
-    system "make", "check" if build.with? "check"
+    system "make", "check" if build.with? "test"
     system "make", "install"
   end
 
   test do
-    cp Dir.glob("#{share}/examples/xrmc/cylind_cell/*"), "."
-    system "#{bin}/xrmc", "input.dat"
+    cp_r (pkgshare/"examples/cylind_cell").children, testpath
+    system bin/"xrmc", "input.dat"
   end
 end
```

tschoonj:
What is the point of this particular change?

````diff
   def install
+    inreplace Dir.glob("{examples,test}/*/Makefile.am"),
+      "$(datadir)/examples/xrmc/", "$(datadir)/examples/"
+
     args = %W[
````

ilovezfs:
@tschoonj that's so the examples end up in `share/xrmc/examples` (i.e., `pkgshare/"example"`) instead of `share/examples/xrmc` or `share/xrmcs/examples/xrmc`. But feel free to drop that if you like, as long as you also drop `--datarootdir=#{pkgshare}` (if you want `share/"examples/xrmc"`) or leave the `--datarootdir=#{pkgshare}` (if you want `share/"xrmcs/examples/xrmc"` instead of `share/"examples/xrmc"`), and update the test block accordingly.

tschoonj:
ok pushed

ilovezfs:
@tschoonj cool, thanks. Not sure what's up with upstream's generated files, but it might be worth trying to get a new tarball with the stuff regenerated with recent Autotools. Also, the pkgshare thing is all weird since upstream should have datarootdir defaulting to share/xrmc not share, and then the Makefile.am files can just use `(datadir)/examples/` instead of being over-protective with the xrmc subdir, but I digress...

tschoonj:
I am actually a developer of this package and the one who makes the releases. I will try to come up with a new release soon.

ilovezfs:
@tschoonj ah OK. Then maybe you'll want to investigate what's wrong with those generated files further. The test was failing because it's actually segfaulting.

tschoonj:
The problem is that this `xrmc` is getting a bit old, and was generated with a version of `libtool` that doesn't support Yosemite and El Capitan and was in fact providing linker options for the shared library that were suitable for Puma...

This is why `autoreconf` fixes this issue.

ilovezfs:
Puma hehe ... nice

tschoonj:
@ilovezfs would you mind merging this one in and afterwards restarting #4229 ?

Thanks!

ilovezfs:
@tschoonj done!

ilovezfs:
@tschoonj thanks for this PR! I'm just sorry we didn't notice this breakage sooner but luckily you came to our rescue :)

Issue #4232, opened 2016-09-07
GalSim depends on pyfits instead of astropy


josePhoenix:
The `galsim` formula includes a line

```
depends_on "pyfits" => :python
```

However, PyFITS has been [deprecated](http://www.stsci.edu/institute/software_hardware/pyfits) in favor of `astropy.io.fits`. People with `astropy` available should not be required to install a deprecated package. It should be possible to depend on either PyFITS *or* Astropy, but I'm not familiar enough with the formula syntax to say whether that dependency can be expressed.

I just replaced "pyfits" with "astropy" in my local copy of the formula, and it installed with passing tests, so it should be fine.

dpo:
Thanks for pointing that out. Would you like to submit a pull request?

josePhoenix:
Is there a way to depend on either pyfits **or** astropy? I think I'd be trading one problem for another (for users of PyFITS) if I just changed the name of the dependency, but I would be happy to open a PR with that change.

PR #4233, opened 2016-09-07
galsim: Depend on astropy.io.fits for FITS I/O

josePhoenix:
~~Check for `astropy` availability, but only if `pyfits` is not available.~~

Depend on Astropy instead of PyFITS. Addresses #4232.

### DIFF
--- a/galsim.rb
+++ b/galsim.rb
@@ -15,10 +15,8 @@ class Galsim < Formula
   depends_on "boost-python"
   depends_on "tmv-cpp"
 
-  # FITS support should come from astropy.io.fits
-  # but don't complain if PyFITS is already installed
-  check_pyfits = system "python -c 'import pyfits' > /dev/null 2>&1"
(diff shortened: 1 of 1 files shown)


galsim.rb
```diff
@@ -15,8 +15,10 @@ class Galsim < Formula
   depends_on "boost-python"
   depends_on "tmv-cpp"
 
-  # pyfits should come from pip
-  depends_on "pyfits" => :python
+  # FITS support should come from astropy.io.fits
+  # but don't complain if PyFITS is already installed
+  check_pyfits = system "python -c 'import pyfits' > /dev/null 2>&1"
```
dpo: I don't think we want that. This code will be executed each time brew scans through the formulae. If `pyfits` is phased out, then let's just drop it and go with `astropy`.

josePhoenix: Darn, okay. I was really pleased with myself for figuring out a way to do this 😛

dpo:
Thanks!

PR #4249, opened 2016-09-10
URL updates

tschoonj:
Just changing the URLs of a couple of software package I distribute...


ilovezfs:
@tschoonj thanks. standard commit subject line for this would be `foo: update url` I'll fix them in post.

tschoonj:
Ok, thanks!

ilovezfs:
@tschoonj :rocket:'d thanks for the updates! Maybe you can get certbot/let's encrypt SSL certs for that box! :)

tschoonj:
I am trying! I am in the process of moving all my projects to Cloudfront but I cannot seem to register an SSL certificate with Amazon Route 53 because I want to keep my domain's WHOIS details private 😄

ilovezfs:
Someone Somewhere
123 Lane Road
The Nowherelands, Villageville, 123456

Issue #2545, opened 2015-07-22
openmeeg fails to build on 10.10.3


bfontaine:
```
==> Installing openmeeg from homebrew/homebrew-science
==> Downloading https://github.com/openmeeg/openmeeg/archive/release-2.1.tar.gz
==> Downloading from https://codeload.github.com/openmeeg/openmeeg/tar.gz/release-2.1
######################################################################## 100,0%
==> cmake . -DUSE_PROGRESSBAR=ON -DCMAKE_C_FLAGS_RELEASE= -DCMAKE_CXX_FLAGS_RELEASE= -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/openmeeg/2.1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_V
==> make install
                 ^
1 error generated.
make[2]: *** [libs/MatLib/MatLib/CMakeFiles/OpenMEEGMaths.dir/sparse_matrix.cpp.o] Error 1
make[1]: *** [libs/MatLib/MatLib/CMakeFiles/OpenMEEGMaths.dir/all] Error 2
make: *** [all] Error 2
```

Logs: https://gist.github.com/ca776a3bb0c13f837a5c

dpo:
```
/tmp/openmeeg20150722-36570-1vhitez/openmeeg-release-2.1/libs/MatLib/MatLib/MatLibConfig.h:72:18: fatal error: 'cblas.h' file not found
        #include <cblas.h>
                 ^
```
I don't see a CMake option to specify what BLAS you want to link against. What does `locate cblas.h` say on your system?

bfontaine:
```
$ locate cblas.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/Headers/cblas.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/Headers/cblas.h
[...]

PR #4276, opened 2016-09-11
openmeeg: revision for hdf5

ilovezfs:


dpo:
@BrewTestBot test this please

dpo:
I looked at https://github.com/Homebrew/homebrew-science/issues/2545 yesterday, which is still an issue on 10.10. Adding `-framework Accelerate` to `CMAKE_CXX_FLAGS` gets past that issue, but then I ran into another problem where `FC.h` can't be found. I think that's a header generated by CMake for C/Fortran interoperability. Unfortunately, Fortran support seems broken in `openmeeg`; I wasn't able to have it discover the Fortran compiler.

See https://github.com/openmeeg/openmeeg/issues/139

iMichka:
There seem to be a new version here: http://openmeeg.github.io (2.3.0).
Though it is not tagged on GitHub: https://github.com/openmeeg/openmeeg

The formula has been broken since at least July 2015. This does not really block the hdf5 migration, it's just that if this gets fixed on day, maybe the next failures could come through the hdf5 changes. No way to test that now, but as say, it does not prevent us to ship a newer hdf5.

ilovezfs:
Closing this for now.

PR #4282, opened 2016-09-11
shogun: revision for hdf5

ilovezfs:


ilovezfs:
Closing in favor of #4375.

PR #4296, opened 2016-09-12
vigra: fix build on Mavericks

ilovezfs:


ilovezfs:
gcc-6 enblend-enfuse trouble is fixed in enblend-enfuse 4.2, but that seems to have a hard `:tex` dependency :(

dpo:
Is Mavericks officially not supported anymore now?

ilovezfs:
@dpo correct. this can go bye-bye. I see upstream decided not to fix this when you reported the problem.

PR #4304, opened 2016-09-14
octave: fix case-sensitive build failure on macOS

ilovezfs:
lvecLibFort not lveclibFort, which fixes segfault in "make check" since
octave requires OpenBLAS or vecLibFort to work properly, and configure
simply wasn't finding libvecLibFort.dylib


dpo:
Good catch! I remember fixing this in other formulae.

PR #4326, opened 2016-09-20
soplex 2.2.1 (new formula)

syarra:
### Have you:

### New formula: have you

### DIFF
--- a/soplex.rb
+++ b/soplex.rb
@@ -6,21 +6,20 @@ class Soplex < Formula
 
   option "without-test", "Skip build-time tests (not recommended)"
 
-  depends_on "gcc"
   depends_on "zlib" if OS.linux?
   depends_on "gmp"
 
   def install
+    make_args = ["SHARED=true"]
     if OS.mac?
-      File.open("make/make.darwin.x86_64.gnu", "a") do |f|
+      File.open("make/make.darwin.x86_64.clang", "a") do |f|
         f.puts "LIBBUILDFLAGS+= -m64 -L/usr/local/lib -lgmp -lz"
       end
+      make_args += ["COMP=clang"]
     end
-    system "make", "SHARED=true", "USRCXXFLAGS=-I/usr/local/include", "USRLDFLAGS=-L/usr/local/lib"
-    system "make", "test" if build.with? "test"
-    lib.install Dir["lib/libsoplex*"]
-    bin.install Dir["bin/soplex*"]
-    include.install Dir["src/*.h"]
+    system "make", *make_args, "USRCXXFLAGS=-I#{Formula["gmp"].opt_include}", "USRLDFLAGS=-L#{Formula["gmp"].opt_lib}"
+    system "make", "test", *make_args if build.with? "test"
+    system "make", "install", "INSTALLDIR=#{prefix}", *make_args
     pkgshare.install "src/example.cpp"
   end
 
--- a/soplex.rb
+++ b/soplex.rb
@@ -7,16 +7,32 @@ class Soplex < Formula
   option "without-test", "Skip build-time tests (not recommended)"
 
   depends_on "gcc"
-  depends_on "zlib"
+  depends_on "zlib" if OS.linux?
   depends_on "gmp"
 
   def install
-    system "make", "SHARED=true"
+    if OS.mac?
+      File.open("make/make.darwin.x86_64.gnu", "a") do |f|
+        f.puts "LIBBUILDFLAGS+= -m64 -L/usr/local/lib -lgmp -lz"
+      end
+    end
+    system "make", "SHARED=true", "USRCXXFLAGS=-I/usr/local/include", "USRLDFLAGS=-L/usr/local/lib"
     system "make", "test" if build.with? "test"
     lib.install Dir["lib/libsoplex*"]
     bin.install Dir["bin/soplex*"]
     include.install Dir["src/*.h"]
     pkgshare.install "src/example.cpp"
+  end
+
+  def caveats; <<-EOS.undent
+      SoPlex is distributed under the ZIB Academic License
+      (http://scip.zib.de/academic.txt).
+      You are allowed to retrieve SoPlex for research purposes as a member of
+      a non-commercial and academic institution.
+
+      We agreed to this license for you.
+      If this is unacceptable you should uninstall.
+    EOS
   end
 
   test do
--- a/soplex.rb
+++ b/soplex.rb
@@ -13,11 +13,11 @@ class Soplex < Formula
     make_args = ["SHARED=true"]
     if OS.mac?
       File.open("make/make.darwin.x86_64.clang", "a") do |f|
-        f.puts "LIBBUILDFLAGS+= -m64 -L/usr/local/lib -lgmp -lz"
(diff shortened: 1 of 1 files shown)


dpo:
Excellent, thank you! If I remember well, SoPlex comes with a specific license. There should probably be a `caveats` method alerting the user that we've accepted the license terms for them, and they should uninstall if that's not agreeable. See, e.g., `gmt.rb` for an example.

soplex.rb
```diff
@@ -0,0 +1,27 @@
+class Soplex < Formula
+  desc "The Sequential object-oriented simPlex"
+  homepage "http://soplex.zib.de"
+  url "http://soplex.zib.de/download/release/soplex-2.2.1.tgz"
+  sha256 "c2fffca4095c56f7261820dc2bcc2a31a6a6ff91446a9da083814a913a2a086f"
+
+  option "without-test", "Skip build-time tests (not recommended)"
+
+  depends_on "gcc"
+  depends_on "zlib"
```
dpo: I think this is only necessary on Linux.

soplex.rb
```diff
@@ -7,18 +7,34 @@ class Soplex < Formula
...
   depends_on "gmp"
 
   def install
-    system "make", "SHARED=true"
+    if OS.mac?
+      File.open("make/make.darwin.x86_64.gnu", "a") do |f|
+        f.puts "LIBBUILDFLAGS+= -m64 -L/usr/local/lib -lgmp -lz"
+      end
+    end
+    system "make", "SHARED=true", "USRCXXFLAGS=-I/usr/local/include", "USRLDFLAGS=-L/usr/local/lib"
```
dpo: Those paths should already be on the include and library paths.

soplex.rb
```diff
@@ -7,18 +7,34 @@ class Soplex < Formula
...
+  def caveats; <<-EOS.undent
+      SoPlex is distributed under the ZIB Academic License
+      (http://scip.zib.de/academic.txt).
+      You are allowed to retrieve SoPlex for research purposes as a member of
+      a non-commercial and academic institution.
+
+      We agreed to this license for you.
+      If this is unacceptable you should uninstall.
+    EOS
+  end
```
dpo: Thank you!

soplex.rb
syarra: They should. But somehow `brew` doesn't find them.

soplex.rb
```diff
@@ -0,0 +1,43 @@
+class Soplex < Formula
+  desc "The Sequential object-oriented simPlex"
+  homepage "http://soplex.zib.de"
+  url "http://soplex.zib.de/download/release/soplex-2.2.1.tgz"
+  sha256 "c2fffca4095c56f7261820dc2bcc2a31a6a6ff91446a9da083814a913a2a086f"
+
+  option "without-test", "Skip build-time tests (not recommended)"
+
+  depends_on "gcc"
```
dpo: You don't want to depend on `gcc`. Does this not build with Clang? If not, then there should be a `fails_with` block (see, e.g., `graph-tool`, which uses them extensively).

soplex.rb
dpo: What is `make` supposed to find at those locations? It would be ok to indicate `Formula[...].opt_include`, for instance.

soplex.rb
```diff
@@ -0,0 +1,43 @@
...
+      File.open("make/make.darwin.x86_64.gnu", "a") do |f|
+        f.puts "LIBBUILDFLAGS+= -m64 -L/usr/local/lib -lgmp -lz"
+      end
+    end
+    system "make", "SHARED=true", "USRCXXFLAGS=-I/usr/local/include", "USRLDFLAGS=-L/usr/local/lib"
+    system "make", "test" if build.with? "test"
+    lib.install Dir["lib/libsoplex*"]
+    bin.install Dir["bin/soplex*"]
+    include.install Dir["src/*.h"]
+    pkgshare.install "src/example.cpp"
```
dpo: SoPlex doesn't have a `make install`?

soplex.rb
```diff
@@ -0,0 +1,43 @@
...
+      a non-commercial and academic institution.
+
+      We agreed to this license for you.
+      If this is unacceptable you should uninstall.
+    EOS
+  end
+
+  test do
+    libs = ["-lsoplex", "-lz", "-lgmp"]
+    system ENV["CXX"], "#{pkgshare}/example.cpp", "-oexample", *libs
```
dpo: This uses Clang (see my comment above about gcc).

soplex.rb
syarra: No it doesn't have one!

soplex.rb
syarra: Ok, I get it!

soplex.rb
```diff
@@ -0,0 +1,42 @@
...
+  depends_on "zlib" if OS.linux?
+  depends_on "gmp"
+
+  def install
+    make_args = ["SHARED=true"]
+    if OS.mac?
+      File.open("make/make.darwin.x86_64.clang", "a") do |f|
+        f.puts "LIBBUILDFLAGS+= -m64 -lgmp -lz"
+      end
+      make_args += ["COMP=clang"]
```
dpo: `ENV["CXX"]`? A user can change the compiler, e.g., with `--gcc=...`.

syarra: Then how should I call the makefile? In SoPlex the default is `gcc` but we can specify another  compiler with the `COMP` keyword.

dpo: I think you should be able to set `make_args << "COMP=#{ENV["CXX"]}"`

soplex.rb
goxberry: I've built SoPlex with Clang. This formula shouldn't need to depend on `gcc`.

soplex.rb
```diff
@@ -0,0 +1,42 @@
+class Soplex < Formula
+  desc "The Sequential object-oriented simPlex"
+  homepage "http://soplex.zib.de"
+  url "http://soplex.zib.de/download/release/soplex-2.2.1.tgz"
+  sha256 "c2fffca4095c56f7261820dc2bcc2a31a6a6ff91446a9da083814a913a2a086f"
+
+  option "without-test", "Skip build-time tests (not recommended)"
+
+  depends_on "zlib" if OS.linux?
```
goxberry: Is `zlib` required at all? I think UG needs a SoPlex build with `zlib`, but I'm not sure SoPlex requires it.

goxberry:
Out of curiosity, are you planning on writing packages for the whole SCIP optimization suite? I could put you in touch with some developers on the SCIP team, if you're interested.

syarra:
Not for now. But it will be nice if we could do it.

I’m already in touch with Ambros Gleixner for SoPlex.

soplex.rb
dpo: Will this be updated?

dpo:
Thank you. There's just a small audit warning:
```
==> brew audit homebrew/science/soplex --new-formula
==> FAILED
Error: 1 problem in 1 formula
homebrew/science/soplex:
  * C: 24: col 3: Use empty lines between method definitions.
```
I'll fix it.

Issue #4303, opened 2016-09-13
OpenCV install fails on MacOS Sierra (10.12) at 16%


barronlroth:
### Indicate troubleshooting steps taken

- [ YES ] Ran `brew update` and retried your prior step?
- [ YES ] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [ YES ] Checked that a similar issue hasn't already been filed?
- [ YES ] Looked up the software bug tracker if reporting a build error?
- [ YES ] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

#### Brief summary of issue:
OpenCV installation fails at `[ 16%] Built target opencv_video`

#### The complete command that you issued:
brew install opencv

#### Gist with output of `brew doctor`:
In Gist below 

#### Link output by `brew gist-logs <formula>`:
https://gist.github.com/barronlroth/2860981929c4ebbdb2dffbd1f72bbebd

#### Output of `brew config`:
In gist above^

### Formula Requests:

**The quickest way to get a new formula into Homebrew is to file a [Pull Request](https://github.com/Homebrew/homebrew-science/pulls)**.

We will consider your request but it may be closed if it's something we're not actively planning to work on.

dpo:
For reference, the error is
```
[...]

PR #4336, opened 2016-09-23
opencv 2.4.13.1

JCount:
### Have you:

-----
Fix building stable on macOS Sierra and on El Capitan using Xcode 8

Fixes: https://github.com/Homebrew/homebrew-science/issues/4303

_Note: This should be ready to be bottled for Sierra._


opencv.rb
```diff
@@ -39,7 +53,7 @@ class Opencv < Formula
   depends_on "gstreamer"  => :optional
   depends_on "gst-plugins-good" if build.with? "gstreamer"
   depends_on "jasper"     => :optional
-  depends_on :java        => :optional
+  depends_on java: :optional
```
JCount: This is purely a style question. When RuboCop auto-corrected the ruby hash syntax on this line, it removed the whitespace formatting. I decided to leave it that way, but is there any established preferred practice regarding whitespace and the more modern hash syntax at this time?

ilovezfs: in homebrew/core the style is strictly single spaces, not alignment in a column, but in homebrew/science that particular rule is not enforced

JCount: thank you

JCount: Reverted to using hash rockets again

dpo: Aligning in a column seems to be a legacy of old scientific software, probably dating back to the old Fortran days.

JCount:
**IMPORTANT:**

A hold should probably be put on merging this for the time being (a day at most). 

This is because when I generated this PR to fix the building issue, I accidentally overlooked the fact that there has been a `.1` bug-fix release since the current stable of `2.4.13`. There isn't much in it that directly pertains to building on macOS (unless you are attempting to build with `gcc` instead of `clang`), but since the revision would be bumped it probably makes sense to remove the revision and move stable to the `2.4.13.1` release, which does predate the build fix. I would like to do a little glancing over the code changes/testing with some of the optional args before actually opening a new PR to include both the upstream back-ported build patch and the version bump just to be safe. I can however confirm that a basic version bump did build on my system in the default configuration with no problem.

barronlroth:
@JCount Excuse my noob question, but what does the `head` version omit/do differently than the default version? Also, are users that are upgrading to Sierra / Xcode 8 that previously installed OpenCV successfully  before the QTKit deprecation getting errors now?

JCount:
@barronlroth The `head` version already has the default video I/O set to the  AVFoundation framework, with both QuickTime and QTKit disabled by default. It is still possible to pass args to `CMake` to build either of these as a video I/O as long as the necessary headers, and corresponding frameworks, are present. However, on Sierra the QuickTime framework has been removed and what remains of QTKit is only a class to act as a bridge to convert legacy formats into forms that AVFoundation can handle. Therefore upgrading to Sierra would (highly probably) break an installation of OpenCV that relied on either of these two frameworks because the linking would be broken. 

In the case of upgrading to Xcode 8 on 10.11, I honestly do not know if it would break an existing installation, but it no longer includes the header files needed to do a new build of OpenCV with either QuickTime or QTKit. 

Trying to build the stable version of OpenCV from source with Xcode 8 will fail, but a `head` build should succeed as long as `--with-quicktime` is not passed.

barronlroth:
@JCount Thanks for the explanation, OpenCV was installed successfully with `--HEAD` and my code is now functional on Xcode 8!

JCount:
@barronlroth Glad to hear it!

dpo:
@JCount I'm tempted to think that we should update `opencv` (and `opencv3`) according to your changes. How much would your patch change if `stable` pointed to the newer 2.4.13.1? Could we add an option to build with QuickTime support if Xcode < 8?!

JCount:
I can certainly do that. I mainly became stuck trying to determine what would make the most sense.

dpo:
Ok, let me know if that's reasonable because I'm not an `opencv` user. I'd suggest QuickTime support should be either built-in or `:recommended` on platforms where it's available, so as to stick the the previous behavior there.

JCount:
Unfortunately, I am personally unable to cherrypick the commit in master that fixes the issue in `opencv3` to fix the stable build. `3.1.0` was cut on Dec 18, 2015 and there were 1,211 commits before the fix was merged into master, and the code has changed so much it may not possible to apply the fix to stable on our own before the next release of `opencv3`. However, if anyone more skilled and knowledgable of the code than me wants to try, I wish you luck. The commit is https://github.com/opencv/opencv/commit/f85e33f4af15638cdff6ffb50e7669370cdd7f5b

dpo:
Sounds like a job for `git bisect`! log<sub>2</sub>(1,211) is 10.24, so with a good script, it should be wrapped up after 11 builds :-).

JCount:
I've been working on my `git bisect` skills, but they are currently pretty bad... 😑

dpo:
There's an issue with the `openimageio` test on 10.10 and 10.11:
```
==> brew test --verbose openimageio
==> FAILED
Testing homebrew/science/openimageio
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20161016-29385-1vqo0dz.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-science/openimageio.rb --verbose
==> /usr/local/Cellar/openimageio/1.6.16/bin/oiiotool --info /usr/local/Homebrew/Library/Homebrew/test/fixtures/test.jpg
dyld: Symbol not found: __ZNK5boost16re_detail_10610031cpp_regex_traits_implementationIcE17transform_primaryEPKcS4_
  Referenced from: /usr/local/Cellar/openimageio/1.6.16/lib/libOpenImageIO.1.6.dylib
  Expected in: /usr/local/opt/boost/lib/libboost_regex-mt.dylib
 in /usr/local/Cellar/openimageio/1.6.16/lib/libOpenImageIO.1.6.dylib
Error: homebrew/science/openimageio: failed
Failed executing: /usr/local/Cellar/openimageio/1.6.16/bin/oiiotool --info /usr/local/Homebrew/Library/Homebrew/test/fixtures/test.jpg
```

JCount:
@dpo `openimageio` never received a revision bump for the latest version of `boost`, so the 10.10 and 10.11 bottles seem to be failing. (That's my theory at least.) I'll try building them from source on 10.11 and see if it works; if so, I'll open a PR to bump the revision of `openimageio`.

ilovezfs:
@JCount exactly zero of the science formulae have been tested or rebuilt against 1.62 unless they've since been updated/upgraded for other reasons.

JCount:
I figured as much.

JCount:
A source build of `openimageio` against `boost 1.62.0` and this PR of `opencv` is fine on my test machine of 10.11 & Xcode 8. I think this means I should open a PR to bump the revision of `openimageio` so it is no longer blocking.

dpo:
@JCount That would be great, thank you!

ilovezfs:
@JCount looks like there's a new release, which you can do instead of a revision, if you're feeling ambitious today.

JCount:
I just noticed that myself @ilovezfs :stuck_out_tongue_closed_eyes:

JCount:
@ilovezfs It looks like `openimageio` may have to be put on ice for a bit. It relies quite heavily on CMake's "native" FindBoost.cmake for both `boost` and `boost-python` support. Unfortunately, support for `boost 1.62.0` was added after `3.6.2` was cut. Support is present in `3.7.0-rc1` and `homebrew's` `HEAD` build, so I pretty much have a completed PR to update/fix `openimageio` that seems to work fine, but it cannot be built with the stable release of `cmake`.

Therefore, my feeling is that we should go forward with this PR, since it will at least allow `opencv` to work on with Xcode 8. (Someone else is free to take a crack at `openimageio` but personally I think we may need to wait on it until the next release of `cmake`.)

_One last thought:_ I could open a WIP PR for `openimageio` with a slightly cleaned up version of what I currently have. Thereby both making it available for feedback and testing.

dpo:
@JCount Sorry for making you update the entire tree here! A preliminary PR would be fantastic, with a note that it builds with the `devel` version of `cmake`. Hopefully, there will be a new release before long.

JCount:
@dpo WIP `openimageio 1.7.7` https://github.com/Homebrew/homebrew-science/pull/4486

ilovezfs:
Thanks @JCount! This PR was much needed. :bowing_man:

JCount:
No problem :bowtie:!

astiob:
Why does this check the compiler version to enable or disable AVFoundation and QuickTime? Shouldn’t the Xcode or OS X version be used?

JCount:
The clang build version is intrinsically tied to which version of Xcode is being used, and therefore whether or not the headers are present to support building with Quicktime. Support for AVFoundation is being added to the stable release through an (upstream) back-ported patch, so I tried to avoid changing the default behavior of the actual stable release unless it was necessary, as an attempt avoid breaking things. Once the next stable is released, this will become a moot point, but I did not want to, in effect, create a new stable release version.

astiob:
Isn’t it possible for Clang to come not from Xcode? (In particular, but not only, in Linuxbrew.)

ilovezfs:
returning false from such comparisons automatically is already being worked on ;)

astiob:
I don’t understand why you don’t just compare the version numbers that are actually relevant. If you need Xcode 8, check for Xcode 8. Consider [how the GnuTLS formula in core does it](https://github.com/Homebrew/homebrew-core/blob/master@%7B2016-10-30%7D/Formula/gnutls.rb#L31), for example. The compiler version has nothing to do with whether a certain framework is present, and the fact that a certain compiler is bundled with a certain framework is essentially a coincidence; this kind of knowledge can be used in framework/platform-specific back-end code in the absence of better ways to detect the framework, but I really think it shouldn’t be assumed in a formula. If nothing else, it obfuscates the formula code. To a reader, it doesn’t make sense why AVFoundation code is enabled for Clang build 800 and not in build 799. Does build 799 miscompile it? Well, no, actually, it has nothing to do with Clang at all.

PR #4355, opened 2016-09-27
Update to GAP 4.8.5

alex-konovalov:
GAP 4.8.5 has been released today (http://mail.gap-system.org/pipermail/forum/2016/005347.html). This is straightforward update of the details of the new source archive and version number.


jonchang:
Thanks @alex-konovalov!

PR #4357, opened 2016-09-27
RepeatScout: add mirror url

nextgenusfs:
Download url was broken, added mirror url hosted at RepeatMasker

### DIFF
--- a/repeatscout.rb
+++ b/repeatscout.rb
@@ -3,8 +3,7 @@ class Repeatscout < Formula
   # doi "10.1093/bioinformatics/bti1018"
   # tag "bioinformatics"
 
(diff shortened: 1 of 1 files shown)


dpo:
Looks good, thanks. Let's see if it builds!

dpo:
The download fails every time. The RepeatScout home page has a different link: http://bix.ucsd.edu/repeatscout/RepeatScout-1.0.5.tar.gz

nextgenusfs:
Okay, I will try that, so just edit and then do another PR?

dpo:
What you did is good.

dpo:
Merged. Thank you!

sjackman:
🍼🐧

PR #4408, opened 2016-10-05
Use --without-test instead of --without-check

HadrienG:
### Have you:

The `--without-check` option is deprecated. I updated CONTRIBUTING.md to reflect that contributors should use `--without-test` instead.

example output of `brew audit --strict --online` on a formula with the `--without-check` option:
```
* Use '--without-test' instead of '--without-check'. Migrate '--without-check' with deprecated_option
```


dpo:
Thank you!

PR #4412, opened 2016-10-06
Sundials 2.7.0

### LINKED ISSUE #2687: Sundials formula is out of date
It seems that the sundials formula available here is quite out of date. 
The formula installs sundials 2.5.0, yet the latest available version of sundials is 2.6.2 as it can be seen in the [the official page](https://computation.llnl.gov/casc/sundials/download/download.php).

schoeps:
### Have you:

### Comments
fixes #2687 and will hopefully allow me to compile GetDP (#4405) on the build servers. Finally....


schoeps:
The dependecies (Petsc,cantera) compile without problem on my machine. However, they will need a revision bump. Shall I add this here are in a second PR?

ilovezfs:
Let's do them in subsequent PRs so that we don't blow up Jenkins.

schoeps:
so we are good?

ilovezfs:
@schoeps :ship:'d! thanks for another contribution to homebrew!

two things:
1) commit subjects are case-sensitive
```
[joe:/usr … ps/homebrew/homebrew-science] master ± brew pull --bottle https://github.com/Homebrew/homebrew-science/pull/4412
==> Fetching patch 
Patch: https://github.com/Homebrew/homebrew-science/pull/4412.patch
==> Applying patch
Applying: Sundials 2.7.0
==> Patch closes issue #4412
Warning: Nonstandard bump subject: Sundials 2.7.0
Warning: Subject should be: sundials 2.7.0
==> Fetching patch (bottle commit)
Patch: https://github.com/BrewTestBot/homebrew-science/compare/homebrew:master...pr-4412.patch
```

2) `brew audit --strict` will still complain about this formula because `without-check`/`with-check` options are supposed to be deprecated in favor of `without-test`/`with-test` options. We can worry about that next time for this formula. Note that when you change these, any `build.with? "check"` in the formula has to be changed to `build.with? "test"` and any `build.without? "check"`, changed to `build.without? "test"`

sundials.rb
```diff
@@ -11,777 +10,49 @@ class Sundials < Formula
...
-            # Parallel FIDA examples
-            system "mpiexec", "-np", "4", "./ida/fcmix_parallel/fidaHeat2D_kry_bbd_p"
-
-            # Parallel FKINSOL examples
-            # TODO: Fix bug in parallel FKINSOL example or ask upstream for fix
-            system "mpiexec", "-np", "4", "./kinsol/fcmix_parallel/fkinDiagon_kry_p"
+    if build.with? "petsc"
+      args << "-DPETSC_ENABLE=ON"
+      args << "-DPETSC_LIBRARY_DIR=#{Formula["slepc"].opt_lib}"
+      args << "-DPETSC_INCLUDE_DIR=#{Formula["slepc"].opt_inc}"
```
dpo: Should this have been `petsc` instead of `slepc`?

sundials.rb
```diff
@@ -11,777 +10,49 @@ class Sundials < Formula
...
   end
 
-  depends_on "openblas" => :optional
   depends_on :fortran => :optional
   depends_on :mpi => [:cc, :f77, :recommended]
 
-  option "without-check", "Skip build-time checks and examples (not recommended)"
+  depends_on "cmake" => :build
+  depends_on "openblas" => :optional
+  depends_on "petsc" => :optional
```
dpo: There's a circular dependency here because `petsc` depends on `sundials`...

sundials.rb
schoeps: Yes, thanks. I will make a revision, also taking care of the other comments above

sundials.rb
schoeps: I was aware and therefore I made it optional. I cannot comment why you would want `sundials` to depend on `petsc`. I checked that this the package compiles with that switch enabled but I did not use it for anything. Shall I remove it? (I don't mind)

dpo: I can see why one would want `sundials` to depend on `petsc` (e.g., to solve linearized discretized ODEs with a Krylov method), but I'm not sure why one would want `petsc` to depend on `sundials`.

schoeps: `petsc` has several time steppers and apparently also the ones from `sundials`, see e.g. https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/TS/

PR #4414, opened 2016-10-06
petsc: revision for sundials

jonchang:


schoeps:
may I suggest to also bump `cantera` since it depends on sundials, too.

PR #4415, opened 2016-10-06
beast 1.8.4

jonchang:
Also install examples to `pkgshare` and add a neglected
`:java` requirement


sjackman:
@sjackman 🐧

sjackman:
@jonchang Thanks for updating Beast! When you update a bioinformatics formula, or any formula with a Linux bottle, can you give me a poke (as above), and I'll update the Linux bottle. Thanks!

jonchang:
@sjackman will do. I tried to figure out how to bottle for Linux but couldn't get it to work. Is there a guide somewhere?

PR #4389, opened 2016-09-30
nixio 1.2.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #4417, opened 2016-10-07
Nixio 1.2.0 --- version update

jgrewe:
extension of the original pull request #4389 

Sorry, wasn't able to create a pull request to the original fork, so this PR has to be seen as a replacement for #4389


jonchang:
Thanks @jgrewe and @ilovezfs!

PR #4428, opened 2016-10-10
Test odd gatk error. Do not merge

sjackman:


sjackman:
@Homebrew/maintainers I'm experiencing a very strange error. I've tried to debug it, and I'm completely confused.

```ruby
class Gatk < Formula
  url "NA"
  version "NA"
end
```

`brew audit` for this dummy formula is giving the error message:
```
Error: You must first download GATK from
http://www.broadinstitute.org/gatk/download
to /Users/brew/Jenkins/pr-science/version/sierra/home/Downloads/GenomeAnalysisTK-2.8-1.tar.bz2
```

This error message is from a buggy version of this formula that was committed to `master` two years ago in e059df972fb156b1d59a0df9bd2d9a448ca831b8 and then reverted by Mike in 8e3a540.
See [the BrewTestBot log](https://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/6056/version=sierra/testReport/junit/brew-test-bot/sierra/audit_homebrew_science_gatk___new_formula/)

Why am I seeing this error message from this zombie formula now?

MikeMcQuaid:
> Why am I seeing this error message from this zombie formula now?

The `audit` crawls through old versions looking for `revision`s etc. I'm guessing it's not catching whatever exception or exit you used.

sjackman:
Yep, that's definitely what's happening. Does it check only the most recent commit, or how far back does it go?

MikeMcQuaid:
It goes through the entire history of the formula. It'll need a Homebrew/brew change to catch whatever the exception is.

sjackman:
The exception is raised by `abort` at
https://github.com/Homebrew/homebrew-science/commit/e059df972fb156b1d59a0df9bd2d9a448ca831b8#diff-a7435792b64b816b433326eaa2e605f3R11

PR #4433, opened 2016-10-10
metis: fix openmp option, audits

dpo:


dpo:
```
dyld: Library not loaded: /usr/local/opt/sundials/lib/libsundials_cvode.1.dylib
  Referenced from: /usr/local/opt/slepc/real/lib/libslepc.3.7.dylib
  Reason: image not found
```

dpo:
@BrewTestBot test this please

PR #4436, opened 2016-10-10
viennarna: fix openmp option, audits

dpo:


sjackman:
Hi, Dominique. When you update a bioinformatics formula, could you please tag me with a `@sjackman 🐧` so that I can update the bottle for Linuxbrew?

dpo:
Sure, no problem!

PR #4439, opened 2016-10-11
libbi: prevent perl version conflicts

sbfnk:
### Have you:

This is to prevent a perl version conflict arising because the package installs the "Template" perl library which uses compiled code. Because of this, if libbi is run with a different perl version from the one used for installing, it will throw a segmentation fault. This commit fixes this by prepending the PATH containing perl at install time to the PATH at execution time. Are there better ways of resolving this? Should it be left to the user to check that the first perl in the PATH is the same at installation and execution?


jonchang:
Yeah, this is tricky. I think this fix is fine; you may want to add a caveat stating that `libbi` must be run with the same version of perl it was installed with, and a reinstall would be necessary if changing perl versions?

sbfnk:
Have done that and added a test that will fail if there is a perl version conflict. Thanks!

jonchang:
Thanks @sbfnk!

PR #4443, opened 2016-10-12
mad-x 5.02.11 (new formula)

Eothred:
### Have you:

- [x ] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [ x] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [x ] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [x ] Built your formula locally prior to submission with `brew install <formula>`?

### New formula: have you

- [x ] Written a sensible test? (The best test is to compile and run a sample program.)


Eothred:
I get this complaint from the audit which I was not able to understand:

```
% brew audit --strict --online mad-x
homebrew/science/mad-x:
  * C: 13: col 32: Use hash rockets syntax.
Error: 1 problem in 1 formula
```

mad-x.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  url "http://svn.cern.ch/guest/madx/tags/5.02.11/madX/"
+  head "http://svn.cern.ch/guest/madx/trunk/madX/"
+
+  head "http://svn.cern.ch/guest/madx/trunk/madX"
+
+  depends_on "cmake" => :build
+  depends_on :x11
+
+  conflicts_with "cern-ndiff", because: "mad-x provides ndiff"
```
dpo: Use `:because => "mad-x provides ndiff"`.

mad-x.rb
```diff
@@ -0,0 +1,40 @@
...
+  patch :DATA
+
+  def install
+    system "cmake", "-DBUILD_SHARED_LIBS=ON", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"hello.txt").write("print,text='Hello,Mad-X';stop;")
+    system "madx_dev < hello.txt"
```
dpo: Does this yield an expected result that could be checked?

Eothred: it should print the phrase "Hello,Mad-X", but it will also print some header info including execution time and version number.
```
% madx_dev < hello.mad

  ++++++++++++++++++++++++++++++++++++++++++++
  +     MAD-X 5.02.11  (64 bit, Darwin)      +
  + Support: mad@cern.ch, http://cern.ch/mad +
  + Release   date: 2016.10.11               +
  + Execution date: 2016.10.13 08:47:21      +
  ++++++++++++++++++++++++++++++++++++++++++++
print,text='Hello,Mad-X';

Hello,Mad-X
stop;


  Number of warnings: 0

  ++++++++++++++++++++++++++++++++++++++++++++
  +          MAD-X finished normally         +
  ++++++++++++++++++++++++++++++++++++++++++++
```

Eothred:
as I already had fortran compiler installed I forgot to add the dependency

mad-x.rb
Eothred: There is a giant test system that can be executed with 'ctest' which also checks that you are getting correct results in the simulations that are set up. However I had understood that the main purpose of the test in Homebrew was to check that the binary executes without problems? I will check your assert suggestion, if that works I completely agree it does a more useful check

mad-x.rb
```diff
@@ -0,0 +1,39 @@
...
+  homepage "https://cern.ch/mad"
+
+  url "http://svn.cern.ch/guest/madx/tags/5.02.11/madX/"
+  head "http://svn.cern.ch/guest/madx/trunk/madX/"
+
+  depends_on "cmake" => :build
+  depends_on :fortran
+  depends_on :x11
+
+  conflicts_with "cern-ndiff", :because => "mad-x provides ndiff"
```
Eothred: not sure what you mean? to not build cern-ndiff in this project directly and rather depend on cern-ndiff? Currently I do not believe the build scripts are written in such a way, ndiff is always installed when you run 'make install'. I could create another small patch to exclude it:
```
--- CMakeLists.txt	(revision 5952)
+++ CMakeLists.txt	(working copy)
@@ -59,7 +59,6 @@
 # Add sources subdirectory:
 add_subdirectory(libs/ptc)
 add_subdirectory(src)
-add_subdirectory(tools)
 add_subdirectory(syntax)

 # Install documentation as well:
@@ -71,4 +70,3 @@
 # Setting up testing:
 include(setupTesting)
 # Add test folders:
-include(ndifftests)
```

Eothred:
While we discussed they released another version, so I have updated that. I also added the assert_match suggestion from @dpo to the test

mad-x.rb
Eothred: Should I include this patch and remove the dependency?

mad-x.rb
```diff
@@ -0,0 +1,53 @@
+class MadX < Formula
+  desc "Methodical Accelerator Design"
+  homepage "https://cern.ch/mad"
+
+  url "http://svn.cern.ch/guest/madx/tags/5.02.12/madX/"
+  head "http://svn.cern.ch/guest/madx/trunk/madX/"
+
+  depends_on "cmake" => :build
+  depends_on :fortran
+  depends_on :x11
```
dpo: Is there no need to indicate the dependence on `cern-ndiff`?

Eothred: No, the dependency is only for testing the code, mad-x does not use this directly in any situation.

dpo:
Thanks!

PR #3754, opened 2016-06-16
vtk: Depends on libxml2 for Linuxbrew

sjackman:
- [x] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [x] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [x] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [ ] Built your formula locally prior to submission with `brew install <formula>`?
- [x] Removed the `revision` line, if any, when bumping the version number?
- [x] Added/bumped the `revision` number if the changes affect the precompiled bottles?
- [x] Not altered the `bottle` section?
- [ ] Checked that the tests still pass?

Fix error:
`Could NOT find LibXml2 (missing: LIBXML2_LIBRARIES LIBXML2_INCLUDE_DIR)`

No need to rebuild the Mac bottles when this PR is merged. Please pull the Linux bottle.

PR #4446, opened 2016-10-12
vtk: improve python path setup

### LINKED ISSUE #3577: VTK forces the directory `.../include/python3.5m`

- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [X] Checked that a similar issue hasn't already been filed?
- [?] Looked up the software bug tracker if reporting a build error?
- [NA] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

### Bug reports / build errors:

Building VTK 7.0.0 fails with Python 3.4.3 installed through Homebrew.

```
CMake Error at CMake/FindPythonLibs.cmake:183 (FILE):
  file STRINGS file
  "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/incl
ude/python3.5m/patchlevel.h"
  cannot be read.
  CMake/vtkWrapPython.cmake:164 (find_package)
  Utilities/Python/CMakeLists.txt:4 (include)


-- Found PythonLibs: /usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework
/Versions/3.4/lib/libpython3.5.dylib  
CMake Error at CMake/FindPythonLibs.cmake:183 (FILE):
  file STRINGS file
  "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/incl
ude/python3.5m/patchlevel.h"
  cannot be read.
  CMake/vtkPythonWrapping.cmake:1 (find_package)
  Wrapping/PythonCore/CMakeLists.txt:2 (include)


CMake Warning at CMake/vtkPythonWrapping.cmake:7 (message):
  Python library .  mismatches python 3.4.
  Wrapping/PythonCore/CMakeLists.txt:2 (include)

```

I traced this down to the lines
```
        args << "-DPYTHON_INCLUDE_DIR='#{`python3-config --prefix`.chomp}/include/python3.5m'"
        args << "-DPYTHON_LIBRARY='#{`python3-config --prefix`.chomp}/lib/libpython3.5.dylib'"
        args << "-DVTK_INSTALL_PYTHON_MODULE_DIR='#{lib}/python3.5/site-packages'"
```
in `vtk.rb`. This produces the wrong include dir -- because with 3.4.3, the corresponding include dir is `/python3.4m`.

It seems like the best approach should be to detect which Python is installed, and to adapt these directories accordingly?

iMichka:
### Have you:

This fixes #3577
The changes are inspired from the gdcm formula

- The check for simultaneous python2 and python3 flags was put first, to bail out earlier from that situation.
- The path setup is now done by looking what is the actual python. This removes the hard-coded dependencies on python 3.5 (python 3.6 is coming soon)
- The path setup was made more uniform for python 2 and 3

Note: this does not change the situation about the linking hack which is currently in the VTK formula.
This change just makes sure all the values passed to cmake are consistent.

### DIFF
--- a/vtk.rb
+++ b/vtk.rb
@@ -4,7 +4,7 @@ class Vtk < Formula
   url "http://www.vtk.org/files/release/7.0/VTK-7.0.0.tar.gz"
   mirror "https://fossies.org/linux/misc/VTK-7.0.0.tar.gz"
   sha256 "78a990a15ead79cdc752e86b83cfab7dbf5b7ef51ba409db02570dbdd9ec32c3"
-  revision 4
+  revision 5
 
   head "https://github.com/Kitware/VTK.git"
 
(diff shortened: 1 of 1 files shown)


iMichka:
BTW I also saw that this should fix the problem encountered in #3754

jonchang:
Thanks for this; I'll merge once CI passes.

iMichka:
Brew bot failed due to audit error "python modules have explicit framework links". This is the reason we have the temporary (ugly) workaround in the formula.
Also, the test failed due to a timeout, but when I run "brew test vtk" the test correctly passes on my system.

jonchang:
Hmm, looks like the test failed because the test-bot uninstalls build-time-only dependencies prior to running tests. As the test uses `cmake` it will fail on the CI servers. Do you know if `cmake` should also be a runtime dependency for vtk?

iMichka:
Makes sense.

As long as you know how to write a correct makefile, with all the right flags and include folders, you do not need cmake. But most of the people use cmake in combination with VTK because the whole VTK structure is made to work with cmake out of the box.

99% of the people I see trying to build without cmake on the VTK mailing list encounter build problems at one point or another, and the answer is often just: "use cmake". But strictly speaking it's not a runtime dependency.

You could also have people using the python wrappings, and these do not need cmake at all.

jonchang:
Thanks @iMichka. I created a pull request for an updated test that does not depend on cmake in #4451.

PR #4449, opened 2016-10-12
harry: option for openmp, fix audits

dpo:


dpo:
Strangely, three tests fail on 10.11/Jenkins but they pass on 10.11/Travis:
```
FAIL: check_configs.sh
FAIL: check_measures.sh
FAIL: check_options.sh
```

dpo:
@sjackman @tseemann any ideas?

PR #2460, opened 2015-06-26
ArUco 1.3.0 (new formula)

pierre-rouanet:
Add support for the ArUco library : a minimal library for Augmented Reality applications based on OpenCV
http://www.uco.es/investiga/grupos/ava/node/26

PR #4456, opened 2016-10-14
aruco 2.0.14 (new formula)

kosma:
# ArUco library : a minimal library for Augmented Reality applications based on OpenCV

Homepage: http://www.uco.es/investiga/grupos/ava/node/26
### Notes:
1. There's an existing pull request #2460 for an outdated version that hasn't been touched in a year.
2. The package benefits from (optional) OpenMP support and the formula reflects this, but I couldn't quite test it due to the gcc multilib/openmp conflict. Should I rebuild gcc without multilib (ugh) and try again, remove it from the formula or just leave it as-is? FWIW, the OpenMP support builds and works perfectly with gcc under Linux.
3. `pkg-config` support is broken upstream, so I patched it. I submitted to patch to the authors; library is under active development so we should expect it to be fixed upstream for some value of _soon_.
4. [Proof of a working build](https://www.dropbox.com/s/0y7o05g97itrgw4/Screenshot%202016-10-14%2002.14.41.png?dl=0) included. ;)
### Have you:
### New formula: have you


aruco.rb
```diff
@@ -0,0 +1,60 @@
+class Aruco < Formula
+  desc "minimal library for Augmented Reality applications based on OpenCV"
+  homepage "http://www.uco.es/investiga/grupos/ava/node/26"
+  url "https://downloads.sourceforge.net/project/aruco/2.0.13/aruco-2.0.13.zip"
+  sha256 "8d4f24fd2b88264908d1d8e7d366eaf814e999a6e1bf41a33be8ca9153245a7a"
+
```
dpo: Is there also a `head` to point to?

aruco.rb
```diff
@@ -0,0 +1,60 @@
...
+  option "with-openmp", "Build with OpenMP support"
+  needs :openmp if build.with? "openmp"
+
+  depends_on "cmake" => :build
+  depends_on "opencv"
+
+  patch :DATA # upstream pkg-config is horribly broken, fix it
```
dpo: could you please point to the issue you submitted upstream? That will help remind us to remove the patch when upgrading.

dpo:
Oops, there's an issue with the patch:
```
==> Patching
patching file CMakeLists.txt
Hunk #1 FAILED at 324.
1 out of 1 hunk FAILED -- saving rejects to file CMakeLists.txt.rej
patching file aruco.pc.in
Error: Failure while executing: /usr/bin/patch -g 0 -f -p1
```
Could it be that the file you're patching is not at the top level?

dpo:
> Proof of a working build included. ;)

Beautiful! Thank you!

aruco.rb
kosma: None that I know of, unfortunately. I'm trying to work something out with the authors but the library is being developed in an university setting so I wouldn't be surprised if there was no version control at all.

kosma:
> Oops, there's an issue with the patch

@dpo It looks like a whitespace issue - I'm patching two files and one of them uses `\r\n` while the other just `\n`, and `patch` chokes on that. I'm not sure about the underlying issue as it could be either git's `autocrlf` messing with things, `patch` getting confused about different line endings or something else mangling the patch along the way. I'm afraid the best course of action might be to wait for upstream to merge the patch so we can just avoid this issue altogether.

kosma:
@dpo Regarding the failing CI build, do you have any advice? The`opencv` formula is currently broken (due to Xcode 8 release removing QTKit) and until it's fixed, the only workaround is to manually `brew install opencv --HEAD` - but depending on `HEAD` is unsupported (and a bad idea anyway).

dpo:
@kosma, about the patch, could you check that it's not your text editor changing tabs to spaces or something of that sort?

dpo:
@kosma Is it a possibility to depend on `qt5`? Or to drop QT support altogether?

kosma:
@dpo It's QuickTime, not QT - see #4303, we're pretty much blocked by it. Regarding the patch - tabs/spaces are correct. It's different newline styles that cause trouble. Let's just hope the patch gets merged by the time the opencv situation clears.

dpo:
Sorry, trying to go too fast.

dpo:
@BrewTestBot test this please

kosma:
I'm still waiting for library authors to fix it upstream. As to why the patch sometimes applies and sometimes not (you can see it succeed on one CI server but fail in another) - my hunch is that git's `core.autocrlf` settings are different on those buildservers.

dpo:
Sounds good. Thanks for the update. Let's know when the patch has been merged upstream.

kosma:
@dpo The pkg-config issue has been fixed upstream, I've updated the formula with the new version. The only remaining problem is the `brew test` which is failing because `pkg-config` is not available at test time. This line is the culprit:

```ruby
system ENV.cc, "test.cpp", *`pkg-config --cflags --libs aruco`.chomp.split, "-o", "test"
```

What's the recommended way to solved it? Should I add `pkg-config` as a run-time dependency?

jonchang:
Is it expected that users will need `pkg-config` to use this formula? If so you can depend on it with `depends_on "pkg-config" => :run`

iMichka:
Closing due to inactivity. If somebody wants to work on this, feel welcome to pick this PR up. Thanks.

kosma:
If someone could pick up this work, it would be great. Due to personal reasons I'm unable to give this formula enough attention and love to keep it Homebrew-quality and up to date. :(

PR #4457, opened 2016-10-14
tmv-cpp 0.73

dpo:


dpo:
The build fails `--with-openmp`: https://github.com/rmjarvis/tmv/issues/22

PR #4465, opened 2016-10-14
cube: fails with clang, audit fixes

dpo:
The home page doesn't seem to exist anymore. @sjackman @tseemann do you know if `cube` has a new home page somewhere?


dpo:
```
==> Installing cube from homebrew/science
qt: This formula either does not compile or function as expected on macOS
versions newer than El Capitan due to an upstream incompatibility.
Error: An unsatisfied requirement failed this build.
```

dpo:
@sjackman 🐧

PR #4473, opened 2016-10-16
omcompiler 1.9.6 (new formula)

SimplyDanny:
### Have you:

### New formula: have you

Modelica is an object-oriented, equation based language to conveniently model complex physical systems. OpenModelica is one implementation of the language providing a modeling and simulation environment. OMCompiler is the base of OpenModelica. It translates Modelica code into C code.


omcompiler.rb
```diff
@@ -0,0 +1,69 @@
+class Omcompiler < Formula
+  desc "Modelica compiler translating Modelica to C code"
+  homepage "https://www.openmodelica.org"
+  url "https://github.com/OpenModelica/OMCompiler.git",
+      :tag => "v1.9.6",
+      :revision => "d3c35c90b382998275f42eb5da05c0f25e6b63a8"
+
+  # Options
+  option "with-boost",      "Use boost"
+  option "with-sundials",   "Use sundials to add more numerical solvers to the simulation runtime"
```
dpo: These two options are redundant with lines 29 and 30, and should be removed.

omcompiler.rb
```diff
@@ -0,0 +1,69 @@
...
+  depends_on "automake"     => :build
+  depends_on "cmake"        => :build
+  depends_on "libtool"      => :build
+  depends_on "pkg-config"   => :build
+  depends_on "gnu-sed"      => :build
+  depends_on "xz"           => :build
+
+  # Essential dependencies
+  depends_on :fortran
+  depends_on "gcc"
```
dpo: There's no need to depend on `gcc`, the `:fortran` dependency takes care of it. If you mean that this library only builds with gcc and not with clang, then there should be a `fails_with` block.

omcompiler.rb
```diff
@@ -0,0 +1,69 @@
...
+  depends_on "homebrew/science/lp_solve"
+  depends_on "homebrew/science/hwloc"
```
dpo: There's no need to fully qualify formulae from this tap.

SimplyDanny:
All annotations are put into practice. I hope I did not forget anything.

dpo:
Thank you!!!

PR #4475, opened 2016-10-16
octave: fix for qt5

schoeps:
### Have you:

### Comments
Compiles with new qt5 from https://github.com/Homebrew/homebrew-core/pull/5951, see also https://github.com/Homebrew/homebrew-science/pull/4327


schoeps:
@dpo @ilovezfs Now the question: shall we make `--with-qt5` default again or not. Currently, I tend to say no....

schoeps:
@ilovezfs second question for you: shall we add symlinks in the lib dir of qscintilla2 to `lib*-qt`? Octave and other packages might like this.

octave.rb
```diff
@@ -78,51 +80,49 @@ class Octave < Formula
...
-  depends_on "ghostscript"    => :recommended  # ps/pdf image output
-  depends_on "gl2ps"          => :recommended
-  depends_on "graphicsmagick" => :recommended  # imread/imwrite
-  depends_on "transfig"       => :recommended
+  depends_on "arpack"              => :recommended
+  depends_on "curl"                => :recommended
+  depends_on "epstool"             => :recommended
+  depends_on "fftw"                => :recommended
+  depends_on "fltk"                => :recommended
+  depends_on "ghostscript"         => :recommended  # ps/pdf image output
```
ilovezfs: This conflicts with the `:tex` requirement in HEAD so should be a stable block (or use `:tex` for stable too but usually that's considered too heavy to require unconditionally for stable)

octave.rb
```diff
@@ -78,51 +80,49 @@ class Octave < Formula
...
-  depends_on "qscintilla2"     if build.with? "qt5"
-  depends_on "qt5"             if build.with? "qt5"
-  depends_on "suite-sparse"    if build.with? "suite-sparse"
-  depends_on "veclibfort"      if build.without?("openblas") && OS.mac?
-
-  depends_on "openblas" => (OS.mac? ? :optional : :recommended)
+  depends_on "libsndfile"          if build.with? "audio"
+  depends_on "llvm"                if build.with? "jit"
+  depends_on "portaudio"           if build.with? "audio"
+  depends_on "pstoedit"            if build.with? "ghostscript"
```
ilovezfs: same comment as ghostscript, and I think it may apply to epstool too (any others?)

octave.rb
```diff
@@ -211,6 +211,11 @@ def install
 
     system "./bootstrap" if build.head?
 
+    # Octave does strict assumptions on the name of qscintilla2
+    # see http://savannah.gnu.org/bugs/?48773
+    inreplace "m4/acinclude.m4", "qscintilla2-qt5 qt5scintilla2", "qscintilla2-qt5 qt5scintilla2 qscintilla2"
```
ilovezfs: Although the upstream fix should patch the m4 file itself, I think we can avoid the Autotools dependencies for stable (and the autoreconf below) by making the corresponding string replacements in the generated configure file instead (whether it was generated by upstream and distributed in the tarball or created just-in-time during bootstrap shouldn't matter).

octave.rb
```diff
@@ -78,51 +80,49 @@ class Octave < Formula
...
-  depends_on "qhull"           if build.with? "qhull"
-  depends_on "qrupdate"        if build.with? "qrupdate"
-  depends_on "qscintilla2"     if build.with? "qt5"
-  depends_on "qt5"             if build.with? "qt5"
-  depends_on "suite-sparse"    if build.with? "suite-sparse"
-  depends_on "veclibfort"      if build.without?("openblas") && OS.mac?
-
-  depends_on "openblas" => (OS.mac? ? :optional : :recommended)
+  depends_on "libsndfile"          if build.with? "audio"
+  depends_on "llvm"                if build.with? "jit"
```
ilovezfs: Should be `depends_on "llvm" => :optional` and deprecate the "with-jit" option

ilovezfs:
>@ilovezfs second question for you: shall we add symlinks in the lib dir of qscintilla2 to lib*-qt? Octave and other packages might like this.

Given that we're trying to kill off qt4 entirely, probably not, but I don't feel strongly one way or the other.

ilovezfs:
>@dpo @ilovezfs Now the question: shall we make --with-qt5 default again or not. Currently, I tend to say no.... 

Do people generally want to use Octave with or without the GUI?

I think it should probably be default on otherwise CI won't check if it's actually building properly. However, if at some point Octave moves to homebrew/core, then ideology will dictate it be made optional.

dpo:
Why would octave ever move to core? This is definitely its home.

Personally, I never use the GUI, but I may not be the typical user. Is that info available from the analytics?

ilovezfs:
>Is that info available from the analytics?

Not really since the GUI has been the default for a while.

octave.rb
ilovezfs: I should say they are very likely to be in conflict on macOS because the default way the requirement is fulfilled is by installing MacTeX, which creates a mess for everyone by installing ghostscript, etc. too. See https://github.com/Homebrew/homebrew-core/issues/4732. It's also come up in homebrew/science issues. I don't know what the best solution here is other than begging MacTex to stop doing that lol or creating a fancy :ghostscript requirement.

octave.rb
```diff
@@ -78,51 +80,49 @@ class Octave < Formula
   option "with-jit",               "Use the experimental just-in-time compiler (not recommended)"
   option "with-openblas",          "Use OpenBLAS instead of native LAPACK/BLAS"
   option "with-osmesa",            "Use the OSmesa library (incompatible with opengl)"
-  option "with-qt5",               "Compile with graphical user interface"
 
   # build dependencies
-  depends_on "gnu-sed"         => :build
-  depends_on "pkg-config"      => :build
+  depends_on "gnu-sed"             => :build
```
schoeps: Octave maintainers decided that gnu sed is an explicit requirement. I will add a link to the source code.

ilovezfs: Thanks!

octave.rb
```diff
@@ -78,51 +80,49 @@ class Octave < Formula
...
-  depends_on "transfig"       => :recommended
+  depends_on "arpack"              => :recommended
+  depends_on "curl"                => :recommended
+  depends_on "epstool"             => :recommended
+  depends_on "fftw"                => :recommended
+  depends_on "fltk"                => :recommended
+  depends_on "ghostscript"         => :recommended  # ps/pdf image output
+  depends_on "gl2ps"               => :recommended
+  depends_on "glpk"                => :recommended
+  depends_on "gnuplot"             => :recommended
```
schoeps: This is not a technical necessity; it's just a "good" choice. I don't think we should enforce this.

octave.rb
schoeps: I have installed both and I had no problems so far. I would prefer things as they are. Also because  `--HEAD` user are anyway on their own...

octave.rb
ilovezfs: I could see making a (possibly default) option (since you're saying it's the "good" choice) for it because it seems sad to require qt5 for the octave gui and require gnuplot, but not give them the qt5 gnuplot too :(

octave.rb
schoeps: Can I do a `if :tex` construction for ghostscript?

octave.rb
ilovezfs: I guess the problem with making it an option is that it will trigger an octave source build, which is undesirable. I'll leave it up to you, but it does seem unfortunate to require qt5 and then "cheap out" here.

schoeps: Ok, I will add `qt5` to `gnuplot` since I do not have a strong opinion on that matter (as discussed in the other thread: `qt5` is the most usable `gnuplot` backend on the Mac by far).

octave.rb
ilovezfs: There is indeed a problem because they try to install files in the same location. If the MacTex files are installed then when you reinstall Octave brew will error out because it will say ghostscript is not linked. Then you're forced to link brew ghostscript and overwrite the MacTex files, which will not necessarily cause any subsequent bugs, but is really bad UX and will have to be redone every time MacTex is upgraded or reinstalled.

>Also because --HEAD user are anyway on their own...

It's a problem for stable too
```
  # build the pdf docs
  if build.with? "docs"
    depends_on :tex             => :build
    depends_on "librsvg"        => :build
  end
```

Is there any way to remove all options that use tex?

schoeps: `:tex` is only needed if installed `--with-docs`. I think we should allow the user to build the (pdf) docs if he or she really wants?

octave.rb
ilovezfs: You can do
```
qt5dep = (build.with? "qt5") ? ["with-qt5"] : []
depends_on "gnuplot" => [:recommended] + qt5dep
```

schoeps: why not just `depends_on "gnuplot" => [:recommended] + ["with-qt5"] if build.with? "qt5"`?

schoeps: that's how I did it now

ilovezfs: The order of operations is very unclear in the above. If you want something more straightforward:
```
if build.with? "qt5"
  depends_on "gnuplot" => [:recommended, "with-qt5"]
else
  depends_on "gnuplot" => :recommended
end
```
which is what we almost always do in core.

ilovezfs: If you want it all on one line, people do this:
```
depends_on "gnuplot" => [:recommended] + ((build.with? "qt5") ? ["with-qt5"] : [])
```

schoeps:
@ilovezfs I pushed a new version, otherwise I cannot keep track of all the changes. It's audited but not yet tested. It contains all issues discussed so far.

octave.rb
```diff
@@ -101,29 +104,39 @@ class Octave < Formula
...
+  # recommended blas dependencies
+  depends_on "openblas"            => (OS.mac? ? :optional : :recommended)
   depends_on "veclibfort"          if build.without?("openblas") && OS.mac?
 
+  # recommended ghostscript dependencies
+  depends_on "ghostscript"         => :recommended # ps/pdf image output
+  depends_on "pstoedit"            if build.with? "ghostscript"
+
+  # recommended qt5 dependencies
+  depends_on "qt5"                 => :optional
```
ilovezfs: I thought we discussed this being default. No?

schoeps: I thought the opposite? :) I guess 'power' users tend to use command-line, beginners rather the gui. I am afraid that maintaining with gui is a lot more effort....

schoeps: I thought the opposite? :) I guess 'power' users tend to use command-line, beginners rather the gui. I am afraid that maintaining with gui is a lot more effort....

ilovezfs: I think the maintenance burden will actually be lower if gui is default on. The reason is that we will be sure that it works when the formula goes through CI and the bottles are generated. If not, then users who want the GUI will face a long build, and then if/when it fails they will file issues here (usually multiple duplicates) and we get to play wack-a-mole trying to fix it after it has already shipped.

ilovezfs: @schoeps if it's installed with the gui / with-qt5, is the command line octave available in the installation too?

schoeps: yes, but it's slightly slower with GUI :) However, I will turn on the GUI but if there is a problem it's not necessarily me who is going to fix it. 

I am now investing a lot of time since I want to have a working `octave` for my lectures (starting on Thursday) but afterwards I can probably not keep up that level.

ilovezfs:
```
    args << "--with-portaudio"           if build.without? "audio"
    args << "--with-sndfile"             if build.without? "audio"
    args << "--disable-java"             if build.without? "java"
    args << "--enable-jit"               if build.with? "jit"
```
The new option names have to be used (portaudio and llvm) otherwise this code will fail to work as expected.

schoeps:
fixed. Thanks for spotting.

schoeps:
@dpo: @ilovezfs made me clean up `octave.rb`. It has become a rather huge changeset. It's more consistent with homebrew now but less backward compatible (which I think is a small price). Opinions?

schoeps:
is there a way for me to download and test the bottles?

ilovezfs:
@schoeps https://bintray.com/homebrew/bottles-science/download_file?file_path=octave-4.2.0-rc2_1.sierra.bottle.tar.gz

ilovezfs:
sha256 should be 48ff4fefeb6077484abfd7330d47fb9cb713903ce051b66a7763cee14c707638

schoeps:
is there a convenient command line interface or do I decompress manually?

ilovezfs:
Just replace the formula file with the one from the PR, and update the Sierra checksum in the bottle block. Then brew uninstall octave, brew install octave, and should download and pour it.

schoeps:
I compiled locally and also verified the bottle and I am still happy. Even one qt5-bug resolved itself. 
I think we can ship if @dpo and @ilovezfs are happy, too.

ilovezfs:
Any trouble pouring the bottle?

Hehe with any luck, the 4.2.0 update will just be bumping the URL!

I'm happy.

schoeps:
> Any trouble pouring the bottle?

I had to to remove `rebuild 1` from the bottle section, that took me a minute. Besides that, it was a pleasure :)

dpo:
I'll test soon and get back to you. Thanks much for all the hard work here!!

schoeps:
@dpo I start my computational engineering lecture tomorrow, so it would be great if students could install `octave` easily via homebrew, i.e., please merge :)

ilovezfs:
@schoeps merged. Thanks for all the hard work on this. If @dpo finds any issues, I'm sure we can address them in a follow-up PR.

dpo:
Just tested. Works perfectly for me. Thanks @schoeps!

PR #4480, opened 2016-10-17
root6: Adding option `all`

leoredi:
### Have you:

- [ ] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?

### New formula: have you

Adding the option to install root6 with all the optional flags set to `ON`

### DIFF
--- a/root6.rb
+++ b/root6.rb
@@ -53,6 +53,7 @@ class Root6 < Formula
       -Dbuiltin_freetype=ON
       -Droofit=ON
       -Dminuit2=ON
+      #{config_opt("all")}
       #{config_opt("python")}
       #{config_opt("ssl", "openssl")}
       #{config_opt("xrootd")}


root6.rb
```diff
@@ -53,6 +53,7 @@ def install
       -Dbuiltin_freetype=ON
       -Droofit=ON
       -Dminuit2=ON
+      #{config_opt("all")}
```
dpo: Right now, this builds `root6` with `-Dall=OFF` (look at the definition of `config_opt` above. There needs to be an option. See
https://travis-ci.org/Homebrew/homebrew-science/jobs/168233791#L511

What components exactly does `all` bundle in?

leoredi: Hi, thanks for looking. If I understood the code correctly, you option is set to `OFF` when you install with the usual `brew install root6`, which is exactly what you want. In some occasions (like here at CERN where we are heavy users of other options) one would use the flag `--with-all` in order to use activate the flag. You can have a look [here](https://root.cern.ch/building-root#options) to understand what you will install.

dpo: What I'm saying is that `brew install root6 --with-all` will still build with `-Dall=OFF` because the formula doesn't declare the `with-all` option.

Looking at the link you provided, I'm not very keen to include an option to build with `-Dall=ON` because many of those optional components are available separately as Homebrew formulae (`fftw3`, `freetype`, `zlib`, `gsl`, `llvm`, `tbb`, ...). We prefer to rely on the formulae instead of having `root6` download and install its own versions of those libraries. I'd be very happy to add options to `root6` to build with extra components though, e.g, `with-gsl`, `with-tbb`, etc.

leoredi: OK sure, so since I see that RooFit and minut2 are already set by default on `ON` why don't we add:

* `cocoa`
* `fortran `
* `gdml`
* `mt`

dpo: I don't work with `root6`. Are you one of the developers?

leoredi: Nope, I work at CERN where root6 is developed. In all honesty since I feel you are not convinced by these changing just live it as it is and I will modify my own version. Thanks for the time spent on this.

dpo: I was just saying that I'm not sure what other packages are most useful to users. I'm happy to trust you on this. If you'd rather hold, please close the pull request.

PR #4496, opened 2016-10-20
opencv: don't compare against nil

rwhogg:
### Have you:

DevelopmentTools.clang_build_version can return nil in some cases, which lacks
a >= method.

See https://github.com/Linuxbrew/brew/issues/140


dpo:
> Upstream is working on a general fix for this problem (which is that certain methods return nil, which can't be compared against), but until that fix lands we'll just have to fix individual formulae.

Could you please point to the upstream issue/pr?

rwhogg:
https://github.com/Homebrew/brew/issues/729 appears to be one instance of the problem; there may be others too.

@mistydemeo might have more information on this one.

dpo:
Thanks!

sjackman:
Thanks for this fix, Bob!

PR #4498, opened 2016-10-20
travis/linux setup

dpo:
Revisit the Linux setup on Travis to closely match OSX. What apt packages should be installed by default? I should probably remove `libblas-dev` and `liblapack-dev` from the list?!

cc: @ilovezfs @sjackman @rwhogg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,8 @@ matrix:
       rvm: system
     - os: linux
       rvm: system
-      services: docker
       sudo: required
-  allow_failures:
-    - os: linux
+      dist: trusty
 
 before_install:
   - if [ -f ".git/shallow" ]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,55 +7,47 @@ matrix:
...
+addons:
+  apt_packages:
+    - build-essential
+    - curl
+    - git
+    - python-setuptools
+    - m4
+    - texinfo
+    - libblas-dev
+    - liblapack-dev
```
rwhogg: I would add `libxml-parser-perl` to that list. I don't think `libxml-sax-perl` is necessary; the only formula we've needed it for on core is `docbook2x` so far.

We have a bottle for LAPACK now, so I think you should remove `liblapack-dev`. Likewise for `texinfo` - better that we see missing dependencies on that and fix them.

I don't think you need to explicitly state ones like `git` - you can see which ones are available by default from the Docker image quay.io/travisci/travis-ruby. That said, if you want to keep them there to avoid unexpected surprises, that makes sense too.

dpo:
Updated. Here is an example working PR: https://travis-ci.org/dpo/homebrew-science/builds/169654301.

sjackman:
Looks great. Thanks, Dominique.

We use GCC 5 to build the Linuxbrew bottles. We haven't yet migrated to GCC 6. I'd be inclined to use GCC 5 as the default compiler for `.travis.yml` until Linuxbrew does make the switch.

dpo:
@sjackman Done. Successful PR: https://travis-ci.org/dpo/homebrew-science/jobs/169846333

sjackman:
:+1: Great. Let's merge it.

dpo:
I'm having some issues with formulae that `depends_on :fortran`. It seems Linuxbrew wants to pour `gcc` even though `gfortran` is already available. See the (long) discussion on Slack.

sjackman:
When a formula is installed from bottle (rather than from source), the `default_formula` is always installed, completely ignoring the return value of `satisfy`. This behaviour is the same on both Homebrew and Linuxbrew. Here's the line of logic that implements this behaviour: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula_installer.rb#L349

dpo:
If I try to preempt that behavior and [install `gcc` up front](https://travis-ci.org/dpo/homebrew-science/jobs/169980375#L238), then brewing a formula that `depends_on :fortran` [installs it again!](https://travis-ci.org/dpo/homebrew-science/jobs/169980375#L463)

dpo:
Ok, I'm happy with this one. On Linux, we always `brew install gcc` preemptively so bottles will be linked against a known `gcc`. All formulae that have `depends_on :fortran` or have a dependency that `depends_on :fortran`, or `depends_on :mpi => :f77/f90` will pour `gcc` anyways.

.travis.yml
```diff
@@ -7,55 +7,42 @@ matrix:
...
-        linuxbrew
-        brew test-bot --tap=homebrew/science;
-    fi
+  - brew test-bot --tap=homebrew/science;
 
 notifications:
   email:
     on_success: never
-    on_failure: always
+    on_failure: never
```
ilovezfs: I certainly won't mind less Travis "SPAM" but not sure how @sjackman will feel about this change.

dpo:
Updated. No gcc preinstalled. Let's move this along.

sjackman:
:party: Thanks, Dominique!

PR #4506, opened 2016-10-21
abyss 2.0.2

sjackman:
Fix compile errors with gcc-6 and boost-1.62.


sjackman:
```
==> brew install cerulean
==> Installing cerulean from homebrew/science
cerulean: Unsatisfied dependency: numpy
Homebrew does not provide special Python dependencies; install with:
  `pip install numpy`
Error: An unsatisfied requirement failed this build.
==> FAILED
```
https://travis-ci.org/Homebrew/homebrew-science/jobs/171540694#L1534

PR #4511, opened 2016-10-22
simpleitk 0.10.0

iMichka:
### Have you:

- Bumped version to 0.10.0
- Updated description to comply with brew audit
- Updated position of options to comply with brew audit
- Deprecated examples option to comply with brew audit
- Added option for python 3 support (inspired from the gdcm and vtk formulae)


iMichka:
Setting up all these python paths, envs and so on is tedious. We are already doing that in gdcm, vtk, and now simpleitk. The insighttoolkit formula could also benefit from it.

It would be great if there was a way to grab that information from homebrew, so that there is a single place where these things get defined. Each time something needs a change, at least the 4 mentioned formulae need to be tweaked.

simpleitk.rb
```diff
@@ -12,54 +12,77 @@ class Simpleitk < Formula
     sha256 "b09618f1801915dd6cdba0c25af6685c5ddbb7dc21c8ad11e178c4e9600ae5f8" => :mavericks
   end
 
+  option "with-examples", "Compile and install various examples"
+  option "with-csharp", "Compile for C sharp"
+  option "with-java", "Compile for Java"
```
dpo: This could be removed actually; line 27 takes care of it.

dpo:
It's possible to make things somewhat generic using constructs such as
```ruby
Language::Python.each_python(build) do |python, version|
  system python, *Language::Python.setup_install_args(prefix)
end
```
The formulae over at homebrew/python are instructive. You can use similar constructs in the `test` method.

simpleitk.rb
iMichka: Right, updated

iMichka:
I will have a look at the more generic construct for later (when I find some time for it). I want to see if I can adapt it for multiple formulae first.

iMichka:
I updated the PR by changing the tar.gz, there is now a SimpleITK-0.10.0-1.tar.gz version that contains the missing file. The maintainers just repackaged the 0.10.0 version. The patch was therefore removed.

dpo:
Are you happy with this?

iMichka:
Yes. The formula was update and all is green. Improving the python setup (also for other formulae) is something I may look at later, if I can find some spare time :)

dpo:
Thank you!

dewet22:
@iMichka After the `-1` patch, brew now misinterprets the version number as just "1". I opened #4820 to address this.

PR #4517, opened 2016-10-25
graph-tool: depend on boost160

ilovezfs:
graph-tool does not yet have boost 1.62.0 compatibility

CC @guglielmo


ilovezfs:
Upstream issue is https://git.skewed.de/count0/graph-tool/issues/347

PR #4524, opened 2016-10-26
ceres-solver.rb update to match eigen.rb

ahundt:
### Have you:
### New formula: have you

When you tried to use ceres the following happens, so recompiling ceres-solver will fix it:

```
-- Failed to find Ceres - Found Eigen dependency, but the version of Eigen found (3.2.10) does not exactly match the version of Eigen Ceres was compiled with (3.2.9). This can cause subtle bugs by triggering violations of the One Definition Rule. See the Wikipedia article http://en.wikipedia.org/wiki/One_Definition_Rule for more details
-- Looking for Ceres (optional)... - not found
CMake Error at /usr/local/share/modules/CommonTools.cmake:826 (message):


  Module grl was requested to be build with Ceres.  Please ensure that the
  package is installed in a standard system location or set DEPENDS_Ceres_DIR
  to the installation prefix (i.e., root directory of the installation).  To
  disable features which require this optional dependency, set the WITH_Ceres
  option to OFF and try again.

  The DEPENDS_Ceres_DIR variable can alternatively be set to the directory
  containing a CeresConfig.cmake or ceres-config.cmake file.  If no such file
  exists, contact either the developer of this project or CMake BASIS to
  provide a FindCeres.cmake file.

  /usr/local/share/modules/ProjectTools.cmake:2135 (basis_find_package)
  /usr/local/share/modules/ProjectTools.cmake:2422 (basis_find_packages)
  /usr/local/share/modules/ProjectTools.cmake:2696 (basis_project_begin)
  modules/grl/CMakeLists.txt:84 (basis_project_impl)
```

### DIFF
--- a/ceres-solver.rb
+++ b/ceres-solver.rb
@@ -3,7 +3,7 @@ class CeresSolver < Formula
   homepage "http://ceres-solver.org/"
   url "http://ceres-solver.org/ceres-solver-1.11.0.tar.gz"
   sha256 "4d666cc33296b4c5cd77bad18ffc487b3223d4bbb7d1dfb342ed9a87dc9af844"
-  revision 3
+  revision 4
 
   head "https://ceres-solver.googlesource.com/ceres-solver.git"
 


dpo:
Every single test fails on linux.

dpo:
This formula isn't linux ready. There are explicit references to dylibs.

dpo:
[This line](https://github.com/Homebrew/homebrew-science/pull/4524/files#diff-afa72fa346dd0dd8a0de92aa0c674df3R32) should be updated. `suite-sparse` has a mandatory dependency on `metis` (not `metis4`). The `suite-sparse-options` function should go.

ahundt:
@dpo true... but right now things are also broken on OS X! This change is definitely less broken than the current tree! If it is possible try triggering a test on the current master with the bottled version on OS X and that too will fail. I think addressing linux is definitely necessary, but it is out of scope and should be a separate issue/pull request from this fix.

ahundt:
Created #4543 for what you mentioned.

dpo:
How is OSX broken? Jenkins built bottles successfully.

ahundt:
@dpo try using ceres-solver in a dependency. It fails immediately because the bottled eigen version does not match the version used by ceres-solver. There is (correctly) a specific check for this in the CMake scripts. This is still another instance of that issue where changing dependant libraries breaks dependencies silently.

PR #4527, opened 2016-10-26
graph-tool: fix openmp dependency

dpo:


Lucidyan:
@dpo 
Before installation, I clean all dependencies with command 
`brew deps --include-optional --include-build graph-tool | xargs -tn1 brew uninstall -f`
And Install graph-tool:
`brew install --build-from-source graph-tool --without-python --with-python3 --with-openmp  --without-scipy --without-matplotlib`
It's installs fine, compille with `gcc` and `openmp` flag, but when I trying to import, I see this:
```
In [1]: from graph_tool.all import *
dyld: lazy symbol binding failed: Symbol not found: __ZN5boost6python7objects23register_dynamic_id_auxENS0_9type_infoEPFSt4pairIPvS2_ES4_E
  Referenced from: /usr/local/lib/python3.5/site-packages/graph_tool/libgraph_tool_core.so
  Expected in: flat namespace

dyld: Symbol not found: __ZN5boost6python7objects23register_dynamic_id_auxENS0_9type_infoEPFSt4pairIPvS2_ES4_E
  Referenced from: /usr/local/lib/python3.5/site-packages/graph_tool/libgraph_tool_core.so
  Expected in: flat namespace

fish: 'ipython3' terminated by signal SIGABRT (Abort)
```

Without flags all importing correct (but compills with openmp-free `clang`)

Maybe the key in this message (@ilovezfs add this package in deps, coz graph-tool not working with 1.62 version of boost)
```
Warning: homebrew/science/graph-tool dependency homebrew/versions/boost160 was built with a different C++ standard
library (libc++ from clang). This may cause problems at runtime.
```

dpo:
Yes I think that's the key. You might have to rebuild `boost` 1.60 with gcc.

Lucidyan:
Rebuild it with gcc-6, same results here

dpo:
@Lucidyan Thanks for carrying out those time-consuming experiments! The problem you're hitting is https://github.com/Homebrew/homebrew-science/issues/3259. Unfortunately, I don't know the solution.

dpo:
@BrewTestBot test this please

dpo:
`libepoxy` fails to build on Linux because EGL is missing: https://travis-ci.org/Homebrew/homebrew-science/jobs/172767019#L5937
The `home page for `libepoxy`](https://github.com/anholt/libepoxy#building-autotools) recommends to install `libegl1-mesa-dev`.

PR #4528, opened 2016-10-27
Create siril.rb

lock042:
-Don't have MacOS to really try it.
-Would like to add
   brew link --force gettext
   brew link pcre jpeg
   brew link glib shared-mime-info gdk-pixbuf atk gsettings-desktop-schemas gtk+3
but don't know how to do it.
-’depends_on "gcc"’ apparently gives error


siril.rb
```diff
@@ -0,0 +1,45 @@
+class Siril < Formula
+  desc "Astronomical image (pre-)processing application"
+  homepage "http://free-astro.org/index.php/Siril"
+  url "https://free-astro.org/download/siril-0.9.4.tar.bz2"
+  sha256 "3c90111faac63f2d0f17f1a05891ef9e97a2819e5073db2d26e8c3ab7072d91e"
```
dpo: Is there a `head` repository?

siril.rb
```diff
@@ -0,0 +1,45 @@
...
+  depends_on "automake"
+  depends_on "intltool"
+  depends_on "gettext"
+  depends_on "ffms2"
+  depends_on "fftw"
+  depends_on "freetype"
+  depends_on "gnome-icon-theme"
+  depends_on "gsettings-desktop-schemas"
+  depends_on "openjpeg"
+  depends_on "homebrew/science/cfitsio"
```
dpo: There's no need to fully qualify formulae in this tap.

dpo:
I think this would be a great addition, thanks!

siril.rb
lock042: Yes indeed. We use a svn server: https://free-astro.org/svn/siril/

siril.rb
lock042: Do you mean I can remove "homebrew/science" for cfitsio ?

siril.rb
```diff
@@ -0,0 +1,45 @@
...
+    system "autoreconf", "-fi", "-Wno-portability"
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+    bin.install "src/siril"
+  end
+  test do
+    system "siril", "-v"
```
lock042: The thing is for now we need gui for processing images. So I have to say I have no idea on how I can write the test.

siril.rb
dpo: Ok, you can add it here as a `head`.

siril.rb
dpo: Exactly.

siril.rb
dpo: Ok, let's leave it for now.

siril.rb
lock042: OK, I will make the changes. But I have to say I don't know how to proceed with github :)

dpo:
Thanks. Why did you think you needed `depends_on "gcc"`?

lock042:
@dpo . Gcc it's just because clang does not compile with openmp if the version is too old. Am I wrong ? This is the only reason.
Does my rb file file does respect the rules now ? Because I see that all checks have failed :).

dpo:
@lock042 It seems that OSX is happy building `siril` with Clang, but without OpenMP: https://travis-ci.org/Homebrew/homebrew-science/jobs/171730235#L3179

To add OpenMP support, you can do the following:
1. add `option "with-openmp", "Enable OpenMP multithreading"` before the dependencies
2. add `needs :openmp if build.with? "openmp"` after the dependencies, before `install`
3. set a configure option if necessary to build with OpenMP in `install`, e.g.,
```ruby
args = ["--disable-debug",
        "--disable-dependency-tracking",
        "--disable-silent-rules",
        "--prefix=#{prefix}"]
args << "--with-openmp=yes" if build.with? "openmp"
system "./configure", *args
```

It would also be great if you could rebase your branch so the Linux build has a chance to go through (I made changes to the build settings after you opened this PR).

lock042:
@dpo : I fixed the issue (I hope)

siril.rb
```diff
@@ -0,0 +1,52 @@
...
+  needs :openmp if build.with? "openmp"
+
+  def install
+    system "intltoolize", "-f", "-c"
+    system "autoreconf", "-fi", "-Wno-portability"
+    args = ["--disable-debug",
+        "--disable-dependency-tracking",
+        "--disable-silent-rules",
+        "--prefix=#{prefix}"]
+    args << "--with-openmp=yes" if build.with? "openmp"
```
dpo: I was just using this as an example. Is this indeed the correct option to pass to `configure`?

siril.rb
```diff
@@ -0,0 +1,52 @@
+class Siril < Formula
+  desc "Astronomical image (pre-)processing application"
+  homepage "http://free-astro.org/index.php/Siril"
+  url "https://free-astro.org/download/siril-0.9.4.tar.bz2"
+  head "https://free-astro.org/svn/siril/"
+  sha256 "3c90111faac63f2d0f17f1a05891ef9e97a2819e5073db2d26e8c3ab7072d91e"
```
dpo: There's an audit warning that `sha256` should appear before `head`.

siril.rb
```diff
@@ -0,0 +1,52 @@
...
+        "--disable-dependency-tracking",
+        "--disable-silent-rules",
+        "--prefix=#{prefix}"]
+    args << "--with-openmp=yes" if build.with? "openmp"
+    system "./configure", *args
+    system "make", "install"
+    bin.install "src/siril"
+  end
+  test do
+    system "siril", "-v"
```
dpo: There's a second audit warning here that `system "siril"` should be `system "#{bin}/siril`.

siril.rb
lock042: Oh sorry. In fact, with Gcc to disable openmp we write --disable-openmp. enamble-openmp would enable but this is not required as it is the default behavior.
For clang, I don't know yet.

lock042:
@dpo : I tried to fix issues.

lock042:
@dpo : Where can we see the output of the configure ? I see nothing on the report. I would like to know if Siril well compiled with openmp this time.

dpo:
@lock042 It doesn't compile with OpenMP by default because the default compiler is clang. That's good because otherwise, users would necessarily have to also install `gcc`. If a user wants OpenMP support, they can compile from source using `brew install siril --with-openmp`.

lock042:
@dpo : OK. So now my file is ok ?

By the way, thank you for your help :).

lock042:
@dpo : I tried on a mac. No problem for simple installation without openmp, although I would prefer that "brew install siril" gives openmp option by default.
I tried
`brew install siril --with-openmp`
and I have an error:
`Undefined symbols for architecture x86_64:
  "std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_dispose()", referenced from:
      c_pipp_avi_write_dib::~c_pipp_avi_write_dib() in pipp_avi_write_dib.o
  "operator delete(void*, unsigned long)", referenced from:
      c_pipp_avi_write_dib::~c_pipp_avi_write_dib() in pipp_avi_write_dib.o
make[1]: *** [siril] Error 1
make: *** [install-recursive] Error 1
`

dpo:
@lock042 You're very welcome. Here are a few more comments:

1. in order to rebase (which would be good to see if the build goes through on Linux), you can proceed as follows
   1. `brew update`
   2. go to your `patch-2` branch: `git checkout patch-2`
   3. rebase: `git rebase master` (if you haven't changed anything else, this should go through without trouble; if you run into trouble, `git rebase --abort`)
   4. if the rebase worked, force push your branch: `git push -f git@github.com:lock042/homebrew-science.git patch-2`
2. `brew install siril --with-openmp` uses gcc, so it seems there's a compilation error in that case. You'll find the output of `configure` under `~/Library/Logs/Homebrew/siril`. If you want to share it, paste here the output of `brew gist-logs siril`.

lock042:
@dpo : 
Hello. Thanks for your last message.
1 - I did follow these steps. So I think that rebase is done.

2 - Yes, there are errors, but these errors should not occurs. Because This code compile on a lot of system with gcc. Even on os-x following this : http://free-astro.org/index.php/Siril:install#Installation_via_homebrew_.28recommended_method_since_0.9.5.29.
The difference is that we use gcc with the option `--without-multilib` in Os-x procedure.
However, I wrote a patch that remove one error. But I'm still looking for the other.

lock042:
"Funny things", now the test fails because of 
`checking for GTK... no
configure: error: Package requirements (gtk+-3.0 >= 3.6.0) were not met:

Package 'pango', required by 'gdk-3.0', not found`

But everything was ok just before. I've just added ` => :build` on each dependencies

lock042:
@dpo : I think that could help me : http://stackoverflow.com/questions/11852568/gcc-4-8-on-mac-os-x-10-8-throws-undefined-symbols-for-architecture-x86-64

How I can add the linker "-lstdc++" to my formula ?

lock042:
@dpo : Ok, It seems the only difference with the two method this is that g++ don't have `-std=c++11` option when it fails on homebrew. How I can add this option ??

I tried to add: `args << "cxxflags=-std=c++11"` but it failed.

lock042:
@dpo : I fixed the issue.
I've tested on an OSX machine and this is working well.

siril.rb
```diff
@@ -0,0 +1,68 @@
...
+  patch :DATA
+
+  needs :openmp if build.with? "openmp"
+  needs :cxx11
+
+  def install
+    ENV.cxx11
+
+    ENV.append "CXXFLAGS", "-std=c++11"
+    ENV.append "CXXFLAGS", "-stdlib=libc++" if ENV.compiler == :clang
```
dpo: I don't think these two lines are necessary; `ENV.cxx11` already performs those actions.  Does `siril` not build if you only leave `ENV.cxx11`?

lock042: @dpo : I've copied these lines from homebrew-core/Formula/znc.rb

dpo: @lock042 If you can just remove those two lines and push to your branch, the bots will do the testing.

dpo: ps: I thought you said you had the cxx11 troubles only when using `--with-openmp`. In that case, you can use `needs :cxx11 if build.with? "openmp"` and `ENV.cxx11 if build.with? "openmp"`.

lock042: OK, it's done. I haven't seen the test with the option `--with-openmp` with the bot.

Hope this time it'll be ok and could be merged :).

dpo:
Many thanks. It's minute work to get it all right.

lock042:
@dpo : is it ok for you?

dpo:
Yes, it seems to build for me `--with-openmp`. Here's just a simple fix:
```
needs :cxx11
```
should be
```
needs :cxx11 if build.with? "openmp"
```

Also you could greatly simplify your list of dependencies because `gtk-mac-integration` will install `gtk+3`, which will in turn install `gettext`, `freetype`, `gsettings-desktop-schemas` and `gdk-pixbuf`. So you can remove all of those from your list and only leave `gtk-mac-integration`.

When I launch `siril`, I get
```
Unable to read file: Failed to open file 'siril3.glade': No such file or directory
Unable to read file: Failed to open file '/usr/local/Cellar/siril/0.9.Resources/siril3.glade': No such file or directory
```
That file is here:
```
$ ls /usr/local/Cellar/siril/0.9.4/share/siril/
gtk.css      gtk_old.css  pixmaps      siril3.glade
```
Homebrew refers to that location as `pkgshare`. I'm not sure how to tell Siril where to find it.

lock042:
@dpo : thx for the tip. I did it.

Indeed, you may have some "error" messages at start. This is normal, do not worry ;)

dpo:
Great. Thank you for your patience and contribution!

lock042:
@dpo : no problem. Very happy to see siril in homebrew/science (even if one check failed?).
Moreover, next release will be more integrated to os-x :).

dpo:
The Linux failure is related to `libepoxy`, not to `siril`. I've seen it before.

lock042:
@dpo : I think we should add this:

`depends_on "gcc"  if build.with? "openmp"`

Don't you think ?

dpo:
Homebrew will automatically select gcc if you build with openmp.

lock042:
@dpo : I've experienced the case where:
`brew install siril --with-openmp`
is not possible because we need to install gcc first. Could we make it automatic ?

PR #4550, opened 2016-11-01
transdecoder 3.0.1

ilovezfs:
vendor URI::Escape
install TransDecoder.lrgTests submodule
run the tests


ilovezfs:
CC @sjackman

ilovezfs:
@sjackman :penguin:

PR #4561, opened 2016-11-03
ampl-mp: rename from asl

dpo:
The name `ampl-mp` would be more correct as libasl is only a part of this package.


tkralphs:
I guess we should change the name in all the recipes [here](https://github.com/coin-or-tools/homebrew-coinor)?

dpo:
We should. The `depends_on` is still ok but for some reason, the `Formula["asl"]` isn't handled properly.

PR #4569, opened 2016-11-04
ome-files 0.2.3

rleigh-codelibre:
New patch releases for ome-xml and ome-files.

### Have you:

- [*] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [*] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [*] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [*] Built your formula locally prior to submission with `brew install <formula>`?

- [N/A] Removed the `revision` line, if any, when bumping the version number?
- [N/A] Added/bumped the `revision` number if the changes affect the precompiled bottles?
- [*] Not altered the `bottle` section?
- [*] Checked that the tests still pass?


rleigh-codelibre:
Note that the travis linuxbrew failure is most likely due to homebrew automatically adding `-DCMAKE_BUILD_WITH_INSTALL_RPATH=1 `.  Why is that added when it's stripped on install?  This is resulting in the test failures since the test binaries can't find the shared libraries in the build tree.

dpo:
`ome-common` has a bunch of failed tests on Linux: https://travis-ci.org/Homebrew/homebrew-science/jobs/173396623#L1636

rleigh-codelibre:
@dpo I commented on this failure above.  homebrew on linux is passing a cmake option which alters the linker behaviour.  This looks like something which changed in the last week, and is I think a bit questionable.

rleigh-codelibre:
@dpo I've opened https://github.com/Linuxbrew/brew/issues/171 regarding the linuxbrew build failure.  I'm a bit limited in what I can do to work around that in this package.  I could ignore the default cmake settings when building on Linux, but it would be better for Linuxbrew to fix this to have a sane build/test environment.

sjackman:
See PR #172.

rleigh-codelibre:
@sjackman Unfortunately, https://github.com/Linuxbrew/brew/pull/172 goes too far the other way I think.  This doesn't help with this PR (or existing formulae, e.g. ome-common).

Ideally, we want to set the RPATH when we perform "make install", by setting `CMAKE_INSTALL_RPATH`.  But we don't want this to take effect until that point.  That is to say, to avoid using `CMAKE_BUILD_WITH_INSTALL_RPATH`.  At build time, we need the RPATH to be using the cmake default of the build tree, and set `DYLD_FALLPATH_LIBRARY_PATH` when running the build and tests so that the external library dependencies can be satisfied.  I had an attempt at doing this, but didn't succeed so far.

If you try building the `ome-common` formula currently in homebrew-science, you'll see the problem straight off.

ilovezfs:
The resource blocks should be upgraded.

rleigh-codelibre:
@ilovezfs The python sphinx/docutils blocks?  There's a 1.2.3 sphinx dependency here, which prevent upgrading with this source release.  It's lifted in upstream git for the upcoming 0.3.0 release.

ilovezfs:
```
  resource "Sphinx" do
    url "https://files.pythonhosted.org/packages/fd/87/c4dbc660a2c36ccd7cbbeb6cd611167f052ab4f4d32f542bdcaabef99276/Sphinx-1.2.3.tar.gz"
    sha256 "8e6a77a20b2df950de322fc32f3b508697d9d654fe984e3cc88f446a5b4c17c5"
  end

  resource "docutils" do
    url "https://files.pythonhosted.org/packages/05/25/7b5484aca5d46915493f1fd4ecb63c38c333bd32aa9ad6e19da8d08895ae/docutils-0.13.1.tar.gz"
    sha256 "718c0f5fb677be0f34b781e04241c4067cbd9327b66bdd8e763201130f5175be"
  end

  resource "Jinja2" do
    url "https://files.pythonhosted.org/packages/5f/bd/5815d4d925a2b8cbbb4b4960f018441b0c65f24ba29f3bdcfb3c8218a307/Jinja2-2.8.1.tar.gz"
    sha256 "35341f3a97b46327b3ef1eb624aadea87a535b8f50863036e085e7c426ac5891"
  end

  resource "MarkupSafe" do
    url "https://files.pythonhosted.org/packages/c0/41/bae1254e0396c0cc8cf1751cb7d9afc90a602353695af5952530482c963f/MarkupSafe-0.23.tar.gz"
    sha256 "a4ec1aff59b95a14b45eb2e23761a0179e98319da5a7eb76b56ea8cdc7b871c3"
  end

  resource "Pygments" do
    url "https://files.pythonhosted.org/packages/b8/67/ab177979be1c81bc99c8d0592ef22d547e70bb4c6815c383286ed5dec504/Pygments-2.1.3.tar.gz"
    sha256 "88e4c8a91b2af5962bfa5ea2447ec6dd357018e86e94c7d14bd8cacbc5b55d81"
  end
```

ilovezfs:
Also, minimum os requirement can be modernized as `depends_on :macos => :mavericks`

sjackman:
> Ideally, we want to set the RPATH when we perform "make install", by setting CMAKE_INSTALL_RPATH. But we don't want this to take effect until that point. That is to say, to avoid using CMAKE_BUILD_WITH_INSTALL_RPATH. At build time, we need the RPATH to be using the cmake default of the build tree, and set DYLD_FALLPATH_LIBRARY_PATH when running the build and tests so that the external library dependencies can be satisfied. I had an attempt at doing this, but didn't succeed so far.

If you have a question regarding Linuxbrew that I can help answer, please fire away.

rleigh-codelibre:
Closing this PR to replace with a more up to date set of releases.

rleigh-codelibre:
Superseded by https://github.com/Homebrew/homebrew-science/pull/5964

PR #4593, opened 2016-11-13
graph-tool: Bump version to 2.19

count0:


graph-tool.rb
```diff
@@ -4,15 +4,8 @@ class GraphTool < Formula
   revision 4
```
jonchang: I suspect the audit failure is due to this line, since the [contribution guidelines](https://github.com/Homebrew/homebrew-science/blob/master/.github/PULL_REQUEST_TEMPLATE.md) ask you to remove the `revision` on new versions.

jonchang:
Thanks @count0!

PR #4300, opened 2016-09-13
hdf5 1.10.0-patch1

ilovezfs:
Created with `brew bump-formula-pr`.

PR #4599, opened 2016-11-16
Fix broken hdf5.rb to point to latest release

chrisidefix:
Formula is currently pointing to non-existing HDF5 source file: `hdf5-1.8.17.tar.bz2`, which has been removed from the HDF5 server and was replaced with hdf5-1.8.18.tar.bz2 on 2016-11-14.

I am aware of: https://github.com/Homebrew/homebrew-science/pull/4300 but that PR seems idle and requires additional work. It would be worth merging this in the mean-time to continue having a working HDF5 formula.

### DIFF
--- a/hdf5.rb
+++ b/hdf5.rb
@@ -1,8 +1,8 @@
 class Hdf5 < Formula
   desc "File format designed to store large amounts of data"
   homepage "http://www.hdfgroup.org/HDF5"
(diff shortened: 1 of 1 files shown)


jonchang:
Thanks @chrisidefix. I also fixed the audit warning, please be sure to run `brew audit --strict --online hdf5` as our audit checks are always getting updated.

sjackman:
@sjackman 🐧

PR #4605, opened 2016-11-17
Therion: add tcl-tk dependency

un1x86:
Sorry for the mess. I closed the last pull-request and created a new one.

### DIFF
--- a/therion.rb
+++ b/therion.rb
@@ -10,7 +10,7 @@ class Therion < Formula
   depends_on :tex
   depends_on "vtk"
   depends_on "wxmac"
-  depends_on "homebrew/dupes/tcl-tk" if MacOS.version == :sierra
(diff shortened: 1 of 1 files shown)


therion.rb
```diff
@@ -9,6 +10,7 @@ class Therion < Formula
   depends_on :tex
   depends_on "vtk"
   depends_on "wxmac"
+  depends_on "homebrew/dupes/tcl-tk" if MacOS.version == :sierra
```
jonchang: Should this be `> :sierra`? Assuming that Apple won't ship tcl-tk in the future?

un1x86: Could be >= :sierra if Apple won't ship a newer version of tcl-tk in the future I guess.

jonchang:
Thanks @un1x86!

PR #4622, opened 2016-11-18
trinity 2.3.1-beta (devel)

ilovezfs:
CC @sjackman

### DIFF
--- a/trinity.rb
+++ b/trinity.rb
@@ -1,11 +1,21 @@
 class Trinity < Formula
   desc "RNA-Seq de novo assembler"
   homepage "https://trinityrnaseq.github.io"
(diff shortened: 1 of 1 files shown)


sjackman:
I'm not inclined to add `devel` unless you yourself have a need for it or someone has requested it. I'm not particularly against it though. Looks good to me.

ilovezfs:
@sjackman nope, I don't need it. This is just to dust it off and make sure it works before the new release actually ships. I already hit one issue in 2.3.0-prerelease, which upstream quickly fixed (https://github.com/trinityrnaseq/trinityrnaseq/issues/181), so with any luck 2.3.1-prerelease will soon be promoted.

ilovezfs:
We can dig this PR up once 2.3.1 ships as stable.

sjackman:
Superseded by #4629

PR #4623, opened 2016-11-20
mummer: audit fixes

ilovezfs:
also, use assert_match and pipe_output in the test instead of grep


ilovezfs:
@sjackman :penguin:

PR #4624, opened 2016-11-20
spades: apply jemalloc fix for Sierra

ilovezfs:
This fix is already in the latest release of jemalloc, but hasn't made
its way over to the vendored version of jemalloc in SPAdes yet.


ilovezfs:
Reported upstream: https://github.com/ablab/spades/issues/9

PR #4628, opened 2016-11-21
ceres-solver 1.12.0rc4

### LINKED ISSUE #4603: ceres-solver bottle uses old Eigen version

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Checked that a similar issue hasn't already been filed?

#### Brief summary of issue:
This appears to be the same as issue 3195 and 3454, but now the latest version of Eigen is 3.3.0, and ceres-solver is built against 3.2.10.

#### The complete command that you issued:
brew install ceres-solver

ilovezfs:
use release candidate as stable for the sake of eigen
see https://github.com/ceres-solver/ceres-solver/issues/240.

re-enable "make test" on Linux

Fixes #4603.

CC @dpo @jonchang @ayonsen

### DIFF
--- a/ceres-solver.rb
+++ b/ceres-solver.rb
@@ -1,11 +1,10 @@
 class CeresSolver < Formula
   desc "C++ library for large-scale optimization"
   homepage "http://ceres-solver.org/"
-  url "http://ceres-solver.org/ceres-solver-1.11.0.tar.gz"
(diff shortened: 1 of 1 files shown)


ilovezfs:
Audit passed on Linux, so I guess Linuxbrew doesn't have the audit check for "rc" in stable URLs.

PR #4629, opened 2016-11-21
trinity 2.3.2

ilovezfs:
java requirement is now 1.8
reparallelize the build
gcc6 fix is no longer needed
scope -fopenmp fix to Linux

CC @sjackman


ilovezfs:
@sjackman :penguin:

PR #4657, opened 2016-11-29
New upstream version siril 0.9.5

lock042:
New upstream version available.

New dependencies to libffmpeg but it seems that ffms2 also download these dependencies ?


jonchang:
Thanks @lock042!

PR #4674, opened 2016-12-05
[RFC] Fix travis failure

equal-l2:
Copied almost all from [Linuxbrew/homebrew-core](https://github.com/Linuxbrew/homebrew-core/blob/master/.travis.yml).


equal-l2:
Hmm... Why?

equal-l2:
The cause (line 767)
```
Removing Library/Homebrew/vendor/portable-ruby/
```

PR #4684, opened 2016-12-07
phlawd: revision bump for gcc 6

jonchang:
The current bottle for Sierra fails `brew linkage`:

```
      Missing libraries:
/usr/local/lib/gcc/6/libgcc_s.1.dylib
/usr/local/opt/gcc/lib/gcc/6/libgomp.1.dylib
/usr/local/opt/gcc/lib/gcc/6/libstdc++.6.dylib
```


ilovezfs:
The libraries haven't changed, so this revision bump isn't needed. The question would be why aren't you (or test-bot?) not being prompted to install gcc-6 in the particular scenario.

jonchang:
Thanks for the pointer! I figured it out. I updated `muscle` in #4673 and saw a build failure with that message. Looks like `phlawd` must build with and link against `gcc` due to `needs :openmp`. `muscle` can use `clang` and therefore won't `brew install gcc`, but `test-bot` doesn't install the reverse dependencies' own dependencies when doing linkage tests and therefore the linking fails.

ilovezfs:
Yup, happens quite a bit. :(

PR #4701, opened 2016-12-11
cmdstan: depend on eigen32

ilovezfs:


ilovezfs:
@BrewTestBot test this please

PR #4702, opened 2016-12-11
cppad: depend on eigen32 optionally

ilovezfs:
instead of eigen


ilovezfs:
let's wait for the next Homebrew tag to ship first, since it will fix option deprecations which are currently broken

ilovezfs:
@BrewTestBot test this please

PR #4703, opened 2016-12-11
dgtal: depend on eigen32 optionally for stable

ilovezfs:
retain homebrew/core/eigen for head, which has eigen 3.3.x support


ilovezfs:
let's wait for the next Homebrew tag to ship first, since it will fix option deprecations which are currently broken

ilovezfs:
@BrewTestBot test this please

PR #4705, opened 2016-12-11
travis.yml: don't tap anything.

MikeMcQuaid:
These are immediately removed by test-bot because it's replicating end-user behaviour where we cannot set up dependent taps. If this breaks formulae they need dependencies to be fully scoped to e.g. `homebrew/versions/eigen32`.


ilovezfs:
This causes all `depends_on :x11` to fail for Linux because the default formula for :x11 is `linuxbrew/xorg/xorg` and for some reason

```
depends_on "linuxbrew/xorg/xorg"
```
will work but
```
depends_on :x11
```
won't. So apparently the logic for tapping the fully scoped formulae breaks down for requirements. Not sure what to advise here, other than moving linuxbrew/xorg/xorg to Linuxbrew/homebrew-core.

PR #4708, opened 2016-12-11
Update siril.rb, make openmp enabled by default

lock042:
I want openmp enabled by default. Indeed, openmp is more than an option it increases the speed of siril a lot.


lock042:
@ilovezfs: I saw you tried some commits with no success. Does my patch will be pulled anyway ?
Cheers.

ilovezfs:
@lock042 indeed. I was trying to determine the source of the Linux breakage. But it's preexisting. I filed

https://github.com/Linuxbrew/homebrew-xorg/issues/228

which will need to be resolved first unless someone feels like doing local testing.

lock042:
So this update of siril can't be released yet ?

ilovezfs:
It can.

ilovezfs:
need to revision bump it too. i'll do that now.

lock042:
Thank you so much @ilovezfs

lock042:
@ilovezfs: that strange because this time the other test also failed with:
` * The homepage is not reachable (curl exit code 7)`
If this is the siril homepage, I do not understand as the website is currently reachable.

ilovezfs:
@lock042 probably just a fluke. We can ignore it.

lock042:
@ilovezfs glad to hear that.
Thanks again for your help, I'm very happy to have siril by default with opencv.
That's a great improvement !!!
Cheers,

ilovezfs:
@lock042 shipped! Thanks for the PR. You can brew update and install the new version to test the speed difference!

PR #4719, opened 2016-12-13
biomake 0.1.0 (new formula)

sjackman:
GNU-Make-like utility for managing builds and complex workflows
https://github.com/evoldoers/biomake


sjackman:
`GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)`

PR #4721, opened 2016-12-13
sratoolkit: move examples from share to pkgshare

ilovezfs:


ilovezfs:
@BrewTestBot test this please

ilovezfs:
@sjackman :penguin:

PR #4727, opened 2016-12-15
Openfst 1.5.4

stevenbedrick:
Just a version bump for `openfst` and `opengrm-ngram`.

### Have you:

### DIFF
--- a/openfst.rb
+++ b/openfst.rb
@@ -1,7 +1,8 @@
 class Openfst < Formula
+  desc "Open-source library for working with weighted finite-state transducers."
   homepage "http://www.openfst.org/"
-  url "http://openfst.org/twiki/pub/FST/FstDownload/openfst-1.5.3.tar.gz"
-  sha256 "9b09e457aeab87f613508b92a0f9f820140c9e18d05584e3f1ae384396b5dcbd"
+  url "http://openfst.org/twiki/pub/FST/FstDownload/openfst-1.5.4.tar.gz"
+  sha256 "acc115aaaa33de53de62dae44120ab368fabaea06f52606b77714081ecd32657"
 
   bottle do
     cellar :any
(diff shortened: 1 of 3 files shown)


stevenbedrick:
OK, the Jenkins output shows that a dependency of `openfst`, `opengrm-thrax`, is failing to build with the updated version of `openfst`- but I checked that, and it's building fine from source on my El Capitan system. So I'm not sure what's going on...

jonchang:
Looks like you need to do a revision bump in `opengrm-thrax`. Just add a `revision 1` line in `opengrm-thrax.rb`

stevenbedrick:
Oh, thanks- I'd read the docs on what that `revision` business was about but must've misunderstood it. Let's see if that fixes it!

stevenbedrick:
OK, it's failing on the `brew audit opengrm-thrax` step, but on my local machine, that step works just fine. Any suggestions would be very appreciated...

jonchang:
Just have to update the test block. 

```
homebrew/science/opengrm-thrax:
  * fully scope test system calls e.g. system "#{bin}/thraxmakedep"
  * fully scope test system calls e.g. system "#{bin}/thraxrandom-generator"
```

stevenbedrick:
I see, thanks- but what about line 30, which is :

    system "make"

My understanding is that `#{bin}` will refer to the Homebrew bin directory; how do I tell it to point at the system `make` executable? Sorry, I know it's a simple question...

jonchang:
It'll only check for unscoped calls to binaries that the formula itself installs, so don't worry calls to system tools like `make` and such.

stevenbedrick:
OK! Thanks so much for the answer, that makes sense. Let's see if that does the trick...

jonchang:
Thanks! This is merged.

PR #2770, opened 2015-09-23
Netcdf 4.3.3.1 (fix)

davydden:
Fix for linuxbrew when compiling with fortran.

PR #4735, opened 2016-12-16
netcdf: revision for hdf5

ilovezfs:


sbfnk:
brew audit fails as discussed before: https://github.com/Homebrew/homebrew-science/pull/2770#issuecomment-142831686

ilovezfs:
@sbfnk yes davydden is quite correct lol.

PR #4740, opened 2016-12-19
blast: remove install workaround

iMichka:
The patch is already merged upstream and will be present
in release 2.6.0. This removes the need for the install
workaround.

### Have you:


iMichka:
Brew audit complains about:
> 'fails_with :llvm' is now a no-op so should be removed

@MikeMcQuaid This is due to that change in https://github.com/Homebrew/brew/pull/1673
The thing is, I have no clue what should be used to replace these lines :) Could you maybe give me a hint?

jonchang:
You can just take the whole `fails_with :llvm` block out and it should work the same.

iMichka:
Done. I just double-checked, 2335 was an older build (under 10.7); so it is no more supported anyway by homebrew. I was troubled because I was thinking: how can the formula fail with that llvm version if I remove that information. But if it is not supported anyway, removal is the way to go :). Thanks

MikeMcQuaid:
> But if it is not supported anyway, removal is the way to go :). Thanks

Yep, we never use that compiler now.

iMichka:
The build failures seem unrelated, some other formula could not build but for different reasons. I think this is good to go for now.

jonchang:
@BrewTestBot test this please

blast.rb
```diff
@@ -38,6 +33,12 @@ class Blast < Formula
   depends_on "pcre" => :recommended
   depends_on :python if MacOS.version <= :snow_leopard
 
+  patch do
+    # Fixed upstream in future version 2.6.0
+    url "https://gist.githubusercontent.com/iMichka/81ac6de72010c7812a47fc277dd21cb3/raw/80acfe731d2e4373b33f76b3da9aec2192b559b2/blast25.patch"
```
ilovezfs: Please open a PR to https://github.com/Homebrew/formula-patches with this :)

iMichka:
Done. The build errors are unrelated (these are other packages failing to install due to missing dependencies).

jonchang:
Thanks @iMichka!

PR #4748, opened 2016-12-22
cgatools 1.8.0 (new formula) osx&linux

BenjaminHCCarr:
This is the Complete Genomics Analysis Tools, 1.8.0 is current:
http://cgatools.sourceforge.net/
http://www.completegenomics.com/public-data/analysis-tools/cgatools/

However it is impossible to compile this from source as the source code specifically calls for`BOOST_FILESYSTEM_VERSION 2`

This was a known issue as `cgatools-1.8.0.1-source/cgatools/core.hpp:[43-46]` notes:
```// Use version 2 of the Boost Filesystem library.
// This stopped being the default as of Boost 1.44 and
// will no longer be supported in Boost 1.48.
#define BOOST_FILESYSTEM_VERSION 2
```
boost is currently at `1.62.0`. `1.48.0` was released in 2011.

However Complete Genomics data are just as common now as they were then, if not more so. This binary version is the only way to install `cgatools` on a modern machine without also using a version of `boost-1.47.0` around simply to compile this tool.

While binaries are discouraged in `homebrew-core` I am requesting that this be merged into `homebrew-science` as a binary distribution only to simplify those of us doing bioinformatics.

### Have you:

- [*] Built your formula locally prior to submission with `brew install <formula>`?

### New formula: have you


jonchang:
Hi @BenjaminHCCarr, thanks for contributing to Homebrew! I think we would prefer to not have binary-only formulae. There are two ways to build from source, (preferring the first):

1. Port `cgatools` to use boost filesystem v3. Looking through the source code and the boost docs it seems pretty straightforward to do so.
2. Build a custom copy of boost and link to it. `cd $(brew --repo homebrew/core) && grep 'resource "boost"' -r .` for examples. 

Feel free to ask questions if you need help with either of these steps.

BenjaminHCCarr:
@jonchang thanks for at least commenting; I have not coded against `boost` before; butI am going to take a shot at it: [BenjaminHCCarr/cgatools](https://github.com/BenjaminHCCarr/cgatools/)

It appears even @galaxyproject is just pulling in the binary tarballs in their deployment so there is definitely a need.

I tweeted about it here: https://twitter.com/BenjaminHCCarr/status/812025627935207424 but so far no traction or interest. If I have to go it alone I will, it will just take longer.

I'm going to close this and work on the port in my spare-spare-time.

Issue #4399, opened 2016-10-04
regression: brew upgrade homebrew/science/opencv fails


timotheecour:
(used to work)
ubuntu 16.04:
brew gist-logs homebrew/science/opencv
https://gist.github.com/5949e4f2a2b1e304e82dc28b83514239

ubuntu 14.04:
brew gist-logs homebrew/science/opencv
https://gist.github.com/5ed499c3abfd54df8871fa3206dec251

==> Upgrading homebrew/science/opencv

```
==> Upgrading homebrew/science/opencv
==> Downloading https://github.com/opencv/opencv/archive/2.4.13.tar.gz
Already downloaded: /home/timothee/.cache/Homebrew/opencv-2.4.13.tar.gz
Usage: /home/timothee/.linuxbrew/bin/python-config [--prefix|--exec-prefix|--includes|--libs|--cflags|--ldflags|--help]
==> cmake .. -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFIX=/home/timothee/.linuxbrew/Cellar/op
==> make
Last 15 lines from /home/timothee/.cache/Homebrew/Logs/opencv/02.make:
[100%] Built target opencv_perf_stitching
[100%] Linking CXX executable ../../bin/opencv_test_contrib
cd /tmp/opencv-20161004-8360-8p02mm/opencv-2.4.13/macbuild/modules/contrib && /home/timothee/.linuxbrew/Cellar/cmake/3.6.2/bin/cmake -E cmake_link_script CMakeFiles/opencv_test_contrib.dir/link.txt --verbose=1
[...]

PR #4750, opened 2016-12-22
fix issue #4399: opencv regression

timotheecour:
brew install opencv has been broken for months. 
Found the culprit:
py_lib = OS.linux? ? `python-config --configdir`.chomp : "#{py_prefix}/lib"
didn't seem to work on my system (as well as for a few other ppl).
This should explain why:
which -a python-config
/home/timothee/.linuxbrew//bin/python-config
/usr/bin/python-config

/usr/bin/python-config --configdir
/usr/lib/python2.7/config-x86_64-linux-gnu

python-config --configdir
Usage: /home/timothee/.linuxbrew//bin/python-config [--prefix|--exec-prefix|--includes|--libs|--cflags|--ldflags|--help]

Without this fix, `brew install opencv` would succeed or fail depending on whether brew had installed python-config


jonchang:
Very odd, as `--configdir` is not an option in Python 2. Thanks for tracking this down @timotheecour!

PR #4767, opened 2016-12-26
sratoolkit fix sha256 error for ncbi-vdb

cmdcolin:
The sha256 checksum was failing on ncbi-vdb and seemed to need updating. I manually checked that by downloading and running shasum -a 256 locally


ilovezfs:
Please open an issue on their tracker asking why this happened and why a new tag was not created instead. As a matter of policy, we do not update checksums without understanding the situation first because upstream may have been hacked.

cmdcolin:
Sounds good!

cmdcolin:
Oh wait, why would it be an issue on their tracker? I guess we can cross reference with them but it seems a little confusng

ilovezfs:
@cmdcolin because it's their tarball ...

cmdcolin:
Looks like they can confirm the retagged tht led to the checksum difference

PR #4772, opened 2016-12-28
kollector: Build a bottle for Linuxbrew

sjackman:


sjackman:
I unintentionally pushed this formula without bottles. Bottling now with this PR.

PR #4775, opened 2016-12-28
anvio 2.1.0 (new formula)

ozcan:
### Have you:

### New formula: have you


ozcan:
Hello Homebrew Science Maintainers,

Can someone review this formula please?

Thank you very much,

meren:
Hi everyone,

Are there any comments/suggestions for this formula? :)

Thanks,

jonchang:
We're trying to move away from providing formulae just for python packages, except in unusual cases (i.e., `numpy` and `scipy`). Is there a reason why `brew install <dependencies> && pip install anvio` is inadequate?

meren:
Hi Jonathan,

Thanks for the response. Having one anvio formula in the brew repository not only will make things much more simpler for users who already use brew to setup their environment, but will make it easier for other developers to add anvio as a dependency in their formulae for seamless integration of the platform into other projects, making things tidier. 

We tried many things to provide our users with an effortless and working installation of anvi'o via pip, Docker, or even a make-shift Mac OS X installer. While each one of them failed one way or another, we have been very happy with homebrew, and we would like to maintain anvio this way.

We are already suggesting our users to follow the homebrew path by doing this temporarily:

http://merenlab.org/2016/06/26/installation-v2/#painless-installation-with-homebrew

All these aside, we understand if there is a strategic decision on the homebrew side that makes our PR unsuitable, in which case we can take our formula and go fly a kite :)



Thanks,

sjackman:
@meren You can simplify:
```
brew tap homebrew/homebrew-science
brew install wget
wget https://raw.githubusercontent.com/ozcan/homebrew-science/eda73cbcf0d7eabed8eba258ce13540ff1d3a648/anvio.rb -O /usr/local/Homebrew/Library/Taps/homebrew/homebrew-science/anvio.rb
brew install anvio
```
to
```sh
brew install https://github.com/ozcan/homebrew-science/raw/eda73cbcf0d7eabed8eba258ce13540ff1d3a648/anvio.rb
```

sjackman:
Can you expand on why it's not possible to install `anvio` using `pip`?

tdsmith:
This looks like a good candidate for Homebrew to me; it depends on a bunch of hairy binary dependencies and it's being used as an application, not a library. Homebrew routinely accepts applications written in Python.

You should try to rewrite this to use the virtualenv support; jrnl.rb in core is a good example.

anvio.rb
```diff
@@ -0,0 +1,99 @@
...
+  depends_on "homebrew/homebrew-science/muscle"
+  depends_on "homebrew/homebrew-science/blast"
+
+  resource "scipy" do
+    url "https://github.com/scipy/scipy/releases/download/v0.18.1/scipy-0.18.1.tar.xz"
+    sha256 "406d4e2dec5e46ad9f2ab08620174d064d3b5aab987591d1acd77eda846bd95e"
+  end
+
+  resource "numpy" do
+    url "https://pypi.python.org/packages/d0/30/36316dadbbc71cbdca0aceb5e098ae21601d6f2336a390c34082cd2bf0c5/numpy-1.12.0rc1.zip"
```
tdsmith: Is a released version acceptable?

ozcan:
Hi @sjackman ,

We have not been happy with pip because even though pip works well small libraries it often fails for applications with many dependencies, and it becomes impossible for end users to get away from conflicts. To minimize those conflicts we suggest our users to install anvi'o using virtualenv, but many of them are not familiar with virtualenv, and source/deactivate workflow confuses them.

For instance we have a C extensions which requires numpy and Cython to be installed before setup.py is run. Even though numpy and Cython are listed as dependencies, pip way of resolving dependencies does not guarantee that they will be installed with the correct order, and we keep getting questions from uses asking about failed pip installations.

We came to the conclusion that brew is really best option to install anvi'o for both the end users and for us.

Thank you very much,

tdsmith:
osc in core is another example of a formula that uses virtualenv encapsulation in a more flexible way: https://github.com/tdsmith/homebrew-core/blob/24d8ef1db525317e3005bcb85a8a0d3db22a35b2/Formula/osc.rb

jonchang:
Thanks for contributing to Homebrew @ozcan! This has been merged.

meren:
Thank you very much for all the input and help, everyone!

Issue #4788, opened 2017-01-01
brew install opencv3: error: '_mm_cvtph_ps' was not declared in this scope


timotheecour:
```
brew install opencv3:
opencv-3.2.0/modules/core/include/opencv2/core/hal/intrin_sse.hpp:1721:42: error: '_mm_cvtph_ps' was not declared in this scope
```

Even after doing the same fix to opencv3.rb as I did to opencv.rb here:

https://github.com/Homebrew/homebrew-science/commit/b0350e62de1deb3ccf20face10e11c1df1bd4820

This is therefore a new error different from https://github.com/Homebrew/homebrew-science/issues/4400

timotheecour:
@jonchang please reopen, if you re-read the comments https://github.com/Homebrew/homebrew-science/commit/b0350e62de1deb3ccf20face10e11c1df1bd4820 was necessary (for a different reason as this bug) but does not fix this bug, I am still getting:

```
/tmp/opencv3-20170102-23654-1onu84u/opencv-3.2.0/modules/core/include/opencv2/core/hal/intrin_sse.hpp:1721:42: error: '_mm_cvtph_ps' was not declared in this scope
```

Issue #4400, opened 2016-10-04
regression: brew upgrade homebrew/science/opencv3 fails


timotheecour:
brew gist-logs homebrew/science/opencv3
https://gist.github.com/e00f6e3f21eebb9f5c022f241612ec44

(used to work)

```
Usage: /home/timothee/.linuxbrew/bin/python-config [--prefix|--exec-prefix|--includes|--libs|--cflags|--ldflags|--help]
==> Downloading https://raw.githubusercontent.com/opencv/opencv_3rdparty/81a676001ca8075ada498583e4166079e5744668/ippicv/ippicv_linux_20151201.tgz
Already downloaded: /home/timothee/.cache/Homebrew/opencv3--icv-linux-20151201.tgz
==> cmake .. -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFIX=/home/timothee/.linuxbrew/Cellar/opencv3/3.1.0_4 -DCMAKE_BUILD_TYPE=Release -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_VERBOSE_MAKEFILE=ON -Wno-dev -DCMAK
==> make
Last 15 lines from /home/timothee/.cache/Homebrew/Logs/opencv3/02.make:
make[2]: Leaving directory '/tmp/opencv3-20161004-6085-1tszp1q/opencv-3.1.0/macbuild'
[100%] Built target opencv_traincascade
cd /tmp/opencv3-20161004-6085-1tszp1q/opencv-3.1.0/macbuild && /home/timothee/.linuxbrew/Cellar/cmake/3.6.2/bin/cmake -E cmake_depends "Unix Makefiles" /tmp/opencv3-20161004-6085-1tszp1q/opencv-3.1.0 /tmp/opencv3-20161004-6085-1tszp1q/opencv-3.1.0/modules/python/python2 /tmp/opencv3-20161004-6085-1tszp1q/opencv-3.1.0/macbuild /tmp/opencv3-20161004-6085-1tszp1q/opencv-3.1.0/macbuild/modules/python2 /tmp/opencv3-20161004-6085-1tszp1q/opencv-3.1.0/macbuild/modules/python2/CMakeFiles/opencv_python2.dir/DependInfo.cmake --color=
Scanning dependencies of target opencv_python2
[...]

PR #4789, opened 2017-01-01
fix issue 4400 (analog as fix for 4399)

timotheecour:
Same as https://github.com/Homebrew/homebrew-science/commit/b0350e62de1deb3ccf20face10e11c1df1bd4820#diff-ebaa128f3fee9eff3ca3d844be60b7fd that fixed issue #4399 [opencv2] but for https://github.com/Homebrew/homebrew-science/issues/4400 [opencv3]

However, due to a recent change in opencv3 brew formula, this is necessary but not sufficient, see: https://github.com/Homebrew/homebrew-science/issues/4788


jonchang:
Thanks @timotheecour.

PR #4797, opened 2017-01-02
Delete patch for bug fixed (bug #41027)

rubenrua:
See http://savannah.gnu.org/bugs/?41027


jonchang:
Hi @rubenrua, thanks for your contribution, but I'm not certain I understand. Does you mean that the bug was already fixed in a previous version, but the patch in the formula wasn't properly removed?

rubenrua:
Yes. I found the issue installing Octave tonight. And the patch inside the formula is not needed because it was fixed in the source of Octave.

jonchang:
Thanks @rubenrua!

PR #2407, opened 2015-06-13
uproc 1.2 (version update + fix Issue #2406)

tseemann:

PR #4798, opened 2017-01-02
uproc 1.2

jonchang:
Tarballs on the website are down so using Github releases. Now also needs autotools to build.


jonchang:
See also https://github.com/Homebrew/homebrew-science/pull/2407 @tseemann

PR #4801, opened 2017-01-04
update armadillo.rb to 7.600.2

conradsnicta:
@jonchang  @ilovezfs @dpo update armadillo.rb to use Armadillo 7.600.2, which fixes a memory leak when using C++11/C++14.


jonchang:
Thanks for contributing to Homebrew, @conradsnicta! This has been merged.

PR #4819, opened 2017-01-05
cudainfo 0.9 (new formula)

pwnall:
This is a `cpuinfo` equivalent for CUDA cards whose output happens to be both machine-readable and (sort of) human-friendly.

I wrote this tool to automate the step of figuring out what compute version a CUDA card has, which is needed e.g., to build [tensorflow](https://www.tensorflow.org/). The manual instructions included in most CUDA libraries involve going to [a nvidia page](https://developer.nvidia.com/cuda-gpus) and finding the computer's GPU card in one of the tables there. I hope that this tool will eventually mean that no scientist ever has to visit that page again to get an ML framework set up.

So, while this tool isn't scientific on its own, I expect that it will be used to install other software that belongs in homebrew-science, such as caffe and tensorflow. For example, I am using it in my [work-in-progress tensorflow formula](https://github.com/pwnall/homebrew-science/blob/tensorflow/tensorflow.rb).

I hope that you'll accept this formula, as I think it will make people's lives easier. I look forward to your feedback!

### Have you:

### New formula: have you

### DIFF
--- a/cudainfo.rb
+++ b/cudainfo.rb
@@ -1,0 +1,27 @@
+require File.expand_path("../Requirements/cuda_requirement", __FILE__)
+
+class Cudainfo < Formula
+  desc "CLI tool that shows the properties of installed CUDA GPUs"
+  homepage "https://github.com/pwnall/cudainfo"
+
+  stable do
+    url "https://github.com/pwnall/cudainfo/archive/v0.9.tar.gz"
+    sha256 "bf58ca69df86d9b7ff4948fd5fda249410c7ee0677fb2bef6664f8b0399b33fc"
+  end
+
+  head do
+    url "https://github.com/pwnall/cudainfo.git"
+  end
+
+  depends_on CudaRequirement
+
+  def install
+    system "make", "cudainfo"
+    bin.install "cudainfo"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


fxcoudert:
There are several projects or source files named "cudainfo" throughout github, and this one is does not seem widely used or famous in any respect. It's a self-submission of a repo with 1 commit 20 days ago, no other activity. How did it pass the `brew audit` online tests?

pwnall:
@fxcoudert Sorry, while explaining why the submission is in homebrew-science, I forgot that I also need to explain that it doesn't meet the popularity threshold yet.

FWIW, I wrote the whole thing to automate the installation of tensorflow ([my formula here](https://github.com/pwnall/homebrew-science/blob/tensorflow/tensorflow.rb)) and I imagine that others will need the tool.

Would it be more correct for me to bundle this in the PR I'll use for tensorflow?

Thank you very much for your help!

jonchang:
Yeah if it's only used in one formula in homebrew/science you might just want to bundle it instead. You might also consider [starting and maintaining your own tap](http://docs.brew.sh/How-to-Create-and-Maintain-a-Tap.html) if you are going to be making a lot of cuda-related formulae.

jonchang:
I don't think we'll be taking this formula at this time.

PR #4841, opened 2017-01-11
added needed dependency htslib because of tabix

StephanJanosch:
PR because of missing dependency

### DIFF
--- a/vcftools.rb
+++ b/vcftools.rb
@@ -6,7 +6,7 @@ class Vcftools < Formula
 
   url "https://github.com/vcftools/vcftools/archive/v0.1.13.tar.gz"
   sha256 "0e241da57bc7048161d3751a1be842ad36e6a43f803c91cc9ef18aa15b3fc85e"
-
+  revision 1
   head "https://github.com/vcftools/vcftools.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


iMichka:
Hi. Thanks you for your contribution to Homebrew.

This does not seem to be a build dependency? Does vcftools run without it?

iMichka:
Just for ref, #4774 is about updating to v0.1.14, but this is a different topic.

StephanJanosch:
vcf-concensus needs tabix. It fails with a not so easy to read error message of tabix is missing. It's a perl script. So you won't see a build dependency.

iMichka:
Thanks for the explanation.

Per http://docs.brew.sh/Formula-Cookbook.html, could you rewrite this as?:
`depends_on "htslib" => :run`
so we make clear that this a runtime dependency.

Per http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html, take the opportunity to update the commit message:
- First line should be `vcftools: add htslib runtime dependency`

You sort of took a shortcut here by removing all the pull request template, which was there to remind you to follow the submitting conventions.

Please add a line in the commit message too, describing why we need htslib (because of tabix). This helps in case somebody looks at the git log later.

You can `git commit --amend` your changes and push them to GitHub with `git push --force` to have only a single commit.

StephanJanosch:
> You sort of took a shortcut here by removing all the pull request template

Sorry for that, but I was super busy that day. And I am still super busy. I had the choice of simply doing it with my minimal git knowledge or not to do it. Let's see when I will have time for this.

iMichka:
I took over in #5261 and added the dependency. Closing here.

Issue #4843, opened 2017-01-11
netCDF and authentication/authorization


lesserwhirls:
When accessing remote data with `netCDF` tools, like `ncdump`, users occasionally need to be able to use client side certificates for authentication and/or authorization. Currently, the homebrew build of netCDF links to the system copy of `libcurl`, which isn't built against `openssl` (in favor of `SecureTransport` alone). The result is that certificates, in PEM format, are not picked up properly by the `SecureTransport` libs and authentication/authorization fails.

If `netCDF` is built against the homebrew version of `libcurl` (supplied by the `curl` package), then we should be good, as it is built against `openssl`. Note that the homebrew version of `curl` is built against `openssl`, and curl commands to servers requiring the authorization/authentication work as expected.

Not sure if this is as simple as adding a dependency on `curl`, as I am not familiar with homebrew recipes.

edited - I meant curl, not wget.

iMichka:
Hi. Thanks for the explanation.

> Not sure if this is as simple as adding a dependency on curl, as I am not familiar with homebrew recipes.

You can build with the `-v` flag to see what is going on. Maybe the logs will help you see with which curl netcdf gets compiled. And did you try this yourself? You seem not sure that this will fix the problem.

lesserwhirls:
Hi @iMichka - I wasn't sure it would work until I tried it locally, and it does indeed work. I thought a patch would be more involved than one line (plus a revision bump), but when it turned out to be that easy, I went ahead and made a PR. I forgot to update this issue to make clear that the PR I submitted links to the correct (i.e. homebrew) version of curl.

iMichka:
For ref PR opened here: #4847

iMichka:
Closing here, see #4847 for discussion and possible solutions.

PR #4847, opened 2017-01-12
netcdf: fixes Homebrew/homebrew-science#4843

lesserwhirls:
### Have you:

This PR does not introduce any new errors.

### DIFF
--- a/netcdf.rb
+++ b/netcdf.rb
@@ -4,6 +4,7 @@ class Netcdf < Formula
   url "ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-4.4.1.1.tar.gz"
   mirror "http://www.gfd-dennou.org/library/netcdf/unidata-mirror/netcdf-4.4.1.1.tar.gz"
   sha256 "4d44c6f4d02a8faf10ea619bfe1ba8224cd993024f4da12988c7465f663c8cae"
+  revision 1
 
   bottle do
     cellar :any
@@ -24,6 +25,7 @@ class Netcdf < Formula
 
   depends_on :fortran => :optional
   depends_on "hdf5"
+  depends_on "curl" => "with-openssl"
 
   resource "cxx" do
     url "https://github.com/Unidata/netcdf-cxx4/archive/v4.3.0.tar.gz"


iMichka:
@BrewTestBot test this please

iMichka:
Hi. Thanks for your contribution. You will need to bump the revision to 1, so that new bottles are built.

lesserwhirls:
Sorry about forgetting that - I didn't see a revision value in the original ruby file, so I was a little confused.

ilovezfs:
depending on curl => "with-openssl" is not really an acceptable practice except under extraordinary circumstances. are we sure there's no other solutions here?

lesserwhirls:
It looks like it is possible to import the certificates into the keychain app, and `SecureTransport` will use it. The particular certs I was using when I found this issue expire every three days, which is a little bit of a pain. There appears to be a bug that's been reported to Apple dealing with using certs specified by a path (as opposed to certs in the keychain app), but I am not sure what the status is:

https://github.com/curl/curl/issues/976#issuecomment-241797370

@petemoore, have you had any correspondence with Apple regarding the `SecureTransport` bug report you made?

petemoore:
Hi @lesserwhirls,

I've been advised by Apple Engineering that my bug (27968953) was a duplicate of bug 27968221, and my bug was promptly closed by Apple on 8 Sep 2016. Unfortunately, I do not currently have access to bug 27968221, so I've no idea what the status is. :/

Today I have asked for access to bug 27968221 from my closed bug 27968953. I'll let you know if I get it. Otherwise, you might be able to request access to these two bugs too.

I could not see obvious contact information to reach Apple to discuss the matter directly.

petemoore:
For reference, the contents of bug 27968953  that I raised:

## Summary
When curl is built with OpenSSL, the SSL_CERT_FILE environment variable when set is interpreted as the CA certificate file to use. However, the curl that Apple ships is built with SecureTransport rather than OpenSSL transport, yet still attempts to use this SSL_CERT_FILE, which results in failure for the command to complete.

This causes problems when a system has more than one version of curl installed, and one requires a certificate bundle, and therefore e.g. sets the SSL_CERT_FILE environment variable e.g. in the user profile. The native version of curl on OS X then consistently fails.

I raised this bug originally against curl in https://github.com/curl/curl/issues/976 but it seems this might be a problem with SecureTransport instead.

## Steps to Reproduce
1. curl -L https://curl.haxx.se/ca/cacert.pem > ~/cacert.pem

2a. curl --cacert ~/cacert.pem -I https://storage.googleapis.com/golang/go1.7.src.tar.gz
 * OR *
2b. SSL_CERT_FILE=~/cacert.pem /usr/bin/curl -I https://storage.googleapis.com/golang/go1.7.src.tar.gz


## Expected Results
Result from step 2a/2b should instead be:
```
HTTP/1.1 200 OK
X-GUploader-UploadID: AEnB2UplWlz4rniGH-7K0_WwhXYZci9pdEYGrZDSyEwTLv9X3rgeLfEDM2NLPZdQ_9slzjYOD8jaRDtNBDnk3Onswn0M05lVIA
Expires: Tue, 23 Aug 2016 09:52:59 GMT
Date: Tue, 23 Aug 2016 08:52:59 GMT
Cache-Control: public, max-age=3600
Last-Modified: Mon, 15 Aug 2016 23:20:30 GMT
ETag: "a30c3bd1a7fcc6a48acfb74936a19b4c"
x-goog-generation: 1471303230615000
x-goog-metageneration: 1
x-goog-stored-content-encoding: identity
x-goog-stored-content-length: 14091954
Content-Type: application/x-gzip
x-goog-hash: crc32c=B1onUw==
x-goog-hash: md5=oww70af8xqSKz7dJNqGbTA==
x-goog-storage-class: STANDARD
Accept-Ranges: bytes
Content-Length: 14091954
Server: UploadServer
Alternate-Protocol: 443:quic
Alt-Svc: quic=":443"; ma=2592000; v="35,34,33,32,31,30"
```
## Actual Results
Result from step 2a/2b is actually:
```
curl: (51) SSL: certificate verification failed (result: 5)
```
## Version
Apple OS X El Capitan Version 10.11.6
```
pmoore@Petes-iMac:~ $ uname -a
Darwin Petes-iMac.local 15.6.0 Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64 x86_64
```

lesserwhirls:
Just a note on this PR: adding the certificates to the KeyChain app no longer works (not sure why). It looks like the only option, at this point, for using netCDF with authorization/authentication is to make sure curl is built with openssl support.

iMichka:
I am pretty new to the openssl problems on mac, I hope I will be clear and I don't mixup things here, but we decided that we will not accept this PR.

We will not add the `with-openssl` flag. For multiple reasons.
- We prefer to use the curl provided by mac. The curl provided by homebrew is keg-only, so to be used only in last resort or only if you really know what you are doing.
- We want to reduce the numbers of optional flags that get called like this to reduce rebuild times. There is even a buildbot check for this now which will complain about you doing this.

Your options here are:
- let the netcdf devs add support for `SecureTransport`?
- tweak you certificates configuration? Get the bug fixed in Apple's curl?
- carry that patch around and rebuild your stack with it, but this is up to you to keep this working for later

PR #4863, opened 2017-01-17
OpenNI/OpenNI2

xzcvczx:
I have ignored a lot of the problems bought up by audit currently as this is just to make the fomulas more correct

### Have you:

### New formula: have you


iMichka:
This fails with:

```
homebrew/science/openni2:
  * Stable version URLs should not contain beta
  * The installation was broken.
    Broken dylib links found:
      libMWClosestPoint.dylib
```

The url of the openni2 project looks weird. It is the url of a fork. Seems the correct url would be: https://github.com/OpenNI/OpenNI2

But as OpenNI2 has been in beta since 2014, you could use the https://github.com/occipital/OpenNI2/releases/tag/v2.2.0-debian release, which is not tagged as beta.

The two projects look abandoned though (looking at their git history). They could at least use a stable tag for v2.2.0.

I do not know what the second error is. I'll let you investigate this first. Tell us if there is any progress or questions.

xzcvczx:
well i haven't actually touched most of the stuff in the file, Just clarified that java is required, So i am really unsure as to why the dylib is broken but i will take a look. 

Same with the beta, just leaving it where it was pointed at, Also there is a commit from last last year on one of the branches i think, but openni2 does seem relatively dead where as openni is very dead.

Issue #3597, opened 2016-04-27
Formulas that might need to depend on veclibfort


dpo:
In order to avoid errors and segfaults when using single precision and/or Fortran, `veclibfort` must be interposed between the library and the Accelerate framework.
- [ ] `armadillo.rb`
- [ ] `dealii.rb`
- [ ] `elemental.rb`
- [ ] `hypre.rb`
- [ ] `ipopt.rb`
- [x] `octave.rb`
- [ ] `p4est.rb`
- [ ] `pastix.rb`
- [ ] `petsc.rb`
- [ ] `plink2.rb`
- [ ] `r.rb`
- [ ] `slepc.rb`
- [ ] `suite-sparse.rb`
- [ ] `sundials.rb`
- [x] `superlu_dist.rb`
- [ ] `trilinos.rb`

PR #4866, opened 2017-01-17
hypre 2.11.1

iMichka:
Fixes following audit failures

  * `option` (line 19) should be put before `depends_on` (line 17)
    Length is calculated as hypre + desc. (currently 144)
  * Description shouldn't start with an indefinite article (A)
  * Use '--without-test' instead of '--without-check'. Migrate '--without-check' with `deprecated_option`.
  * C: 83: col 24: Closing array brace must be on the same line as the last array element when opening brace is on the same line as the first array element.
  * C: 104: col 9: Favor a normal if-statement over a modifier clause in a multiline statement.
  * C: 156: col 13: Avoid more than 3 levels of block nesting.

- Added a simple test
- Added veclibfort depency as suggested in #3597
- Depend on openblas for linux to prevent linking errors

### Have you:

### New formula: have you

### DIFF
--- a/hypre.rb
+++ b/hypre.rb
@@ -1,10 +1,9 @@
 class Hypre < Formula
-  desc "A library of high performance preconditioners that features parallel multigrid methods for both structured and unstructured grid problems"
+  desc "Library featuring parallel multigrid methods for grid problems"
   homepage "http://computation.llnl.gov/casc/hypre/software.html"
-  url "http://ftp.mcs.anl.gov/pub/petsc/externalpackages/hypre-2.10.0b-p2.tar.gz"
-  version "2.10.0b-p2"
-  sha256 "76414e693e5381e352e759c851c8cd5969dba7001b1dc153fe0f1ff60a5bb168"
-  revision 1
+  url "http://ftp.mcs.anl.gov/pub/petsc/externalpackages/hypre-2.11.1.tar.gz"
+  mirror "http://computation.llnl.gov/projects/hypre-scalable-linear-solvers-multigrid-methods/download/hypre-2.11.1.tar.gz"
+  sha256 "6bb2ff565ff694596d0e94d0a75f0c3a2cd6715b8b7652bc71feb8698554db93"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


iMichka:
I'll try to come up with a sensible test. And see if I can get the linux build to run.

iMichka:
@sjackman This is ready for a small review.

I read that long thread about openblas (https://github.com/Linuxbrew/legacy-linuxbrew/issues/1075), would be worth document this somewhere (we could have a homebrew-science specific formula cookbook?).

@ilovezfs I added the veclibfort dependency but did not investigate more (it did build without before).

sjackman:
Here's what we have that's closest to a formula cookbook for Homebrew-Science: https://github.com/Homebrew/homebrew-science/wiki/Maintainers

hypre.rb
```diff
@@ -13,18 +12,21 @@ class Hypre < Formula
...
   option "with-mli", "Use internal MLI routines"
   option "without-accelerate", "Build without Accelerate framework (use internal BLAS routines)"
   option "with-debug", "Build with debug flags"
   option "with-bigint", "Build with 64-bit indices"
 
+  deprecated_option "with-check" => "with-test"
+
+  depends_on "veclibfort" if build.without?("openblas") && OS.mac?
+  depends_on :fortran => :recommended
+  depends_on :mpi => [:cc, :cxx, :f90, :f77, :recommended]
```
sjackman: Are all four of these needed?

iMichka: @sjackman I added veclibfort due to #3597. But I am not sure that just adding the dependency is enough?

The other two were already there so I didn't touch them (just moved the around). MPI is probably needed because when I look at their first example tutorial on their website, they do use some MPI stuff. Not sure this is needed during compile time though, but some things are run in this example folder at build time.

I am not really happy with it because I have no clue what it is doing, but I trust the original author knew what he was doing.

sjackman: I was wondering whether all four languages `[:cc, :cxx, :f90, :f77]` were required.

iMichka: No clue about those 4. I just left them untouched. But it looks like a lot of requirements.

Issue #4854, opened 2017-01-15
viennarna is outdated


twogee:
viennarna in mac brew is version 2.1.9, but the latest version is 2.3.1 and the URL is different.
I tried "brew edit viennarna" and change the URL + checksum to rebuild from source, but the source archive is sent as an attachment and brew seems unable to handle it properly.
There's a dmg available for download, so maybe viennarna shoould move to the caskroom.
What do you think?

iMichka:
Hi. Thanks for letting us know.

Looking at their homepage (http://www.tbi.univie.ac.at/RNA/#download), the new url seems to be:
`http://www.tbi.univie.ac.at/RNA/download/sourcecode/2_3_x/ViennaRNA-2.3.1.tar.gz`

I am not sure what you mean by: `but the source archive is sent as an attachment and brew seems unable to handle it properly`? Do you have a specific error message that you could share?

If you want to contribute, you can follow the instructions on how to open a pull request [here](http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html).

twogee:
What I meant was that if I try to run brew install on updated formula, the contents of ViennaRNA-2.3.1.tar.gz are those of the web page, that is, it contains the HTML code. I believe the reason for this is that the actual tar.gz file is sent as an attachment to the web page.

twogee:
I tried again and noticed that brew is getting redirected

```
==> Upgrading homebrew/science/viennarna 
==> Downloading https://www.tbi.univie.ac.at/RNA/download/sourcecode/2_3_x/ViennaRNA-2.3.1.tar.gz
==> Downloading from https://www.tbi.univie.ac.at/RNA/
```

but wget is not... is Vienna University blocking homebrew?

iMichka:
I had a look at this.

Homebrew tries to setup the url here:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/download_strategy.rb#L401-L408

Here is the output:
```
curl -I -L https://www.tbi.univie.ac.at/RNA/download/sourcecode/2_3_x/ViennaRNA-2.3.1.tar.gz 

HTTP/1.1 302 Found
Date: Tue, 17 Jan 2017 18:09:19 GMT
Server: Apache/2.2.22
X-Powered-By: PHP/5.3.10-1ubuntu3.25
Location: https://www.tbi.univie.ac.at/RNA
Content-Type: text/html

HTTP/1.1 301 Moved Permanently
[...]

PR #4867, opened 2017-01-17
viennarna 2.3.1

twogee:


twogee:
Closes issue #4854 
Thanks for guidance!

iMichka:
The mac build failed with:

>  * python modules have explicit framework links
      /usr/local/opt/viennarna/lib/python2.7/site-packages/_RNA.so

This is quite known problem. Here is some documentation.
http://blog.tim-smith.us/2015/09/python-extension-modules-os-x/

The linking flags need to be adapted somewhere in the upstream code (they need to be ```-undefined dynamic_lookup```, the link I gave explains it all).

Do you feel up to report this upstream and try to get their code updated?

twogee:
I will give it a try.

RaumZeit:
This is ViennaRNA Package upstream @ViennaRNA
We just released version 2.3.2 of the ViennaRNA Package about a minute ago. Unfortunately just before I got note of this particular problem. I've already created a patch that will take care of passing
```
-undefined dynamic_lookup
```
as linker flags for the python extension. But it will only be officially available with the next version.
I suggest that for the time being you could patch interfaces/Python/Makefile replacing the line
```
_RNA_la_LDFLAGS = -avoid-version -module -shared -export-dynamic -shrext $(PYTHON2_SO) $(PYTHON2_LDFLAGS)
```
by something like
```
_RNA_la_LDFLAGS = -avoid-version -module -bundle -undefined dynamic_lookup -export-dynamic -shrext $(PYTHON2_SO) 
```
This should do the trick...

iMichka:
@RaumZeit Thanks for the reactivity. We can live with a patch for some time, it's not a problem.

The ultimate test will be to run (to check if the solution works):
> otool -L /usr/local/opt/viennarna/lib/python2.7/site-packages/_RNA.so

There should be no mention of the Python library path in the output.

If this is OK, @twogee, you will need to create a new pull request containing the patch here:
https://github.com/Homebrew/formula-patches
Just create a new folder in that project and add the diff in a .patch file.

Once the patch is merged, you can update your formula by adding a patch block. See the patch documentation here: http://docs.brew.sh/Formula-Cookbook.html

Don't hesitate if you have question, I can guide you through this but it should be pretty easy.

twogee:
Sorry about the break; I'll try 2.3.3, see #4917

iMichka:
No problem. Thanks. We can continue the discussion in the other pull request. Closing here.

PR #4870, opened 2017-01-18
OTPS 2 (new formula)

zimonkaizoku:
This is a new formula for the OSU Tidal Prediction Software (OTPS2).
The software can be used for tidal prediction and should be a
recommended dependency for MB-System (mbsystem.rb).

`brew audit` returns an error : `* 'DATA' was found, but no '__END__'`
regarding the directory 'DATA' that is used for the optional tidal model.
This is not ruby, it's just the name of the directory.

There is no make install as the modules are either meant to be
executed in the OTPS folder, or, as intended with this request, used
as a recommended dependency for mbsystem's module mbotps.


zimonkaizoku:
Looks like all checks failed. One reason is the mentioned 'DATA' directory, but I don't think that there's a way around this.

The other thing seems to be the test. The test worked well on my machine. It's running one of the modules (extract_HC) with an example input file (setup.inp) that is provided in the original tarball. This should create and output file (sample.out). I don't know why Travis said `sh: 1: cannot open setup.inp: No such file`.

I just realized, the test will most likely fail if OTPS is installed without the tpxo8 option. But I don't know any alternative test and making the download of tpxo8 (1.2GB) mandatory seems to be a bad idea for people who want to use their own model...

iMichka:
@BrewTestBot test this please

iMichka:
Mac audit:

> homebrew/science/otps:
> * 'DATA' was found, but no '__END__'

otps.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def install
+    system "make", "extract_HC"
+    system "make", "predict_tide"
+    system "make", "extract_local_model"
+
+    prefix.install Dir["*"]
+
+    if build.with? "tpxo8"
+      (prefix/"DATA").install resource("tpxo8")
```
iMichka: Can this not be `data` in lowercase? We have a "bug" in our audit, so this may be a workaround.

otps.rb
```diff
@@ -0,0 +1,50 @@
+class Otps < Formula
+  desc "OTPS: OSU Tidal Prediction Software"
+  homepage "http://volkov.oce.orst.edu/tides/tpxo8_atlas.html"
+  url "ftp://anonymous:anonymous%40homebrew.com@ftp.oce.orst.edu/dist/tides/OTPS2.tar.Z"
+  sha256 "5ac90789e4765da9efeeb2823ac713b12b30c0ad1e8d8b10514f591137f56329"
+
+  option "with-tpxo8", "Install TPXO8-atlas-compact tide model"
```
iMichka: We do not like options, because they are not tested by the build bots, and people need to rebuild from source when using them. Would it make sense to alway have this ON? Does it impact the build time a lot or breaks the formula in any other way?

otps.rb
```diff
@@ -0,0 +1,50 @@
...
+      link.write <<-EOS.undent
+        #{prefix}/DATA/hf.tpxo8_atlas_30
+        #{prefix}/DATA/uv.tpxo8_atlas_30
+        #{prefix}/DATA/grid_tpxo8atlas_30
+      EOS
+      link.close
+    end
+  end
+
+  def caveats
```
iMichka: Is this something that is accessible in the documentation elsewhere? I think we prefer caveats for really different/specific things. If it is just the same hint as in the manual/doc of the formula, I would prefer to not add a caveat.

iMichka:

PR #4874, opened 2017-01-19
lammps 2016.11.17

fxcoudert:


fxcoudert:
The previous version was not an official "stable" release from the LAMMPS team. I am adding a comment so that we may avoid that in the future.

jonchang:
Merged. Thanks @fxcoudert!

Issue #4865, opened 2017-01-17
R -- any reason not to use CRAN's build env variables?


gwerbin:
The ["R Installation and Administration" manual](https://cran.r-project.org/doc/manuals/r-release/R-admin.html) says that the official CRAN binary is built using the following environment variables:

```shell
CC=clang
CXX=clang++
F77=gfortran-4.8
FC=$F77
OBJC=clang
CFLAGS='-Wall -mtune=core2 -g -O2'
CXXFLAGS='-Wall -mtune=core2 -g -O2'
OBJCFLAGS='-Wall -mtune=core2 -g -O2'
F77FLAGS='-Wall -g -O2'
FCFLAGS=$F77FLAGS
```

See https://cran.r-project.org/doc/manuals/r-release/R-admin.html#macOS for more.

It looks like this change would be easy; I think you'd just have to add

```ruby
# Use the same build variables as the official CRAN binary
args << "CFLAGS=-Wall -mtune=core2 -g -O2"
args << "CXXFLAGS=-Wall -mtune=core2 -g -O2"
args << "OBJCFLAGS=-Wall -mtune=core2 -g -O2"
args << "F77FLAGS=-Wall -g -O2"
args << "FCFLAGS=-Wall -g -O2"
```

somewhere before the line

```ruby
system "./configure", *args
```

Then the only difference being that instead of always using `gfortran-4.8`, Homebrew would just use the latest version installed with GCC.

Is there any reason this shouldn't be in the R formula? I made this change to my local Homebrew version and it certainly didn't seem to break anything.

greg-rocketrip:
Sorry to double-post (on a different account no less), but this is what the change might look like:

https://github.com/Homebrew/homebrew-science/compare/master...greg-rocketrip:r-cran-build-parity

iMichka:
Closing here too. The question was answered in the PR. We already pass the right flags.

PR #4882, opened 2017-01-22
Add CRAN build env vars

gwerbin:
PR accompanying https://github.com/Homebrew/homebrew-science/issues/4865.

Note that `brew audit --strict --online` gives a false positive error for a part of the Formula that this PR doesn't touch.

### Have you:

### DIFF
--- a/r.rb
+++ b/r.rb
@@ -112,6 +112,13 @@ class R < Formula
     args << "--without-tcltk" if build.without? "tcltk"
     args << "--without-x" if build.without? "x11"
 
+    # Use the same environment variables as the CRAN binary
+    args << "CFLAGS=-Wall -mtune=core2 -g -O2"
+    args << "CXXFLAGS=-Wall -mtune=core2 -g -O2"
+    args << "OBJCFLAGS=-Wall -mtune=core2 -g -O2"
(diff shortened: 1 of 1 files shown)


iMichka:
Hi. Thanks for you contribution.

`-Wall` is just turning on more compiler warnings. Probably useful when debugging the compilation process, but this is probably not something we will do often here. It could be added but I do not see any gain doing so.
`-g -O2` is already passed to the compiler (I just double checked locally by compiling the current formula with -v). What makes you think this is not the case?
`-mtune=core2`: I think this will be filtered out here: https://github.com/Homebrew/brew/blob/f3460f642c0cafd3da8a938f56394b02025c0ba2/Library/Homebrew/shims/super/cc#L155

Bottles are already built for core2, see the [docs](http://docs.brew.sh/Bottles.html). The day we drop core2, there is only one place to change this. Not for each formula separately.

Is there any specific reason you wanted to add these flags? Any error you saw? Or just that you wanted to align these flags with those in R's documentation?

gwerbin:
@iMichka just to align with the docs. I didn't think to check the existing flags with `-v`, and good to know about `-mtune=core`.

Issue #4830, opened 2017-01-08
opencv3 compilation error


remia:
### Indicate troubleshooting steps taken

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Checked that a similar issue hasn't already been filed?
- [ ] Looked up the software bug tracker if reporting a build error?
- [ ] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

### Bug reports / build errors:

> /tmp/opencv3-20170108-86584-18679kx/opencv-3.2.0/modules/python/src2/cv2.cpp:6:10: fatal error: 'Python.h' file not found

#### Brief summary of issue:

The issue occurs only when installing opencv3 with python3 support, the error seems to indicate a missing header and/or header directory compilation setting. 

The complete command invoking clang++ is : 

[...]

PR #4885, opened 2017-01-23
OpenCV3 failed to build for Python3 from source

CarlosRDomin:
This PR solves issue #4830:
Building from source, the flag `--with-python3` would fail to build OpenCV3 for Python3. By analyzing the `--verbose` log, I realized the issue was related to the ENV variable `PYTHONPATH` being set (and pointing to Python2 instead). All I had to do was **unset `PYTHONPATH`** in the formula, right before it looks for the Python3 installation.

This was the actual error in the `--verbose` log:
> Your PYTHONPATH points to a site-packages dir for Python 2.x but you are running Python 3.x!
You should unset PYTHONPATH to fix this.

### Have you:

### New formula: have you


iMichka:
Thanks for your contribution.

I think @tdsmith has already seen a bunch of these errors during the last months/years. I looked a little bit around in the different homebrew projects, found a lot of similar errors. But there were never clear answers about what was the problem.

I am not sure if we should reset that env variable manually. I mean, we would have to do it for more formulae than this one. Maybe a more general fix could be found. I want to first understand what is going on instead of fixing each formula one by one.

CarlosRDomin:
Yup, absolutely agree. I also wonder what the root cause of the problem is. It seems like at some point during the execution of the script, the ENV var must be set to point to Python 2 (because I explicitly unset it before brew installing, and `puts "PYTHONPATH is #{PYTHONPATH}"` still prints my Python 2 `site-packages` dir). If I have some time this week I'll try to figure out which specific line triggers this issue :)

PS: My PR failed the Travis test with a weird error, but it was passing all the `brew tests` and `brew test opencv3` locally, any idea why?

tdsmith:
Blah; the :python requirement sets PYTHONPATH for dubious reasons; it expects you'll use `each_python` if you need to build for Python 3 as well, though that's obviously an incorrect expectation here. This fix seems fine, although the root cause remains annoying.

iMichka:
Don't worry about the Travis build. It is the Linux build for LinuxBrew. It is not mandatory that it passes, we will look at this later.

jonchang:
I suppose I'm kind of confused as to why this is necessary as I'm not familiar enough with Python 3 to understand why it would work differently. Reading [python_requirement.rb](https://github.com/Homebrew/brew/blob/c9c7f462d37500549127efba96c7a25e5c70de4a/Library/Homebrew/requirements/python_requirement.rb#L20-L27) it seems that `PYTHONPATH` would be set to `python#{short_version}/site-packages`, which I think would be `python3.4/site-packages` or similar for Python 3.

iMichka:
I built opencv3 `--with-python3`, but without this patch, and it worked for me. There were some changes in homebrew which fixed similar issues with weird python paths lately. Could you please try again? If it fails, could we see the `brew gist-logs` for your build?

iMichka:
Thinking especially at #5247. Looks similar.

CarlosRDomin:
Unfortunately, this still doesn't work for me... :(
This are extracts from the `cmake` step:
```sh
-- Found PythonInterp: /usr/local/opt/python/bin/python (found suitable version "2.7.13", minimum required is "2.7") 
-- Found PythonLibs: /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/libpython2.7.dylib (found suitable exact version "2.7.13") 
-- Could NOT find PythonInterp: Found unsuitable version "1.4", but required is at least "3.4" (found /usr/local/bin/python3)
-- Could NOT find PythonInterp: Found unsuitable version "1.4", but required is at least "3.2" (found /usr/local/bin/python3)
```
```sh
--   Python 2:
--     Interpreter:                 /usr/local/opt/python/bin/python (ver 2.7.13)
--     Libraries:                   /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/libpython2.7.dylib (ver 2.7.13)
--     numpy:                       /usr/local/lib/python2.7/site-packages/numpy/core/include (ver 1.12.1)
--     packages path:               lib/python2.7/site-packages
-- 
--   Python 3:
--     Interpreter:                 NO
--     Libraries:                   NO
--     numpy:                       NO (Python3 wrappers can not be generated)
--     packages path:               -
-- 
--   Python (for build):            /usr/local/opt/python/bin/python
```

Full gist-logs are here https://gist.github.com/b28413ecf91c8407f7e47bde1842d4c2
After taking a quick look at `01.cmake`, it seems like Python3 paths (`PYTHON3_LIBRARY` and `PYTHON3_INCLUDE_DIR`) are not correctly found/specified by the `opencv3` formula:
```sh
-DPYTHON2_LIBRARY=/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/libpython2.7.dylib
-DPYTHON2_INCLUDE_DIR=/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/include/python2.7
-DPYTHON3_LIBRARY=/libpython3.6.dylib
-DPYTHON3_INCLUDE_DIR=
```

CarlosRDomin:
Also, by reading the verbose output of the installation (`brew install -v opencv3 [...]`), I found where the error is. The `opencv3` formula, when trying to find the Python3 variables (the `if build.with? "python3"` block on lines 200-206), crashes with this error:
```sh
Failed to import the site module
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 544, in <module>
    main()
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 530, in main
    known_paths = addusersitepackages(known_paths)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 282, in addusersitepackages
    user_site = getusersitepackages()
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 258, in getusersitepackages
    user_base = getuserbase() # this will also set USER_BASE
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 248, in getuserbase
    USER_BASE = get_config_var('userbase')
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/sysconfig.py", line 601, in get_config_var
    return get_config_vars().get(name)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/sysconfig.py", line 580, in get_config_vars
    import _osx_support
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/_osx_support.py", line 4, in <module>
    import re
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/re.py", line 142, in <module>
    class RegexFlag(enum.IntFlag):
AttributeError: module 'enum' has no attribute 'IntFlag'
Failed to import the site module
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 544, in <module>
    main()
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 530, in main
    known_paths = addusersitepackages(known_paths)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 282, in addusersitepackages
    user_site = getusersitepackages()
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 258, in getusersitepackages
    user_base = getuserbase() # this will also set USER_BASE
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site.py", line 248, in getuserbase
    USER_BASE = get_config_var('userbase')
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/sysconfig.py", line 601, in get_config_var
    return get_config_vars().get(name)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/sysconfig.py", line 580, in get_config_vars
    import _osx_support
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/_osx_support.py", line 4, in <module>
    import re
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/re.py", line 142, in <module>
    class RegexFlag(enum.IntFlag):
AttributeError: module 'enum' has no attribute 'IntFlag'
```

A quick Google search pointed to an issue on homebrew-core (https://github.com/Homebrew/homebrew-core/issues/9229) where, surprise, the issue is gone by unsetting `PYTHONPATH`...

I went back to my patch and at least now I have OpenCV3 again, let me know if you figure out a potential fix and want me to try again.

iMichka:
I just read the topic.

What does `echo $PYTHONPATH` give for you? You need to double check your .config files, there may be a hardcoded path that imports the python2 path into your python3 build. This is why unsetting it works.

We are working on a more general solution, so that we do not need to unset the path in each formula. See https://github.com/Homebrew/brew/issues/932

CarlosRDomin:
`echo $PYTHONPATH` gives an empty string (` `), I already checked to make sure that it wasn't set to Python 2 in my environment, so that's not the cause of the problem.
Could you specify which `.config` files (path) you want me to check?
I understand this is a common problem across multiple formulas and it makes total sense trying to find a general fix, I hope we can all find a good solution. I'm going to try and find where in the `opencv3` formula `PYTHONPATH` is set to point to Python 2, maybe that'll give us a hint...

CarlosRDomin:
Hmmm, so my `$PYTHONPATH` is unset, but when I install `opencv3` (`brew reinstall -v opencv3 [...]`), it's like Homebrew is internally setting it to Python2 somewhere.
If I add the following line after line 100 of `opencv3.rb`:
```ruby
def install
    puts "PYTHONPATH is #{ENV["PYTHONPATH"]}"
```
it outputs: `PYTHONPATH is /usr/local/lib/python2.7/site-packages`, but this is even before the installation script starts... 🤔

iMichka:
> Could you specify which .config files (path) you want me to check?

I was thinking about some .profile, .bashrcv or .zshrc files in your home folder.

> it's like Homebrew is internally setting it to Python2 somewhere.

This is what I was looking for. And this is wrong.

CarlosRDomin:
I don't think it's an issue with my config files. The only file related to terminal config in my home folder is ~/.bash_profile, and these are the contents:
```sh
$ cat ~/.bash_profile 
export LC_ALL=en_US.UTF-8
export LANG=en_US.UTF-8

export HOMEBREW_EDITOR=atom
export PATH="/usr/local/opt/opencv3/bin:$PATH"
```

As for Homebrew internally setting `PYTHONPATH`, where could I read some documentation about the internal Homebrew code? Or at least, which file (and from which repo) calls the `install` method on a formula? I unsuccessfully tried to take a look at it last night, but I don't have a Ruby debugger (nor Ruby programming experience) to really be able to figure it out on my own. I'll be happy to help though :)

iMichka:
The "package manager" code itself can be found here: https://github.com/Homebrew/brew

I will maybe have some time tomorrow to look at this too.

jonchang:
See [this comment](https://github.com/Homebrew/homebrew-science/pull/4885#issuecomment-275814956), you want to look at `python_requirement.rb`

iMichka:
If I use `puts "PYTHONPATH is #{ENV["PYTHONPATH"]}"` in my install block the PYTHONPATH is empty for me.

This is because for me, the whole env block is never called: [python_requirement.rb](https://github.com/Homebrew/brew/blob/c9c7f462d37500549127efba96c7a25e5c70de4a/Library/Homebrew/requirements/python_requirement.rb#L17-L28).

I was also thinking about something: there is only one PYTHONPATH for Python 2 and Python 3. This means that we can't build both at the same time.

@jonchang when is this env block supposed to be called?

tdsmith:
It is set if you have a dependency on :python or :python3.

tdsmith:
We should probably get rid of the PYTHONPATH twiddling in the :python requirements and do something more useful here: https://github.com/Homebrew/brew/blob/1ba401039eeafec6dca464e9be18a5257a3d89bc/Library/Homebrew/formula.rb#L1726

jonchang:
My read on this ([aided by this Python bug](https://bugs.python.org/issue2375)) is that there's no `PYTHON3PATH` or equivalent, so right now you can't build opencv with both `:python` and `:python3`. OpenCV will need to be taught to not rely on `PYTHONPATH`, or at least permit us to build against both at the same time, since (I believe‚ we're trying to avoid setting `PYTHONPATH` inside formulae.

tdsmith:
> I believe‚ we're trying to avoid setting PYTHONPATH inside formulae.

I don't think I know a reason for that, if setting it helps you build something.

iMichka:
@CarlosRDomin Could you please update this PR. I would like to merge this as a workaround so that people can work. There seem to be a mixup in your commits. Could you squash this, maybe rebase, and remove the changes around jpeg turbo?

iMichka:
I added the workaround myself in #5424 . This has now been merged. It leaves time so that whoever wants can have a look at the homebrew core code later, while users can install opencv3 with Python3.

@CarlosRDomin You can open a new separate pull request for the jpeg-turbo option. Thanks

PR #4887, opened 2017-01-23
salmon 0.8.0

rob-p:
### Have you:

### New formula: have you

### DIFF
--- a/salmon.rb
+++ b/salmon.rb
@@ -3,8 +3,8 @@ class Salmon < Formula
   homepage "https://github.com/COMBINE-lab/salmon"
   # tag "bioinformatics"
 
-  url "https://github.com/COMBINE-lab/salmon/archive/v0.7.2.tar.gz"
-  sha256 "d35147663d349a6c28bcabf51e85d5a45f24273be1c4cda76173ffa15bd68d0a"
+  url "https://github.com/COMBINE-lab/salmon/archive/v0.8.0.tar.gz"
+  sha256 "6aedc4d518f12f2023c97db751b5cea6c8c617f2f1635f59a510820702983de1"
 
   head "https://github.com/COMBINE-lab/salmon.git"
 


ilovezfs:
On 10.11,
```
==> FAILED
Testing homebrew/science/salmon
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20170122-15315-1604sxc.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-science/salmon.rb --verbose
==> /usr/local/Cellar/salmon/0.8.0/bin/salmon --version
dyld: lazy symbol binding failed: Symbol not found: _os_unfair_lock_lock
  Referenced from: /usr/local/Cellar/salmon/0.8.0/bin/salmon
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _os_unfair_lock_lock
  Referenced from: /usr/local/Cellar/salmon/0.8.0/bin/salmon
  Expected in: /usr/lib/libSystem.B.dylib

Error: homebrew/science/salmon: failed
Failed executing: /usr/local/Cellar/salmon/0.8.0/bin/salmon --version
```

iMichka:
Hi. Thank you for your submission.

The El Capitan test failed with:
```
==> /usr/local/Cellar/salmon/0.8.0/bin/salmon --version
dyld: lazy symbol binding failed: Symbol not found: _os_unfair_lock_lock
  Referenced from: /usr/local/Cellar/salmon/0.8.0/bin/salmon
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _os_unfair_lock_lock
  Referenced from: /usr/local/Cellar/salmon/0.8.0/bin/salmon
  Expected in: /usr/lib/libSystem.B.dylib
```

ilovezfs:
This is caused by the vendored jemalloc. The bug has been fixed upstream but because the formula doesn't depend on the brew version of jemalloc, and instead downloads its own copy, its not getting the benefit of the fix yet. So on a connected note, it's not really permissible for formulae to download anything unless it's checksummed (is it?), and even then it's not something we prefer if the same dependency is available as a formula.

ilovezfs:
It looks like it picks up the brew jemalloc if it's added as a dependency
```
iMac-TMP:homebrew-science joe$ git diff -- salmon.rb 
diff --git a/salmon.rb b/salmon.rb
index b402bc4..8a24487 100644
--- a/salmon.rb
+++ b/salmon.rb
@@ -23,6 +23,7 @@ class Salmon < Formula
   depends_on "automake" => :build
   depends_on "cmake" => :build
   depends_on "boost"
+  depends_on "jemalloc"
   depends_on "tbb"
   depends_on "xz"
   depends_on "zlib" unless OS.mac?
```
which avoids the problem.

However, that still leaves the questions about the other vendored dependencies. Are they being verified with sha256 checksumming or better? And are any of the others replaceable by the brew versions?

rob-p:
@ilovezfs : Thanks for the heads-up about jemalloc.  I've bumped the version upstream in `salmon` as well.  The other vendored dependencies are fetched in the same way as jemalloc --- from GitHub tarballs using CMake's `ExternalProject_Add` functionality.  This does not, I believe , do a checksum on the fetched files by default, though I can look into the best way to do this.

The other dependencies that exist in homebrew-science or homebrew are:
  * bwa
  * Jellyfish
  * libdivsufsort
  * cereal

The problem with `bwa` and `Jellyfish`.  Honestly, the `bwa` dependency is likely to go away soon, but the problem with both of these right now is that I rely on them at the source rather than strictly library level.  The problem with `libdivsufsort` is that the version currently in homebrew-science is too old, so maybe I could help prepare a version bump for that one, and there is no problem with `cereal`, it's just a header only library and so is trivial to grab and place in my include directory at build time.  To depend on a system installation, I have to find or fashion a CMake module to check for the system version.

iMichka:
FYI  [ExternalProject_Add](https://cmake.org/cmake/help/v3.2/module/ExternalProject.html) can do a md5 check (with `URL_MD5`). Better than nothing. (It may be worth asking the cmake people if they plan to add sha256 one day).

The insighttoolkit (ITK) is for example using this. And there is always a cmake flag, like USE_SYSTEM_HDF5, that allows to switch between the downloaded library or a system provided one. On homebrew we then use the system/homebrew libraries.

iMichka:
@BrewTestBot test this please

rob-p:
It looks like CMake might already support computing other hashes.  Take a look at the `URL_HASH ALOG=value` option [here](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html).

ilovezfs:
@rob-p any update in circumstances here? I noticed there's a 0.8.1 now!

rob-p:
Hi @ilovezfs,

  Yes and no.  On one hand, I figured out how to use CMake's `ExternalProject_add` to verify a SHA1 of downloaded content --- and on my server, that seems to work swimmingly.  Unfortunately, that solution doesn't currently work on the VM I use to build & automatically test Salmon (the issue is related to the VM's `libcurl` not properly supporting `https`).  So, the solution will have to rely on finding a reliable and easy way to get a newer and compatible `libcurl` for an ancient VM (I build using HolyBuildBox for wide ranging backward compatibility), or using some external mechanism for verifying the downloaded content.  I already implemented the latter solution for the source tarball that is fetched by a bash script (and not within CMake) [see here](https://github.com/COMBINE-lab/salmon/blob/master/scripts/fetchRapMap.sh#L28).

ilovezfs:
Note that any verification mechanism should be sha256 or better at this point.

rob-p:
Sure --- CMake's solution (as mine with `shasum`) is generic to the actual hash used.

rob-p:
Ok, I just made a [commit](https://github.com/COMBINE-lab/salmon/commit/d25835e8f25257898ac7831ebe3a035b9d41cfe5) on our develop branch that fixes this.  It does, however, require that the client have `sha256sum` for my solution to work (otherwise it doesn't verify).  Can we add `sha256sum` as a vendored dependency within homebrew-science?

ilovezfs:
On macOS you should use `/usr/bin/shasum -a 256`

rob-p:
Ok, [this commit](https://github.com/COMBINE-lab/salmon/commit/e90d9c58ad57472e1604a5dee64ed1df03c70b5c) uses `shasum -a256` if `sha256sum` is aliased.  I left out the full path in case there is a reason to have `shasum` come from somewhere else.

iMichka:
Closing in favour of #5284

PR #4892, opened 2017-01-23
orocos-kdl: use eigen@3.2

ilovezfs:


ilovezfs:
@BrewTestBot test this please

PR #4894, opened 2017-01-23
Import formulae from homebrew/python.

MikeMcQuaid:
The rest are going to homebrew/core but these have hard homebrew/science dependencies.

Also see https://github.com/Homebrew/homebrew-python/pull/388 and https://github.com/Homebrew/homebrew-core/pull/9193.


jonchang:
There are 8 formulae that say `depends_on "homebrew/python/numpy"` or similar. Should these be migrated to `depends_on "numpy"`?

MikeMcQuaid:
Yes but that isn't a blocker for this being merged. It sounds like there may be some others that need to get imported here as we're going to import fewer to homebrew/core.

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
@BrewTestBot test this please

PR #4901, opened 2017-01-24
gfan: use gcc@5

ilovezfs:


ilovezfs:
@BrewTestBot test this please

fxcoudert:
If no use, no release since 2011 and does only build with specific version of the compiler, why keep it? https://github.com/Homebrew/homebrew-science/pull/4898

ilovezfs:
Because we can?

fxcoudert:
My bad, I am new and I thought there was a requirement in the project of integrating software that is at least reasonably maintained upstream (I see that even with GCC 5 the source needs patching) and build on some sort of "default" configuration.

ilovezfs:
That's a better description of policy in core than in science.

jonchang:
@BrewTestBot test this please

ilovezfs:
@jonchang we have a brew test-bot bug that makes it impossible to use gcc@5. So we'll either need to disable the bottle here or wait till it's fixed. The problem doesn't affect local user installation, just test-bot.

ilovezfs:
I've gone ahead and added an explicit bottle disable for now.

jonchang:
Ok thanks! I was literally just trying to figure out why bottles weren't building for this.

PR #4918, opened 2017-01-27
octave: depend on qt@5.7

ilovezfs:
See Homebrew/homebrew-core#9297 and Homebrew/homebrew-core#9308.


ilovezfs:
@schoeps FYI, shipped!

schoeps:
Cool, thanks! I am already compiling :) With every new qt release I hope that the Octave bug [50025](https://savannah.gnu.org/bugs/?50025) will disappear :)

ilovezfs:
@schoeps Yeah, keep hope alive until Riverbank catches up and we can test Octave with 5.8.0!

PR #4582, opened 2016-11-08
sumo 0.28.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #4923, opened 2017-01-27
sumo: rename fox dependency

ilovezfs:


dunn:
Fox is failing to build but the logs are getting truncated and I can't see the error.

dunn:
Maybe this should be combined with #4582?

PR #4936, opened 2017-01-29
cmor: revision for netcdf

ilovezfs:


iMichka:
Two builds failed with an audit error:
> * The homepage http://www2-pcmdi.llnl.gov/cmor is not reachable (HTTP status code 000)

Funnily the yosemite build did not have the problem. I can confirm that for the moment that website is not working. Maybe we could use: `http://cmor.llnl.gov`?

ilovezfs:
>Funnily the yosemite build did not have the problem.

We skip the homepage check on Yosemite because its curl is old and causes false positives.

ilovezfs:
@BrewTestBot test this please

PR #4946, opened 2017-01-29
moab: revision for netcdf

ilovezfs:


iMichka:
Let's just double check with upstream what happened. I will open an issue.

ilovezfs:
great thanks

iMichka:
Ok so upstream repackaged the tarball with less files because they had some stale configuration files in there ... We can push this.

PR #4949, opened 2017-01-29
ncview: revision for netcdf

ilovezfs:


ilovezfs:
@BrewTestBot test this please

ilovezfs:
```
Configuration error: You specified that the "clang" C compiler should be
used to build ncview, but the netcdf library was compiled with the "/usr/local/Homebrew/Library/Homebrew/shims/super/clang"
compiler.  Here is the path where I found the netcdf library:
     -L/usr/local/Cellar/netcdf/4.4.1.1_1/lib -lnetcdf -lhdf5_hl -lhdf5 -lsz -lz -ldl -lm -lcurl
You must use the same compiler for ncview as was used to build the netcdf library!
```
ridiculous.

ilovezfs:
also needs this revision due to a libpng upgrade

PR #4955, opened 2017-01-30
netcdf: use -lmpi if hdf5 was built with mpi

ilovezfs:
Also,

- fix formula bugs in the new CMake-based formula
- use configure with the cxx-compat resource since there is no CMake-based build
  system for that option at this time

CC @iMichka

See https://github.com/Homebrew/homebrew-science/pull/4932.


ilovezfs:
shipped!

PR #4978, opened 2017-02-02
getdp: revision for netcdf

ilovezfs:


iMichka:
Linux circle-ci build timed out. I relaunched the build once more, to give it a chance. If it didn't work we can merge that just in as-is.

PR #4982, opened 2017-02-03
libSBOL

bbartley:
### Have you:

- [ x ] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [ x ] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [ x ] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [ x ] Built your formula locally prior to submission with `brew install <formula>`?

### New formula: have you

- [ x ] Written a sensible test? (The best test is to compile and run a sample program.)

- [ x ] Added/bumped the `revision` number if the changes affect the precompiled bottles?
- [ x ] Not altered the `bottle` section?
- [ x ] Checked that the tests still pass?


libsbol.rb
```diff
@@ -2,7 +2,8 @@ class Libsbol < Formula
   desc "Read and write files in the Synthetic Biology Open Language (SBOL)"
   homepage "https://synbiodex.github.io/libSBOL"
   url "https://github.com/SynBioDex/libSBOL/archive/v2.1.1.tar.gz"
-  sha256 "66bb18d752f98f4ea53f774de7d28dfea41b8c8896bd01bce1ee1231f9cc7249"
+  sha256 "85703a675d2b2335d130989d78ca7437b8b64ea1cf7d829b58e4e4ee847661ea"
```
iMichka: Wait, the sha256 can't change like this. Did somebody fiddle with the initially released tar.gz? This has serious security implications, and people should never do this. Without knowing what has happened, I tend to default to: the server was hacked and the tar.gz was modified. Could you double-check this with upstream?

bbartley: Hi Michka

Sorry for the irregularities, I'm still learning this. I deleted the original v2.1.1 tag and replaced it on a more recent commit, hence the sha256 has changed.  I'm trying to re-release libSBOL without bumping the version.  The previous release was intended to be a "dry-run", now this is the official release.  Please advise.

iMichka: Okay, I get it now. This is not how you should do it.

As soon as you pushed your tag to GitHub, people will rely on it. If it's a dry run, use a beta or alpha tag. Or make release candidates.

If there was an error with the tag, for example the wrong commit was tagged, or you realise that wrong files have been delivered, or are not happy with the release, you need to provide a new tag. Version 2.1.1-1 would have made it, with a new tag on a different commit. Or 2.1.2. It is totally valid to make a new release due to a packaging error. It happened to me multiple times, you just bump the version number, to keep things consistent.

I just realised that the formula got already merged here: https://github.com/Homebrew/homebrew-science/commit/5519c4b609e8a621fe2a05be4d20818e4b303033
This means that the sha256 is wrong in the current formula, which makes it un-installable. We have the sha256 check in there to make sure the upstream tar.gz didn't get hacked, so that we do not deliver viruses or other things to our users.

So please, if possible, tag a new version. If not we will accept the new sha256 (because now we have no other choice). Please make a single commit for that, on a new branch, checked out from master (if origin is the "original" homebrew-science repo):
`git reset --hard origin/master`
`git pull origin master`
`git checkout -b libsbol-fix`
Make your changes to libsbol.rb
`git commit`
`git push <yourforkName> libsbol-fix`
Open a new pull request which should contain 1 single commit with only the changes in version number and sha256.

Thanks

bbartley: OK thanks for the help and sorry for the noob mistakes!  I think I learned a couple things. 
1) Don't delete tags
2) Start a new feature branch with each version bump
3) The definition of a 'revision' is a change in the Formula file (as opposed to the source repository) 

The new pull request is here: https://github.com/Homebrew/homebrew-science/pull/4985

iMichka:
Thanks. Let's continue on the other pull request then (#4985). I'll close this one.

PR #4989, opened 2017-02-03
netcdf: enable cxx and cxx compat options

iMichka:
netcdf is used by more than 10 formulae, which all
require different options. This makes the CI
rebuild netcdf often, slowing down version bumps.

Fortran can not be enabled because it breaks some
other packages.

### Have you:

### New formula: have you

### DIFF
--- a/dealii.rb
+++ b/dealii.rb
@@ -3,7 +3,7 @@ class Dealii < Formula
   homepage "http://www.dealii.org"
   url "https://github.com/dealii/dealii/releases/download/v8.4.2/dealii-8.4.2.tar.gz"
   sha256 "ec7c00fadc9d298d1a0d16c08fb26818868410a9622c59ba624096872f3058e4"
-  revision 5
+  revision 6
 
   head "https://github.com/dealii/dealii.git"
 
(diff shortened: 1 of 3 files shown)


iMichka:
@ilovezfs

iMichka:
@ilovezfs I think this could go in like this.

We could put fortran as recommended (it was optional initially), which still allows people to build it `--without-fortan`, giving an escape route for those which will get problems due to the mixed libstd++/libc++ builds. Would that maybe be a better strategy?

ilovezfs:
I think this is worth trying even if we end up having to be backtrack and make `:fortran` optional again.

ilovezfs:
shipped!

ilovezfs:
- [ ] cmor https://github.com/Homebrew/homebrew-science/pull/4997

ilovezfs:
It looks like this PR may have to be reverted. The gnudatalanguage PR https://github.com/Homebrew/homebrew-science/pull/4984 was green on all three CIs before I merged this, but now it's red on both Linux CIs. So that means that anything that depends on netcdf that wasn't using the with-fortran option now potentially cannot be built.

ilovezfs:
I've reverted all three commits (PR commit + the two bottle commits).

ilovezfs:
@iMichka if we do this at all, I think we'll need to leave fortran as a non-default option.

iMichka:
Ok I left the fortran option as-is, but enabled the other two. Small win, it only will help for `radx` (and `vtk` later). But that's still better. Thanks for having played around with this yesterday.

PR #5003, opened 2017-02-04
abinit: remove netcdf "with-fortran" build option

ilovezfs:
netcdf no longer has a "with-fortran" option since it's always enabled

also drop libxc dependency since abinit does not work with 3.0.0


ilovezfs:
Blocked by #4999.

PR #5026, opened 2017-02-05
vigra 1.11.0.1-alpha1

iMichka:


vigra.rb
```diff
@@ -1,7 +1,8 @@
 class Vigra < Formula
   desc "Image processing and analysis library"
   homepage "https://ukoethe.github.io/vigra/"
-  url "https://github.com/ukoethe/vigra/releases/download/Version-1-11-0/vigra-1.11.0-src.tar.gz"
+  url "https://github.com/ukoethe/vigra.git", :revision => "ee6b17e4efc91cba6d3e350f98a5c4f9db6e9514"
+  version "1.11.0.1-alpha1"
```
jakirkham: Why the `.1-alpha1`?

iMichka: Because upstream was broken; didn't tag a new release. So we improvised.

iMichka: See also: https://github.com/ukoethe/vigra/issues/404

jakirkham: Interesting. Thanks for the context.

So why not use the recommended `1.11.0.post1` for the release?

iMichka: Recommended by who? We just took that decision. You may be right, and next time we will consider the `post` option.

My initial move would have been to delete the formula, because it didn't build anymore. But we were so nice and saved it from it's certain death for the next 2-3 years :)

PR #5036, opened 2017-02-05
openmeeg: revision for hdf5

iMichka:


iMichka:
Not working, but this was known. The real fix will be done here: #4276

PR #4254, opened 2016-09-11
blasr 4.0.0

ilovezfs:
also revision for hdf5

PR #5049, opened 2017-02-05
pbsuite: revision for hdf5

iMichka:


iMichka:
Needs blasr first; blocked by #4254

iMichka:
The blasr update was abandoned because upstream is slow to respond. Will do the same here. Closing.

PR #5065, opened 2017-02-06
poretools: Build a bottle for Linuxbrew: h5py: cc: error: unrecognized command line option '-R'

sjackman:


sjackman:
Circle
```
    cc -pthread -shared -L/home/linuxbrew/.linuxbrew/lib -Wl,--dynamic-linker=/home/linuxbrew/.linuxbrew/lib/ld.so -Wl,-rpath,/home/linuxbrew/.linuxbrew/lib -L/home/linuxbrew/.linuxbrew/lib -Wl,--dynamic-linker=/home/linuxbrew/.linuxbrew/lib/ld.so -Wl,-rpath,/home/linuxbrew/.linuxbrew/lib -Os -w -pipe -march=core2 -isystem/home/linuxbrew/.linuxbrew/include build/temp.linux-x86_64-2.7/tmp/pip-9a2PpU-build/h5py/defs.o -L/home/linuxbrew/.linuxbrew/opt/hdf5/lib -L/opt/local/lib -L/usr/local/lib -L/home/linuxbrew/.linuxbrew/Cellar/python/2.7.13/lib -R/home/linuxbrew/.linuxbrew/opt/hdf5/lib -R/opt/local/lib -R/usr/local/lib -lhdf5 -lhdf5_hl -lpython2.7 -o build/lib.linux-x86_64-2.7/h5py/defs.so
    cc: error: unrecognized command line option '-R'
```
https://circleci.com/gh/Homebrew/homebrew-science/3214

ilovezfs:
Same error with the older hdf5: https://github.com/Homebrew/homebrew-science/pull/5076

sjackman:
It's an error when building the Python module `h5py`. Its build script is using the command line option `-R` when it should be using `-Wl,-rpath`.

ilovezfs:
right. this is the sort of thing superenv handles.

sjackman:
Yep. Linuxbrew still doesn't have superenv. One day…

ilovezfs:
Could throw a little one-off shim at this one.

ilovezfs:
I'm baffled that upstream doesn't seem to have ever had this reported. I wonder if we're doing something weird that makes it think -R is a good idea on the platform.

ilovezfs:
Oh look! Now they've had it reported hehe

sjackman:
I've got a feeling that the issue is that the `LD` environment variable is being set to `cc`, and they expect it to be set to `ld`. `-R` is a non-standard synonym for `-rpath` for `ld`.

>            For compatibility with other ELF linkers, if the -R option is
           followed by a directory name, rather than a file name, it is
           treated as the -rpath option.

ilovezfs:
maybe we should try ENV["LD"] = "ld" ?

sjackman:
Yep, or simply
```
ENV["LD"] = nil
```

sjackman:
Funny thing is, I don't think Homebrew or Linuxbrew actually sets the `LD` environment variable.

ilovezfs:
Yeah something weird is going on or you'd think this would have been reported by now.

sjackman:
In any case, `cc -R` is broken. It should either be `cc -Wl,-rpath` or `ld -rpath`.

ilovezfs:
I think cc -Wl,-R may work too

ilovezfs:
@sjackman maybe needs to be reported to Cython instead now?

sjackman:
Yeah, I think that's likely. I wanted to first confirm that it's a Cython issue before I reported the bug. I won't have time for that myself though. =/

PR #5069, opened 2017-02-06
omcompiler 1.11.0

SimplyDanny:
### Have you:


SimplyDanny:
Since I have no possibility to test the Formula on Yosemite, I decided to use `depends_on :macos => :el_capitan`.

iMichka:
What was the build failure on Yosemite? It's annoying to get a version bump but loose Yosemite support due to that. But this is probably okay, people on Yosemite will need to roll back manually to the old version ...

ilovezfs:
:-1: to dropping Yosemite unless upstream explicitly says they no longer want to support it.

StefanD986:
@SimplyDanny Was there a specific reason why the (optional) `boost` dependency was removed?

SimplyDanny:
@iMichka As far as I can remember there was a problem with the makefile. Unfortunately, I'm not experienced enough to see how to fix it only with some terminal output and without testing. I remove the `depends_on :macos` line to obtain the output of the build process again.

@StefanD986 Boost is only needed if one wants to build the C++ simulation runtime additionally. The OMCompiler instructions claim that `./configure --with-cppruntime` is all one needs to do so. Actually, this is only correct for the `./configure` in the main OpenModelica repository, not for OMCompiler exclusively. That's the reason why I removed it.

SimplyDanny:
Closing the PR since there is a newer version of OMCompiler running on all supported macOS versions in #6421.

PR #5071, opened 2017-02-06
vtk: improve libPythonX.Y.so detection

iMichka:
Under Ubunutu Xenial (but probably other distributions),
when no brewed python is installed, the actual logic
was picking up the wrong path for the libPythonX.Y.so
file.

The file is located under /usr/lib/x86_64-linux-gnu.

When no file is found, an error is now thrown,
to bail out early with a clear error message,
instead of having to read cryptic error messages
in the build log the next time this happens.

To my knowledge there seem to be no way to get this
path programatically. sysconfig.get_config_var() could
be used (checking for LIBDIR, LIBPC), but these values
seem inconsistent on OS X, so it is probably better
to handle this by trial and error.

Also, the caveats message was removed, as this is not true
anymore. We do not link against a specific python.
This should have been removed in commit
fdbb385236a86153444847eecb5831cc433eafe0
but was forgotten.

### Have you:

### New formula: have you

### DIFF
--- a/vtk.rb
+++ b/vtk.rb
@@ -146,8 +146,12 @@ class Vtk < Formula
           args << "-DPYTHON_LIBRARY='#{python_prefix}/Python'"
         elsif File.exist? "#{python_prefix}/lib/lib#{python_version}.a"
           args << "-DPYTHON_LIBRARY='#{python_prefix}/lib/lib#{python_version}.a'"
-        else
+        elsif File.exist? "#{python_prefix}/lib/lib#{python_version}.#{dylib}"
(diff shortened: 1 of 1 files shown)


iMichka:
No need to rebuild bottles for this.

@sjackman This may interest you :)

iMichka:
The Sierra build cleanup failed, but that is unrelated.
> Recording test results
ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?
Archiving artifacts

It even produced bottles but as said before, I won't pull them, as they are not needed for that change.

sjackman:
Thanks, Michka! 🐧

PR #4751, opened 2016-12-23
astrometry-net 0.70

ilovezfs:
Created with `brew bump-formula-pr`.

PR #5085, opened 2017-02-07
astrometry-net 0.70

jonchang:
Also don't vendor `numpy`.


iMichka:
@jonchang this is a duplicate of #4751 
But you seem to have made more changes and we will see what happens with circle-ci.

You'll need to sort out with @ilovezfs which one you want to keep :)

jonchang:
Whoops, lol. Let's use the one that doesn't vendor `numpy` (this one) so the bottles are smaller.

jonchang:
Assuming that CI passes.

jonchang:
Merged!

dstndstn:
thanks from upstream :)

PR #5118, opened 2017-02-12
veclibfort: Error: Use openblas instead of veclibfort on Linux

sjackman:


sjackman:
```
Error: Use openblas instead of veclibfort on Linux
```
https://travis-ci.org/Homebrew/homebrew-science/builds/200895542#L1097

This formula needs to be modified in such a way that CI succeeds without building anything on Linuxbrew, probably using an unfulfilled requirement.

sjackman:
Resolved by #5197

PR #5121, opened 2017-02-12
qrupdate: Build a bottle for Linuxbrew

sjackman:


sjackman:
TravisCI
```
Error: No available formula with the name "homebrew/dupes/ncurses" 
Please tap it and then try again: brew tap homebrew/dupes
```
https://travis-ci.org/Homebrew/homebrew-science/builds/200913661#L1187

PR #5143, opened 2017-02-14
sumo 0.28.0

jonchang:
Closes #4582.


jonchang:
Looks like `fox` is failing on linuxbrew: 

```
/home/linuxbrew/.linuxbrew/lib/../lib/libX11.so: undefined reference to `xcb_wait_for_reply64'
```

PR #5163, opened 2017-02-18
matplotlib: Removed unnecessary resource numpy

kyungminlee:
Homebrew package `numpy` is already specified as a dependency.


jonchang:
Thanks @kyungminlee!

PR #5178, opened 2017-02-22
urh: add head-only

davidhubbard:
### Have you:

### New formula: have you


iMichka:
@davidhubbard Why not use the officially tagged release (v1.3.7)?

jonchang:
This formula has a few things that will need to be fixed, but I wonder if it's a good fit for Homebrew/science? I think Homebrew/core might be a better place for it. Any thoughts?

davidhubbard:
I'm happy with Homebrew/core. (I think Homebrew/science is more appropriate for all SDR applications, but all current SDR apps are in Homebrew/core...) Please do share the cleanups, I might as well get your review and fix those issues first before I create the PR on Homebrew/core.

I will add the official v1.3.7 release, good suggestion.

urh.rb
```diff
@@ -0,0 +1,40 @@
+class Urh < Formula
+  desc "Universal Radio Hacker"
+  homepage "https://github.com/jopohl/urh"
+  revision 1
```
jonchang: `revision` is not needed unless there is an update to the formula but the version doesn't change.

urh.rb
```diff
@@ -0,0 +1,40 @@
...
+  head "https://github.com/jopohl/urh.git", :using => :git, :branch => "master"
```
jonchang: `head "https://github.com/jopohl/urh.git"` is fine

urh.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  depends_on "pkg-config" => :build
+
+  depends_on :python3
+  depends_on "pyqt5" => ["with-python"] # pyqt5 has --without-python3, please make sure that was not selected
```
jonchang: This is unusual, seeing as how the other dependencies require `python3`. Do you mean `without-python` or `with-python3`?

urh.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python#{xy}/site-packages"
+    system "python3", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    assert File.exist?("#{bin}/urh"), "I'm an untestable GUI app."
```
jonchang: I didn't notice this before; Homebrew/core might not like that it's a GUI only app. Are there no command-line utilities that this comes with, or does this provide any functions as a library?

jonchang: I actually don't know their policy for GUI apps but it seems like there's a fair number of them in there, so maybe this isn't as big of a deal as I thought.

urh.rb
davidhubbard: pyqt5 was pretty confusing, yeah. I executed these two commands:

```
$ brew info pyqt5
pyqt5: stable 5.7.1
Python bindings for v5 of Qt
https://www.riverbankcomputing.com/software/pyqt/download5
homebrew/Cellar/pyqt5/5.7.1_1 (904 files, 47.2M) *
  Built from source on 2017-02-21 at 18:47:14 with: --with-python
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pyqt5.rb
==> Dependencies
Required: qt@5.7 (installed), sip --with-python3 (installed)
==> Requirements
Recommended: python3 (installed)
Optional: python (installed)
==> Options
--with-debug
        Build with debug symbols
--with-docs
        Install HTML documentation and python examples
--with-python
        Build with python support
--without-python3
        Build without python3 support

$ brew reinstall pyqt5 --with-python --with-python3
==> Reinstalling pyqt5 --with-python
Warning: pyqt5: this formula has no --with-python3 option so it will be ignored!
==> Using the sandbox
```
I'm confused about whether `--with-python3` is needed.

I see these lines in `pyqt5.rb`: [source](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pyqt5.rb#L19)
```
    if build.without?("python3") && build.without?("python")
      odie "pyqt5: --with-python3 must be specified when using --without-python"
    end
```

So I conclude I should `depends_on "pyqt5" => ["with-python3"]` but can I get confirmation on that? :)

urh.rb
davidhubbard: The only symlinked file is bin/urh, but there are other things I could test within `libexec/"python#{xy}/site-packages/urh"`. I'll try to improve the tests.

urh.rb
jonchang: It's probably just `depends_on "pyqt5"`. The warning is correct since there is no `--with-python3` option (it's the default)

```
 $ brew options pyqt5
--with-debug
	Build with debug symbols
--with-docs
	Install HTML documentation and python examples
--with-python
	Build with python support
--without-python3
	Build without python3 support
```

urh.rb
davidhubbard: Done

urh.rb
davidhubbard: Done

urh.rb
davidhubbard: Done

urh.rb
davidhubbard: Done

davidhubbard:
Here's an updated formula that uses the official "1.3.8" release.

jonchang:
Looks like the tests are failing:

```
Testing homebrew/science/urh
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20170224-27085-1fxd00t.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-science/urh.rb --verbose
==> python3 -c import util.crc;c = util.crc.crc_generic();assert(c.crc([1, 2, 3]) == [True, False, False, False, False, False, True, True, False, False, False, False, False, False, True, True])
ModuleNotFoundError: No module named 'util'
Error: homebrew/science/urh: failed
```

davidhubbard:
Ok, I got the import fixed. Tests pass now.

davidhubbard:
There is still the question of whether I should move this formula to `homebrew-core` or stay in `homebrew-science`.

My understanding is I should move it to `homebrew-core`. Can someone please confirm?

jonchang:
This should probably be in core because, as you say, all other SDR formula are in homebrew/core. Good luck!

PR #5180, opened 2017-02-22
racon 0.5.0 (new formula)

sjackman:
Ultrafast consensus module for raw de novo genome assembly
of long uncorrected reads.
http://genome.cshlp.org/content/early/2017/01/18/gr.214270.116


sjackman:
```
```
Ignoring because it's published.

PR #5183, opened 2017-02-23
acado 1.2.2beta (new version)

SimplyDanny:
ACADO is a software library and algorithm collection to formulate and solve
problems that occur in the field of dynamic optimization. These include model
predictive control, state and parameter estimation and robust optimization.

### Have you:

The problem with `brew audit --strict --online acado` is that the project only uses `*beta` tags. So there is no way to satisfy `brew audit` at the moment.


iMichka:
Hi. Thank you for your contribution.
This fails with:

> homebrew/science/acado:
> * Stable version URLs should not contain beta

Could upstream tag a release without beta?

SimplyDanny:
I don't think so. All tags contain a `beta`, maybe for cautiousness reasons. I didn't know that a `beta`-tag leads to a problem in Homebrew.

iMichka:
It is not making problems per se, but for us it is often a warning that it may be more work to support the formula on the long run. By experience beta versions are unstable; and only published for testing, not for packaging. For homebrew-science the rule is not so strict, we will make an exception here.

Please add the doi on top of the formula, below the sha256, # doi "DOI: 10.1002/oca.939"
I got it from that paper, I do not know if it is the original one, but that should do: http://sist.shanghaitech.edu.cn/faculty/boris/paper/acado.pdf

acado.rb
```diff
@@ -0,0 +1,50 @@
...
+  desc "Toolkit for Automatic Control and Dynamic Optimization"
+  homepage "https://acado.github.io/index.html"
+  url "https://github.com/acado/acado/archive/v1.2.2beta.tar.gz"
+  version "1.2.2beta"
+  sha256 "cf0779e64dd5d20989e97340c04ecccf542fe8d993c96b53e5c465693cb354b7"
+
+  depends_on "cmake" => :build
+  depends_on "gnuplot" => :recommended
+  depends_on "doxygen" => :optional
+  depends_on "graphviz" => :optional
```
iMichka: Are the recommended dependencies really needed?

`:recommened` generates a `--without-gnuplot` option. What happens in that case?

And for the other two, doxygen looks like the docs get build. Is there no way to disable the docs building ? I mean, I can use `--with-doxygen`, what happens then?

SimplyDanny: You are right. doxygen is only needed to build the documentation. This isn't concerned to Homebrew. I will delete the doxygen dependency.
Gnuplot is nice to have in order to plot the results although it isn't necessary for the software to build/run. I wouldn't use ACADO without it, but it may be omitted actually. What is Homebrew's philosophy in this case?
As far as I can see, Graphviz is also only used for documentation generation. I will delete it, too.

fxcoudert:
Thanks for your contribution @SimplyDanny!

mvukov:
:+1: Thanks guys!

PR #5184, opened 2017-02-23
netcdf: rewrite @rpath to make system Python happy

ilovezfs:
CC @tdsmith @iMichka @woodruffw


ilovezfs:
@tdsmith Is it overkill to be doing this for libnetcdf-cxx4.dylib also?

Issue #4825, opened 2017-01-07
brew install opencv3: error: '_mm_cvtph_ps' was not declared in this scope


timotheecour:
same as https://github.com/Homebrew/homebrew-science/issues/4788 but but @jonchang incorrectly closed this issue; and because of https://github.com/isaacs/github/issues/583 we can't reopen issues. My fix https://github.com/Homebrew/homebrew-science/pull/4750 was enough to fix opencv3 a couple weeks ago but this new error crept up in the meantime

timotheecour:
the culprit: https://github.com/Homebrew/homebrew-science/commit/42b72dda3e8507148159d642b69ce2d53a533602 introduced the regression

timotheecour:
related: https://aur.archlinux.org/packages/opencv-git/?comments=all also mentions same error

jonchang:
Hi @timotheecour, do you know if this error only occurs on 3.2, or did the removed patches in the 3.2 formula cause this error?

timotheecour:
no idea, but could we revert to previous state where things weren't broken?

timotheecour:
is this normal?

```
brew reinstall opencv3 --verbose --HEAD

-- Could NOT find PythonLibs (missing:  PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS) (Required is exact version "3.5.2")
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/katrina/home/timothee/ubuntu1604/brew/lib/python2.7/site-packages/numpy/__init__.py", line 142, in <module>
    from . import add_newdocs
  File "/katrina/home/timothee/ubuntu1604/brew/lib/python2.7/site-packages/numpy/add_newdocs.py", line 13, in <module>
    from numpy.lib import add_newdoc
  File "/katrina/home/timothee/ubuntu1604/brew/lib/python2.7/site-packages/numpy/lib/__init__.py", line 8, in <module>
    from .type_check import *
  File "/katrina/home/timothee/ubuntu1604/brew/lib/python2.7/site-packages/numpy/lib/type_check.py", line 11, in <module>
    import numpy.core.numeric as _nx
  File "/katrina/home/timothee/ubuntu1604/brew/lib/python2.7/site-packages/numpy/core/__init__.py", line 14, in <module>
    from . import multiarray
[...]

PR #5186, opened 2017-02-24
opencv3: fix for long broken linuxbrew (3 unrelated errors)

timotheecour:
@sjackman  @jonchang

fixes https://github.com/Homebrew/homebrew-science/issues/4825 which was just the tip of the iceberg...

### DIFF
--- a/opencv3.rb
+++ b/opencv3.rb
@@ -115,7 +115,7 @@ class Opencv3 < Formula
     ]
 
     # cf https://github.com/Homebrew/homebrew-science/pull/5185
-    args << "-DBUILD_OPENEXR=" + (if OS.linux? then "ON" else "OFF" end)
+    args << "-DBUILD_OPENEXR=" + (OS.linux? ? "ON" : "OFF")
 
     if OS.linux?
       # http://answers.opencv.org/question/121651/fata-error-lapacke_h_path-notfound-when-building-opencv-32/
--- a/opencv3.rb
+++ b/opencv3.rb
@@ -107,13 +107,24 @@ class Opencv3 < Formula
       -DBUILD_JASPER=OFF
       -DBUILD_JPEG=OFF
       -DBUILD_TIFF=OFF
-      -DBUILD_OPENEXR=OFF
       -DBUILD_PNG=OFF
       -DBUILD_ZLIB=OFF
       -DCMAKE_OSX_DEPLOYMENT_TARGET=
       -DJPEG_INCLUDE_DIR=#{jpeg.opt_include}
       -DJPEG_LIBRARY=#{jpeg.opt_lib}/libjpeg.#{dylib}
     ]
+
+    # cf https://github.com/Homebrew/homebrew-science/pull/5185
+    args << "-DBUILD_OPENEXR=" + (if OS.linux? then "ON" else "OFF" end)
+
+    if OS.linux?
+      # http://answers.opencv.org/question/121651/fata-error-lapacke_h_path-notfound-when-building-opencv-32/
+      args << "-DWITH_LAPACK=OFF"
+
+      # avoid error: '_mm_cvtps_ph' was not declared in this scope; cf https://github.com/RoboSherlock/robosherlock/issues/78#issuecomment-274469830
+      args << "-DOPENCV_EXTRA_CXX_FLAGS='-march=core-avx2'"
+    end
+
     args << "-DBUILD_opencv_java=" + arg_switch("java")
     args << "-DBUILD_opencv_python2=" + arg_switch("python")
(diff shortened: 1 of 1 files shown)


sjackman:
It built bottles for Linuxbrew on CircleCI. That's good. `brew test` failed:
```
==> brew test homebrew/science/opencv3 --verbose
Testing homebrew/science/opencv3
==> /usr/bin/g++-4.8 test.cpp -I/home/linuxbrew/.linuxbrew/Cellar/opencv3/3.2.0/include -L/home/linuxbrew/.linuxbrew/Cellar/opencv3/3.2.0/lib -o test

==> python -c 'import cv2; print(cv2.__version__)'
Illegal instruction (core dumped)
Error: homebrew/science/opencv3: failed
<0> expected but was
<132>.
==> FAILED
```
`Illegal instruction` usually indicates that an instruction was used that is not available on this machine, for example using an SSE 4.2 instruction on a machine that does not have them. If you have access to a Linux machine, you can test the bottle out on that machine:
https://circleci.com/gh/Homebrew/homebrew-science/3596#artifacts/containers/0
https://3596-4106792-gh.circle-artifacts.com/0/tmp/circle-artifacts.7q95bYj/opencv3-3.2.0.x86_64_linux.bottle.1.tar.gz

timotheecour:
```
brew2 test homebrew/science/opencv3
Error: Testing requires the latest version of homebrew/science/opencv3
```

`brew2 help test` doesn't show a way to test from a file; what command do u want me to run? and I guess docs should be updated for that

opencv3.rb
```diff
@@ -107,13 +107,24 @@ def install
...
+
+    # cf https://github.com/Homebrew/homebrew-science/pull/5185
+    args << "-DBUILD_OPENEXR=" + (OS.linux? ? "ON" : "OFF")
+
+    if OS.linux?
+      # http://answers.opencv.org/question/121651/fata-error-lapacke_h_path-notfound-when-building-opencv-32/
+      args << "-DWITH_LAPACK=OFF"
+
+      # avoid error: '_mm_cvtps_ph' was not declared in this scope; cf https://github.com/RoboSherlock/robosherlock/issues/78#issuecomment-274469830
+      args << "-DOPENCV_EXTRA_CXX_FLAGS='-march=core-avx2'"
```
sjackman: This line looks good, assuming that there's no way to configure opencv3 not to use AVX2.
Note to me: `-march=core-avx2` is a synonym for `haswell`.
I found these useful resources for `-march` flags:
https://gcc.gnu.org/onlinedocs/gcc-6.3.0/gcc/x86-Options.html#x86-Options
https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/driver-i386.c#L772
https://raw.githubusercontent.com/gcc-mirror/gcc/master/gcc/config/i386/i386.c

opencv3.rb
```diff
@@ -107,13 +107,24 @@ def install
...
-      -DBUILD_OPENEXR=OFF
       -DBUILD_PNG=OFF
       -DBUILD_ZLIB=OFF
       -DCMAKE_OSX_DEPLOYMENT_TARGET=
       -DJPEG_INCLUDE_DIR=#{jpeg.opt_include}
       -DJPEG_LIBRARY=#{jpeg.opt_lib}/libjpeg.#{dylib}
     ]
+
+    # cf https://github.com/Homebrew/homebrew-science/pull/5185
+    args << "-DBUILD_OPENEXR=" + (OS.linux? ? "ON" : "OFF")
```
sjackman: Note to self: this line was merged for `opencv` in #5185.

sjackman:
> could u suggest another way? still, it's an improvement over the current status where nothing builds at all; so maybe we could merge this PR in the meantime and keep the bug open until the test is fixed

Two options:
1. Is it possible to configure `opencv3` not to use AVX2?
2. Modify brew test to check whether the CPU supports AVX2 and return if it does not.

```ruby
test do
  unless Hardware::CPU::avx2?
    owarn "brew test opencv3 requires AVX2"
    return
   end
  …
end

iMichka:
I would also be in favour to merge this as soon as possible and open a new issue for the failing test, so we have more time to look at it later.

sjackman:
Merged without a bottle because of the failed test. I think this formula needs two things:
1. `depends_on "openblas" unless OS.mac?` and remove `-DWITH_LAPACK=OFF`
2. Skip the test `unless Hardware::CPU::avx2?` (see https://github.com/Homebrew/homebrew-science/pull/5186#issuecomment-283423918)

iMichka:
Closed by d829670a5d2647338fcbc2384017ac2341cab2f8

I'll have a look at the missing things to get bottles built.

PR #5204, opened 2017-03-01
Add python3 option to enable pyROOT support

david-july1991:

### DIFF
--- a/root6.rb
+++ b/root6.rb
@@ -20,7 +20,8 @@ class Root6 < Formula
   depends_on "xrootd" => :optional
   depends_on "fftw" => :optional
   depends_on "openssl" => :recommended # use homebrew's openssl
-  depends_on :python => :recommended # make sure we install pyroot
+  depends_on :python => :recommended
+  depends_on :python3 => :optional
   depends_on :x11 => :recommended if OS.linux?
   depends_on :fortran => :recommended # enabled by default since 6.08.00
   depends_on "gsl" => :recommended
@@ -34,6 +35,8 @@ class Root6 < Formula
   end
 
   def install
+    dylib = OS.mac? ? "dylib" : "so"
+
     # brew audit doesn't like non-executables in bin
     # so we will move {thisroot,setxrd}.{c,}sh to libexec
     # (and change any references to them)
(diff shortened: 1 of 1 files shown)


iMichka:
@BrewTestBot test this please

david-july1991:
Two checks were timed out and not successful. I don't have any idea to improve it. Is this request still available?

root6.rb
```diff
@@ -16,11 +16,17 @@ class Root6 < Formula
     sha256 "da8baf9db5a25c18edbc634794cbca3b773fe22b517ab1a734ff4d5bb1ae2253" => :yosemite
   end
 
+  option "with-python3", "Enable python3 support instead of python2"
```
iMichka: Don't add this option, it will be added automatically by the depends-on.
See: `Specifying other formulae as dependencies` in http://docs.brew.sh/Formula-Cookbook.html

root6.rb
```diff
@@ -16,11 +16,17 @@ class Root6 < Formula
...
   depends_on "cmake" => :build
   depends_on "xrootd" => :optional
   depends_on "fftw" => :optional
   depends_on "openssl" => :recommended # use homebrew's openssl
-  depends_on :python => :recommended # make sure we install pyroot
+  if build.with? "python3" # make sure we install pyroot
+    depends_on :python3
+  else
+    depends_on :python => :recommended
+  end
```
iMichka: This can be just:
> depends_on :python => :recommended
depends_on :python3 => :recommended

root6.rb
```diff
@@ -33,6 +39,16 @@ def config_opt(opt, pkg = opt)
     "-D#{opt}=#{(build.with? pkg) ? "ON" : "OFF"}"
   end
 
+  def py_args
+    if build.with? "python3"
+      %W[-Dpython=ON -DPYTHON_EXECUTABLE=#{HOMEBREW_PREFIX}/bin/python3]
```
iMichka: You can get the current python executable in the path with code like this:
https://github.com/Homebrew/homebrew-science/blob/master/vtk.rb#L133-L134

Because for the moment you would only use brewed python, excluding people using system python. You can also have a look at the vtk formula if you need to provide include and library paths.

david-july1991:
Thank @iMichka for your suggestion. Please check my update.

iMichka:
You need to add `dylib = OS.mac? ? "dylib" : "so"` somewhere at the beginning of the install block, so that the variable is defined.

david-july1991:
Sorry for being late to reply. I've update the code.

iMichka:
Hi. Don't know what happened to your branch but you just should have done:
`git checkout master`
`git pull origin master`
`git checkout develop` (your branch)
`git rebase master`
Then make changes, and then:
`git commit --amend`
`git push --force <forkremotename> develop`

david-july1991:
@iMichka, sorry about that. I didn't know if I could force to push. I've update it.

iMichka:
You have still the old `def py_args` code block in your PR that need to be removed.

We can ignore the travis/circle timeouts for the time being.

iMichka:
Also, could you align the style of the args addition with what we have for the vtk formula?
Using `args <<`. And if python is set to OFF by default, you do not need to explicitly set it again to OFF, which would make the code slightly shorter.

david-july1991:
Thank you for your advice. I've update it.

root6.rb
```diff
@@ -20,7 +20,8 @@ class Root6 < Formula
   depends_on "xrootd" => :optional
   depends_on "fftw" => :optional
   depends_on "openssl" => :recommended # use homebrew's openssl
-  depends_on :python => :recommended # make sure we install pyroot
+  depends_on :python => :optional
```
iMichka: Please leave python2 as recommended, because we want to build with python2 by default. This means that you will need to use the `--without-python --with-python3` syntax to install with python3; it's a little bit cumbersome but that is how it is.

See also the docs about this: http://docs.brew.sh/Formula-Cookbook.html

root6.rb
```diff
@@ -55,14 +58,44 @@ def install
...
+      else
+        odie "No libpythonX.Y.{dylib|so|a} file found!"
+      end
+      
+      args << "-Dpython=ON"
+      args << "-DPYTHON_EXECUTABLE='#{python_executable}'"
+      args << "-DPYTHON_INCLUDE_DIR='#{python_include}'"
+      args << "-DPYTHON_LIBRARY='#{python_library}'"
+    else
+      args << "-Dpython=OFF"
```
iMichka: Looking at the source code of root, there are 2 python flags: https://github.com/root-project/root/blob/1fb4972abda0568f0094ed8996b05216cae9b1f1/cmake/modules/RootBuildOptions.cmake#L139-L140

So this could be rewritten as:
> args << "-Dpython=" + ((build.with? "python") ? "ON" : "OFF")
args << "-Dpython3=" + ((build.with? "python3") ? "ON" : "OFF")

So you can move this to the `elsif` block

iMichka:
Two more changes. I think we should be good then.

david-july1991:
Oh, I didn't notice that there is python3 flag; let me add the code. And I have fixed python2 option default.

iMichka:
You can remove the else block completely, you have duplicated code now.
And move these lines out of the if/else logic:
> args << "-Dpython=" + ((build.with? "python") ? "ON" : "OFF")
args << "-Dpython3=" + ((build.with? "python3") ? "ON" : "OFF")

david-july1991:
Sorry for being late. You're right; I have modified the part.

iMichka:
@david-july1991 Thanks for your contribution and your patience.

This is now merged. No new bottles have been built because the old ones are still good.
I will maybe have a look at what takes so much time to build on Linux and try to build some Linux bottles later.

BetterWang:
On my OSX, no libPyROOT is built when I tried "brew install --with-python3 --without-python root6".
I have both python2 and python3 installed on my mac.
Could this be a problem?

schleimschein:
@BetterWang  I have the same problem. Are there any solutions?

iMichka:
Please don't post on closed issues, and open new ones. Your problem may be different. And fill out the issue template so we can have a look at your log files. Thanks.

PR #5246, opened 2017-03-09
GDCM: update formula to v2.6.7

waldyrious:
### Have you:

### New formula: have you


iMichka:
@waldyrious Thanks for your contribution.
The mac build fails with a lot of errors like this:

> compilation error: file http://docbook.sourceforge.net/release/xsl-ns/current/html/autoidx.xsl line 702 element with-param
XSLT-with-param: Failed to compile select expression '//d:index[][1]'
XPath error : Invalid expression
translate(., , )

waldyrious:
Hmmm... that seems to be an error with the Docbook manpages, not gdcm per se. I'm not sure what's going on -- I see lots of "document is empty" and "unable to load" errors, but I just checked, those links *are* available in http://docbook.sourceforge.net/release/xsl-ns/current/. Would it be ok if we changed the formula to pass the `-DGDCM_BUILD_DOCBOOK_MANPAGES=FALSE` argument by default to cmake?

iMichka:
@BrewTestBot test this please

jonchang:
Probably docbook is trying to verify the xml schema but is unable to. I don't remember if the Jenkins test-bots are forbidden from accessing the network in the install step but that might be the case.

waldyrious:
@jonchang I don't think that's the cause, because I tried building GDCM locally and also got a bunch of similar errors. I'm at a loss regarding how find out what's causing that, though...

iMichka:
I just built it locally and it worked.
I also found this discussion: https://sourceforge.net/p/gdcm/mailman/message/34866202/

We could:
- Install `docbook-xsl-ns-stylesheets`. I guess that the `dockbook` formula would work, but this needs to be investigated.
- `ENV.deparallelize` to check if this is some transient network problem due to multiple parallel downloads.

waldyrious:
@iMichka I just build gdcm after `brew install docbook` (using `make -j8`) and it worked with no issues. Unfortunately, I forgot to attempt a clean build before that, but I did `brew uninstall docbook` and re-built several times, without getting the error even once. So installing docbook does not seem to affect this issue.

Maybe we could restart the CI build and see if this works now? That would suggest it was a temporary problem with the servers the docbook files were being fetched from.

iMichka:
@BrewTestBot test this please

iMichka:
Still failing. I think @malaterre is the maintainer. He may have an idea.
Here is a link to a failing gdcm build: https://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/7530/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_homebrew_science_gdcm/

malaterre:
I would add -DGDCM_BUILD_DOCBOOK_MANPAGES:BOOL=OFF just to get rid of the issue. I have failed to debug an issue where parallel build + xsltproc accessing remote http file fails to work sometimes. So if you do not care about the man page on macOS the setting is just fine. If you really want the man page, you'll need to pull in the docbook dependencies as other linux distros are doing to build the manpages.

iMichka:
Thanks. Let's do this. @waldyrious Please update the PR accordingly.

It's just weird that it reproducibly fails on the mac buildbots. Wondering what is different there.

waldyrious:
@iMichka I gave it a shot, but I am quite new to this, so I'm not sure that approach works, or is the recommended way to achieve this. Looking forward to your feedback.

waldyrious:
At last, all tests pass! @iMichka @malaterre please take another look whenever you can.

gdcm.rb
```diff
@@ -11,9 +11,12 @@ class Gdcm < Formula
   end
 
   option "with-check", "Run the GDCM test suite"
+  option "with-manpages", "Build man pages from XML docbook"
+  
   depends_on :python => :optional
   depends_on :python3 => :optional
   depends_on "openssl" => :optional
+  depends_on "docbook" => :build if build.with? "manpages"
```
iMichka: I think I just suggested to test with that dependency to see if it helps, but it is not mandatory, because the formula built without if for me locally. You should remove this line.

waldyrious: Makes sense -- done.

iMichka:
@waldyrious Thanks for your contribution!

waldyrious:
@iMichka my pleasure -- thank you for the patience and guidance :)

PR #5263, opened 2017-03-13
scram 0.12.0

rakhimov:
### Have you:

The new release.
Testing with the new cxx14 config feature.


jonchang:
`needs :cxx14` won't be on Linux until Linuxbrew/brew#294 is merged.

rakhimov:
Thanks @jonchang ! I will try to revert the script.

rakhimov:
@jonchang I believe this is not blocked anymore.

rakhimov:
Ubuntu14.04 based CIs are broken with cxx14 due to gcc 4 vs 5 linking issues.

scram.rb
```diff
@@ -1,3 +1,4 @@
+
```
sjackman: Please remove this blank line.

scram.rb
```diff
@@ -25,7 +26,7 @@ class Scram < Formula
   depends_on "libxml++"
   depends_on "gperftools" => :recommended
 
-  needs :cxx11
+  needs :cxx14
```
sjackman: Is this a change from 0.11.6 to 0.12.0 that `scram` now requires C++14?

sjackman: Is this a change from 0.11.6 to 0.12.0 that `scram` now requires C++14?

rakhimov: scram required cxx14 in 0.11.6 as well,
but that option was not available in homebrew at the time,
so cxx11 was used instead and cmake config of scram would anyway check and pass c++14 flags.

sjackman:
Fails on TravisCI and CircleCI
```
ccache /home/linuxbrew/.linuxbrew/Library/Homebrew/shims/linux/super/g++-5  -DNDEBUG  -rdynamic CMakeFiles/scram.dir/scram.cc.o  -o ../bin/scram ../lib/libscramcore.a -lboost_program_options -lboost_filesystem -lboost_system -lboost_date_time -lboost_random -ltcmalloc_minimal -lxml++-2.6 -lxml2 -lglibmm-2.4 -lglib-2.0 -lsigc-2.0 -lboost_program_options -lboost_filesystem -lboost_system -lboost_date_time 
CMakeFiles/scram.dir/scram.cc.o: In function `(anonymous namespace)::ConstructOptions()':
scram.cc:(.text+0x173): undefined reference to `boost::program_options::options_description::options_description(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int, unsigned int)'
```
https://travis-ci.org/Homebrew/homebrew-science/builds/228059709#L3314

`needs :cxx14` will use g++ 5.3 which is not ABI compatible with the g++ 4.8 used to build the bottles of its dependencies, particularly `boost`. This formula won't work on LInuxbrew until we migrate to the new ABI.

scram.rb
```diff
@@ -1,32 +1,25 @@
...
     sha256 "42623f102a91c5253080664d53f3b54cde1f92a56e524b85a0bb5ef1f8f564b9" => :sierra
     sha256 "21c99f723de2fa25850b1aaf4cc17640170d5cead1b9858ead3bb376fd28610a" => :el_capitan
     sha256 "308d9f66e32bf77b36afbc7e77893a05f2ae5b4e67b46867a3582015055e5947" => :yosemite
   end
 
-  depends_on MacRequirement
+  needs :cxx14
+
+  # C++14 uses GCC 5.3, which is not ABI compatible with GCC 4.8
+  depends_on :macos
```
sjackman: Ack. Sorry. I forgot that this line only works on Linux, not on macOS Homebrew. You'll need:
```ruby
depends_on :macos unless OS.mac?
```

rakhimov:
Sadly, at this point, I feel linuxbrew is one big mistake, pain in the a**. (no offense intended).

sjackman:
Uh, rough. Every other Linux distro had to go through the ABI migration, and it wasn't easy for any them. Linuxbrew has the same challenge.

rakhimov:
I think the proper solution for the linuxbrew would be to actually detect the system abi version,
and use gcc-4.9 or clang for C++14 builds for gcc 4 abi on Ubuntu 14.04, for example.

sjackman:
Using GCC 4.9 is an option to support C++14 on Linuxbrew. I had an old PR open here:
https://github.com/Linuxbrew/homebrew-core/pull/1857
Another option is using GCC 5.3 with `-D_GLIBCXX_USE_CXX11_ABI=0`.

scram.rb
rakhimov: I think github diffs are now confused which empty lines are which.

rakhimov:
Closing due to upcoming new release.

PR #5280, opened 2017-03-16
NEST: fix Cython dependency, add as build resource

physicalist:
Added Cython as build-time resource, i.e. temporarily installed in buildpath during compilation.
Previously brew would fail with "Unsatisfied dependency: cython" for all versions
of Python, despite cython having been installed via pip/pip3 before
`brew nest --with-python` or `brew nest --with-python3`.
(Cython handling adapted from homebrew-core/nghttp2.rb)

### Have you:

### New formula: have you


iMichka:
This will be blocked by #5293. The upstream sha256 has changed for that release. Once #5293 has been merged you will be able to rebase and we will have a second look at your changes.

What I

physicalist:
OK, hopefully this will be resolved quickly.

If you have any idea why Homebrew might fail the Cython test in the first place, let me know. I didn't spot any obvious reason in Homebrew's core files (/usr/local/Homebrew/Library/Homebrew) yet.

tammoippen:
@physicalist I found issues later aswell and my first fix was to just remove the `build` argument/option/... in `depends_on "cython" => [:python3, :build]`. Afterwards `brew` was able to find the pip'ed`cython`. But I think your solution is more elegant.

iMichka:
Okay the other PR has been merged, this can be rebased.

Could you sort out what solution is the more elegant? :)
If we could avoid having to add resources that would be great.

physicalist:
OK I just checked, even with the latest Homebrew and homebrew-science it still complains about the missing Cython dependency. So I would rather include the resource, as I don't understand why it works at all for @tammoippen and I agree that it's more elegant ;)

@iMichka you mean I should rebase it locally in my repository, then open a new Pull Request? I tried the Github Conflict resolver, but actually screwed up by leaving the `depends_on "cython"` line in the  conflicting file. Oops :S

iMichka:
We have some docs here: http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html
See the section: `To make changes based on feedback`

I will test this locally.

physicalist:
OK, thanks for your pointers @iMichka. I've followed the instructions, the modified pull request is up. & ready for the merge.The `git push --force` was what I naïvely would have tried but was too afraid that would mess up the pull request :P

iMichka:
I am -1 on the change. Looking at this, I think it is a bug in brew.

The requirement error is thrown in the `LanguageModuleRequirement` class.

When it is initialised, `module_name` == `build`, and `import_name` == `cython`. See:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/language_module_requirement.rb#L6-L9

Due to `@import_name = import_name || module_name`, the check is then:
`/usr/bin/env python -c import build`
which of course fails. Removing the :build from `depends_on "cython" => [:python, :build]` makes it work, and the formula builds.

So I would prefer we fix this bug in brew.

iMichka:
I opened a bug in the brew repository. Let's wait for feedback.

nest.rb
```diff
@@ -62,6 +67,21 @@ def install
...
+      # then add local cython in buildpath to paths
+      ENV.prepend_create_path "PATH", buildpath/"cython/bin"
+      ENV.prepend_create_path "PYTHONPATH", buildpath/"cython/lib/python#{pyver}/site-packages"
+
+      resource("Cython").stage do
+        system "python3", *Language::Python.setup_install_args(buildpath/"cython")
+      end
+
+      # superenv kills the current python's include path, so we need to add it back
+      ENV.append_to_cflags `python3-config --includes`.chomp
```
iMichka: I think this is not needed anymore. I removed all the ENV modifiers and it compiled fine. Could you try without it?

nest.rb
```diff
@@ -70,6 +90,21 @@ def install
...
+      # then add local cython in buildpath to paths
+      ENV.prepend_create_path "PATH", buildpath/"cython/bin"
+      ENV.prepend_create_path "PYTHONPATH", buildpath/"cython/lib/python#{pyver}/site-packages"
+
+      resource("Cython").stage do
+        system "python", *Language::Python.setup_install_args(buildpath/"cython")
+      end
+
+      # superenv kills the current python's include path, so we need to add it back
+      ENV.append_to_cflags `python-config --includes`.chomp
```
iMichka: Same here

physicalist:
OK, I had a look at your bug report at  [brew/#2381](https://github.com/Homebrew/brew/issues/2381) now (link added for future reference).
I'll update my pull request ASAP, just want to make sure your suggested changes don't break it on my side either :)

Regarding the report itself - my original intention was to file a similar bug report as well, but I couldn't figure out how to properly debug this, I'm pretty new to the Ruby environment. Debugging a Formula was as far as I got, but not Homebrew proper. Is there a brew development-specific CONTRIBUTING.md somewhere that I missed?

nest.rb
physicalist: Is it possible you had Cython manually installed? In that case, it will build without any of the above. I was surprised that it worked until I realised that I had installed Cython via pip while testing different options. Otherwise:
- If I remove `ENV.prepend_create_path "PATH", buildpath/"cython/bin"`, the build fails complaining not having found Cython
(the same happens, obviously, when I remove the `resource("Cython").stage do` part)
- If I remove _only_ `ENV.prepend_create_path "PYTHONPATH", buildpath/"cython/lib/python#{pyver}/site-packages"`, it fails complaining, oddly enough, about the Cython version being <0.19 when it's definitely 0.25.2 as that's hardcoded in the Formula.
- The only flag that I could remove without it causing any issues was `ENV.append_to_cflags`; it failed to build without that line though when I originally wrote the Formula. No idea why. The changes in Homebrew/brew in all of March seemed unrelated to this issue.

physicalist: Btw do you want to get rid of comments in general?

iMichka: For the comments I was only speaking of those above the `resource "Cython" do` block. I realised that this was pretty common in the core formulae, so most of people know what this means. And no need to cross-reference another unrelated formula.

iMichka: And you are right, setting the paths is needed. And this seems to be done in some other formula too. I could also reproduce the behaviour you describe.

iMichka:
@physicalist Thanks your contribution and your patience; this has now been merged.

PR #5288, opened 2017-03-19
r: Don't hint openssl

tdsmith:
The openssl package is smarter than we are and we're confusing it. I
don't think there are any other popular packages that need to link to
openssl.

cf. https://github.com/jeroen/openssl/issues/39

### DIFF
--- a/littler.rb
+++ b/littler.rb
@@ -5,7 +5,7 @@ class Littler < Formula
 
   url "http://dirk.eddelbuettel.com/code/littler/littler-0.2.3.tar.gz"
   sha256 "98cd741c68a5c8f65b06c96d2f56d3b44979b3990335e7869b002c005ef80ba7"
-  revision 8
+  revision 9
   head "https://github.com/eddelbuettel/littler.git"
 
   bottle do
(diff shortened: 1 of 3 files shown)


ilovezfs:
It would be nice to actually fix this.

tdsmith:
Fix what?

ilovezfs:
>The openssl package is smarter than we are and we're confusing it.

You'd have to tell me.

tdsmith:
Our CPPFLAGS are stepping on their discovery process because of the way R
orders compiler arguments; this is a fix.

tdsmith:
Pulling without the plr commit.

ilovezfs:
So if you're on a version of macOS that comes with system openssl, will it use our openssl or the system openssl?

tdsmith:
Ours.

ilovezfs:
and what's with the nonsense for readline and gettext then? Why is that necessary still? This looks like the purview of superenv anyway.

tdsmith:
Superenv is no longer relevant; these hints are to help third-party extension modules, shipped out-of-tree, build correctly.

ilovezfs:
So how would those extensions know where to find our openssl in a non-standard prefix?

tdsmith:
It invokes `brew`. https://github.com/jeroen/openssl/blob/master/configure#L37

ilovezfs:
Ah OK. So presumably it could/should be doing something like that for gettext and readline, too, then?

tdsmith:
Yeah, whatever packages link against gettext and readline could theoretically do that, but probably don't; I was (pleasantly) surprised that the openssl package did this.

ilovezfs:
Yep.

PR #5334, opened 2017-03-26
hdf@4: import from homebrew/core (and versions).

MikeMcQuaid:
Keeping the name as `hdf@4` to avoid it changing back to the older name yet again.


ilovezfs:
Closing in favor of https://github.com/Homebrew/homebrew-science/pull/5335.

PR #5335, opened 2017-03-26
hdf4: import from homebrew/core

JCount:
restore name from hdf@4 to hdf4


JCount:
@dpo I am hesitant to simply rework the patch so that it applies, because with this release it appears the `HDF4_LIB_CORENAME` variable in CMakeLists.txt was removed. Leading me to worry if changing `HDF4_SRC_LIB_CORENAME` from `hdf` to `df`, essentially carrying that patch forward, may cause issues.

If the patch in `hdf4` is removed, `nexusformat` seems to link against it fine once rebuilt. However, I cannot get `gnudatalanguage` to detect it, no matter what I pass to cmake as `-DHDFDIR=`. I think this is simply due to incompatibilities with the `FindHDF.cmake` being used.

At this point my instinct is to remove the patch and the optional dependency in `gnudatalanguage`, your thoughts or those of any one else in homebrew/science would be most valued.

Thanks

JCount:
Update to `4.1.12` tabled for now.

MikeMcQuaid:
Given this was a dupe of https://github.com/Homebrew/homebrew-science/pull/5334 I'm not sure why it was pulled?

ilovezfs:
because it is more correct.

MikeMcQuaid:
I don't agree. Regardless, in future a comment or even discussion would be nice and I hope it was tested for this keg to be renamed twice.

ilovezfs:
hdf5@1.8 preexisted your PRs and was incompatible with them. There was nothing to discuss.

MikeMcQuaid:
Sure there is: why/how are they incompatible?

MikeMcQuaid:
@ilovezfs (I believe that they are incompatible FWIW but it would be nice for me and others looking at the PR to know why/how they are for future reference?)

ilovezfs:
There are two pieces of software:
hdf4
hdf5

hdf5 has two supported versions:
1.8.x
1.10.x

hdf4 has one supported version:
4.2.x

The import into core incorrectly renamed hdf4 to hdf@4. It should have been imported into science as hdf4.

Then it was realized by chdiza that there was no corresponding un-@-ed "hdf" in core. Of course that is the case because there is no such software "hdf," and moving it to science retaining the name hdf@4 would be nonsensical since there is already hdf5 and hdf5@1.8 in science.

ilovezfs:
Note the same error was made with `rebar@3`, which is a separate piece of software named `rebar3` and still needs to be fixed.

MikeMcQuaid:
@ilovezfs Thanks for explaining ❤️

PR #5348, opened 2017-03-29
opencv3: fixes for Linux

iMichka:
- Added openblas dependency for Linux, so that WITH_LAPACK can be ON.
- Removed workaround for build error on Linux, fixed by setting
the CMAKE_CXX_FLAGS='-march=core2' flag.
- Disabled the WITH_QUICKTIME option which is only working under MacOS.


iMichka:
@alalek @sjackman I didn't use `CPU_BASELINE` as this option was not available with the opencv3 version we are delivering. When building with head all the ENABLE_ARCH flags can be dropped, as opencv3 will automatically detect what is right.

Let's see if we are able to build some bottles.

opencv3.rb
```diff
@@ -75,6 +75,7 @@ class Opencv3 < Formula
   depends_on "qt5" => :optional
   depends_on "tbb" => :optional
   depends_on "vtk" => :optional
+  depends_on "openblas" if OS.linux?
```
sjackman: I prefer `depends_on "openblas" unless OS.mac?`

opencv3.rb
```diff
@@ -212,8 +204,11 @@ def install
       args << "-DENABLE_SSE41=ON" if Hardware::CPU.sse4?
       args << "-DENABLE_SSE42=ON" if Hardware::CPU.sse4_2?
       args << "-DENABLE_AVX=ON" if Hardware::CPU.avx?
+      args << "-DENABLE_AVX2=ON" if Hardware::CPU.avx2?
     end
 
+    args << "-DCMAKE_CXX_FLAGS='-march=core2'" if OS.linux?
```
sjackman: Add a comment describing why this line is necessary? Should it be…
```ruby
args << "-DCMAKE_CXX_FLAGS='-march=core2'" if OS.linux? && build.bottle?
```

PR #5354, opened 2017-03-31
mothur: Build a bottle for Linuxbrew: undefined reference to boost::iostreams::detail::zlib_base::reset

sjackman:


sjackman:
```
mothurout.cpp:(.text._ZN5boost6detail17sp_counted_impl_pINS_9iostreams16symmetric_filterINS2_6detail22zlib_decompressor_implISaIcEEES6_E4implEE7disposeEv[_ZN5boost6detail17sp_counted_impl_pINS_9iostreams16symmetric_filterINS2_6detail22zlib_decompressor_implISaIcEEES6_E4implEE7disposeEv]+0x20): undefined reference to `boost::iostreams::detail::zlib_base::reset(bool, bool)'
```
https://travis-ci.org/Homebrew/homebrew-science/builds/217232805#L1632

tseemann:
i just did a brew install of the current mothur and it did failed

https://gist.github.com/359bb1f5618769dd11340861dd256577

```
 In function `boost::detail::sp_counted_impl_p<boost::iostreams::symmetric_filter<boost::iostreams::detail::zlib_decompressor_impl<std::allocator<char> >, std::allocator<char> >::impl>::dispose()':
```

is this related?

sjackman:
Yeah, looks like the same error to me. Undefined symbols at link time. Possibly due to mismatched C++ ABI, but it needs further troubleshooting.
You could start by checking whether the symbol `boost::iostreams::detail::zlib_base::before(char const*&, char const*, char*&, char*)` exists in the library `libboost_iostreams.so` using `nm` and `c++filt`.
It's possible it's also due to mismatched header and library files from different versions of boost.

sjackman:
I'm trying a rebuild now with `--env=super`. That could help the latter case.

sjackman:
This chap has the same error message, and says he can resolve it by using the `gold` linker rather than `ld`. https://bugs.launchpad.net/ubuntu/+source/binutils/+bug/1415317

PR #5387, opened 2017-04-04
Migration from homebrew-dupes to core

fxcoudert:
Update requirements for formulas that have been migrated from `homebrew-dupes` to core.


iMichka:
@fxcoudert Maybe 5 different pull request would be better so we can go through them one by one?

iMichka:
Let's track the different PR's in this issue: #5455
I have some time right now, I'll try to get these formula green.

PR #5413, opened 2017-04-10
iqtree: Depends on zlib for Linuxbrew

sjackman:


sjackman:
```
==> brew linkage homebrew/science/iqtree
System libraries:
  /lib/x86_64-linux-gnu/libgcc_s.so.1
  /lib/x86_64-linux-gnu/libz.so.1
  /usr/lib/x86_64-linux-gnu/libgomp.so.1
  /usr/lib/x86_64-linux-gnu/libstdc++.so.6
==> brew linkage --test homebrew/science/iqtree
System libraries:
  /lib/x86_64-linux-gnu/libgcc_s.so.1
  /lib/x86_64-linux-gnu/libz.so.1
  /usr/lib/x86_64-linux-gnu/libgomp.so.1
  /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No broken dylib links
==> FAILED
```
False positive on `libgomp.so.1`

maxim-belkin:
my only question would be: why "Linux" label is in black??? is it like a "Black label"? 😄

ilovezfs:
probably because it looks a bit like a penguin

sjackman:
🐧 So it does!

PR #5431, opened 2017-04-15
simpleitk 1.0.0

iMichka:
SimpleITK_EXPLICIT_INSTANTIATION=OFF is needed to make this build,
else this fails with:
-- Configuring done
CMake Error in Code/Explicit/src/CMakeLists.txt:
Target "SimpleITKExplicit" INTERFACE_INCLUDE_DIRECTORIES property contains
path:
"/tmp/simpleitk-20170405-63774-kll36r/sitk-build/ITK-prefix/include/ITK-4.11"
which is prefixed in the source directory.
-- Generating done


iMichka:
Build time is around 1h10, so I will try to reduce the number of cores for bottling on Linux. I am still working on some tweaks to reduce the size of these bottles, there are a bunch of files that get installed which are not needed.

iMichka:
Ok `ENV["MAKEFLAGS"] = "-j1" if ENV["CIRCLECI"]` worked for circle, but in fact I need `ENV.deparallelize if OS.linux? && build.bottle?` so that circle and travis pass (I hope :)).

iMichka:
Ok the circle memory problem seems to be transient. I was able to get bottles once, so let's see if I can do it again. Ignoring the travis failure though, build should be OK, but failing due to the same memory problem.

iMichka:
> homebrew/science/simpleitk:
  * C: 50: col 36: Only use parentheses for ternary expressions with complex conditions.
  * C: 52: col 33: Only use parentheses for ternary expressions with complex conditions.
  * C: 53: col 31: Only use parentheses for ternary expressions with complex conditions.
  * C: 54: col 30: Only use parentheses for ternary expressions with complex conditions.
  * C: 55: col 28: Only use parentheses for ternary expressions with complex conditions.
  * C: 56: col 31: Only use parentheses for ternary expressions with complex conditions.
  * C: 57: col 30: Only use parentheses for ternary expressions with complex conditions.

iMichka:
Giving this a last change for Linux. Rebased. If Linux bottles do not build due to the CI limitation, I'll just merge this in as-is with the mac bottles. That PR has been open since forever

iMichka:
Merged. I'll try to build the bottle manually.

iMichka:
Linux bottle was manually built and added with f04196a1c23f931228cd6175751b57019d7a20c4

PR #5437, opened 2017-04-17
nusmv 2.6.0

bfontaine:


jonchang:
Looks like they've switched their build system to cmake. There's also a few things, this formula accesses the network during build, which needs to be fixed. It downloads https://github.com/niklasso/minisat/archive/37dc6c67e2af26379d88ce349eb9c4c6160e8543.zip as well as http://www.princeton.edu/~chaff/zchaff/zchaff.64bit.2007.3.12.zip

PR #5442, opened 2017-04-18
rstudio-server 1.0.143

### LINKED ISSUE #5486: upgrade request - rstudio-server.rb
I'm a user of the excellent formula for rstudio-server. 

Currently it installs rstudio-server 1.0.44

Rstudio-server current release is v1.0.143

It would be nice if it would be upgraded.  I tried to edit the formula myself but failed!

randy3k:
Bump to version ~1.0.136~ 1.0.143 and also added linuxbrew support.

close #5486 

### Have you:


randy3k:
Linuxbrew builds failed because the dependency `libprelude` cannot be installed. The formula was tested locally and should work just fine.

randy3k:
It seems that `libprelude` requires `ruby.h` but it cannot be found. I notice that `ruby2.0-dev` was installed at least on CircleCI, not 100% why `ruby.h` cannot be found.

randy3k:
Cannot exactly figure out why on Linux, it used to be okay a few month ago..

```
[ 92%] Linking CXX executable rsession
CMakeFiles/rsession.dir/modules/SessionFiles.cpp.o: In function `boost::detail::sp_counted_impl_p<boost::iostreams::symmetric_filter<boost::iostreams::detail::zlib_compressor_impl<std::allocator<char> >, std::allocator<char> >::impl>::dispose()':
SessionFiles.cpp:(.text._ZN5boost6detail17sp_counted_impl_pINS_9iostreams16symmetric_filterINS2_6detail20zlib_compressor_implISaIcEEES6_E4implEE7disposeEv[_ZN5boost6detail17sp_counted_impl_pINS_9iostreams16symmetric_filterINS2_6detail20zlib_compressor_implISaIcEEES6_E4implEE7disposeEv]+0x28): undefined reference to `boost::iostreams::detail::zlib_base::reset(bool, bool)'
```

randy3k:
I guess we should just merge this as it is as (1) I don't have time to fix the linuxbrew issue right now, (2) no many people will install rstudio-server via linuxbrew anyway.

ilovezfs:
@BrewTestBot test this please

randy3k:
Build failed due to https://github.com/Homebrew/brew/issues/2530, btw, I am upgrading it to v1.0.143.

randy3k:
I noted that R has updated to 3.4 and unfortunately `rstudio-server v1.0.xxx` doesn't build against R 3.4 since the R API has changed.....

We need to wait for a new release of RStudio which supports R 3.4. Meanwhile, I can make this formula build against the `master` branch of RStudio.

iMichka:
Or you could create a new R formula, R@OldVersionNumber, and use that. Look at some formula like llvm for an example. This strategy is probably cleaner, because we will be reluctant to rely on a master branch build.

Or if you are able to extract the patches to make this work with the current `rstudio-server` version, that would also be OK. But I do not know how many changes are needed to make this work.

randy3k:
It wouldn't be an easy patch, the master branch of rstudio is very different from v1.0.143...

The formula has just been updated, the master branch build can be installed with `--HEAD`.

randy3k:
Actually, the patch was easy...running on my machine now.

screen shot 2017-04-22 at 7 17 48 pm

randy3k:
Finally fixed linuxbrew build.
screen shot 2017-04-22 at 9 30 33 pm

randy3k:
The travis build installed but failed the linkage test.
```
==> brew linkage --test homebrew/science/rstudio-server
System libraries:
  /lib/x86_64-linux-gnu/librt.so.1
  /lib/x86_64-linux-gnu/libtinfo.so.5
  /lib/x86_64-linux-gnu/libutil.so.1
  /usr/lib/x86_64-linux-gnu/libedit.so.2
Missing libraries:
  libR.so
Unwanted system libraries:
  libedit.so.2
  libtinfo.so.5
```

It complained that `libR` is not found but it is a kind of intended.  The path to `libR` is controlled by the env `R_HOME` which gives us the ability to switch different version of R. Is there any way to skip or bypass the `brew linkage` test?

randy3k:
Just repeat what I said earlier.

> I guess we should just merge this as it is because
> 1. I don't have time to fix the linuxbrew issue right now,
> 2. no many people will install rstudio-server via linuxbrew anyway.

rstudio-server.rb
```diff
@@ -92,99 +113,133 @@ def install
...
+      (common_dir/"rmarkdown").install resource("rmarkdown")
+    end
+
     resource("pandoc").stage do
-      (common_dir/"pandoc/1.17.2/").install "mac/pandoc"
-      (common_dir/"pandoc/1.17.2/").install "mac/pandoc-citeproc"
+      if OS.linux?
+        arch = Hardware::CPU.is_64_bit? ? "x86_64" : "i686"
+        if File.exist?("/etc/redhat-release") || File.exist?("/etc/centos-release")
+          distritbuion = "rpm"
```
tdsmith: nit: "distritbuion" looks misspelled, albeit consistently; might be nice to factor this out into a separate method since you use it a couple of times

tdsmith:
I have no specific objections but I don't understand the Linuxbrew stuff that's happening in post-install and it would be nice to get eyes on that.

The situation seems to be that the formula works on Linuxbrew but fails the linkage test for reasons that aren't relevant at runtime?

rstudio-server.rb
randy3k: fixed, i have refactored it into a function.

randy3k:
@tdsmith 
the postinstall stuff ensures the rpath's of the files are correctly set. Ideally, we should run `brew patchelf rstudio-server`, however it is prohibited to run `brew` command in a formula. Therefore we need to use `keg.relocate_dynamic_linkage`. This solution is from @sjackman https://github.com/Linuxbrew/homebrew-core/issues/1205

sjackman:
The apparently missing `libR.so` is fine, as you said. The `R` shell script sets the `LD_LIBRARY_PATH` at run time. To silence the warning, you'd have to add `r/3.3.3_1/lib/R/lib` to the `RPATH` of the executables the require `libR.so`. With `--env=super`, you could probably accomplish that by appending that directory to the environment variable `HOMEBREW_RPATH_PATHS` (untested).

sjackman:
```
Unwanted system libraries:
  libedit.so.2
  libtinfo.so.5
```
This issue is more of a real problem. Linuxbrew provides neither of these libraries. Unless the host provides these libraries, it won't work, and we prefer not to depend on host libraries.

`libedit.so.2` is provided on Ubuntu by [`libedit`](http://packages.ubuntu.com/trusty/libedit2) 3.1-20130712.
`libedit.so.0` is provided on Homebrew/Linuxbrew by `libedit` 20160903-3.1.
I'm not sure why the SONAME don't match. The version numbers don't match. Perhaps that's why.

`libtinfo.so.5` is provided on Ubuntu by [`ncurses`](http://packages.ubuntu.com/trusty/libedit2).
Linuxbrew symlinks `libtinfo.so` to `libncursesw.so`.

Creating symlinks to provide these two SONAME would likely work. Alternatively, using `patchelf --replace-needed` to modify the `NEEDED` libraries of the executable would work as well.

sjackman:
Bumping `libedit` version to `20170329-3.1` doesn't change the SONAME. Why does Ubuntu use `libedit` from `20130712`!? License issues? In any case, I'm guessing that Ubuntu is intentionally changing the SONAME for some reason.
Debian uses `3.1-20160903-3` https://packages.debian.org/sid/libedit2 and also uses the `SONAME` of `libedit.so.2`.

sjackman:
Being that these are pretty core libraries on Debian/Ubuntu, and to provide binary compatibility, I think it's pretty reasonable for Linuxbrew to provide symlinks for both `libedit.so.2` and `libtinfo.so.5`.

sjackman:
If the fix for these two library symlinks falls to the `libedit` and `ncurses` formulae, then you can ignore the error for now in this formula.

sjackman:
Ah, there's one gotcha. Homebrew/Linuxbrew provides `ncurses` 6, which should provide `libtinfo.so.6` not `libtinfo.so.5`. There's currently no `ncurses@5` formula.

sjackman:
I can create a formula for `ncurses@5` for Linuxbrew.

sjackman:
CircleCI `Your build has exceeded the memory limit of 4G on 1 container.`
The bottle did not build because CircleCI exceeded the available memory.
See https://4331-4106792-gh.circle-artifacts.com/0/tmp/memory-usage.txt
They start two Java processes with `java -Xmx1536M` which consume 3 GB. `ant` consumes another 0.8 GB. Two GCC `cc1plus` processes consumes the remaining memory.

ilovezfs:
Shipped! Thanks @randy3k

sjackman:
I didn't notice that this formula was missing the dependencies on `libedit` and `ncurses`. I've opened https://github.com/Homebrew/homebrew-science/pull/5502 to hopefully fix the bottling for Linuxbrew.

PR #5443, opened 2017-04-18
pillow: Depends on python for Linuxbrew

sjackman:


sjackman:
```
libImaging/ImPlatform.h:10:20: fatal error: Python.h: No such file or directory
```
https://travis-ci.org/Homebrew/homebrew-science/builds/223269524#L1682

sjackman:
Possibly related to `--env=super`. Two possible solutions/workarounds:
```ruby
depends_on :python if OS.linux? && build.bottle?
```
or
```ruby
env :std if OS.linux?
```

sjackman:
Fixed with
```ruby
depends_on :python unless OS.mac?
```

PR #5447, opened 2017-04-19
xmi-msim: add support for HDF5 1.10.0

tschoonj:


tschoonj:
@ilovezfs any idea why `:fortran` no longer drags in gfortran on Yosemite and El Capitan?

xmi-msim.rb
```diff
@@ -3,23 +3,30 @@ class XmiMsim < Formula
...
   depends_on "libxslt"
   depends_on "glib"
   depends_on "hdf5"
-  depends_on "pkg-config" => :build
-  depends_on "xraylib" => "with-fortran"
+  depends_on "xraylib"
+
+  # add support for HDF5 1.10.0
+  patch do
+    url "https://github.com/tschoonj/xmimsim/commit/1459971313ea4a3cbbfdc87332b91dfcdfc0f3d7.diff"
```
iMichka: Has this patch been submitted upstream? We do not want to carry this around forever.

Patches need to be submitted here first:
https://github.com/Homebrew/formula-patches

xrmc.rb
```diff
@@ -1,9 +1,8 @@
 class Xrmc < Formula
   desc "Monte Carlo simulation of X-ray imaging and spectroscopy experiments"
   homepage "https://github.com/golosio/xrmc"
-  url "https://xrmc.tomschoonjans.eu/xrmc-6.5.0.tar.gz"
-  sha256 "4995eaaf3b4583d443d0cf2003d73d1855b443938e431a4f758a607f540e026a"
-  revision 2
+  url "https://xrmc.tomschoonjans.eu/xrmc-6.6.0.tar.gz"
```
iMichka: I'm not really happy about the version bump here. Could this be made later in a separate pull request? Just to keep the topic separated. It can also be done before this pull request is merged if this is necessary.

tschoonj: Initially I only planned on fixing an XMI-MSIM bug in this PR, but then I saw #5470. I will do some rebasing and squashing momentarily to clean things up.

tschoonj:
Ok let's wait until `fgsl` gets a rebuild against `gcc7`

xrmc.rb
ilovezfs: In general if there is a newer version, we do that in lieu of a revision bump when possible.

ilovezfs:
@BrewTestBot test this please

Issue #5411, opened 2017-04-10
r-gui El capitan bottle compiled using Sierra sources


tobadia:
### Indicate troubleshooting steps taken

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Checked that a similar issue hasn't already been filed?
- [x] Looked up the software bug tracker if reporting a build error?
- [x] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

### Bug reports / build errors:
R-gui app is correctly installed using `brew install r-gui`. It downloads `https://homebrew.bintray.com/bottles-science/r-gui-1.68.el_capitan.bottle.tar.gz` and pours fine. However when starting the app I get prompted with an OSX error "This requires osx 10.12. You're running 10.11.6".
`shasum -a 256 /Users/tobadia/Library/Caches/Homebrew/r-gui-1.68.el_capitan.bottle.tar.gz` returns the expected sha checksum as can be seen in r-gui.rb (59d782583a0273c4009c29f9bd935eef5837a79c06875e2cd62a63f6a1d8f29a).

This leads me to believe the binary in the bottle has been improperly compiled using Sierra sources instead of El Capitan. Maybe just an mix-up between version sources (might be worth checking if the Sierra version isn't actually the El Capitan one, and the other way around).
#### Brief summary of issue:
The `r-gui-1.68.el_capitan.bottle.tar.gz` bottle seems to have been incorrectly compiled using SIerra sources instead of El Capitan.

#### The complete command that you issued:
brew install r-gui

#### Gist with output of `brew doctor`:
Your system is ready to brew.

#### Link output by `brew gist-logs <formula>`:
Error: No logs.

#### Output of `brew config`:
HOMEBREW_VERSION: 1.1.12
ORIGIN: https://github.com/Homebrew/brew
HEAD: 8ea778f7c7ea790694485030d105a92a9416ef33
Last commit: 8 days ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: afe9867b31dc82e4317e7b5e268dc509acde2dd7
Core tap last commit: 79 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
[...]

PR #5454, opened 2017-04-20
r-gui: add MACOSX_DEPLOYMENT_TARGET

tobadia:
Following the issue discussed [here](https://github.com/Homebrew/homebrew-science/issues/5411) and suggestion for a PR to fix, here's my slight edits.

brew audit complained when passing #{MacOS.version.to_s} directly in the xcodebuild call so I stored its content in a variable similar to r_opt_prefix, to remain consistent.

### Have you:

(there are still two errors but were already there prior to my edit and probably irrelevant here, one is hash rocket syntax on line 3 and second is request for a test do block)

~- [ ] Removed the `revision` line, if any, when bumping the version number?~
- [X] Added/bumped the `revision` number if the changes affect the precompiled bottles?~

### DIFF
--- a/r-gui.rb
+++ b/r-gui.rb
@@ -11,6 +11,7 @@ class RGui < Formula
   homepage "https://cran.r-project.org/bin/macosx/"
   url "https://cran.r-project.org/bin/macosx/Mac-GUI-1.68.tar.gz"
   sha256 "7dff17659a69e3c492fdfc3fb765e3c9537157d50b6886236bee7ad873eb416d"
+  revision 1
 
   head "https://svn.r-project.org/R-packages/trunk/Mac-GUI"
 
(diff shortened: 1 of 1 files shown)


iMichka:
Hi. Nice finding that out.
Please add a `revision 1` line, so that the bottles get rebuild.

Did you try this syntax?:
`"MACOSX_DEPLOYMENT_TARGET=#{MacOS.version}"`
I saw in some formula that it was written like this. Not sure what the audit will say for that case.

tobadia:
I'll add the revision line to rebuild the bottles, thanks for the tip!

At first I used `"MACOSX_DEPLOYMENT_TARGET=#{MacOS.version.to_s}"` as a flag, but it somehow yielded a third warning in audit:
`* W: 40: col 57: Redundant use of `Object#to_s` in interpolation.`
I've just switched to `#{MacOS.version}` and it seems to do the trick. I assumed passing a call to a homebrew variable wasn't that clean, but it's just the `.to_s`that was redundant. Let me know if that's fine this way now.

jonchang:
Thanks for contributing to Homebrew @tobadia, and taking the time to figure this issue out! This has been merged.

PR #5477, opened 2017-04-22
rstudio-server: revision for r

ilovezfs:


ilovezfs:
let's go with https://github.com/Homebrew/homebrew-science/pull/5442

PR #5483, opened 2017-04-22
pymol: migrate tcl-tk from dupes to core

iMichka:


iMichka:
> * `revision` (line 7) should be put before `head` (line 6)
  * Dependency tcl-tk does not define option "with-x11"
  * Dependency tcl-tk does not define option "with-x11"
  * The URL http://pymol.org could use HTTPS rather than HTTP
  * Dependency python should not use option with-tcl-tk
  * Dependency tcl-tk should not use option with-x11

iMichka:
@ilovezfs 
>  * Dependency python should not use option with-tcl-tk
    * Dependency tcl-tk should not use option with-threads

I guess vendoring python is not a solution here :)

ilovezfs:
You should try dropping both tcl* and x11* here.

ilovezfs:
@iMichka the description of the tcl-tk option is
```
 option "with-tcl-tk", "Use Homebrew's Tk instead of macOS Tk (has optional Cocoa and threads support)"
```

So presumably Tk is actually available (at least on macOS) without the option.

tdsmith:
The system version has known issues; I don't know whether they're important here.

iMichka:
@tdsmith I built some combinations locally on Sierra:

- with system python in path, (no brewed python), and no python/tcl-tk dependencies in the formula: crashes 1-2 min after start
- with `depends_on "python"` and `depends_on "tcl-tk" => ["with-threads"]`: crashes 1-2 min after start
- with `depends_on "python"` and `depends_on "tcl-tk": crashes 1-2 min after start

See the full error
https://gist.github.com/iMichka/4636f24d3658bb8bfc30eeab1a56653d

Looks like this error here: https://trac.macports.org/ticket/46007

tdsmith:
Is your brewed python installed with `--with-tcl-tk`? That should be the only possibility that works.

iMichka:
I did not test that flag at all in all in my tests. For the 2 tests with brewed python; this is without `--with-tcl-tk`.

I am not really sure I understand what is the difference between the tcl-tk package, and the tk that is provided by python.

tdsmith:
Python's stdlib has a Tkinter module that binds to Tk (which is a windowing toolkit distributed with the Tcl language). That binding is determined at Python's build time.

I don't think tcl-tk is used by pymol, except through Tkinter.

iMichka:
Interestingly if I build:
`python --with-tcl-tk`, `tcl-tk --with-threads`, I get the exception directly (pymol opens and directly crashes). With the other combinations one needed to wait 1 minute or so. In the long case, one would get the apple report. In the short case, only this is printed:
> /usr/local/bin/pymol: line 4: 34380 Abort trap: 6           "/usr/local/opt/python/bin/python2.7" "/usr/local/Cellar/pymol/1.8.4.0_1/libexec/lib/python2.7/site-packages/pymol/__init__.py" "$@"

I think the last case is what most of the users saw, as this corresponds to the state of the actual formula (with as single difference the migration from dupes to core, but this should play no role here (I hope)).

iMichka:
I also tried to build the new 1.8.6.0 version; which ends up with 183 compilation warnings, and 1 error somewhere hidden in that mess ...

iMichka:
Superseded by #6066

PR #5484, opened 2017-04-22
therion: migrate tcl-tk from dupes to core

iMichka:


iMichka:
@BrewTestBot test this please

iMichka:
The Linux error was:
> extern/getline.c:137:1: error: conflicting types for 'getline'

As we never built bottles for therion, I did not investigate more. There seem to be some activity upstream and a new version (5.4.1) is out. That version fails to build on Mac for the moment, I'll open a separate PR for that.

PR #5158, opened 2017-02-17
augustus: Build a bottle for Linuxbrew: undefined reference to `boost::iostreams::zlib::okay'

sjackman:

PR #5487, opened 2017-04-23
blast: Remove fails_with :gcc => "5"

ntkme:
`fails_with :gcc => "5"` was added in 2015 (https://github.com/Homebrew/homebrew-science/commit/5b0c9db232bac6e64d108c67e55c3ba916de20ea), but it is no longer the case under the current version of blast.


ntkme:
* Travis CI failed due to timeout.
* Circle CI failed due to `augustus`, which is irrelevant here.
    * To be more specific, `augustus` depends on `boost` compiled with c++11, but `boost` is compiled without c++11 by default. I'm not sure about what's the best way to solve this.
* Jenkins failed due to an irrelevant error downloading something.

jonchang:
Builds are timing out due to the last of bottles for `boost`. C.f. https://github.com/Linuxbrew/homebrew-core/pull/2431. Going to see if I can build bottles locally

jonchang:
@BrewTestBot test this please

sjackman:
> To be more specific, augustus depends on boost compiled with c++11, but boost is compiled without c++11 by default. I'm not sure about what's the best way to solve this.

Can you elaborate on this point?

ntkme:
@sjackman Sorry I was wrong. It has nothing to do with `boost`.

It's a [GCC 5 problem](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html), and I will open a PR for that soon.

iMichka:
Should this pull request be abandoned?

ntkme:
@iMichka This PR is needed for Linuxbrew and Linuxbrew CI to build blast with GCC 5.

ntkme:
@sjackman Linuxbrew CI failed again due to `augustus` (GCC 5.1 C++ ABI issue).

The log showed that `blast` was compiled and tested without problem, and `augustus` really has nothing to do with `blast`. They have no dependency on each other. I don't know why the CI was trying to compile `augustus` from source after compiling `blast`. Is there any way to workaround that or `augustus` has to be fixed before this PR?

sjackman:
I've built a bottle for `augustus` for Linuxbrew and merged PR #5158.
I've retriggered CI for this PR.
`busco` is a reverse dependency of `blast`, and `busco` depends on `augustus`.

sjackman:
> By the way, the bottles are built with GCC 4.8 while CI runs GCC 5?

CircleCI runs Ubuntu 14 and GCC 4.8 and is used to build bottles.
TravisCI runs Ubuntu 12 and GCC 5.3 and is used to test interoperability of `glibc` and `gcc`.

ntkme:
All checks have passed. Have anything been changed in the CI environment?

iMichka:
@ntkme Thanks for your contribution!

sjackman:
@ntkme Yes, TravisCI now runs Ubuntu 14 and GCC 4.8, the same environment as CircleCI

PR #5501, opened 2017-04-25
travis: reconcile with linuxbrew/core

jonchang:
Merge changes between the `.travis.yml` in homebrew/science and the copy in [linuxbrew/core](https://github.com/Linuxbrew/homebrew-core/blob/master/.travis.yml). I will comment justifying specific changes below.


.travis.yml
```diff
@@ -9,46 +9,70 @@ addons:
     - csh
     - cython
     - python-astropy
-    - python-matplotlib
     - python-nose
     - python-pip
-    - python-scipy
```
jonchang: `python-matplotlib` and `python-scipy` are removed because we should be depending on the Homebrew bottled variants rather than the system copies.

.travis.yml
```diff
@@ -9,46 +9,70 @@ addons:
...
     - python-sphinx
 
 before_install:
   - sudo apt-get -qq update
   - sudo apt-get install -y libxml-parser-perl libxml-sax-perl
+  - DEBIAN_FRONTEND=noninteractive sudo apt-get autoremove -y --purge
```
jonchang: These system packages are removed so we can depend on the Homebrew variants for portability on bottle pours.

.travis.yml
```diff
@@ -9,46 +9,70 @@ addons:
...
-  # Fix incorrect file permissions (664)
-  - chmod 0644 *.rb
-  - umask 022
   - brew tap linuxbrew/xorg
 
 install:
   - pip install pysam
 
 script:
-  - brew tap homebrew/dupes
```
jonchang: `homebrew/dupes` is 💀 so take this out

.travis.yml
```diff
@@ -9,46 +9,70 @@ addons:
...
+      libwebp-dev
+      libwmf-dev
+      libx11-dev
+      libxml2-dev
+      llvm-3.4
+      mysql-common
+      zlib1g-dev
+  # Fix error: Incorrect file permissions (664)
+  - chmod 0644 Formula/*.rb
+  - umask 022
```
jonchang: `chmod` and `umask` commands moved up from below.

.travis.yml
```diff
@@ -9,46 +9,70 @@ addons:
...
   - if [ -f ".git/shallow" ]; then
       travis_retry git fetch --unshallow;
     fi
-  - sudo mkdir -p /home/linuxbrew
-  - sudo chown "$USER" /home/linuxbrew
-  - LINUXBREW=/home/linuxbrew/.linuxbrew
+  - HOME=/home/linuxbrew
+  - sudo mkdir $HOME
+  - sudo chown "$USER:" $HOME
+  - LINUXBREW=$HOME/.linuxbrew
```
jonchang: These all do the same thing but I've changed this to match what happens in `linuxbrew/core/.travis.yml`

.travis.yml
```diff
@@ -9,46 +9,70 @@ addons:
...
   - HOMEBREW_TAP_DIR="$(brew --repo "$TRAVIS_REPO_SLUG")"
   - rm -rf "$HOMEBREW_TAP_DIR"
   - mkdir -p "$(dirname "$HOMEBREW_TAP_DIR")"
   - ln -s "$PWD" "$HOMEBREW_TAP_DIR"
-  - export HOMEBREW_DEVELOPER="1"
+  - export HOMEBREW_DEVELOPER=1
+  - export HOMEBREW_NO_AUTO_UPDATE=1
+  - export HOMEBREW_VERBOSE=1 HOMEBREW_VERBOSE_USING_DOTS=1
   - env | grep TRAVIS | tee /tmp/travis.env
-  - export HOMEBREW_DEVELOPER="1"
```
jonchang: We oddly had multiple exports of `HOMEBREW_DEVELOPER=1` so fix that up. Also set all the other environment variables, including `HOMEBREW_VERBOSE_USING_DOTS` which hopefully will prevent some timeouts that we've seen.

.travis.yml
```diff
@@ -9,46 +9,70 @@ addons:
...
   - pip install pysam
 
 script:
-  - brew tap homebrew/dupes
   - brew tap linuxbrew/xorg
   - brew test-bot
 
 notifications:
+  slack:
+    secure: PvK4LXijxgFTbT6SvWUveluJdbe7AfnJ+Ip/xb2gVRsq71DHSi96OtaEZq+qAvU6UU07VaCY/55cTb+HZE1twCOE4aE4l6iJONVmwIb+s9bGQfV9+H/r+y+J1EJzIzAfTbsn4mGJpkG9e2326c0vGcM/lFpHH0qarhOEYs4jLt17Ijg4mk+I5jLIOZ2vZsC7IZ0W8oe7Qzomyyfzj5XW2kUa41e8CgZf4byqAvnVpPpfvkm/rODGPREOmcpOr2Pu8fhNLrswXRlhgakZgi5JPPdaRocIhnbRbyqE2Qngnlk/wzno1e3NZYDcxTa5xAgAa5h/TIpyLeB/pXfpc1n5RuP7BjA/mZSYaG5Dg1IIvGPu/23+HlZ9umMn0E0f5j3EsV66CwErJGChm6hN7xPAxLz1Mj9v2E3CIDjyYcg/aZ+O27oaUlGoBGrX51g8poq6LhG+SxHx/VkDAJ+X4T56EQYTCA+ezEIs6KdWZl7AB4ztHqTe7w4YJIIhh509iViFKX4pJdQnajNf97IvRl9hHsLa9k9GuLWOQvxjkEJ/1LSfVkNEfPxHd2btIVdg6DQsVrxyfwufF+KlQDKNHR4Jbz+KEvrWIbXWQAj+2g7xwj5l2/Tyv69gR7FqBpSxh+Ue7wrWrVCIkzn/BEU4H0nHGR5D4dOwmgiOxXVhEZjYqjU=
```
jonchang: This should post to the Linuxbrew #travisci Slack I think.

sjackman: This token will make the Slack notifications come from "Travis CI core", which isn't quite right. I'm having trouble generating a new token for Science right now. I'd suggest removing this chunk for now.

sjackman:
Thanks, Jon! Please squash and merge.

PR #5503, opened 2017-04-26
opencv3: fix name of cv2.so when building with Python 3

iMichka:
Setting the suffix variable overrides the weird logic
of opencv3, which tries to give the .so file the following
name: cv2.cpython-36m-darwin.so
Of course, that module can then not be imported in Python 3
with the standard "import cv2" statement.


tdsmith:
That should work; that's a PEP 3149 name. https://www.python.org/dev/peps/pep-3149/

iMichka:
@tdsmith Thanks, didn't know about that pep. I thought it was broken because of the name, but in fact python3 can't import because the caveats for python3 are missing, so you need to add a .pth to make python3 find cv2. I got confused by this. Some tutorials tell you to rename the cv2 file too, so I naively thought that this was really needed. Closing here.

PR #5512, opened 2017-04-29
augustus: Fix build with GCC 5.1

ntkme:
As mentioned in #5487, `augustus ` fails with GCC 5.1, and this PR fixes the problem.

Reference: [GCC 5.1 Dual ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html)


ntkme:
From the `boost` bottle:

```
$ldd /data/brew/lib/libboost_iostreams.so.1.64.0 
	linux-vdso.so.1 (0x00007ffe092f0000)
	libbz2.so.1.0 => /data/brew/lib/libbz2.so.1.0 (0x00007ff7c09a8000)
	libstdc++.so.6 => /data/brew/lib/libstdc++.so.6 (0x00007ff7c0670000)
	libgcc_s.so.1 => /data/brew/lib/libgcc_s.so.1 (0x00007ff7c0458000)
	libc.so.6 => /data/brew/lib/libc.so.6 (0x00007ff7c00a8000)
	libm.so.6 => /data/brew/lib/libm.so.6 (0x00007ff7bfda8000)
	/data/brew/Cellar/glibc/2.19/lib64/ld-linux-x86-64.so.2 (0x00007ff7c0dd8000)
```

After `--build-from-source`:

```
$ ldd /data/brew/lib/libboost_iostreams.so.1.64.0 
	linux-vdso.so.1 (0x00007ffd683b8000)
	libz.so.1 => /data/brew/lib/libz.so.1 (0x00007efe1bd78000)
	libbz2.so.1.0 => /data/brew/lib/libbz2.so.1.0 (0x00007efe1bb60000)
	libstdc++.so.6 => /data/brew/lib/libstdc++.so.6 (0x00007efe1b828000)
	libm.so.6 => /data/brew/lib/libm.so.6 (0x00007efe1b528000)
	libgcc_s.so.1 => /data/brew/lib/libgcc_s.so.1 (0x00007efe1b310000)
	libc.so.6 => /data/brew/lib/libc.so.6 (0x00007efe1af60000)
	/data/brew/Cellar/glibc/2.19/lib64/ld-linux-x86-64.so.2 (0x00007efe1c1b0000)
```

@sjackman So it turns out the `boost` bottle is not linked with `zlib`. Maybe add `zlib` as a dependency to `boost` and rebuild `boost` bottle?

augustus.rb
```diff
@@ -48,3 +54,18 @@ def caveats; <<-EOS.undent
     system "#{bin}/augustus", "--version"
   end
 end
+
```
iMichka: Patches need to be submitted to https://github.com/Homebrew/formula-patches
You can then use a `patch do` block to apply the patch (there are multiple examples in the science repo). You can use a patch even it has not been merged in formula-patches.

sjackman:
My personal preference is to use `inreplace` for one-line patches if it's possible.

augustus.rb
```diff
@@ -26,7 +25,16 @@ class Augustus < Formula
     depends_on "suite-sparse"
   end
 
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/530822e/augustus/patch-homGeneMapping.diff"
+    sha256 "63f47dedb5775311a90a58ebdbcc9263996c99c857218e7c036bd9a08052cb83"
+  end
```
sjackman: My personal preference is to use `inreplace` for one-line patches when it's possible.

augustus.rb
```diff
@@ -26,7 +25,16 @@ class Augustus < Formula
...
+
   def install
+    # Fix the build for GCC 5.1
+    # See https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
+    ENV.append "CXXFLAGS", "-D_GLIBCXX_USE_CXX11_ABI=0" unless ENV.compiler == :clang
```
sjackman:
```ruby
ENV.append "CXXFLAGS", "-D_GLIBCXX_USE_CXX11_ABI=0" if !OS.mac? && ENV.compiler == :"gcc-5"
```

sjackman:
```
10.00s==> brew linkage --test homebrew/science/augustus
System libraries:
  /usr/lib/x86_64-linux-gnu/libsqlite3.so.0
Missing libraries:
  libcolamd.so.2.9.1
  libgsl.so.19
  libgslcblas.so.0
Unwanted system libraries:
  libsqlite3.so.0
Warning: homebrew/science/augustus dependency boost was built with a different C++ standard
library (libstdc++ from gcc-4.8). This may cause problems at runtime.
==> FAILED
```
https://travis-ci.org/Homebrew/homebrew-science/builds/227571364#L1984

Looks like it may be missing dependencies on `gsl`, `suite-sparse`, and `sqlite`.
The formula says these dependencies are only needed if `--with-cgp` is enabled. Should this option be the default?

sjackman:
> Fix the build for GCC 5.1
See https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

Note that this issue is much bigger than `augustus`. Linuxbrew currently bottles using Ubuntu 14 Trust, GCC 4.8, and the old C++ ABI. Linuxbrew ought to migrate to Ubuntu 16 Xenial, GCC 5 or 6, and the new C++ ABI.

augustus.rb
ntkme: I'm ok with it. However, I think this affects any GCC >= 5.1, include GCC 6 and future releases. Is there any way to tell brew that rather than hard code it to GCC 5?

ntkme:
> Looks like it may be missing a dependency on sqlite.
Also gsl and suite-sparse. The formula says these dependencies are only needed if --with-cgp is enabled. Should this option be the default?

Agreed. I checked and found three binaries in augustus linking to system lib rather than brew lib. I will update this PR when I have time.

> Note that this issue is much bigger than augustus. Linuxbrew currently bottles using Ubuntu 14 Trust, GCC 4.8, and the old C++ ABI. Linuxbrew ought to migrate to Ubuntu 16 Xenial, GCC 5 or 6, and the new C++ ABI.

My problem is that our HPC cluster (and many others) are still running old Linux distros (e.g. CentOS 6), which requires `glibc` and `gcc-5` for Linuxbrew. We have our own gcc-4.9 formula for Linuxbrew, but I found the that due to that fact that only one`libstdc++` can be linked, thing does not work very well when having both GCC >= 5.1 and GCC < 5.1 on Linuxbrew.

augustus.rb
iMichka: My preference would be no patch at all and getting stuff fixed upstream, or adding ways to pass stuff as flags out  of the box without us needing to take care of patches :).

But you are right. If this patch is not going upstream, then an `inreplace` is what we want here.

sjackman:
Yes, I think we agree on the problem. Linuxbrew creates bottles using GCC 4.8, but the version of the `gcc` formula in Linuxbrew is GCC 5.3, which uses a different C++ ABI than GCC 4.8 by default.

ntkme:
I'm closing this PR now and trying to get issues fixed in upstream.

Nevertheless, the C++ ABI problem around Linuxbrew and GCC really needs some attention.

PR #5522, opened 2017-05-03
mira: Add desc and rename google-perftools

sjackman:
Remove `fails_with :llvm` and `fails_with :gcc`.


sjackman:
TravisCI
```
exp_flexer.cc: In member function 'virtual int EXPFlexLexer::yylex()':
exp_flexer.cc:766:9: error: no match for 'operator=' (operand types are 'std::istream {aka std::basic_istream<char>}' and 'std::istream* {aka std::basic_istream<char>*}')
    yyin = & std::cin;
```
https://travis-ci.org/Homebrew/homebrew-science/builds/228502926#L2492

sjackman:
mira 4.9.6 has the same issue.

PR #5547, opened 2017-05-04
petsc: revision for gcc

fxcoudert:


jonchang:
This is failing on due to timeout on Linux builders and a build error on macOS that I can't figure out, but a newer version of `petsc` is available which may fix it? https://www.mcs.anl.gov/petsc/download/index.html

ilovezfs:
Superseded by https://github.com/Homebrew/homebrew-science/pull/5595.

PR #5550, opened 2017-05-04
packmol: revision for gcc

fxcoudert:


jonchang:
Added patch to hopefully fix build error. Also reported upstream: https://github.com/leandromartinez98/packmol/issues/5

PR #5552, opened 2017-05-04
openblas: revision for gcc

fxcoudert:


jonchang:
Dependents `mlpack` and `libbi` failed on Linux due to lack of `boost` bottles and an inability to locate `netcdf`, respectively.

Issue #5470, opened 2017-04-22
audit: ensure default dependencies don't use options


jonchang:
Reference: https://github.com/Homebrew/brew/pull/2482

`brew audit` rejects formulae with any dependencies that require options, including optional dependencies. Looking at `brew uses --recursive <formula>`:

* `hdf5` is used by 4 formula `--with-mpi` (optional/recommended) out of a total of 56
* netcdf is used by 6 formulae `--with-fortran` (4/6 recommended) out of a total of 24
* fftw is used by 2 formulae `with-mpi` and 3 `with-fortran` (2/3 recommended) out of a total of 44
* arpack is used by 3 formula `with-mpi` (recommended) out of a total of 11
* armadillo is used by 1 formula `with-hdf5` (recommended) out of a total of 2
* cgal is used by 1 formula `with-eigen` (optional) out of a total of 6
* doxygen is used by 1 formula `with-graphviz` (optional) out of a total of 21
* gnuplot is used by 1 formula `with-qt@5.7` (recommended) out of a total of 12

I generated this list by pasting this code into `brew irb`:

```ruby
Tap.fetch("homebrew/science").formula_files.each do |fn|
    f = Formula[fn]
    f.deps.each do |d|
        puts "#{f.name}: #{d.name} => #{d.tags}" unless d.tags.empty? or d.tags.all? { |t| t.is_a?(Symbol) }
    end
end
```

|formula|depends_on|options|
|----|----|----|
|mlpack|armadillo|["with-hdf5"]|
|dealii|arpack|[:recommended, "with-mpi"]|
|getdp|arpack|[:recommended, "with-mpi"]|
|slepc|arpack|[:recommended, "with-mpi"]|
|dgtal|cgal|[:optional, "with-eigen"]|
|trilinos|doxygen|["with-graphviz", :optional]|
|gubbins|fasttree|["with-double", :recommended]|
|abinit|fftw|["with-mpi", "with-fortran", :recommended]|
|cp2k|fftw|["with-fortran"]|
|petsc|fftw|["with-mpi", "with-fortran", :recommended]|
|octave|gnuplot|[:recommended, "with-qt@5.7"]|
|dealii|hdf5|[:recommended, "with-mpi"]|
|getdp|hdf5|[:recommended, "with-mpi"]|
|petsc|hdf5|["with-mpi", :recommended]|
|trilinos|hdf5|[:optional, "with-mpi"]|
[...]

PR #5554, opened 2017-05-04
octave: revision for gcc

fxcoudert:


jonchang:
I've updated some audits, but there's still work to be done re: `gnuplot` due to https://github.com/Homebrew/homebrew-science/issues/5470.

PR #5557, opened 2017-05-04
moab: revision for gcc

fxcoudert:


jonchang:
`configure` failed to find `netcdf.so` on Linux. No clue why that happened but I'm restarting the build.

PR #5559, opened 2017-05-04
mad-x: revision for gcc

fxcoudert:


jonchang:
```
==> brew linkage --test homebrew/science/mad-x
System libraries:
  /usr/lib/libblas.so.3
  /usr/lib/liblapack.so.3
No broken dylib links
Unwanted system libraries:
  libblas.so.3
  liblapack.so.3
```

sjackman:
On Linux, use `depends_on "openblas" unless OS.mac?`
The `circle.yml` should remove `libblas-dev liblapack-dev libopenblas-dev` to prevent unintentionally linking against the host library.

PR #5561, opened 2017-05-04
libctl: link against libm

fxcoudert:


jonchang:
Linux builds were failing due to the linker not appending `-lm`. This formula didn't have bottles so I removed the `revision` line.

sjackman:
A bottle commit for Linuxbrew was added in 8506fdff6f680417a1739dcf89e8c2c8923a7fa2, but there's no corresponding bottle on Bintray.
https://bintray.com/linuxbrew/bottles-science/libctl#files
I've reverted that commit with e5208e88b25450e087b7ef340fbd30a43eebbb2d

PR #5566, opened 2017-05-04
hopdm: revision for gcc

fxcoudert:


jonchang:
```
==> Fetching hopdm from homebrew/science
==> Downloading http://dl.dropbox.com/u/72178/dist/hopdm-2.13.tar.gz
==> Retrying download
==> Downloading http://dl.dropbox.com/u/72178/dist/hopdm-2.13.tar.gz
Error: Failed to download resource "hopdm"
Download failed: http://dl.dropbox.com/u/72178/dist/hopdm-2.13.tar.gz
```

The original download link is dead due to Dropbox public URLs going :boom:. I can't figure out why the original commit https://github.com/Homebrew/homebrew-science/blob/60f930f65f81f6442a00669200ca58968d38dd74/hopdm.rb didn't use the [link on the website](http://www.maths.ed.ac.uk/~gondzio/software/get_hopdm.html) as the [Internet Archive](https://web-beta.archive.org/web/20110807153920/http://www.maths.ed.ac.uk/~gondzio/software/get_hopdm.html) reports that the page was alive and well back in 2012. @Sharpie do you have any insight into this?

Sharpie:
Hmm. I can't quite recall why I decided to go the self-hosting route. If I had to guess, I'd say it was probably that the website link isn't versioned, so I went the self-hosting route in order to guard against the checksum changing.

Here's a new public link to the file:

https://www.dropbox.com/s/6m2gpmlg1kf7gsj/hopdm-2.13.tar.gz

Although, if there's a better place to host source snapshots these days we should probably use that.

PR #5570, opened 2017-05-04
einspline: revision for gcc

fxcoudert:


fxcoudert:
Doesn't build even before revision, see PR.

PR #5579, opened 2017-05-04
atomic-pseudopotential-engine: revision for gcc

fxcoudert:


fxcoudert:
On all macOS targets:

```
functionals.F90:173:59:

         call xc_f90_lda_x_set_par(functl%conf, functl%irel)
                                                           1
Error: Type mismatch in argument 'alpha' at (1); passed INTEGER(4) to REAL(8)
```

iMichka:
Superseded by: #5874

PR #5582, opened 2017-05-04
abinit: revision for gcc

fxcoudert:


fxcoudert:
On all macOS targets:

```
m_vcoul.F90:351:24:

    call random_seed(put=seed)
                        1
Error: Size of 'put' argument of 'random_seed' intrinsic at (1) too small (12/33)
```

fxcoudert:
(This is an upstream bug, as `RANDOM_SEED` should be called first to determine the necessary size for it `PUT` argument:

```
  call random_seed(size=n)
  allocate(seed(n))
  seed = …
  call random_seed(put=seed)
```

iMichka:
Superseded by #5682.

I will ask upstream if they can provide a patch.

PR #5586, opened 2017-05-04
gdcm: 2.6.8

davidchall:
Bumped version and tested locally.


jonchang:
You can ignore the second one.

```
homebrew/science/gdcm:
  * C: 22: col 14: Don't use parentheses around a method call.
  * Dependency swig should not use option ) || build.with?("python3
```

iMichka:
@BrewTestBot test this please

iMichka:
@davidchall Thanks for your contribution!

PR #5593, opened 2017-05-05
ci: remove blas and lapack

jonchang:
Removes `liblapack-dev` and `libopenblas-dev` on Circle and Travis to avoid linking against these system libraries. CC @sjackman.


sjackman:
Good plan. These dependencies should be provided by `openblas` on Linuxbrew.

PR #5595, opened 2017-05-05
petsc 3.7.6

ilovezfs:
also serves as revision for gcc


ilovezfs:
@BrewTestBot test this please

ilovezfs:
@BrewTestBot test this please

PR #5612, opened 2017-05-08
insighttoolkit 4.11.1

iMichka:


jonchang:
```
homebrew/science/insighttoolkit:
  * Dependency vtk should not use option without-python
  * Dependency vtk should not use option with-python
```

Time to :boom: the python options?

iMichka:
The only reason I have the python options here is that I am using them :)

A lot of people use vtk with python (2/3). Insighttoolkit not so much. We could probably make Python default for VTK. The risk for both formula is that the build time will explode (I never tried it on the bots). But probably worth trying it. Probably around 2 hours of build time.

There is just a choice to be made: Python2 or Python3? We can not have both, and I do not want to maintain 4 formulas vtk + Python 2/3 and itk + Python2/3. I think we should go for Python 3 if possible, though this is maybe a more general question for all our formulas?

iMichka:
I'm ignoring these audit failure for the moment. Because vtk needs an update too. I'll make a separate commit for this later. Let's ship the new version first.

PR #5613, opened 2017-05-08
r: bump revision for gcc

slhck:
Apparently needed because of a gcc update, since I received errors similar to #2286.

### DIFF
--- a/r.rb
+++ b/r.rb
@@ -9,7 +9,7 @@ class R < Formula
   homepage "https://www.r-project.org/"
   url "https://cran.rstudio.com/src/base/R-3/R-3.4.0.tar.gz"
   sha256 "288e9ed42457c47720780433b3d5c3c20983048b789291cc6a7baa11f9428b91"
-  revision 1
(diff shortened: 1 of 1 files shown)


jonchang:
I updated the `revision` for r already; what does `brew update && brew reinstall r` say?

slhck:
Ahah, thanks. Well, here's the timeline I remember:

- I did a brew upgrade of all outdated packages yesterday and today
- I got these errors similar to #2286 today after opening up R
- Tried `brew reinstall r`, didn't help
- Tried `brew reinstall r --build-from-source`

From the previous issues I gathered that the revision had to be bumped in such a case.

jonchang:
Could you please open an issue for this and fill out the issue template? I can't reproduce this error without more information. Thanks @slhck

slhck:
I think I can't reproduce it myself anymore, since I installed everything from scratch. Should be fine though, I guess, if the revision was already bumped because of a recent gcc change.

PR #5616, opened 2017-05-09
spades: revision for gcc

ilovezfs:


ilovezfs:
@BrewTestBot test this please

sjackman:
```
In file included from /tmp/spades-20170509-95918-4k4qxx/SPAdes-3.10.1/src/projects/spades/main.cpp:14:0:
/tmp/spades-20170509-95918-4k4qxx/SPAdes-3.10.1/src/common/utils/segfault_handler.hpp:17:18: error: 'function' in namespace 'std' does not name a template type
     typedef std::function<void()> callback_t;
                  ^~~~~~~~
```
https://jenkins.brew.sh/job/Homebrew%20Science%20Pull%20Requests/239/version=sierra/testReport/junit/brew-test-bot/sierra/install_homebrew_science_spades/

Looks like a missing include
```c++
#include <functional>
```
http://en.cppreference.com/w/cpp/utility/functional/function

We can probably patch it. It'd be good to open an issue with the upstream authors.
https://github.com/ablab/spades

tseemann:
I'll email the spades crew about this.

sjackman:
Thanks, Torsten. You could also create a GitHub issue: https://github.com/ablab/spades

tseemann:
@sjackman that used to be "fake" repo they started when i complained on twitter about a lack of public site where we could all see current issues, but i seems *now* it may be legitimate. awesome!

asl:
The fix will be included into the next SPAdes release. The repo at https://github.com/ablab/spades contains all the public history and is synced once every release. It's not "live" because we tend to overwrite history and may have some branches, etc. that are might be not ready to be shown :)

sjackman:
Great! Thanks, Anton!

tseemann:
Thanks @asl for responding so quickly to my email.

PR #5620, opened 2017-05-09
nest: Enable concurrent Python 2 & 3 support

physicalist:
It is now possible to have support for both Python 2 and Python 3 in
PyNEST concurrently. Caveat: due to NEST's CMAKE build system,
this implies it needs to be fully built twice.

### Have you:

### New formula: have you

### DIFF
--- a/nest.rb
+++ b/nest.rb
@@ -27,7 +27,6 @@ class Nest < Formula
     depends_on "numpy"
     depends_on "scipy"
     depends_on "matplotlib"
-    depends_on "nose" => :python
   end
 
   depends_on :python3 => :optional
@@ -35,7 +34,6 @@ class Nest < Formula
     depends_on "numpy" => "with-python3"
     depends_on "scipy" => "with-python3"
     depends_on "matplotlib" => "with-python3"
-    depends_on "nose" => :python3
   end
 
   depends_on "libtool" => :run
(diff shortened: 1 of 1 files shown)


iMichka:
> * C: 67: col 32: Use `&&` instead of `and`.
  * python modules should be vendored rather than use deprecated `depends_on "nose" => :python`
  * Dependency numpy should not use option with-python3
  * Dependency scipy should not use option with-python3
  * Dependency matplotlib should not use option with-python3
  * python modules should be vendored rather than use deprecated `depends_on "nose" => :python3`
  * The installation was broken.
    Broken dylib links found:
      /usr/local/lib/gcc/7/libgcc_s.1.dylib

nest.rb
```diff
@@ -63,43 +63,52 @@ def install
...
-      system "make"
-      system "make", "install"
-    end
+  def caveats
+    <<-EOS.undent
+      Due to how NEST's build system is designed, concurrent Python2.x and
+      Python3.x support requires compiling NEST twice.
+      `brew test` runs the PyNEST tests with each python version for which
+      support has been enabled during compilation, so it might take a while.
+    EOS
```
jonchang: This `caveat` probably isn't needed; we usually reserve these for things that the user needs to do manually after installation

physicalist: @jonchang why would it need to have the revision bumped? The change is to the build flags only, i.e. it won't affect existing installations and rebuilding with either `--with-python` or `--with-python3` would produce the same result as before. It only enables that you can have both options at the same time.
Noted regarding the `caveat`, let's drop it then.

physicalist:
@iMichka When I compiled it, it worked fine but it used GCC 6.3 as I hadn't upgraded to GCC@7 yet. Not sure if NEST supports GCC 7.1 yet, I'd have to look into that first. What's the policy on explicitly requiring a previous GCC version if it's known to be working but the newest is not?
The dependency options for numpy, scipy and matplotlib are necessary for compatibility with Python3.x but are only ever invoked if NEST is actually compiled `--with-python3`. I see no real alternative here; vendouring these three de-facto standard libraries would be overkill. Unless they are to be removed from the _homebrew-core_ repo, but I guess there's a good reason they're there, and that they have the `python3` option in the first place.
The `nose` dependency OTOH relies on brew's `pip installed` check to avoid compilation breakage. Maybe that's possible to resolve by vendouring it, I'll have a look.

As for the `and` vs `&&`, I tried to replace it accordingly but that breaks the formula :-{

PR #5634, opened 2017-05-11
cgns: revision for hdf5

iMichka:


iMichka:
Linux test fails with:
> ==> brew test homebrew/science/cgns --verbose
Testing homebrew/science/cgns
==> h5cc -I/home/linuxbrew/.linuxbrew/opt/cgns/include -L/home/linuxbrew/.linuxbrew/opt/cgns/lib -lcgns /tmp/cgns-test-20170511-46198-mykcct/test.c

==> ./a.out
./a.out: error while loading shared libraries: libsz.so.2: cannot open shared object file: No such file or directory
Error: homebrew/science/cgns: failed
<0> expected but was
<127>.
==> FAILED

iMichka:
@sjackman

jonchang:
I'm really confused why this is getting linked twice?

```
bash-4.2$ ldd a.out 
	linux-vdso.so.1 =>  (0x00007ffc99e49000)
	libcgns.so.3.3 => /home/linuxbrew/.linuxbrew/opt/cgns/lib/libcgns.so.3.3 (0x00007f86d6aa4000)
	libsz.so.2 => not found
	libz.so.1 => /lib64/libz.so.1 (0x00007f86d688e000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f86d668a000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f86d6388000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f86d5fc7000)
	libhdf5.so.101 => /home/linuxbrew/.linuxbrew/lib/libhdf5.so.101 (0x00007f86d5a07000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f86d6d8e000)
	libsz.so.2 => /home/linuxbrew/.linuxbrew/lib/libsz.so.2 (0x00007f86d57f3000)
```

iMichka:
There is also a problem with libsz.so.2 on mac here: #5636

May be related?

jonchang:
It might be an issue with `h5cc` passing bad arguments down to the linker. I ran the same command with verbose but I can't make heads or tails of the output...

```
Using built-in specs.
COLLECT_GCC=gcc-4.8
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,java,fortran,ada,go,lto --enable-plugin --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.8.5 20150623 (Red Hat 4.8.5-11) (GCC) 
COLLECT_GCC_OPTIONS='-I' '/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/include' '-I' '/home/linuxbrew/.linuxbrew/opt/zlib/include' '-I' '/home/linuxbrew/.linuxbrew/opt/szip/include' '-c' '-v' '-mtune=generic' '-march=x86-64'
 /usr/libexec/gcc/x86_64-redhat-linux/4.8.5/cc1 -quiet -v -I /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/include -I /home/linuxbrew/.linuxbrew/opt/zlib/include -I /home/linuxbrew/.linuxbrew/opt/szip/include test.c -quiet -dumpbase test.c -mtune=generic -march=x86-64 -auxbase test -version -o /tmp/ccL6xEgb.s
GNU C (GCC) version 4.8.5 20150623 (Red Hat 4.8.5-11) (x86_64-redhat-linux)
	compiled by GNU C version 4.8.5 20150623 (Red Hat 4.8.5-11), GMP version 6.0.0, MPFR version 3.1.1, MPC version 1.0.1
GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/4.8.5/include-fixed"
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../x86_64-redhat-linux/include"
#include "..." search starts here:
#include <...> search starts here:
 /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/include
 /home/linuxbrew/.linuxbrew/opt/zlib/include
 /home/linuxbrew/.linuxbrew/opt/szip/include
 /home/linuxbrew/.linuxbrew/include
 /usr/lib/gcc/x86_64-redhat-linux/4.8.5/include
 /usr/local/include
 /usr/include
End of search list.
GNU C (GCC) version 4.8.5 20150623 (Red Hat 4.8.5-11) (x86_64-redhat-linux)
	compiled by GNU C version 4.8.5 20150623 (Red Hat 4.8.5-11), GMP version 6.0.0, MPFR version 3.1.1, MPC version 1.0.1
GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
Compiler executable checksum: 356f86e67978d665416e07d560c8ba0d
COLLECT_GCC_OPTIONS='-I' '/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/include' '-I' '/home/linuxbrew/.linuxbrew/opt/zlib/include' '-I' '/home/linuxbrew/.linuxbrew/opt/szip/include' '-c' '-v' '-mtune=generic' '-march=x86-64'
 as -v -I /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/include -I /home/linuxbrew/.linuxbrew/opt/zlib/include -I /home/linuxbrew/.linuxbrew/opt/szip/include --64 -o test.o /tmp/ccL6xEgb.s
GNU assembler version 2.28 (x86_64-pc-linux-gnu) using BFD version (GNU Binutils) 2.28
COMPILER_PATH=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/:/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/:/usr/libexec/gcc/x86_64-redhat-linux/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/:/usr/lib/gcc/x86_64-redhat-linux/
LIBRARY_PATH=/home/linuxbrew/.linuxbrew/lib/../lib64/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/:/lib/../lib64/:/usr/lib/../lib64/:/home/linuxbrew/.linuxbrew/lib/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../:/lib/:/usr/lib/
COLLECT_GCC_OPTIONS='-I' '/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/include' '-I' '/home/linuxbrew/.linuxbrew/opt/zlib/include' '-I' '/home/linuxbrew/.linuxbrew/opt/szip/include' '-c' '-v' '-mtune=generic' '-march=x86-64'
Using built-in specs.
COLLECT_GCC=gcc-4.8
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,java,fortran,ada,go,lto --enable-plugin --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.8.5 20150623 (Red Hat 4.8.5-11) (GCC) 
COMPILER_PATH=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/:/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/:/usr/libexec/gcc/x86_64-redhat-linux/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/:/usr/lib/gcc/x86_64-redhat-linux/
LIBRARY_PATH=/home/linuxbrew/.linuxbrew/lib/../lib64/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/:/lib/../lib64/:/usr/lib/../lib64/:/home/linuxbrew/.linuxbrew/lib/:/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../:/lib/:/usr/lib/
COLLECT_GCC_OPTIONS='-I' '/home/linuxbrew/.linuxbrew/opt/zlib/include' '-I' '/home/linuxbrew/.linuxbrew/opt/szip/include' '-v' '-L/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib' '-L/home/linuxbrew/.linuxbrew/opt/zlib/lib' '-L/home/linuxbrew/.linuxbrew/opt/szip/lib' '-mtune=generic' '-march=x86-64'
 /usr/libexec/gcc/x86_64-redhat-linux/4.8.5/collect2 --build-id --no-add-needed --eh-frame-hdr --hash-style=gnu -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 /home/linuxbrew/.linuxbrew/lib/../lib64/crt1.o /home/linuxbrew/.linuxbrew/lib/../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.8.5/crtbegin.o -L/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib -L/home/linuxbrew/.linuxbrew/opt/zlib/lib -L/home/linuxbrew/.linuxbrew/opt/szip/lib -L/home/linuxbrew/.linuxbrew/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.8.5 -L/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/home/linuxbrew/.linuxbrew/lib -L/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../.. test.o -lcgns -lsz /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib/libhdf5_hl.a /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib/libhdf5.a -lsz -lz -ldl -lm -rpath /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/4.8.5/crtend.o /home/linuxbrew/.linuxbrew/lib/../lib64/crtn.o
```

sjackman:
```ruby
    system compiler, "-I#{opt_include}", "-L#{opt_lib}", "-lcgns", testpath/"test.c"
```
should be
```ruby
    system compiler, "-I#{opt_include}", testpath/"test.c", "-L#{opt_lib}", "-lcgns"
```
but I don't think that's the cause of this error.

sjackman:
> I'm really confused why this [`libsz.so.2`] is getting linked twice?

My best guess is that the executable depends on two different shared libraries, each of which depends on `libsz.so.2`, and one of those two shared libraries (possibly `libcgns.so.3.3`) does not have its `RPATH` set correctly.

iMichka:
Just rebased and made the change on the `system compiler` line.

iMichka:
@jonchang @sjackman I was able to make this work by adding this to the compilation line:
> "-Wl,-R#{Formula["szip"].lib}"

I also found a nice article which explains what these flags do:
https://gms.tf/ld_library_path-considered-harmful.html

 * The compiler option -Wl instructs the compiler to pass the option following the first comma directly to the linker. All following commas are interpreted as argument delimiter.
* On Linux, -Wl-path,SOMEDIR and -Wl,-RSOMEDIR is equivalent due to option parsing magic - for compatibility reasons -R is overloaded. If the argument of -R is a filename the option has a different effect.

Would that be OK?

sjackman:
Hi, Michka. `"-Wl,-R#{Formula["szip"].lib}"` should be `"-Wl,-rpath=#{Formula["szip"].lib}"`, which is more portable. `-R` is for compatibility with other non-GNU linkers.

> For compatibility with other ELF linkers, if the -R option is
followed by a directory name, rather than a file name, it is
treated as the -rpath option.

If `cgns` depends on `szip`, then Linuxbrew's superenv automatically adds `"-Wl,-rpath=#{Formula["szip"].lib}"` to calls to either `cc` or `ld`. If that's not working for this formula, it would be good to investigate why.

iMichka:
> superenv automatically adds

Yes but here we are manually compiling a test file in the test block, using the h5cc compiler. superenv has no influence on that I guess. So the flags need to be set manually.

sjackman:
Ah, thank. I hadn't caught that this was in the `test` block, which uses `stdenv`. Or is it a bare environment, not even `stdenv`? `stdenv` adds `$HOMEBREW_PREFIX/lib` to the default `RPATH`. So it should be able to find the library of `szip` is linked.

sjackman:
`Stdenv` adds `$HOMEBREW_PREFIX/lib` to the default `RPATH` by setting the `LD_RUN_PATH` environment variable to `$HOMEBREW_PREFIX/lib`. See https://github.com/Linuxbrew/brew/blob/master/Library/Homebrew/extend/ENV/std.rb#L68

iMichka:
@sjackman I looked at this part of the code. It is correctly called, and if I print the content of `ENV` in the test block, I see for example `-Wl,-rpath,/home/linuxbrew/.linuxbrew/lib"`. I think everything is fine on that side.

@jonchang may be right. The compiler used here is not gcc; it is h5cc. I guess that in that case, whatever env hacks we use, these seem to be ignored. I guess that ENV is only valid for gcc or clang (because these are the compilers Homebrew knows about).

Looking at the output provided by @jonchang, I saw that the `COLLECT_GCC_OPTIONS` part is the critical one. If we add the lib path manually like I propose, it ends up in there and it works.

Interestingly you can also call:
```
h5cc -show 
gcc-4.8 -I/home/linuxbrew/.linuxbrew/opt/zlib/include -I/home/linuxbrew/.linuxbrew/opt/szip/include -L/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib/libhdf5_hl.a /home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib/libhdf5.a -L/home/linuxbrew/.linuxbrew/opt/zlib/lib -L/home/linuxbrew/.linuxbrew/opt/szip/lib -lsz -lz -ldl -lm -Wl,-rpath -Wl,/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib
```

Here you see that there is only a `-Wl,-rpath -Wl,/home/linuxbrew/.linuxbrew/Cellar/hdf5/1.10.1/lib`. So maybe the fix is to be done on the h5cc side. Not sure I have time chasing this down for the moment, so I propose to add my solution as a quick-fix (for Linux only, with a comment). Thoughts?

sjackman:
I have real dislike for compiler wrappers like `h5cc` and `mpicc`. You're right that because the compiler is `h5cc` rather than `gcc`, superenv is not being used. Cue sad music. 🎶😢 A wishlist todo item could be to add superenv support for these two compiler wrappers. A short-term hack is certainly easier in this case.

`brew install --env=std cgns` may work, as well as adding `env :std unless OS.mac?`. If you have time, could you check whether this is true?

I'm happy with your suggested fix of adding
```ruby
"-Wl,-rpath=#{Formula["szip"].opt_lib}"
```

iMichka:
@sjackman I tested your propositions, that didn't work. It would be great if we had support for h5cc and mpicc. Maybe one day :D

sjackman:
Thanks for the tests, Michka.

PR #5643, opened 2017-05-11
nexusformat: revision for hdf5

iMichka:


iMichka:
Ignoring audit:
> homebrew/science/nexusformat:
>  * Dependency 'hdf@4' was renamed; use new name 'hdf4'.

iMichka:
Follow up patch is here: #5692

PR #5646, opened 2017-05-11
radx 20170315

iMichka:


iMichka:
> FATAL: Could not checkout 6058d6d79bb94a12bd606e935802cc4e9e13b82d
hudson.plugins.git.GitException: Command "git checkout -f 6058d6d79bb94a12bd606e935802cc4e9e13b82d" returned status code 128:

@BrewTestBot test this please

iMichka:
Fails on mac:
> type mismatch at 1st parameter ('H5::H5Object &' vs 'H5::H5Location &')
  extended.iterateAttrs(Hdf5Utils::appendAttrNames, NULL, &attrNames);
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/include/H5Object.h:76:42: note: passing argument to parameter 'user_op' here
        int iterateAttrs(attr_operator_t user_op, unsigned* idx = NULL, void* op_data = NULL);
                                         ^
1 error generated.

iMichka:
There is a new version (20170315). Still fails to build with the same error. I will report the error upstream.

iMichka:
Will need hdf5@1.8 for Linux first: https://github.com/Linuxbrew/homebrew-core/pull/3788

iMichka:
@BrewTestBot test this please

iMichka:
Here we go. Uses the newest hdf5, but had to do the dirty work myself and wrote a patch. Patch and build failure has been reported upstream.

radx.rb
```diff
@@ -14,19 +13,29 @@ class Radx < Formula
...
+  depends_on "bzip2" unless OS.mac?
 
-  # Prevents build failure on Mac OS X 10.8 and below
-  # FIXME: Remove when it gets fixed upstream (reported)
   patch do
-    url "https://gist.githubusercontent.com/tomyun/ee3a910e07c9ccc4610e/raw/3b151798488c5dc7091506a25ac704dad6687e97/radx-fix-sockutil-mac.diff"
-    sha256 "19d55b7beb985a6facc75a02a92739c3a4208797eea2cbd0f2353a86c7aa90db"
+    # Fixes compilation with hdf5 1.10.0
+    # Reported https://github.com/NCAR/lrose-core/issues/8
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/3124ede999c9ffd700652d20f0455c97512f84a3/radx/radx-hdf5-20170315.diff?full_index=1"
```
ilovezfs: since it's a raw url the full index thing isn't applicable

iMichka: Didn't knew that. I just put it everywhere because of my paranoia, had already some trouble with spurious sha256 values changing. I hope upstream comes up with a better patch and a new release, then we can drop this ... but I do not have high hopes.

PR #5648, opened 2017-05-11
petsc: revision for hdf5

iMichka:


iMichka:
Linux build fails with:

iMichka:
Ignoring audits:
> homebrew/science/petsc:
>  * Dependency hdf5 should not use option with-mpi
>  * Dependency netcdf should not use option with-fortran
>  * Dependency fftw should not use option with-fortran

Linux build was already broken last time

PR #5650, opened 2017-05-11
moab: revision for hdf5

iMichka:


iMichka:
Linux build fails with:
> configure: error: "NetCDF not found or not working"

PR #5651, opened 2017-05-11
nanopolish 0.6.0

iMichka:


iMichka:
Fails on mac with:
> src/nanopolish_consensus.cpp: In function 'void filter_outlier_data(std::vector<HMMInputData>&, const string&)':
src/nanopolish_consensus.cpp:534:69: error: call of overloaded 'abs(unsigned int)' is ambiguous
         double n_events = abs(rs.event_start_idx - rs.event_stop_idx) + 1.0f;
                                                                     ^
In file included from /usr/local/Cellar/gcc/7.1.0/include/c++/7.1.0/cstdlib:75:0,
                 from /usr/local/Cellar/gcc/7.1.0/include/c++/7.1.0/stdlib.h:36,
                 from src/nanopolish_consensus.cpp:9:
/usr/include/stdlib.h:129:6: note: candidate: int abs(int)
 int  abs(int) __pure2;
      ^~~

iMichka:
There is a new version (0.6.0). This fails to build locally, just double-checking with the bot. I'll notify upstream as soon as the logs are there.

iMichka:
I opened a PR upstream with a patch. Got it to build :)

PR #5110, opened 2017-02-11
gmap-gsnap: make check fails on Linuxbrew

sjackman:

PR #5655, opened 2017-05-11
gmap-gsnap: Fix make check failures

sjackman:
Fix the two make check failures with ENV.deparallelize.

PASS: coords1.test
FAIL: align.test
PASS: iit.test
FAIL: setup1.test

Fixes Homebrew/homebrew-science#5110.


ilovezfs:
:tada:

PR #5659, opened 2017-05-11
mbsystem: revision for hdf5

iMichka:


iMichka:
Linux build error can be ignored, `gdal` fails to build, and needs to bottled for Linux first.

iMichka:
Ignoring for the moment:
> homebrew/science/mbsystem:
> * The URL http://www.mbari.org/data/mbsystem/mb-cookbook/index.html could use HTTPS rather than HTTP

PR #5662, opened 2017-05-11
gmt4: revision for hdf5

iMichka:


iMichka:
Linux builds failed due to lack of `gdal` bottles.

Ignoring this for the moment, there were no Linux bottles here before. I want to get the revisions out for the mac users as fast as possible.

PR #5665, opened 2017-05-11
slepc: revision for hdf5

iMichka:


jonchang:
`dealii` https://github.com/Homebrew/homebrew-science/pull/5677 and `getdp` https://github.com/Homebrew/homebrew-science/pull/5674 will need to be rebuilt after this bottle is merged.

iMichka:
Ignoring:
> homebrew/science/slepc:
>  * Dependency arpack should not use option with-mpi

There was no Linux build before and it is still failing with (during the make test step from scalapack):

PR #5666, opened 2017-05-11
octave: revision for hdf5

iMichka:


iMichka:
Ignoring:
> homebrew/science/octave:
>  * C: 104: col 21: Unnecessary spacing detected.
>  * Dependency gnuplot should not use option with-qt@5.7

iMichka:
dynare will need to be rebased (#5669)

PR #5667, opened 2017-05-11
vtk: revision for hdf5

iMichka:


iMichka:
Pcl (#5676) and statismo (#5672) will need to be bumped.

PR #5668, opened 2017-05-11
openimageio: revision for hdf5

iMichka:


iMichka:
On Linux, `scons` failed to build:
> Last 150 lines from /home/ubuntu/.cache/Homebrew/Logs/field3d/01.scons:
In file included from include/Foundation.h:39:0,
                 from include/IStreams.h:39,
                 from src/IStreams.cpp:36:
include/UtilFoundation.h:71:18: fatal error: half.h: No such file or directory
 #include <half.h>

Ignoring that as there was no Linux bottle here before.

Ignoring the Mac audit too:
> homebrew/science/openimageio:
>  * Dependency boost-python should not use option with-python3

PR #5669, opened 2017-05-11
dynare: revision for hdf5

iMichka:


iMichka:
Rebased on master because octave was updated

iMichka:
@BrewTestBot test this please

iMichka:
Linux fails with:
> In file included from /home/linuxbrew/.linuxbrew/Cellar/gcc/5.3.0/include/c++/5.3.0/cstdint:35:0,
                 from /home/linuxbrew/.linuxbrew/Cellar/octave/4.2.1_3/include/octave-4.2.1/octave/octave-config.h:64,
                 from /home/linuxbrew/.linuxbrew/Cellar/octave/4.2.1_3/include/octave-4.2.1/octave/mex.h:51,
                 from ../../../sources/dynmex.h:27,
                 from ../../../sources/kronecker/sparse_hessian_times_B_kronecker_C.cc:28:
/home/linuxbrew/.linuxbrew/Cellar/gcc/5.3.0/include/c++/5.3.0/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^

iMichka:
Ignoring the Linux build error for the moment.
Also ignoring:
> * Dependency slicot should not use option with-default-integer-8
>  * Dependency slicot should not use option matlab=

PR #5671, opened 2017-05-11
insighttoolkit: revision for hdf5

iMichka:


iMichka:
Ignoring:
> homebrew/science/insighttoolkit:
>  * Dependency vtk should not use option without-python
>  * Dependency vtk should not use option with-python

Statismo will need a bump: #5672

PR #5673, opened 2017-05-11
vigra 1.11.1

iMichka:


iMichka:
Fails with:
```
[100%] Built target test_threadpool
./run_test_classifier.sh: line 1: 50294 Killed: 9               ./test_classifier
make[3]: *** [test/classifier/test_classifier] Error 1
make[3]: *** Deleting file `test/classifier/test_classifier'
make[2]: *** [test/classifier/CMakeFiles/test_classifier.dir/all] Error 2
make[1]: *** [CMakeFiles/check.dir/rule] Error 2
make: *** [check] Error 2
```

ilovezfs:
@iMichka 1.11.1 is out https://github.com/ukoethe/vigra/releases/download/Version-1-11-1/vigra-1.11.1-src.tar.gz

iMichka:
Passing on:
>homebrew/science/vigra:
>  * Dependency boost-python should not use option c++11

ilovezfs:
Did you actually test the python bindings?

vigra.rb
```diff
@@ -40,6 +44,11 @@ class Vigra < Formula
   end
 
   def install
+    if build.with? "python"
+      venv = virtualenv_create(libexec)
+      venv.pip_install resources
```
ilovezfs: @tdsmith is virtualenv suitable for python bindings such as vigra's?

iMichka:
I have no clue. Just trying to get the revision bumps out for hdf5. Pushed the change.

Does this mean that the audit is misleading or wrong?

We can make a second pass for Python, I will need more time to investigate this in detail.

ilovezfs:
No it means I think virtualenv is unsuitable for Python bindings and won't automatically be picked up. So if it works at all, it's going to be using the numpy from system Python not the one your installing.

If you just want to get the revisions for hdf5 out, don't make changes for the audits especially if you've not actually tested them.

iMichka:
> No it means I think virtualenv is unsuitable for Python bindings

Is it possible to vendor bumpy without virtualenv?

>If you just want to get the revisions for hdf5 out, don't make changes for the audits especially if you've not actually tested them.

Tried to do half - half. Bumping without fixing stuff and trying to fix other things here and there. Maybe no the best strategy :)

iMichka:
BTW:
> Warning: python modules have explicit framework links
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/analysis.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/blockwise.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/colors.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/filters.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/fourier.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/geometry.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/graphs.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/histogram.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/impex.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/learning.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/noise.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/optimization.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/sampling.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/utilities.so
  /usr/local/opt/vigra/lib/python2.7/site-packages/vigra/vigranumpycore.so

vigra.rb
tdsmith: No, it isn't. Bindings and their dependencies both need to be visible globally in order to be useful.

tdsmith:
> Is it possible to vendor bumpy without virtualenv?

Yes. The safest strategy is to install numpy and other deps to libexec with setup_install_args and then drop a .pth file into lib. I'll actually document this today.

tdsmith:
(This isn't really vendoring; it depends on why you need to do it. "Depend on `numpy`" seems like probably the right thing.)

PR #5674, opened 2017-05-11
getdp: revision for hdf5

iMichka:


schoeps:
I think one should add a python dependency. At least on my machine I cannot compile getdp without it...

iMichka:
Let's wait for the builds to come back. We will see what needs to be done :)

iMichka:
Rebased on master because `slepc` was updated.

iMichka:
Linux build will need scalapack bottles first: #5699
Ignoring the remaining mac audits for the moment.

@schoeps Please open a new issue and fill out the issue template if you think that the Python dependency is really needed. Thanks.

PR #5677, opened 2017-05-11
dealii: revision for hdf5

iMichka:


iMichka:
Rebased on master because `slepc` was updated.

iMichka:
> homebrew/science/dealii:
>  * Dependency netcdf should not use option with-fortran

Ignoring that one

Linux build failed during the scalapck test step:

iMichka:
The scalapack problem for Linux is being worked on here: #5699

PR #5678, opened 2017-05-11
ncview: revision for hdf5

iMichka:


iMichka:
Fails on Linux with:
> /usr/bin/ld: cannot find -lcurl

Let's try by adding the curl dependency.

iMichka:
Rebasing, pysam was removed from travis; that should fix the second Linux build.

PR #5679, opened 2017-05-11
etsf_io: revision for hdf5

iMichka:


iMichka:
Ignoring this for the time being:
> homebrew/science/etsf_io:
> * Dependency netcdf should not use option with-fortran

PR #5680, opened 2017-05-11
anvio: revision for hdf5

iMichka:


iMichka:
Linux builds fails due to `scipy` failing to build. Ignoring this for the moment, there was no Linux bottle here anyway.

PR #5682, opened 2017-05-11
abinit 8.2.3

iMichka:


iMichka:
Fails on mac:
> m_vcoul.F90:351:24:
>
>    call random_seed(put=seed)
>                        1
> Error: Size of 'put' argument of 'random_seed' intrinsic at (1) too small (12/33)
make[3]: *** [m_vcoul.o] Error 1

iMichka:
Rebased as `etsf_io` was bumped.

iMichka:
Will also serve as a revision for the gcc/gfortran bump.

iMichka:
Reported upstream: https://forum.abinit.org/viewtopic.php?f=3&t=3615

iMichka:
Ignoring for the time being:
```
  * Dependency fftw should not use option with-mpi
  * Dependency fftw should not use option with-fortran
  * Dependency netcdf should not use option with-fortran
```

PR #5684, opened 2017-05-11
shogun: depend on opencv@2

iMichka:


iMichka:
Circle ran out of memory

iMichka:
Rebased. Let's try a rebuild. There were Linux bottles, so this must work. If not I'll try a -j1 call.

PR #5686, opened 2017-05-11
med-file: revision for hdf5

iMichka:


bblacey:
@iMichka - thanks for updating the the med-file formula to build bottles against the latest hdf5 version.  I took a look at the [Linux failure](https://travis-ci.org/Homebrew/homebrew-science/builds/231266040#L3541) and I am not sure what the root cause is because the med-formula does not explicitly set the c++ std (perhaps, hdf5 does?).  That said, without the updated med-file bottles on macOS, the [FreeCAD CI builds are failing](https://travis-ci.org/FreeCAD/FreeCAD/jobs/231437966#L5473).  If the Linux issue isn't a quick fix, would it be possible to make the new macOS bottles available before Linux issue is resolved?

iMichka:
Linux fails with:
> Linux fails with:
> brew linkage --test homebrew/science/med-file
System libraries:
  /lib/x86_64-linux-gnu/libgcc_s.so.1
  /usr/lib/x86_64-linux-gnu/libpython2.7.so.1.0
  /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No broken dylib links
Unwanted system libraries:
  libpython2.7.so.1.0

When we added the formula this error was not there. I think this check is fairly new ? (@sjackman)
I do not know why the library is linked to python, this should not have happened. Let me a few hours to try to debug this. If we are not able to fix it we will skip the Linux bottles for the moment.

bblacey:
@iMichka - sounds good - thanks for the prompt response and action.

sjackman:
@iMichka One option is to remove `python-dev` from `circle.yml`
Another option is to pass `--without-python` to `configure` (or whatever option it requires)

bblacey:
The Python bindings should be "supportable" on Linux and removing python bindings from the med-file formula/bottles will break FreeCAD

sjackman:
Sorry, I misread the formula. I thought the Python binding were optional. They are in fact recommended. My mistake.
Removing `python-dev` from `circle.yml` would likely resolve the issue.
It's unexpected that `cmake` is picking up the host Python library rather than the brewed one. The location of the brewed Python is passed to `cmake`.
```ruby
      python_prefix=`#{HOMEBREW_PREFIX}/bin/python-config --prefix`.chomp
      python_include=Dir["#{python_prefix}/include/*"].first
      python_library=Dir["#{python_prefix}/lib/libpython*" + (OS.mac? ? ".dylib" : ".so")].first
```

iMichka:
I'm annoyed because on my ubuntu 16.10, it picks up the homebrew python. We can't see the full log on travis or circle, I'll try in the linuxbrew docker to see what is going on.

But at the end of the day, I am not happy it links to python at all. On mac we are using '-undefined dynamic_lookup' for this. That should be the default on Linux. Though some libraries will still need a hard link to Python libraries. I am not sure what `brew linkage homebrew/science/med-file` tests. It does not really tell me which .so file is wrong.

iMichka:
@sjackman I tested inside docker, I am getting the following hint:
> Possible undeclared dependencies
>  python

I think we should remove the `option` statements, and use `depends_on :python => :recommended`. This makes python default and adds the `--without-python` flag. What do you think?

bblacey:
@iMichka - I will run a local test using the modified formula to confirm the end result is what is expected/needed.  Should have the results shortly.

iMichka:
All green :) Merged.

iMichka:
I think there is a bug in homebrew here, but anyway the right syntax is the one I used, see http://docs.brew.sh/Formula-Cookbook.html

sjackman:
> I am not sure what brew linkage homebrew/science/med-file tests. It does not really tell me which .so file is wrong.

`brew linkage --reverse homebrew/science/med-file` will give you that information.

sjackman:
The previous syntax was also why I misread the formula. Thanks for changing it.

sjackman:
@iMichka wrote…
> I am not happy it links to python at all. On mac we are using '-undefined dynamic_lookup' for this.

Can you elaborate? Are you saying that it should load the Python library `libpython2.7.so.1.0` as a plugin using `dl_open`?

iMichka:
I'm referring to this article.

The thing is that now on Linux the .so files are hard-linked to brewed Python. So I can't use them with system Python, or Anaconda Python (as long as these are ABI compatible, which would be 2.7.X here). You will get the famous `Abort Trap 6` error if you use a different Python than brewed Python now.

This is why we use the `-undefined dynamic_lookup` flag on OS X. I always thought this was the default on Linux (the flag does not exist but the linker has this behaviour), so I was not excepting to find any references to python when calling `ldd` on the .so files.

sjackman:
When the shared library is linked by `ld`, the shared libraries linked with `-l` are added to the `DT_NEEDED` section of the shared library. Is that the behaviour that you are expecting? What is the `ld` link command of the shared library you're interested in?

sjackman:
You can inspect the `DT_NEEDED` section of an ELF file using either `brew linkage --reverse` or `readelf -d | grep NEEDED` or `patchelf --print-needed`.

sjackman:
The dynamic linker `ld.so` searches the `RPATH` specified in the ELF file for the files (`SONAMES`) specified by the `DT_NEEDED` section of that ELF file.

bblacey:
The recent formula changes look good and work as expected/needed.  Thanks for quickly resolving the blocking issue!

iMichka:
@sjackman I will need some time to read through this and have a look at the details. I think the situation is OK like this. Thanks for the explanation. I will need to experiment a little bit first to see how this influences builds with Python linkage.

PR #5688, opened 2017-05-11
blasr: revision for hdf5

iMichka:


iMichka:
@BrewTestBot test this please

iMichka:
Reported upstream: https://github.com/PacificBiosciences/blasr/issues/355

sjackman:
@tmozgach The blasr formula is at 2.2. There's two upstream releases since then: 3.0 and 4.0.0. Could you please update the blasr formula to version 4.0.0?
https://github.com/PacificBiosciences/blasr/releases
https://github.com/PacificBiosciences/blasr/archive/smrtanalysis-4.0.0.tar.gz

iMichka:
I thought of doing that too, but they are not marked as the latest release (2.2 is the latest one).

But it may worth a try; if somebody can pick this up I would be glad, I have way too many PR's open :)

sjackman:
> but they are not marked as the latest release

I'm not sure why that is, but my guess is that it's unintentional.

PR #5692, opened 2017-05-11
nexusformat: fix hdf4 dependency

iMichka:
Fixes: Dependency 'hdf@4' was renamed; use new name 'hdf4'.


iMichka:
All green. No need for new bottles, this is an optional dependency.

PR #5699, opened 2017-05-12
scalapack: Build a bottle for Linuxbrew: took more than 10 minutes since last output

sjackman:


sjackman:
CircleCI
```
==> make test
...........
command umask 022; PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH"; eval "$(perl -I$HOME/perl5/lib/perl5 -Mlocal::lib)"; brew install patchelf && brew tap homebrew/dupes && brew tap linuxbrew/xorg && brew test-bot --keep-old took more than 10 minutes since last output
```
https://circleci.com/gh/Homebrew/homebrew-science/4800

sjackman:
@iMichka @ilovezfs `make test` is supposed to output `.` every minute, but I've seen a few cases, like this one, where it goes ten minutes without outputting a `.`, and the build times out. It's often reproducible on both CircleCI and TravisCI. Any thoughts?

iMichka:
@sjackman Did you try to set override test timeout in the circle config file?
https://circleci.com/docs/1.0/cabal-test-timeout/
https://circleci.com/docs/1.0/configuration/#modifiers

I guess we could increase the timeout.
Another "workaround" would be to call "ctest --verbose" instead of "make test". Then we would get some output. I'm running the build to see and the longest tests last around 120 seconds.

iMichka:
Test 93 lasted 1418 seconds here. Thats a lot. So even if we enable the verbose output, this won't help. I guess that the test timeout should be raised to something like 30 minutes.

sjackman:
Thanks for the CircleCI `timeout` tip! I didn't know about it. I'm testing it out now. I wonder whether TravisCI has a similar option.

sjackman:
```
==> make test
.................
 took more than 120 minutes to run
```
CircleCI timed out during `make test`. We could disable `make test` on CircleCI.

iMichka:
Too bad. Yes, the test ran forever on my setup. Normally a test suite should run fast. Everything is worthless. +1 on the solution to skip the tests on Linux.

sjackman:
TravisCI
```
==> brew audit homebrew/science/scalapack --online
Error validating server certificate for 'https://icl.cs.utk.edu:443':
 - The certificate hostname does not match.
Certificate information:
 - Hostname: icl.utk.edu
 - Valid: from Jan 25 00:00:00 2017 GMT until Jan 25 23:59:59 2020 GMT
 - Issuer: InCommon, Internet2, Ann Arbor, MI, US
 - Fingerprint: B7:36:A1:46:F7:E8:51:B4:52:87:D2:49:80:F1:A7:01:96:6A:01:4E
(R)eject, accept (t)emporarily or accept (p)ermanently? 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
https://travis-ci.org/Homebrew/homebrew-science/builds/232700910#L2457

iMichka:
screen shot 2017-05-16 at 18 11 04

Seems the sysadmins from Tennessee is sleeping?

iMichka:
Taking over here (#5928), as I need bottles for 2-3 other formulae.

sjackman:
Superseded by PR #5928.

PR #5701, opened 2017-05-12
sundials: Build a bottle for Linuxbrew

sjackman:


iMichka:
Has been migrated to core. Work can be picked up when it is merged to linuxbrew-core.

PR #5704, opened 2017-05-12
qrupdate: Fix Linux bottle sha256

iMichka:
There was probably a problem during the publishing
of the bottle, as it was not on bintray. The bottles
were rebuilt manually, and published manually.
There is a slight 1Kb difference in size between
the new bottles and the old ones (which never got uploaded)
but could still be fetched from circle.ci.

This fixes the sha256 for the Linux bottle.


sjackman:
Thanks, Michka!

PR #5675, opened 2017-05-11
libbi: revision for hdf5

iMichka:

PR #5705, opened 2017-05-13
qrupdate: revision for new bottles

iMichka:
The Linux bottle is a rebuild (has a 1 in his name)
and the mac bottles are the original ones.

Commit 8fd654de62a5b3fd1993f5d44716028315469945 did
not really fix the situation.

With this commit a full new set of bottles will
be built and cleanly uploaded for Mac and Linux.


iMichka:
Dynare and libbi will need a bump: #5669 and #5675

PR #5712, opened 2017-05-15
freec: fix the missing bottle for Linuxbrew

tmozgach:


sjackman:
Thank you for your first contribution to Homebrew/science, Tanya! :tada:

PR #5722, opened 2017-05-17
getdp 2.11.1

ilovezfs:
also serves as revision for gmsh


iMichka:
@BrewTestBot test this please

iMichka:
Ignoring for the time being:
> homebrew/science/getdp:
  * Dependency arpack should not use option with-mpi
  * Dependency hdf5 should not use option with-mpi

No Linux bottles need to be pulled, there were none before.

PR #5243, opened 2017-03-08
Plink2

jmannau:
I have updated the plink2 formula to build the latest stable version (1.90 b3.46) as described by this link https://www.cog-genomics.org/plink2
Note: there isn't a publish tar.gz archive for the 3.46 beta source, so I have linked to the appropriate commit source archive https://github.com/chrchang/plink-ng/commit/96f3637b62f4fddcf1e5a16929b77c637867d8cd
This does result in a audit error, but as there is not an official tarsal archive of this version available.

### Have you:

- [X] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [X] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [X] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [X] Built your formula locally prior to submission with `brew install <formula>`?

### New formula: have you

- [X] Written a sensible test? (The best test is to compile and run a sample program.)

### Updates to existing formula: have you

- [X] Removed the `revision` line, if any, when bumping the version number?
- [X] Added/bumped the `revision` number if the changes affect the precompiled bottles?
- [X] Not altered the `bottle` section?
- [X Checked that the tests still pass?

PR #5729, opened 2017-05-18
Updated plink formula to include 1.90b4 as a stable build and 1.90b43…

jmannau:
I have updated the plink2 formula to build the latest tagged version (1.90 b4) as described by this link https://www.cog-genomics.org/plink2
I have also included a developer option to install 1.90b43 from git.
This is an updated version of a previous pull request https://github.com/Homebrew/homebrew-science/pull/5243 that got a but messy.

### Have you:

### DIFF
--- a/plink2.rb
+++ b/plink2.rb
@@ -1,14 +1,11 @@
 class Plink2 < Formula
   desc "Free, open-source whole genome association analysis toolset."
   homepage "https://www.cog-genomics.org/plink2"
+  url "https://github.com/chrchang/plink-ng/archive/v1.90b4.tar.gz"
+  version "1.90b4"
+  sha256 "2ab909fc6c04062be56be86cf00ad6b3927316a34ae4ceb5fc4c250074cb3d26"
   # doi "10.1186/s13742-015-0047-8"
   # tag "bioinformatics"
-
-  stable do
-    version "1.90b4"
-    url "https://github.com/chrchang/plink-ng/archive/v1.90b4.tar.gz"
-    sha256 "2ab909fc6c04062be56be86cf00ad6b3927316a34ae4ceb5fc4c250074cb3d26"
-  end
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


plink2.rb
```diff
@@ -1,15 +1,14 @@
...
   # doi "10.1186/s13742-015-0047-8"
   # tag "bioinformatics"
-  homepage "https://www.cog-genomics.org/plink2"
-  sha256 "2f4afc193c288b13af4410e4587358ee0a6f76ed7c98dd77ca1317aac28adf0d"
-  revision 1
 
-  depends_on :fortran
-  depends_on "openblas" => :optional
-  depends_on "zlib"
+  stable do
```
iMichka: No need for a stable block: we use that only if we need to add a patch to stable. Just update the url's on top.

plink2.rb
```diff
@@ -18,15 +17,26 @@ class Plink2 < Formula
     sha256 "540b6a0f57c322040f1338655456a32c7971a3320fdc17d09fba582930c5722e" => :yosemite
   end
 
+  devel do
```
iMichka: Not sure we want to babysit a devel block here, that may be a moving target. People with want to bleeding-edge can build with `--HEAD`, that may be enough, no?

iMichka:
Could you also add:
`depends_on "openblas" unless OS.mac?` for the Linux build. Seems it is needed as it looks for the `cblas.h` file.

plink2.rb
jmannau: Hi, there are a few release channels for plink2. There are the official tagged releases (https://github.com/chrchang/plink-ng/releases). There are also stable development builds (http://www.cog-genomics.org/plink2/). These aren't tagged in github but do correspond to specific commits (i.e.: 1.9b4.3 - https://github.com/chrchang/plink-ng/commit/49bcb9168478aad0ae047055f52ebbd9593392f3 and 1.9b4.4 - https://github.com/chrchang/plink-ng/commit/6690ddffae5543dd4756694c3fcf6c89acef67d0 )
I have used the devel block to allow people to install the stable development builds or the official tagged releases.

-- HEAD will allow people to install the bleeding edge.

plink2.rb
jmannau: I have removed the stable block as requested

jmannau:
@iMichka I have added the unless macOS for the openblas dependancy as well.

PR #5766, opened 2017-06-03
r: do not use deprecated ENV.Og flag

iMichka:
Looking at the manual, -O (ENV.O0) is all what is needed.
http://valgrind.org/docs/manual/manual-core.html


ilovezfs:
`-O0` means no optimization. We almost never want that unless the build is very broken, in which case it needs to be reported.

iMichka:
@ilovezfs see http://valgrind.org/docs/manual/manual-core.html

> On rare occasions, compiler optimisations (at -O2 and above, and sometimes -O1) have been observed to generate code which fools Memcheck into wrongly reporting uninitialised value errors, or missing uninitialised value errors. We have looked in detail into fixing this, and unfortunately the result is that doing so would give a further significant slowdown in what is already a slow tool. So the best solution is to turn off optimisation altogether. 

It's just for some valgrind debugging, and it is an option here. So it will not impact anybody.

Anyway valgrind has `Required: macOS <= 10.11`, so the option is also pretty useless on the current OS version ...

ilovezfs:
OK.

>Anyway valgrind has Required: macOS <= 10.11, so the option is also pretty useless on the current OS version ...

They're working on it: https://bugs.kde.org/show_bug.cgi?id=365327

PR #5771, opened 2017-06-05
libmatio: build with hdf5 by default

iMichka:
and fix audits.

Removing the option is needed as the openmeeg needs
libmatio with hdf5.


libmatio.rb
```diff
@@ -27,14 +24,10 @@ def install
       --prefix=#{prefix}
       --with-zlib=/usr
       --enable-extended-sparse=yes
+      --with-hdf5=#{HOMEBREW_PREFIX}
```
ilovezfs: Formula["hdf5"].opt_prefix

iMichka:
Thanks for the review @ilovezfs

PR #5783, opened 2017-06-07
libmatio: add missing hdf5 dependency

iMichka:
With commit ffa1416a4d29fb2879285275fa0aed2fff8826b0,
the dependency was removed, but it is still needed
because hdf5 needs to be installed first.


agramfort:
indeed. I think this is it as I have the same error message with old version with omitting the  --with-hdf5 argument.

I guess will be happy again tomorrow. Thanks @iMichka

PR #5479, opened 2017-04-22
gnudatalanguage: migrate hdf4 from versions to science tap

iMichka:

PR #2496, opened 2015-07-07
gnudatalanguage: support for plplot 5.11.0

tschoonj:
The last release of plplot (Homebrew/homebrew#41187) introduces several backwards
incompatibilities that need patching in gnudatalanguage.

This commit should take care of that...

PR #5799, opened 2017-06-11
gnudatalanguage: Conflicts with plplot

rwhogg:
Both install a pltek binary.

Signed-off-by: Bob W. Hogg <rwhogg@linux.com>

`plplot` has been vendored since #5479 and only one formula (homebrew/science/scamp) needs `plplot`, so I don't think this is a big deal, but as I don't use these formulae I'm no expert on the subject.

Note that I didn't add a conflict to `plplot`, since it seemed weird to me to add a conflict with a tap to a core formula.

LMK if you have other plans around this; https://github.com/Homebrew/homebrew-science/pull/2496#issuecomment-120086566 suggested moving `plplot` to science, which might make sense whether you decide to merge this or not.

### Have you:


iMichka:
@rwhogg The only reason why I vendored plplot is because it is built with x11 (DPLD_xwin=ON), and we were able to get rid of `depends_on "plplot" => "with-x11"`, which was flagged as red on the buildbot.

Conflicts are maybe a limitation when vendoring other formulae to get rid of options.

I think moving plplot to science is not a good idea, we are moving stuff in the other direction: we are slowly starting to move some formulae from science to core. And I think some core plotting libs like plplot are fine in core.

I do not know if `gnudatalanguage` needs `pltek` (probably not), but the binary could also be renamed as `pltek-gdl-vendored` or whatever.

rwhogg:
> I do not know if gnudatalanguage needs pltek (probably not), but the binary could also be renamed as pltek-gdl-vendored or whatever.

As I'm not a user of the formula in question, I think it's probably safest to close this and let someone with greater knowledge of the situation handle this. Probably should have submitted this as an issue rather than a PR, my bad.

iMichka:
We should merge this. At least it is clear what the situation is and there is no surprise for our users. I do not know much about these formulas. So I am +1 on your PR, I didn't tell it in my previous message. I just wanted to trigger a discussion about the drawback of the vendoring we did here; and to explore alternative solutions.

PR #5812, opened 2017-06-13
dynare 4.5.0

houtanb:
### Have you:


houtanb:
Hi, I also have a quick question. Is there anything specific that needs to be done to have a linuxbrew bottle created?

ilovezfs:
>Is there anything specific that needs to be done to have a linuxbrew bottle created?

The build must succeed on circleci.

houtanb:
@ilovezfs ok, we can forget that for now. I'll work on it when I have more time. Best I can tell the above, failing tests were run for linuxbrew. The formula should pass on homebrew. Have I misunderstood?

ilovezfs:
The audit is failing
```
homebrew/science/dynare:
  * Dependency slicot should not use option with-default-integer-8
```

houtanb:
@ilovezfs this seems like a preexisting issue that is independent of the version bump. See #5470

ilovezfs:
indeed.

houtanb:
@ilovezfs so that will not prevent the version bump here? Also, I have not followed the conversation. Why can't a formula be dependent on another formula created with an option?

To explain, compiling slicot with 8-byte integer support is needed when we compile mex files for Matlab where 64-bit matrix indexing is used. However, it is not needed when we compile mex files for Octave where only 32-bit indexing of matrices is supported. If we cannot pass an option to the formula, how can we support mex files for both of these platforms?

ilovezfs:
@houtanb no it doesn't need to block the version bump.

>Why can't a formula be dependent on another formula created with an option?

For a few different reasons, but the main one is that it exposes builds to breaking when the source build of a dependency breaks, even if the dependency has a bottle.

There are a few ways to address the audit here:
1) make 8-byte integer support the default for slicot. Is there any reason someone would affirmatively want a slicot build without 8-byte support?
2) make a separate version of the slicot formula that has 8-byte-support
3) vendor slicot with 8-byte-support into the formula

ilovezfs:
And indeed the audit has pointed us in the right direction. According to analytics, over 80% of the slicot builds are currently being done with that option, so we should make it the default (or even just always turn it on).

ilovezfs:
@houtanb https://github.com/Homebrew/homebrew-science/pull/5813

ilovezfs:
Thanks @houtanb! Shipped.

houtanb:
@ilovezfs Thanks so much for your work. We must be on different time zones because I would like to have responded before you advanced so much.

There seems to be a problem with the bundle that is distributed for `dynare 4.5.0`, at least on macOS 10.12, as it is missing `/usr/local/Cellar/dynare/4.5.0/lib/dynare/mex/octave/kalman_steady_state.mex`. As this is the mex file associated with `slicot`, I wonder if the changes made caused the configure script when run by BrewBot to say that slicot was not present and hence not build this file for Octave. Do you save the BrewBot builds somewhere so I can help debug this?

And, though only 20% of users who use `slicot` build without 8-byte integer support, my guess is that those are Octave users. So I wonder if forcing 8-byte integer support is not creating a problem for them.

ilovezfs:
Note that the change to slicot is to install both library files not just one library file.

> As this is the mex file associated with slicot, I wonder if the changes made caused the configure script when run by BrewBot to say that slicot was not present and hence not build this file for Octave

It used to be
```
depends_on "slicot" => "with-default-integer-8" if build.with? "matlab="
```
Now it's
```
depends_on "slicot" if build.with? "matlab="
```

So I'm not sure how changes to slicot are relevant to the bottle given that `if build.with? "matlab="` would be false before and after the change.

>As this is the mex file associated with slicot, I wonder if the changes made caused the configure script when run by BrewBot to say that slicot was not present and hence not build this file for Octave.

Again, slicot is not a dependency unless you pass `--with-matlab=`.


>Do you save the BrewBot builds somewhere so I can help debug this?

https://jenkins.brew.sh/job/Homebrew%20Science%20Pull%20Requests/619/version=sierra/consoleFull

houtanb:
@ilovezfs you are totally right. This is an error that exists in the formula. `slicot` should be required for both Matlab and Octave builds, the difference being that 8-byte integer `slicot` should be used with the Matlab build while 4-byte integer `slicot` should be used for the Octave build, but our configure system takes care of this.

For the fix, see PR #5818

Thanks again for all your help!

ilovezfs:
You're welcome! Have you verified that the change in #5818 leads to `/usr/local/Cellar/dynare/4.5.0/lib/dynare/mex/octave/kalman_steady_state.mex` being present as expected?

houtanb:
@ilovezfs yes, with the small exception of the revision number, `/usr/local/Cellar/dynare/4.5.0_1/lib/dynare/mex/octave/kalman_steady_state.mex` exists

ilovezfs:
Great!

PR #5837, opened 2017-06-18
tensorflow: remove formula

iMichka:
Tensorflow already exists in homebrew-core
under the "libtensorflow" name.


cglong:
It looks to me that `tensorflow` and `libtensorflow` are actually different, the former containing the Python library itself and the latter containing a C-based interface:
```
chrislong@gMac ~> brew list libtensorflow 
/usr/local/Cellar/libtensorflow/1.2.1/include/tensorflow/c/c_api.h
/usr/local/Cellar/libtensorflow/1.2.1/lib/pkgconfig/tensorflow.pc
/usr/local/Cellar/libtensorflow/1.2.1/lib/libtensorflow.so
```

iMichka:
@cglong It's the same source code, coming from https://github.com/tensorflow/tensorflow/

What may be different is that the version in science had python options, and the version in core didn't. If want Python, you will need to submit a pull request in homebrew-core, so that the formula there can be modified.

petr-ujezdsky:
Well, that's a bummer :(. I want to try out the `tensorflow`, but I do not want to go through the `sudo easy_install ...` and `sudo pip ...` hassle on their [official install guide](https://www.tensorflow.org/install/install_mac).

I was really happy to see that there is tensorflow formulae on [brew official site](http://formulae.brew.sh/formula/tensorflow).
Unfortunately the simple
```
pujezdsky@Petr:~$ brew search tensorflow
==> Searching local taps...
libtensorflow
==> Searching taps on GitHub...
==> Searching blacklisted, migrated and deleted formulae...
```
did not work.

After some investigation I have found it here. _But it has been removed from this tap and left as core's "unusable" `libtensorflow` C library._

I think that those two should coexist. `libtensorflow` as C library to be used in other programs and `tensorflow` as the "main one".

**Can I ask you politely to add `tensorflow` with the Python functionality to the core?** Or at least add it back to this tap, because it is not duplicate.

cglong:
I've been attempting to merge `libtensorflow` with this formula, but have unfortunately run into tensorflow/tensorflow#13080 (caused by the GitHub checksum changes). Hopefully that gets resolved soon and we can tentatively get a PR open 😄

ilovezfs:
@cglong I'd just use inreplace :)

cglong:
Haha, I didn't expect to find you here! 😄 I was going to wait until I opened the PR, but I may as well get this question out of the way now:

As part of the Bazel build process, [around 20 dependencies are downloaded](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl). It looks like Bazel is as rigorous about deps as Homebrew (versioning tarballs, comparing checksums, etc.). Is it okay that these are downloaded outside of the Homebrew `resource` model?

ilovezfs:
>Haha, I didn't expect to find you here!

I am everywhere :)

>Is it okay that these are downloaded outside of the Homebrew resource model?

If they're being built from source, yes.

PR #5843, opened 2017-06-19
vcftools: fix audit

iMichka:


iMichka:
Yes, you are right. That was not the intent. I was checking versions with `live-check`, but I did not double-check the release page. I changed the PR to just fix the audit :)

PR #5869, opened 2017-06-22
siril: revision for gsl

iMichka:


iMichka:
No Linux bottles here before; ignoring Linux build failure.

PR #5873, opened 2017-06-22
astrometry-net 0.72

iMichka:


iMichka:
Failing on Linux with:

> Error: homebrew/science/astrometry-net contains conflicting version recursive dependencies:
  openssl@1.1, openssl
View these with `brew deps --tree homebrew/science/astrometry-net`.

ilovezfs:
http://astrometry.net/downloads/astrometry.net-0.71.tar.gz is out

mwcraig:
http://astrometry.net/downloads/astrometry.net-0.72.tar.gz is now out....

dstndstn:
It looks like this is a dependency clash between openssl and openssl@1.1.  The openssl dependency comes via python and numpy, while openssl@1.1 comes from wget.  The wget "dependency" is extremely weak; wget is not used at all in the main astrometry.net processing tasks; it only appears in some very peripheral python code (for retrieving SDSS data), and even there, there is also the option to use "curl".  So I think you'd be fine to drop the "wget" dependency from the formula.

iMichka:
Yes. After some digging lately, we came to the conclusion that we should avoid to use wget if possible. Ideally all the formulae should use `openssl@1.1`, but that is not the case ... because people are really slow at updating their libraries.

I'll try to rebuild this without wget.

iMichka:
Done with e5168de17082317031aba5e6d1e8ade7ac7e3953

@dstndstn Thanks for finding out that wget was not needed here.

PR #5868, opened 2017-06-22
xmi-msim: revision for gsl

iMichka:

PR #5878, opened 2017-06-22
fgsl: revision for gsl

iMichka:


iMichka:
Needs xmi-msim first #5868

iMichka:
My mistake, it is xmi-msim that needs fgsl. This build was all green and built bottles for mac and Linux.

PR #5879, opened 2017-06-22
getdp: revision for gsl

iMichka:


iMichka:
Ignoring (for the moment):
> homebrew/science/getdp:
  * Dependency arpack should not use option with-mpi
  * Dependency hdf5 should not use option with-mpi

No previous Linux bottles; needs slepc first.

PR #5884, opened 2017-06-22
libeemd: revision for gsl

iMichka:


iMichka:
>/usr/lib/gcc/x86_64-linux-gnu/4.8/../../../x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
make: *** [libeemd.1.4.dylib] Error 1

That can maybe be fixed, there are different solutions:
https://stackoverflow.com/questions/11116399/crt1-o-in-function-start-undefined-reference-to-main-in-linux

iMichka:
Rebased. Let's see if it works better with gcc5.

PR #5885, opened 2017-06-22
mathgl: revision for gsl

iMichka:


iMichka:
> homebrew/science/mathgl:
  * C: 45: col 35: Only use parentheses for ternary expressions with complex conditions.
  * C: 46: col 36: Only use parentheses for ternary expressions with complex conditions.

PR #5888, opened 2017-06-22
nest: revision for gsl

iMichka:


iMichka:
> homebrew/science/nest:
>  * C: 139: col 5: Favor modifier `if` usage when having a single-line body. Another good alternative is the usage of control flow `&&`/`||`.
>  * python modules should be vendored rather than use deprecated `depends_on "nose" => :python`
>  * Dependency numpy should not use option with-python3
>  * Dependency scipy should not use option with-python3
> * Dependency matplotlib should not use option with-python3
>  * python modules should be vendored rather than use deprecated `depends_on "nose" => :python3`

PR #5891, opened 2017-06-22
root: revision for gsl

iMichka:


iMichka:
> homebrew/science/root:
>  * W: 28: col 4: `end` at 28, 3 is not aligned with `stable do` at 8, 2.
>  * The URL http://root.cern.ch could use HTTPS rather than HTTP

iMichka:
And fails on Linux with: `configure: libX11 MUST be installed`

iMichka:
I was not able to get the Linux build to run. Even by adding the libx11 dependency and using the ` --with-x11-libdi` flag. Passing on this for the moment.

PR #5892, opened 2017-06-22
root6: revision for gsl

iMichka:


iMichka:
On Linux:
>  brew test-bot took more than 120 minutes to run

Ignoring that as there were no Linux bottles here before

PR #5893, opened 2017-06-22
visp: revision for gsl

iMichka:


iMichka:
>   * C: 27: col 5: Only use parentheses for ternary expressions with complex conditions.
>  * Stable: The URL http://gforge.inria.fr/frs/download.php/latestfile/475/visp-3.0.1.tar.gz could use HTTPS rather than HTTP

PR #5897, opened 2017-06-24
cgns: update comment and improve syntax

iMichka:


iMichka:
@sjackman Just FYI, I updated the comment and improved the syntax.

PR #5911, opened 2017-06-29
gmt: update gmt to 5.4.2 and gshhg to 2.3.7

seisman:


iMichka:
@seisman Thanks for your contribution!

As there were no Linux bottles here before, we can ignore the Linux failure.

PR #5912, opened 2017-06-29
gmt4: update gmt4 to 4.5.16 and gshhg to 2.3.7

seisman:


iMichka:
@seisman Thanks for your contribution!

As there were no Linux bottles here before, we can ignore the Linux failure.

PR #5914, opened 2017-06-29
vtk 8.0.0

iMichka:
Also fixes:
* C: 76: col 36: Only use parentheses for ternary expressions with complex conditions.
* Dependency 'qt5' was renamed; use new name 'qt'.
* Dependency 'qt5' is an alias; use the canonical name 'qt'.


iMichka:
There will be other audit failures that I am ignoring for the time being:
```
* python modules should be vendored rather than use deprecated `depends_on "matplotlib" => :python`
  * Dependency pyqt5 should not use option with-python
  * Dependency pyqt5 should not use option without-python3
  * Dependency sip should not use option with-python3
  * Dependency sip should not use option without-python
```

I will bump the reverse deps in separate commits.

PR #5918, opened 2017-06-29
pcl: revision for vtk

iMichka:


iMichka:
```
homebrew/science/pcl:
  * Dependency 'qt5' was renamed; use new name 'qt'.
  * Dependency 'qt5' is an alias; use the canonical name 'qt'.
  * Dependency 'qt5' was renamed; use new name 'qt'.
  * Dependency 'qt5' is an alias; use the canonical name 'qt'.
  * Dependency pyqt5 should not use option with-python
  * Dependency pyqt5 should not use option without-python3
  * Dependency vtk should not use option with-qt5
  * perl modules should be vendored rather than use deprecated `depends_on "XML::Parser" => :perl`
```

iMichka:
And the Linux build failed due to a too high memory usage.

iMichka:
> * Dependency pyqt5 should not use option with-python
>  * Dependency pyqt5 should not use option without-python3
>  * Dependency vtk should not use option with-qt

Ignoring that for the moment

PR #5929, opened 2017-07-02
vtk: build with qt by default and add needs :cxx11

iMichka:


ilovezfs:
Why qt by default?

iMichka:
Abandoning in favour of https://github.com/Homebrew/homebrew-core/pull/15949, we are migrating vtk to core. We can discuss the qt default there.

PR #5933, opened 2017-07-04
r 3.4.1

erichare:


iMichka:
Hi. Could you fix the audits too?
```
homebrew/science/r:
  * W: 195: col 85: Literal interpolation detected.
  * C: 195: col 85: Prefer double-quoted strings inside interpolations.
  * W: 196: col 118: Literal interpolation detected.
  * C: 196: col 118: Prefer double-quoted strings inside interpolations.
```

erichare:
I've attempted to fix this - I tried to do it in the past as well but was told it was there for specific reason. I removed the interpolations, so then java_home appears in the paths. But that means audit complains that java must be a dependency - i set it as an optional dependency, which I think is accurate given that the RJava functionality does depend on it, but by no means is it necessary.

If you'd prefer this be fixed another way please do let me know.

amoeba:
Is there anything I can do to help get this merged? IIRC these audit issues crop up on every version bump PR so I think they should be addressed in a separate Issue.

iMichka:
> IIRC these audit issues crop up on every version bump PR so I think they should be addressed in a separate Issue.

Yes they do, and it is a good thing. We are trying to keep the code as uniform as possible, and have common rules. Especially as we have hundreds of different contributors. But I agree that they may be addressed in different issues, or can be ignored in certain cases (temporarily). Of course we rely a lot on our contributors to clean up the code when they make a small change.

I'm fine with the solution for java.
For the mac build, you should bump the version of the `littler` formula, which need a new set of bottles. You can do this in a separate commit on top of this one.

erichare:
@iMichka Done. Let me know if there's anything else I can do.

amoeba:
Thanks @iMichka for the helpful response. All of those points make sense to me. And thanks @erichare for the extra commit.

jonchang:
Thanks @erichare, this has been merged. For future reference, our commit style is `littler: revision for r 3.4.1`

erichare:
Sorry @jonchang , i'll keep that in mind about the commit style. Appreciate it. Thanks!

PR #5940, opened 2017-07-06
scram: Fix macOS dependency

### LINKED ISSUE #5726: scram.rb: Missing macOS version (ArgumentError: unknown version nil)

- [x] Ran `brew update` and retried your prior step?
- [x] ~~Ran `brew doctor`, fixed as many issues as possible and retried your prior step?~~
- [x] Checked that a similar issue hasn't already been filed?
- [x] ~~Looked up the software bug tracker if reporting a build error?~~
- [x] ~~If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?~~

### Bug reports / build errors:

During the import for braumeister.org there's an error while importing the scram formula. The problematic code is in [line 16](https://github.com/Homebrew/homebrew-science/blob/0a6b4ba0cfae9edc8cff115de8713dec1028a447/scram.rb#L16) and was introduced in pull request #5694. There's a missing version symbol for the macOS version.

It should look something like this, but I don't know which version is the right one:

```ruby
depends_on :macos => :sierra
```

Additionally, the conditional `unless OS.mac?` seems redundant.

koraktor:
The call to `depends_on` was invalid. Additionally, a redundant `unless OS.mac?` has been removed.
A comment about GCC compatibility has also been removed as there's no obvious connection to macOS.

Closes #5726

@iMichka @rakhimov

### DIFF
--- a/scram.rb
+++ b/scram.rb
@@ -12,8 +12,7 @@ class Scram < Formula
 
   needs :cxx14
 
-  # C++14 uses GCC 5.3, which is not ABI compatible with GCC 4.8
-  depends_on :macos unless OS.mac?
+  depends_on macos: :leopard
 
   depends_on "cmake" => :build
   depends_on "boost"
--- a/scram.rb
+++ b/scram.rb
@@ -12,7 +12,7 @@ class Scram < Formula
 
   needs :cxx14
 
-  depends_on macos: :leopard
+  depends_on :macos => :leopard
 
   depends_on "cmake" => :build
   depends_on "boost"


iMichka:
@BrewTestBot test this please

rakhimov:
The reason for all that fuss code was failures with Linuxbrew:
#5263 
https://github.com/Linuxbrew/brew/issues/366

koraktor:
Looks like `brew audit` failed because of the modern hash syntax. Sorry about that, I currently have no development machine to run `audit` myself.

PR #5944, opened 2017-07-07
r: Depends on cairo for Linux

sjackman:
Change cairo from an optional dependency to a mandatory dependency on Linux.
Cairo is required to render figures to PNG.


sjackman:
The current bottle on Linux is not happy, as it cannot render figures. R 3.3 seemed happy. I believe it could render to PNG without Cairo. That doesn't seem to be the case with R 3.4.
```
❯❯❯ Rscript -e 'capabilities()'
       jpeg         png        tiff       tcltk         X11        aqua 
      FALSE       FALSE       FALSE       FALSE       FALSE       FALSE 
   http/ftp     sockets      libxml        fifo      cledit       iconv 
       TRUE        TRUE        TRUE        TRUE       FALSE        TRUE 
        NLS     profmem       cairo         ICU long.double     libcurl 
       TRUE        TRUE       FALSE       FALSE        TRUE        TRUE 
```

jonchang:
I seem to recall some change to how R builds in 3.4.0, but I can't find it in the official release notes. Regardless, my package manager has cairo as a dependency for R so :+1: assuming tests pass.

sjackman:
With R 3.4 I also had to add to `~/.Rprofile`
```r
options(bitmapType = "cairo")
```
or I saw the error message
```
Error in .External2(C_X11, paste0("png::", filename), g$width, g$height,  : 
  unable to start device PNG
Calls: <Anonymous> ... block_exec -> chunk_device -> dev_new -> do.call -> <Anonymous>
In addition: Warning messages:
1: In grDevices::png(f) : unable to open connection to X11 display ''
2: In (function (filename = "Rplot%03d.png", width = 480, height = 480,  :
  unable to open connection to X11 display ''
```

sjackman:
```
homebrew/science/r:
  * Dependency cairo should not use option with-x11
```
This audit warning is pre-existing and specific to macOS. On linux, `--with-x11` is the default behaviour for `cairo`. I feel like ignoring it is probably the best path forward. Enabling `--with-x11` by default for `cairo` on macOS doesn't seem great, as I believe that adds a user dependency on `XQuartz` (I could be wrong about that).

sjackman:
```
1.00s==> brew linkage --test homebrew/science/r
System libraries:
  /lib/x86_64-linux-gnu/librt.so.1
Missing libraries:
  libR.so
  libRblas.so
  libRlapack.so
No unwanted system libraries
==> FAILED
```
https://travis-ci.org/Homebrew/homebrew-science/builds/250986040#L6448
This error is a false positive, because the R shells script sets `LD_LIBRARY_PATH` to find these libraries.

sjackman:
```
==> brew test --verbose littler
Testing homebrew/science/littler
==> /home/linuxbrew/.linuxbrew/Cellar/littler/0.2.3_12/bin/littler -e 'print(pi)'
.
Last 150 lines from /home/linuxbrew/.cache/Homebrew/Logs/littler/test.01.littler:
2017-07-07 00:51:08 +0000
/home/linuxbrew/.linuxbrew/Cellar/littler/0.2.3_12/bin/littler
-e
'print(pi)'
/home/linuxbrew/.linuxbrew/Cellar/littler/0.2.3_12/bin/littler: error while loading shared libraries: libR.so: cannot open shared object file: No such file or directory
```
https://travis-ci.org/Homebrew/homebrew-science/builds/250986040#L6512

That's weird. `brew test r` passed but `brew test littler` failed. I wonder if the `littler` bottle needs to be rebuilt.

jonchang:
`littler` always needs a revision bump when `r` gets bumped.

sjackman:
Even if it's just a `revision` bump? Can we fix that in the `littler` formula by using `opt_lib` rather than `lib`?

sjackman:
CircleCI successfully built a bottle for `r`.

jonchang:
> Even if it's just a `revision` bump? Can we fix that in the `littler` formula by using `opt_lib` rather than `lib`?

To be fair I don't really understand why `littler` needs a rev bump. From what I can tell in the formula it *is* linking against the `opt` prefix but there's still some sort of linking issue that causes old versions to fail. As you can see `littler` is on revision 12 so it's been happening for a while now

sjackman:
Thanks for the explanation, Jon. I've bumped the revision for `littler`.

sjackman:
I'm going to ignore the `brew audit` and `brew linkage` failures and merge this PR (after CI passes).

iMichka:
> To be fair I don't really understand why littler needs a rev bump. From what I can tell in the formula it is linking against the opt prefix but there's still some sort of linking issue that causes old versions to fail. As you can see littler is on revision 12 so it's been happening for a while now

+1. I've been asking people to bump these revisions each time, because without it old versions fail. But I did that more out of reflex and never really investigated the root cause.

sjackman:
I'd love to look into the root cause when I find some time. Not this month though.

PR #5947, opened 2017-07-08
Update sigrok formulas

denisvm:
### Have you:

### New formula: have you

### DIFF
--- a/libsigrokdecode.rb
+++ b/libsigrokdecode.rb
@@ -1,10 +1,10 @@
 class Libsigrokdecode < Formula
   desc "Library for (streaming) protocol decoding"
-  homepage "http://sigrok.org/"
+  homepage "https://sigrok.org/"
 
   stable do
-    url "http://sigrok.org/download/source/libsigrokdecode/libsigrokdecode-0.4.1.tar.gz"
-    sha256 "065f70c10971173d86e03a8cf5534e600c7a622775113997d8061572135d4a95"
+    url "https://sigrok.org/download/source/libsigrokdecode/libsigrokdecode-0.5.0.tar.gz"
+    sha256 "4aa8579ecea9b421b8ac048a9b18c27e63206839f269374398d89c14a47bd1c1"
 
     resource "librevisa" do
       url "http://www.librevisa.org/git/librevisa.git", :tag => "alpha-2013-08-12",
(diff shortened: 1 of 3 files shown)


pulseview.rb
```diff
@@ -19,10 +20,11 @@ class Pulseview < Formula
 
   depends_on "pkg-config" => :build
   depends_on "cmake" => :build
+  depends_on :x11 => :build
```
iMichka: Is this for Linux only, or for mac too? If it is for Linux only, please add `unless OS.mac?` at the end of the line.

libsigrokdecode.rb
```diff
@@ -70,6 +80,13 @@ class Libsigrokdecode < Formula
     depends_on "automake" => :build
   end
 
+  if build.with?("fx2lafw")
+    depends_on "libtool" => :build
```
iMichka: Are all these dependencies for mac, or are some Linux only? Please add unless OS.mac? at the end of the line if some of these are for Linux only.

Also; these are also added when building with head. Is that right? Could these be regrouped?

denisvm: In README it says that lib tool is required when building 'from git', but when I run `autogen.sh` it disables libtool, so it looks like it's not really required, at least on Mac. I've added your suggestion in all formulas.

Also grouped the common dependencies for HEAD in head section.

pulseview.rb
denisvm: The original formula referenced it through the lines below, but it doesn't really seem to need it. So I'm removing it for now.
```c
-DPNG_INCLUDE_DIR=#{MacOS::X11.include}```

iMichka: Yes, if we can drop x11 stuff, this would be great.

denisvm:
I tried to add a test to libsigrokdecode, but I couldn't find a way to include reference to other libraries in the system. I've commented the test block for now.

```fatal error: 'glib.h' file not found```

iMichka:
You can use something like `Formula["glib"].opt_lib` and `Formula["glib"].opt_include` to pass the paths to the compiler. In the `test` block we do not set the paths, so it needs to be done manually.

denisvm:
Great, thanks for the info. I've just added the library tests :)

denisvm:
Is there anything missing to merge this or should we wait for Qt Linux bottles?

iMichka:
@denisvm Thanks for your contribution, and for the formula cleanup!

PR #5958, opened 2017-07-11
freebayes 1.1.0_1

tseemann:

### Reason for revision

Patches a serious bug that upstream isn't responding to: https://github.com/ekg/freebayes/issues/376


freebayes.rb
```diff
@@ -18,15 +21,21 @@ class Freebayes < Formula
...
+  depends_on "zlib" unless OS.mac?
 
   def install
     ENV.deparallelize
 
     # Build fix: https://github.com/chapmanb/homebrew-cbl/issues/14
     inreplace "vcflib/smithwaterman/Makefile", "-Wl,-s", "" if OS.mac?
 
+    # https://github.com/ekg/freebayes/issues/376
+    fb = "scripts/freebayes-parallel"
```
ilovezfs:
```
    inreplace "scripts/freebayes-parallel" do |s|
      s.gsub! "../vcflib/scripts/vcffirstheader", "vcffirstheader"
      s.gsub! "../vcflib/bin/vcfstreamsort", "vcfstreamsort"
    end

File: freebayes.rb [hunk 2]
```diff
@@ -39,7 +48,7 @@ def install
   end
 
   test do
-    system "#{bin}/freebayes", "--version"
-    system "#{bin}/freebayes-parallel"
+    assert_match "polymorphism", shell_output("#{bin}/freebayes --help 2>&1", 0)
```
ilovezfs: 0 is already the default

freebayes.rb
```diff
@@ -39,7 +48,7 @@ def install
...
+    assert_match "chunks", shell_output("#{bin}/freebayes-parallel 2>&1", 0)
```
ilovezfs: ditto

freebayes.rb
```diff
@@ -18,15 +21,21 @@ class Freebayes < Formula
...
+  depends_on "vcflib"
+  depends_on "zlib" unless OS.mac?
 
   def install
     ENV.deparallelize
 
     # Build fix: https://github.com/chapmanb/homebrew-cbl/issues/14
     inreplace "vcflib/smithwaterman/Makefile", "-Wl,-s", "" if OS.mac?
 
+    # https://github.com/ekg/freebayes/issues/376
```
ilovezfs: Also would be good to note what this fixes and the date of the report. For examples, see https://gist.github.com/ilovezfs/62a9eb715dc044f7202cbf86caaf713b

freebayes.rb
tseemann: I like to explicitly list the exit code, but if the correct style is to omit that parameter then I will follow that style.

ilovezfs: Yep. It should be left out unless it's non-zero, or unless you have to use pipe_output, which doesn't assert its exit code by default.

freebayes.rb
tseemann: i would argue my code is more readable for n=2 patches, but it does write the file twice.  Does your method only write the file once?

freebayes.rb
```diff
@@ -18,15 +21,21 @@ class Freebayes < Formula
...
   depends_on "cmake" => :build
 
-  depends_on "parallel" => :recommended
-  depends_on "vcflib" => :recommended
+  depends_on "parallel"
+  depends_on "vcflib"
+  depends_on "zlib" unless OS.mac?
 
   def install
     ENV.deparallelize
```
ilovezfs: One other thing. Have you checked if this is still necessary? If it is, it should be reported at some point.

freebayes.rb
ilovezfs: Correct. Any time you inreplace more than once on the same file, you should use the block form.

freebayes.rb
tseemann: The github issue in the comment explains the problem?

freebayes.rb
tseemann: vcflib and parallel were originally 'recommended' as not everyone used freebayes-parallel script.  but it's much more popular now, and not having the deps means the script will fail.

ilovezfs:
```
ENV.deparallelize
```
means to pass `-j1` to make, which we consider a bug. It's unrelated to the `parallel` dependency.

tseemann: Oh, sorry YES - the `-j 1` is needed. The makefile is broken wrt multiple threading. I tested it again today.  I'll add a explanation.

ilovezfs:
Shipped!

PR #6009, opened 2017-07-25
ceres-solver.rb update

ahundt:
Eigen changed, so ceres-solver bottle needs to be remade to maintain binary compatibility.

### Have you:

### New formula: have you

### DIFF
--- a/ceres-solver.rb
+++ b/ceres-solver.rb
@@ -3,7 +3,7 @@ class CeresSolver < Formula
   homepage "http://ceres-solver.org/"
   url "http://ceres-solver.org/ceres-solver-1.12.0.tar.gz"
   sha256 "745bfed55111e086954126b748eb9efe20e30be5b825c6dec3c525cf20afc895"
-  revision 4
+  revision 5
 
   bottle do
     cellar :any


jonchang:
Thanks for catching this @ahundt! This has been merged.

ahundt:
is there a way this check / change can be automated? unfortunately this bottle bug happens with every eigen release.

jonchang:
Updates to `eigen` will also test its reverse dependencies by installing the `ceres-solver` bottle and running `brew test ceres-solver`. If the test doesn't raise an error we don't bump its `revision` number. I haven't been involved in any `eigen` updates so I don't recall if `brew test ceres-solver` ever errored out on an update.

ahundt:
I'm pretty sure the test should fail *for the bottle*, but would succeed if recompiled. There is a CMake check to ensure the eigen version matches, and that cmake check is run in the test, so the test should fail on a bottle after the eigen version is updated.

ilovezfs:
@jonchang eigen is in core so the ceres-solver won't be tested since it's in science. This happens every single upgrade because ceres-solver ridiculously requires the exact version of eigen regardless of whether the library upgrade is actually ABI changing or not. I think we should move ceres-solver to core too, as I mentioned here:https://github.com/Homebrew/homebrew-core/issues/15619#issuecomment-317688070

jonchang:
@ilovezfs thanks, I didn't realize eigen got migrated. I'm supportive of that idea since it sounds like it'll make eigen/ceres-solver updates less painful.

ilovezfs:
>thanks, I didn't realize eigen got migrated.

It's always been in core.

PR #6022, opened 2017-07-29
ENH : enable openmp for openmeeg

agramfort:
### Have you:

### New formula: have you

There is already a test but it's not specific to openmp

cc @iMichka

### DIFF
--- a/openmeeg.rb
+++ b/openmeeg.rb
@@ -4,6 +4,7 @@ class Openmeeg < Formula
   url "https://github.com/openmeeg/openmeeg/archive/2.4.prerelease.tar.gz"
   version "2.4.prerelease"
   sha256 "fabf9addfcb99d3b3161acde98104a321ec926a1968e444edfce5334adb1ae87"
+  revision 1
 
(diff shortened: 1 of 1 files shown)


agramfort:
@iMichka would it make sense to have openmp on by default or this is not recommended?

iMichka:
It depends. If it is a feature that everybody may want to use, it should not be an option. Because this means people will need to recompile from source instead of using the pre-compiled binaries. Options are also often untested by the CI so compilation bugs are often hidden away from is when we bump version.

On the other side this "option" adds a dependency more, which may or may not be a good thing.

You could make it "recommended" if you think it is a sensible default. I am not an openmeeg user so I have no strong opinion here.

agramfort:
in practice it gives huge speed ups as internal code is massively parallel.
With 4 real independent CPUs you get almost a 4x speed up. I would
therefore to make it default. It also reduce the chances of bugs with the
use of options.

let me try to turn it on by default.

agramfort:
done

iMichka:
@agramfort Thanks for your contribution!

PR #6053, opened 2017-08-02
gnudatalanguage: test build

iMichka:


iMichka:
Testing with a netcdf bump too. Building `gnudatalanguage` works on the bot, but not locally for me.

The netcdf bottles were built with gcc 4.8 on Linux a few months ago. Trying with gcc 5.3.0 now, to keep everything consistent.

iMichka:
@sjackman I need your help here. We are trying to make the `netcdf` build with the fortran option on by default. Most of other formulae needing `netcdf` are fine with that change, but not `gnudatalanguage`. As soon as I depend on fortran in netcdf, I do get linking errors in `gnudatalanguage`. Looks like some API mismatch due to 2 different gcc versions.

I guess it is a similar problem than the one encountered with `ceres-solver`? You can have a look at the log files of both builds:
- netcdf without fortran
https://circleci.com/api/v1.1/project/github/Homebrew/homebrew-science/5739/output/29/0?file=true
- netcdf with fortran
https://circleci.com/api/v1.1/project/github/Homebrew/homebrew-science/5749/output/29/0?file=true

sjackman:
Likely the same issue as https://github.com/Homebrew/homebrew-science/pull/6055#issuecomment-320051664

You can try the same hack workaround. We need a better fix ultimately, and soon by the look of the number of formulae that it affects.
https://github.com/Homebrew/homebrew-science/commit/db949c3ef88db85951b8f8347463b69a01ea2faa

iMichka:
Why is not everything built with gcc 5.3? That would solve it I guess.

sjackman:
By design we use the same `glibc` and `gcc` version as provided by Ubuntu 14. If we use the brewed `gcc` to build bottles, then we're not able to use CircleCI to build bottles for `gcc` or any of its dependencies.

iMichka:
Closing here as this was just a test. Thanks @sjackman for all the explanations.

PR #6059, opened 2017-08-04
prokka: Fix non-executable file in binaries/linux/

tmozgach:
Use the brewed libidn, zlib and bzip2 rather than the host's.

### Have you:

### New formula: have you

### DIFF
--- a/prokka.rb
+++ b/prokka.rb
@@ -3,6 +3,7 @@ class Prokka < Formula
   homepage "http://www.vicbioinformatics.com/software.prokka.shtml"
   # doi "10.1093/bioinformatics/btu153"
   # tag "bioinformatics"
+  revision 1
 
   url "https://github.com/tseemann/prokka/archive/v1.12.tar.gz"
   sha256 "845e46c9db167fb1fc9d16c934b7cfd89ddfeb6cd44bd8f42204ae7b4e87adba"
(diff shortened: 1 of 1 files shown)


tmozgach:
Use the brewed libidn, zlib and bzip2 rather than the host's.

tmozgach:
@sjackman could merge? There are errors:  
``` perl modules should be vendored rather than use deprecated ```

sjackman:
Merged. Thanks, Tanya!

tseemann:
@tmozgach the `binaries` folder is only meant for people who just `git clone` and need it to run. if installing from Brew, it will use the Brew binaries. The bundled binaries (which are mostly static) are only a "last resort". 

How do they lose these executable bit?  Is that done by the brew installer?

@sjackman is there any information online about 'vendoring' perl modules?  i wish we could use brew perl and `cpanm` to do all the deps!

sjackman:
The problem wasn't that the executable bit was lost. It was that the `RPATH` of the bundled executables was set to use the libraries of the host rather than libraries provided by Linuxbrew, so flagged some errors when `brew linkage prokka` is run. Alternatively, we could remove the bundled executables.

sjackman:
> is there any information online about 'vendoring' perl modules? i wish we could use brew perl and cpanm to do all the deps!

Not that I know of. I think you'll have to crib off of other formula. I'd like to convert all the bioinformatics formula to vendored Perl and Python modules as well so that they can be bottled. Using `depends-on "foo" => :perl` or `:python` prevents bottling, because these dependencies are not satisfied when CI is run.

tseemann:
I would argue that the bundled binaries should be removed by the brew installer completely.

PR #6063, opened 2017-08-04
ceres-solver: build a bottle for Linuxbrew

ilovezfs:


ilovezfs:
@sjackman circle doesn't seem to think that is valid syntax

sjackman:
The command is changing from a string in an array to a string that's the key of a hash. Add line 74 as well:
https://github.com/Linuxbrew/homebrew-core/blob/master/circle.yml#L74-L82
And compare the formatting of the two commands. It may have to change slightly.
You can use a syntax validator like `safe_yaml -f circle.yml` to check that the YAML syntax is correct.

ilovezfs:
@sjackman 

>==> make test
>..
command umask 022; PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew>/sbin:$PATH"; eval "$(perl -I$HOME/perl5/lib/perl5 -Mlocal::lib)"; brew install patchelf && brew tap homebrew/dupes && brew tap linuxbrew/xorg && brew test-bot && brew test-bot >--keep-old took more than 120 minutes to run

sjackman:
Looks like `make test` is hung up. If you have the time and inclination, you could use the CircleCI `Rebuild with SSH` feature to troubleshoot it interactively.

jonchang:
On my local machine `make test` is hung up on the `bundle_adjustment_test`. Poking at it with strace shows that it's spinlocking somewhere. Disabling the test can be done with

```
make test ARGS="-E bundle_adjustment_test"
```

dpo:
Seems worth reporting.

ilovezfs:
appears we're not the only ones with problems with that test https://github.com/ceres-solver/ceres-solver/issues/305#issuecomment-319274693

ilovezfs:
Closing in light of https://github.com/Homebrew/homebrew-science/pull/6073 and https://github.com/Homebrew/homebrew-core/pull/16465.

PR #6066, opened 2017-08-04
pymol: Patched in support for external gui in aqua

### LINKED ISSUE #5505: Pymol crashes on startup
#### Brief summary of issue:
Pymol crashes on stratup main window visible for lesses than a sec
terminal output:

`$pymol`:
               
    PyMOL(TM) Molecular Graphics System, Version 1.8.4.0.
    Copyright (c) Schrodinger, LLC.
    All Rights Reserved.
 
    Created by Warren L. DeLano, Ph.D. 
 
    PyMOL is user-supported open-source software.  Although some versions
    are freely available, PyMOL is not in the public domain.
 
    If PyMOL is helpful in your work or study, then please volunteer 
    support for our ongoing efforts to create open and affordable scientific
    software by purchasing a PyMOL Maintenance and/or Support subscription.

    More information can be found at "http://www.pymol.org".
 
    Enter "help" for a list of commands.
    Enter "help <command-name>" for information on a specific command.

    Hit ESC anytime to toggle between text and graphics.
    
    Detected OpenGL version 2.0 or greater. Shaders available.
    Detected GLSL version 1.20.
    OpenGL graphics engine:
    GL_VENDOR:   Intel Inc.
    GL_RENDERER: Intel(R) Iris(TM) Graphics 6100
    GL_VERSION:  2.1 INTEL-10.18.65
    /usr/local/bin/pymol: line 4: 11399 Abort trap: 6           "/usr/local/opt/python/bin/python2.7" "/usr/local/Cellar/pymol/1.8.4.0/libexec/lib/python2.7/site-packages/pymol/__init__.py" "$@"

When installing homebrew complains that pymol wants dupes/tcl-tk and npt tcl-tk

    Warning: Use tcl-tk instead of deprecated homebrew/dupes/tcl-tk

I have tried to install the regular tcl-tk and the installing pymol with --ignore-dependencies (making sure that i had installed the deps. manually) 

#### The complete command that you issued:
`pymol`
#### Gist with output of `brew doctor`:
`brew doctor`:

    Warning: Use tcl-tk instead of deprecated homebrew/dupes/tcl-tk
    Warning: Use tcl-tk instead of deprecated homebrew/dupes/tcl-tk
    Your system is ready to brew.

#### Link output by `brew gist-logs <formula>`:
brew gist-logs pymol:

Error: No logs.

#### Output of `brew config`:
`brew config`:

    HOMEBREW_VERSION: 1.1.13
    ORIGIN: https://github.com/Homebrew/brew
    HEAD: 46f6dc4c93b96abd528c3a0a808a5ca5a795c7f3
    Last commit: 11 days ago
    Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
    Core tap HEAD: d89bd024e3a14f4ce34a60825e3afc72913b4a51
    Core tap last commit: 35 minutes ago
    HOMEBREW_PREFIX: /usr/local
    HOMEBREW_REPOSITORY: /usr/local/Homebrew
    HOMEBREW_CELLAR: /usr/local/Cellar
    HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
    CPU: quad-core 64-bit broadwell
    Homebrew Ruby: 2.0.0-p648
    Clang: 8.1 build 802
    Git: 2.11.0 => /Library/Developer/CommandLineTools/usr/bin/git
    Perl: /usr/bin/perl
    Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/bin/python2.7
    Ruby: /usr/bin/ruby => /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby
    Java: N/A
    macOS: 10.12-x86_64
    Xcode: N/A
    CLT: 8.3.2.0.1.1492020469
    X11: 2.7.11 => /opt/X11

telamonian:
Also removed brewed `tcl-tk` dependency. Fixes #5505. See the [1.8.6.0_-_osx_gui_fix branch of telamonian/pymol](https://github.com/telamonian/pymol/tree/1.8.6.0_-_osx_gui_fix) for the complete code.

### Have you:

## The external GUI is fixed!

Which should also fix #5505! Hooray!

The big change is I've added a patch that makes the external GUI open and behave correctly in a native OS X window (ie Aqua). So everything should work great now, and there should hopefully be much fewer problems with Mac builds of Pymol going forward. Yaaaay!

Basically, it turns out that for Tkinter built against tcltk that in turn has been built against Aqua (which is the version of tcltk that ships with OS X, and is now also the only version of tcltk that homebrew supports), the root GUI window **has** to be run in the main thread. Any attempt to work around this limitation inevitably result in an exception raised in one of the low-level Core Services libraries (ie the Apple system libraries). Pymol, on the other hand, is designed to run the external GUI in its own thread, which then obviously causes problems when that thread tries to create a root Tkinter Aqua window.

I managed to fix all of that mess by adding some that tests if A) an external GUI has been requested in the launch options, B) Pymol is running on OS X, and C) the windowing system is Aqua. If and only if all three of these things are True, the Pymol execution path is altered so that the external GUI does in fact run on the main thread, and the glut GUI (ie the window with the molecular model display, which is the part of the GUI that runs on the main thread under normal conditions) runs on a child thread.

After a bunch of trial and error I've refined the Aqua execution path to the point where there's no noticeable differences in functionality or performance. There's still a couple of minor graphical glitches (eg some flickering help text), but I've fixed enough of them so they're at the point where they don't affect usability. I'll try to fix the rest in the next revision.

I'll be pushing the patch upstream, or at least attempting to, but I figure that might take a while, so in the meantime I just added the patch to a formula am sharing it with the homebrew community, as many others have done in the past for me.

## No more brewed `tcl-tk` dependency

As a consequence of getting Pymol to play nicely with Aqua Tkinter, the Pymol formula no longer has a dependency on a homebrewed version of `tcl-tk` or on a version of `python` built against a brewed `tcl-tk`. I've removed all of the related dependencies. Overall this should make Pymol much easier to install. All you need for a complete installation should now just be:
```
brew install pymol --build-from-source
```
Once the bottles have been refreshed, you won't even need `--build-from-source`.

If for whatever reason you do want to build against a brewed `tcl-tk` you can still do so simply by following the old installation instructions.

## Minor changes
- Bumped version from 1.8.4.0 => 1.8.6.0
- The installation function now explicitly calls `python2` instead of `python`. This fixed a bug that was related to all of the `python` formula path zaniness that's been going over at homebrew-core recently. Basically, without this fix Pymol may build against the system python even if a brewed python is present.

### DIFF
--- a/pymol.rb
+++ b/pymol.rb
@@ -1,8 +1,8 @@
 class Pymol < Formula
   desc "OpenGL based molecular visualization system"
-  homepage "http://pymol.org"
-  url "https://downloads.sourceforge.net/project/pymol/pymol/1.8/pymol-v1.8.4.0.tar.bz2"
-  sha256 "b6147befe74844dd23550461b831b2fa6d170d4456f0059cf93fb1e8cb43d279"
+  homepage "https://pymol.org/"
+  url "https://downloads.sourceforge.net/project/pymol/pymol/1.8/pymol-v1.8.6.0.tar.bz2"
+  sha256 "7eaaf90ac1e1be0969291cdb1154b3631b5b6403cce1d697133d90cd37a3c565"
   head "https://svn.code.sf.net/p/pymol/code/trunk/pymol"
 
   bottle do
@@ -13,17 +13,51 @@ class Pymol < Formula
   end
 
   depends_on "glew"
-  depends_on "python" => "with-tcl-tk"
-  depends_on "homebrew/dupes/tcl-tk" => ["with-threads", "with-x11"]
+  depends_on "msgpack"
   depends_on :x11
 
-  def install
-    ENV.append_to_cflags "-Qunused-arguments" if MacOS.version < :mavericks
+  if OS.mac?
+    depends_on :python
+  else
+    depends_on "freetype"
+    depends_on "gpatch" # see homebrew/homebrew-science#5102
+    depends_on "tcl-tk"
+    depends_on "libxml2"
+    depends_on "python" => "with-tcl-tk"
+  end
 
-    system "python", "-s", "setup.py", "install",
-                     "--bundled-pmw",
-                     "--install-scripts=#{libexec}/bin",
-                     "--install-lib=#{libexec}/lib/python2.7/site-packages"
+  needs :cxx11
+
+  # Patch that makes the OS X native windowing system (Aqua) and PyMol play nicely together.
+  # Fixes https://sourceforge.net/p/pymol/bugs/187/ (05.09.17) and
+  # https://github.com/Homebrew/homebrew-science/issues/5505 (04.27.17), in which bad GUI calls were causing segfaults.
+  patch do
+    url "https://gist.githubusercontent.com/telamonian/494150cf2992d324272de9a0ded78e37/raw/9abce6d6f97b3943856b63f83ad5b7352c339926/osx_aqua_ext_gui_fix.diff"
+    sha256 "14cef03169290500a8232c9ab78964fa169bcd0854a01ced95ccbe55bd70e2c6"
+  end
+
+  def install
+    args = %W[
+      --bundled-pmw
+      --install-scripts=#{libexec}/bin
+      --install-lib=#{libexec}/lib/python2.7/site-packages
+    ]
+
+    if OS.mac?
+      # clang emits >1e5 lines of nullability warnings for pymol, turn them off
+      ENV.append_to_cflags "-Wno-nullability-completeness"
+
+      # support for older Mac OS
+      ENV.append_to_cflags "-Qunused-arguments" if MacOS.version < :mavericks
+
+      system "python", "-s", "setup.py", "install", *args
+    else
+      # on linux, add the path hint that setup.py needs in order to find the freetype and libxml2 headers
+      ENV.prepend_path "PREFIX_PATH", ENV["HOMEBREW_PREFIX"]
+
+      # because the linux python dep is specified with "python" instead of :python, python2 is needed here
+      system "python2", "-s", "setup.py", "install", *args
+    end
 
     bin.install libexec/"bin/pymol"
   end


telamonian:
Ugghh, although apparently Pymol does now have an explicit dependence on `msgpack`. I'll add that in in a sec.

telamonian:
...and now hilariously, a problem with the `msgpack` formula (see msgpack/msgpack-c#623) is causing the linux build to fail. Wooo

ilovezfs:
@sjackman I'd suggest syncing that change up ASAP to avoid a bunch of reports.

iMichka:
Merging in progress: https://github.com/Linuxbrew/homebrew-core/pull/3607

iMichka:
`msgpack` should be OK now, I restarted the Linux builds.

telamonian:
The Pymol Linux build is failing, and it's a bit of mystery. What's happening is that when Pymol is attempting to compile its C extension, some of the environment variable have been set as:

```
HOMEBREW_CC: gcc-5
HOMEBREW_CXX: g++-5
CFLAGS: -Qunused-arguments
CXXFLAGS: -Qunused-arguments
```

So the problem is that Python is trying to use gcc to compile the Pymol C extension, but the ` -Qunused-arguments` flag is only supported by Clang. The passing of an unknown flag causes gcc to hard crash.

There's no mention of `-Qunused-arguments` in the Pymol source or build files, and that flag is not emitted anywhere during Pymol build on OS X. It seems like somehow that flag is leaking into the environment via something that linuxbrew is doing during the setup of the build-bot. Except that every mention of `-Qunused-arguments` was apparently removed from linuxbrew years ago, and even those were in shim filters to remove the flags.

Have of you seen anything similar before? The only other hint I have to go on is that Pymol's `setup.py` is passing `-Wno-unused-function`, but I can't find anyplace it might be getting transformed into `-Qunused-arguments` .

telamonian:
Also, is it possible to interact with/breakpoint the build on travis-ci? I assume that it's not, but it would be a huge help for debugging the linux environment, especially since I don't have a linux box at hand.

sjackman:
Apparently TravisCI does have a debug in ssh mode, though it may not be enabled for our repo: https://docs.travis-ci.com/user/running-build-in-debug-mode/

CircleCI definitely has a debug in ssh mode, `Rebuild with SSH`

Linuxbrew has a Docker image `linuxbrew/linuxbrew` that you can use to reproduce this build failure. There's Docker images based on Debian, Ubuntu, and Travis: https://github.com/linuxbrew/docker

pymol.rb
```diff
@@ -13,14 +13,18 @@ class Pymol < Formula
...
+  depends_on "msgpack"
+  depends_on "python"
   depends_on :x11
 
+  # This patch enables the external GUI (including support for plugins) when running
+  # pymol using Aqua, the OS X native graphics/windowing system.
+  patch :DATA
+
   def install
     ENV.append_to_cflags "-Qunused-arguments" if MacOS.version < :mavericks
```
sjackman: Change this line to…
```ruby
ENV.append_to_cflags "-Qunused-arguments" if OS.mac? && MacOS.version < :mavericks
```
`MacOS.version` is `0` on Linux. `nil` would probably be better.

telamonian: Doh! Should have checked there first

telamonian: Hah, good catch. Definitely should have checked there first.

pymol.rb
```diff
@@ -12,22 +12,40 @@ class Pymol < Formula
...
+    args << "--no-cxx11" if build.without? "c++11"
+
+    # support for older Mac OS
     ENV.append_to_cflags "-Qunused-arguments" if OS.mac? && MacOS.version < :mavericks
 
-    system "python2", "-s", "setup.py", "install",
-                     "--bundled-pmw",
-                     "--install-scripts=#{libexec}/bin",
-                     "--install-lib=#{libexec}/lib/python2.7/site-packages"
+    system "python2", "-s", "setup.py", "install", *args
```
sjackman: My understanding is that using `python2` explicitly shouldn't be necessary. If you `depends_on :python` then `python` should refer to the brewed `python`.

sjackman: No harm either though.

telamonian: ¯\_(ツ)_/¯ I know that's where you guys want to get the new python pathing stuff to, but my experience has been that the `pymol` formula inevitably ends up building against the system python if `python2` is not explicitly called.

I do have `/usr/local/opt/python/libexec/bin` at the front of my path, so I figure it's down to either superenv doing something weird, or to the fact that `pymol` uses an ancient `distutils` only build that probably doing a lot of weird stuff.

In any case, I'll test it again right now.

telamonian: I changed `pymol2` &rarr; `python` and rebuilt. During its build Pymol puts together an `sh` script that it uses for an executable, and here's the line that script uses to launch python:

```
"/usr/bin/python" "/usr/local/Cellar/pymol/1.8.6.0/libexec/lib/python2.7/site-packages/pymol/__init__.py" "$@"
```

And of course `/usr/bin/python` is the Apple system python, so the problem is still there. Should I submit a issue about that?

sjackman: I'd say not. `python2` is fine by me. Thanks for checking.

telamonian:
The travis-ci build-bot died a while back. Something to with the `tinyscheme` formula being invalid. 

In any case, in the interest of this not taking forever I'm trying to move over to that travis Dockerfile sjackman pointed out to me earlier. Only, the thing is, I've never used docker before. I got `docker` and `docker-machine` up and running, and then I tried running `docker build .` in the same directory as the travis Dockerfile. It definitely did something, and downloaded a few gigs of stuff, but I don't think it did what I needed it too.

Any tips for configuring/executing that Dockerfile so that it actually runs the same tests (hopefully in a more interactive way) as the travis-ci server?

sjackman:
> The travis-ci build-bot died a while back. Something to do with the tinyscheme formula being invalid.

Sorry about that. That issue is resolved now.

sjackman:
> Any tips for configuring/executing that Dockerfile so that it actually runs the same tests (hopefully in a more interactive way) as the travis-ci server?

I'd start with the `linuxbrew/linuxbrew` image.
```sh
docker pull linuxbrew/linuxbrew
docker run -it --name=linuxbrew linuxbrew/linuxbrew
```
If that image isn't sufficient to reproduce your issue, you can use the Travis image like so:
```
cd travis
docker build -t linuxbrew:travis .
docker run -it --name=linuxbrew_travis linuxbrew:travis
```

ilovezfs:
Do not use "python2" unless you're using `depends_on "python"` instead of `depends_on :python`.

iMichka:
Just one note: The standard commit message for this PR should be: `pymol 1.8.6.0`
You can add whatever comments below

telamonian:
> Do not use "python2" unless you're using depends_on "python" instead of depends_on :python

Yep, that fixes the issue. I've removed `python2` from the formula.

> I'd start with the linuxbrew/linuxbrew image.

Okay, so I got the linuxbrew Docker container up and running, and I was able to repro the issue I was having on the travis-ci build-bot, but linuxbrew on the VM is behaving a bit strangely. Whenever I do a `brew install`, linuxbrew hangs for about 60 seconds before starting. During the hang, the debug output has a whole bunch of lines similar to:
```
/home/linuxbrew/.linuxbrew/Library/Homebrew/brew.rb (Formulary::FormulaLoader): loading /home/linuxbrew/.linuxbrew/Library/Taps/homebrew/homebrew-core/Formula/glew.rb
```
Most of these formula loadings take almost no time, but two of them, for `font-util.rb` and `patchelf.rb`, each take about 30 seconds, causing the hang. Is this expected behavior for linuxbrew, or did I screw something up in the build environment when I was setting it up?

iMichka:
> Most of these formula loadings take almost no time, but two of them, for font-util.rb and patchelf.rb, each take about 30 seconds, causing the hang.

@sjackman may know more about that

sjackman:
The very first `brew` command that you run updates `brew` and taps `homebrew/core`, so that'll take a bit of time, about 30 seconds or so. After that should be fast.

`/home/linuxbrew/.linuxbrew/Library/Homebrew/brew.rb (Formulary::FormulaLoader): loading /home/linuxbrew/.linuxbrew/Library/Taps/homebrew/homebrew-core/Formula/glew.rb`

That looks like you have debug mode enabled with either `-d` or `--debug`.

telamonian:
As of `dd4f09b` everything is building/working except for the linux runtime. The linux runtime fails because the default linuxbrew `python` install doesn't install a functioning version of `Tkinter`. 

I've added some extra linux dependencies in the last couple of commits that I think will fix the problem. I can't test it very easily though, since I haven't had any luck getting any flavor of `OpenGL` to work on a virtual machine. I tried with the linuxbrew Dockerfile, and I also tried setting up my own vagrant vm based on that Dockerfile, but no luck. 

It seems though that your build-bot for circleci does have `OpenGL` correctly configured. @sjackman do you have a script or something with the magic setup?

telamonian:
> Do not use "python2" unless you're using `depends_on "python"` instead of `depends_on :python`

@ilovezfs This works for specifying a dependency with an option:
```
depends_on "python" => "with-tcl-tk"
```
but this did nothing when I tried it:
```
depends_on :python => "with-tcl-tk"
```

I think I've seen you mention elsewhere that you shouldn't specify any options on dependencies. To boot, that line also fails the formula audit. Are dependencies with options deprecated? Is the issue discussed anywhere in the brew docs?

I can't think of any other way to get a version of `python` with a working `Tkinter` on linux. Is there an alternative to options with dependencies?

ilovezfs:
>Are dependencies with options deprecated?

Yes.

>Is the issue discussed anywhere in the brew docs?

I believe only here:
https://github.com/Homebrew/brew/pull/2482
https://github.com/Homebrew/homebrew-core/issues/13133
https://github.com/Homebrew/homebrew-science/issues/5470

>depends_on :python => "with-tcl-tk"

Right, regardless of the deprecation, that won't work since :python may not be the default formula.

>I think I've seen you mention elsewhere that you shouldn't specify any options on dependencies. 

Correct.

>To boot, that line also fails the formula audit.

Yup.

>I can't think of any other way to get a version of python with a working Tkinter on linux. Is there an alternative to options with dependencies?

You could try to convince Linuxbrew to make that the default in https://github.com/Linuxbrew/homebrew-core/blob/master/Formula/python.rb

pymol.rb
```diff
@@ -22,6 +22,7 @@ class Pymol < Formula
   
   if OS.linux?
     # no guaranteed tcl-tk on linux, include it here as explicit dependency
+    depends_on "gpatch" # see homebrew/homebrew-science#5102
```
sjackman: Should this be a `:build` dependency?
```ruby
depends_on "gpatch" => :build
```

sjackman: Did you run into this error again?
```
Error: /home/linuxbrew/.linuxbrew/opt/gpatch not present or broken
Please reinstall gpatch. Sorry :(
```
I never really understood why it occurred.

sjackman:
> You could try to convince Linuxbrew to make that the default in https://github.com/Linuxbrew/homebrew-core/blob/master/Formula/python.rb

We want to avovid `:python` to pull in all of `xorg`, which is a pretty big dependency.
I could see possibly adding a `python-tk` package.

ilovezfs:
>I could see possibly adding a python-tk formula.

My two cents is that would be the way to go.

pymol.rb
telamonian: yep. Worse yet, `depends_on "gpatch"` ins't fixing it. Reinstalling `gpatch` and restarting the build does work. Not sure how to package that into the formula, though :)

The `gpath` bug only started happening after I added the python/tcltk dependency lines in:
```
depends_on "tcl-tk"
depends_on "python" => "with-tcl-tk"
```
so maybe it has something to do with how linuxbrew is satisfying that very particular dependency? Like, what exactly happens when you run into this dependency if you already have a non tcl-tk version of `python` installed?

telamonian: On closer inspection, it seems like `gpatch` should be getting treated as a dependency of `python` (due to a patch in its formula), but isn't. Here's the relevant lines from `software_spec.rb`:
```
def patch(strip = :p1, src = nil, &block)
    dependency_collector.add("gpatch" => :build) unless OS.mac?
    p = Patch.create(strip, src, &block)
    dependency_collector.add(p.resource) if p.is_a? ExternalPatch
    patches << p
end
```
The patch is wrapped in a `if build.with? "tcl-tk"`, so maybe the sequencing of the dependency resolution process is getting screwed up? Just a guess, since I don't really know enough about it at a low level.

telamonian: And yep, if you remove the `if build.with? "tcl-tk"` from around the `patch do` block in the `python`, that alone is enough for `gpatch` to get correctly listed as a dependency of `python`.

So it seems like because the `--with-tcl-tk` option is getting passed via `depends_on "python" => "with-tcl-tk"` in another formula, the option isn't being properly set during dependency lookup. 

I'm guessing that this sort of thing is why `depends_on "python" => "with-tcl-tk"` is now deprecated.

sjackman:
@telamonian Would you be interested in looking into submitting a `python-tk` formula to Linuxbrew/core? It should include only those files that are not included in `python` so as not to conflict with the `python` formula.

telamonian:
> Would you be interested in looking into submitting a `python-tk` formula...

@sjackman I'll certainly look into it. Initial research makes it seem unlikely that python's `Tkinter` can be built outside of a complete build of `python` itself. So then one possibility would be to rebuild `python` in a tmp dir and then discard the non-relevant files. Pretty crude, but I think it could work. I don't love it though. Did you have any thoughts as to possible implementations for `python-tk`?

Aside from the `depends_on "python" => "with-tcl-tk"` issue and the `gpatch` issue (both of which would be fixed by a `python-tk` formula), I think the new `pymol` is done and ready to be pulled. It builds and runs on both mac and linux now. The build bots are failing for various reasons*, but they do at least show that `pymol` builds on both mac and linux now, and my own testing has shown that `pymol` runs correctly on mac and (with a little coaxing in the form of graphics drivers) linux. I'll keep working on the linux side of things, but as far as mac users go the formula is definitely ready to be pulled.
<br>
<br>
*e.g. failing formula audits, the travis/circleci virtual machines don't have a correct-enough opengl setup to run `pymol` (it seems like this level of GPU support is a stretch for any vm at the moment), etc.), random problems downloading files, etc.

sjackman:
> I'll certainly look into it. Initial research makes it seem unlikely that python's Tkinter can be built outside of a complete build of python itself. So then one possibility would be to rebuild python in a tmp dir and then discard the non-relevant files.

That's pretty much what I was thinking.

sjackman:
> and the `gpatch` issue

Do you have an error message or a `gist` of the `gpatch` issue?

sjackman:
> the travis/circleci virtual machines don't have a correct-enough opengl setup to run pymol

Does `depends_on "linuxbrew/xorg/mesa" unless OS.mac?` help?

sjackman:
> I'll be pushing the patch upstream, or at least attempting to, but I figure that might take a while, so in the meantime I just added the patch to a formula am sharing it with the homebrew community, as many others have done in the past for me.

Can you submit the patch upstream in a pull request? Then you can use the URL to the `diff` on GitHub in this formula.

telamonian:
> Do you have an error message or a gist of the `gpatch` issue?

From line 4862 in the [latest travis build of pymol](https://travis-ci.org/Homebrew/homebrew-science/builds/261781432):
```
==> Installing python --with-tcl-tk
Error: /home/linuxbrew/.linuxbrew/opt/gpatch not present or broken
Please reinstall gpatch. Sorry :(
```

> Does mesa help?

`mesa` is already getting pulled in as a dependency: `pymol` => `glew` => `mesa`. The furthest I was able to get in a vm was to open the program. Trying to actually load a protein structure file and display a model leads to a segfault. It appears that the circleci bot has the same configuration/problem when it runs `brew test pymol`:
```
 Detected OpenGL version 2.0 or greater. Shaders available.
 Detected GLSL version 1.30.
 OpenGL graphics engine:
  GL_VENDOR:   VMware, Inc.
  GL_RENDERER: Gallium 0.4 on llvmpipe (LLVM 3.9, 256 bits)
  GL_VERSION:  3.0 Mesa 17.0.0
 Detected 32 CPU cores.  Enabled multithreaded rendering.
 CmdLoad: PDB-string loaded into object "pept", state 1.
Segmentation fault (core dumped)
```

> Can you submit the patch upstream in a pull request?

The upstream is a [svn on sourceforge](https://sourceforge.net/p/pymol/code). There's no way to submit a pull request in that case, right?

pymol.rb
```diff
@@ -12,18 +12,58 @@ class Pymol < Formula
     sha256 "217df24e99b9b96b13a9b33b137bb3ba811c407d79896091e1ef443074030df6" => :yosemite
   end
 
+  option "without-c++11", "Disable C++11 std library features.\n" \
+    "\tCompiler support for the 'auto' keyword is required regardless."
```
sjackman: What's the purpose of including this option? I'd suggest removing it.

pymol.rb
```diff
@@ -12,18 +12,58 @@ class Pymol < Formula
...
+
+  # core dependencies
   depends_on "glew"
-  depends_on "python" => "with-tcl-tk"
-  depends_on "homebrew/dupes/tcl-tk" => ["with-threads", "with-x11"]
+  depends_on "freetype"
   depends_on :x11
+  
+  if OS.linux?
```
sjackman: Please use `if OS.mac?` rather than `if OS.linux?` and swap the order of the two code blocks below.

pymol.rb
```diff
@@ -12,18 +12,58 @@ class Pymol < Formula
...
+    # no guaranteed tcl-tk on linux, include it here as explicit dependency
```
sjackman: You can remove this comment.

pymol.rb
```diff
@@ -12,18 +12,58 @@ class Pymol < Formula
...
+    args = %W[
+      --bundled-pmw
+      --install-scripts=#{libexec}/bin
+      --install-lib=#{libexec}/lib/python2.7/site-packages
+    ]
+
+    # pass flags to setup.py based on the options passed to the formula
+    args << "--no-libxml" if build.without? "libxml2"
+    args << "--use-msgpackc=no" if build.without? "msgpack"
+    args << "--no-cxx11" if build.without? "c++11"
```
sjackman: You can remove these three lines.

pymol.rb
```diff
@@ -12,18 +12,58 @@ class Pymol < Formula
...
+    depends_on :python
+  end
+  # libraries for optional features/optimizations
+  depends_on "libxml2" => :recommended
+  depends_on "msgpack" => :recommended
+
+  needs :cxx11 if build.with? "c++11"
+
+  # external GUI Aqua support patch
+  patch :DATA
```
sjackman: Please store the patch on GitHub in a gist.

pymol.rb
```diff
@@ -12,18 +12,58 @@ class Pymol < Formula
...
+
+    # on linux, add the path hint that setup.py needs in order to find the freetype and libxml2 headers
+    ENV.prepend "PREFIX_PATH", ENV["HOMEBREW_PREFIX"] if OS.linux?
+    
+    if OS.linux?
+      # because the linux python dep is specified with "python" instead of :python, python2 is needed here
+      system "python2", "-s", "setup.py", "install", *args
+    else
+      system "python", "-s", "setup.py", "install", *args
+    end
```
sjackman: Please move all this conditional code (line 52 to 66) into one big `if OS.mac? … else … end` block.

sjackman:
>> Can you submit the patch upstream in a pull request?

> The upstream is a svn on sourceforge. There's no way to submit a pull request in that case, right?

Yikes. Not as far as I know, no. We prefer not to accept large patches into Homebrew. With the next release of `pymol` the patch would likely not apply cleanly, and we'd have to drop it. Can you submit the patch upstream?

pymol.rb
telamonian: I added all of the options that the pymol devs exposed/documented in the setup.py. I agree it's somewhat pointless, but I figured there wasn't any particular reason to override their judgement. Should I still get rid of it?

pymol.rb
telamonian: I'm not sure I understand. If I remove the three `args << ...` lines, then none of the formula options will do anything (Pymol's `setup.py` is hardcoded to assume that you have all of the recommended dependencies unless you tell it otherwise). Or did you mean something else?

pymol.rb
sjackman: Yes, please. We include options when they are motivated with a use case, but otherwise avoid them.

pymol.rb
```diff
@@ -12,18 +12,58 @@ class Pymol < Formula
...
+    # no guaranteed tcl-tk on linux, include it here as explicit dependency
+    depends_on "gpatch" # see homebrew/homebrew-science#5102
+    depends_on "tcl-tk"
+    depends_on "python" => "with-tcl-tk"
+  else
+    depends_on :python
+  end
+  # libraries for optional features/optimizations
+  depends_on "libxml2" => :recommended
+  depends_on "msgpack" => :recommended
```
sjackman: Please remove `=> :recommended` from these two lines and move them up to line 21 with the other dependencies.

pymol.rb
sjackman: I had commented on line 33, but evidently didn't actually save the comment. Sorry for the confusion. The intention is to remove all the options that are not motivated by a use case.

telamonian:
@sjackman I made all of the changes you suggested and pushed. Except that I just realized that I left in the `--without-c++11` option. When I remove it, should I leave in the `needs :cxx11` line? Or should I just get rid of that too?

> With the next release of pymol the patch would likely not apply cleanly, and we'd have to drop it. Can you submit the patch upstream?

I'm concerned about that too. I already [submitted a patch upstream](https://sourceforge.net/p/pymol/patches/13/) a couple of days ago. From the only other patch request listed on the sourceforge, I'll probably have to wait about a month before hearing anything back.

pymol.rb
sjackman: Good find! Adding `if OS.mac?` may fix it for Linuxbrew:
```ruby
  if build.with?("tcl-tk") && OS.mac?
    patch do
      url "https://raw.githubusercontent.com/Homebrew/patches/42fcf22/python/brewed-tk-patch.diff"
      sha256 "15c153bdfe51a98efe48f8e8379f5d9b5c6c4015e53d3f9364d23c8689857f09"
    end
  end
```
If so, please submit a PR.

sjackman:
Please leave in `needs :cxx11`.

Thanks for submitting the patch upstream!

Since I don't use `pymol` myself, it'd be great to get a 👍 from a macOS user of `pymol` to indicate that this PR works for them.

telamonian:
My primary machine is a Macbook, and I can definitely confirm that it works (it was the linux stuff that gave me trouble). Or did you mean someone aside from the PR dev?

sjackman:
Heh. Yeah, I meant someone else as a sanity check. It could be a Homebrew maintainer or some other Homebrew user of `pymol`.

cowsandmilk:
👍 can confirm it works for me on OS X.

pymol.rb
sjackman: Once this last minor issue is resolved, I'm okay to merge this PR.

sjackman:
@cowsandmilk Thanks for doing this test, David!

sjackman:
@jonchang @ilovezfs Are you okay with merging this PR?

telamonian:
I just pushed a final (hopefully) version of the PR with that last `without-c++11` option stripped out and all of the commits squashed.

sjackman:
Looks good to me. Thanks for this work, Max, and for sticking with this PR.
Please bump the thread if we don't hear back from the other maintainers in a couple of days.

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
...
+  depends_on "libxml2"
+  depends_on "msgpack"
   depends_on :x11
 
+  if OS.mac?
+    depends_on :python
+  else
+    depends_on "gpatch" # see homebrew/homebrew-science#5102
+    depends_on "tcl-tk"
+    depends_on "python" => "with-tcl-tk"
```
jonchang: I'm pretty uncomfortable with this since we've deprecated depending on formulae with options but if we're eventually migrating to a `python-tk` formula in Linuxbrew then I guess it's ok.

sjackman: Agreed, but for the time being we don't have an alternative suggestion.

sjackman: @telamonian It'd be great if you did have time to work on that proposed `python-tk` formula.

telamonian: @sjackman I am looking into `python-tk`. Is there a mechanism in brew for reusing a formula (or part of it) in another formula? For example, can one formula inherit from another? Can you include another formula as a `resource`?

Basically, for the sake of maintainability, I'd like to be able to reuse part or all of the `python` formula when downloading/setting up the build files.

telamonian:
@sjackman  Okay, so the basics of`python-tk.rb` turned to be a lot easier than I imagined (installs of `python` retain enough of their generated config files that you can skip pretty much directly to building `_tkinter.so`). I'm prepping a version for a pull request right now. The only tricky part that's left is to figure out how to install it as a proper brew formula (instead of just copying a single file into the Cellar of `python`). I'm working it out from the `numpy` formula.

I think the name of the formula should probably be `python-tkinter`, though. There's only one file, `_tkinter.so`, that needs to be built, and that lib is then used to fix the `Tkinter` package that `python` always installs no matter what.

telamonian:
There's a bunch of people over on #5505 right now struggling with the current state of the mac build of `pymol`. My sense is that there's going to be a number of issues to iron out with `python-tk/tkinter`.

Would you consider accepting this PR as is, and saving the removal of `depends_on "python" => "with-tcl-tk"` for the next PR?

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
...
+    else
+      # on linux, add the path hint that setup.py needs in order to find the freetype and libxml2 headers
+      ENV.prepend "PREFIX_PATH", ENV["HOMEBREW_PREFIX"]
 
-    system "python", "-s", "setup.py", "install",
-                     "--bundled-pmw",
-                     "--install-scripts=#{libexec}/bin",
-                     "--install-lib=#{libexec}/lib/python2.7/site-packages"
+      # because the linux python dep is specified with "python" instead of :python, python2 is needed here
+      system "python2", "-s", "setup.py", "install", *args
```
ilovezfs: This won't work with system Python on macOS, which is what a source build will attempt to use if Homebrew's Python is not installed.

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
...
+    depends_on "gpatch" # see homebrew/homebrew-science#5102
+    depends_on "tcl-tk"
+    depends_on "python" => "with-tcl-tk"
+  end
+
+  needs :cxx11
+
+  # patch that makes the OS X native windowing system (Aqua) and PyMol play nicely together
+  patch do
+    url "https://github.com/telamonian/pymol/compare/1.8.6.0...1.8.6.0_-_osx_gui_fix_-_release.diff"
```
ilovezfs: This should use a sequence of patch do blocks, one per commit, together with .patch?full_index=1

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
...
+    depends_on :python
+  else
+    depends_on "gpatch" # see homebrew/homebrew-science#5102
+    depends_on "tcl-tk"
+    depends_on "python" => "with-tcl-tk"
+  end
+
+  needs :cxx11
+
+  # patch that makes the OS X native windowing system (Aqua) and PyMol play nicely together
```
ilovezfs: This needs a link to, and date of, the upstream issue report.

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
   end
 
   depends_on "glew"
-  depends_on "python" => "with-tcl-tk"
-  depends_on "homebrew/dupes/tcl-tk" => ["with-threads", "with-x11"]
+  depends_on "freetype"
+  depends_on "libxml2"
+  depends_on "msgpack"
   depends_on :x11
```
ilovezfs: Isn't this Linux only now?

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
   end
 
   depends_on "glew"
-  depends_on "python" => "with-tcl-tk"
-  depends_on "homebrew/dupes/tcl-tk" => ["with-threads", "with-x11"]
+  depends_on "freetype"
+  depends_on "libxml2"
```
ilovezfs: This is a system duplicate on macOS

pymol.rb
```diff
@@ -1,8 +1,8 @@
 class Pymol < Formula
   desc "OpenGL based molecular visualization system"
-  homepage "http://pymol.org"
-  url "https://downloads.sourceforge.net/project/pymol/pymol/1.8/pymol-v1.8.4.0.tar.bz2"
-  sha256 "b6147befe74844dd23550461b831b2fa6d170d4456f0059cf93fb1e8cb43d279"
+  homepage "https://pymol.org"
```
ilovezfs: Should have trailing slash

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
...
+    ]
+
+    if OS.mac?
+      # clang emits >1e5 lines of nullability warnings for pymol, turn them off
+      ENV.append_to_cflags "-Wno-nullability-completeness"
+
+      # support for older Mac OS
+      ENV.append_to_cflags "-Qunused-arguments" if MacOS.version < :mavericks
+
+      system "python", "-s", "setup.py", "install", *args
```
ilovezfs: Do the standard python setup install args not work?

pymol.rb
```diff
@@ -13,17 +13,49 @@ class Pymol < Formula
...
+    else
+      # on linux, add the path hint that setup.py needs in order to find the freetype and libxml2 headers
+      ENV.prepend "PREFIX_PATH", ENV["HOMEBREW_PREFIX"]
```
ilovezfs: I think you want ENV.prepend_path

pymol.rb
telamonian: `libxml2`? Should I just move that to the linux-only dependencies?

pymol.rb
telamonian: The issue is at least 7 years old. There's grumbling about it on various message boards and the `homebrew-science` issue tracker, but I don't know of any issue report. Would #5505 suffice?

pymol.rb
telamonian: Does `ENV.prepend_path` modify `PATH`? That is not what you want. The `setup.py` for `pymol` has custom behavior. It looks specifically for the `PREFIX_PATH` env var when determining header paths, so that's what you need to set.

pymol.rb
telamonian: That's part of a block that will only run on linux. Is that still an issue?

pymol.rb
telamonian: That sounds complicated. Can I just link to a gist with the patch instead?

pymol.rb
telamonian: done

pymol.rb
telamonian: The program has two windows, the main GUI and the external GUI. The main GUI still runs in X11.

pymol.rb
telamonian: looking at it further, both `freetype` and `libxml2` are `keg_only`. Does that mean that I should remove them from the mac dependencies, or should I make them `recommended` on mac?

pymol.rb
telamonian: I don't think so. Though if you have something in mind I'll be glad to test it out.

pymol.rb
telamonian: just tested it, runs fine without brewed `freetype` and `libxml2` on mac, so I just removed them from the mac dependencies.

pymol.rb
telamonian: I found a recent-ish related ticket on the upstream's issue tracker. I'll just link to that.

pymol.rb
sjackman: `ENV.prepend` uses a default separator of space ` ` and `ENV.prepend_path` uses a default separator of colon `:`.

telamonian:
Just pushed a new commit that addresses the review by @ilovezfs. The only thing I didn't touch was the request to split up the patch into separate patches, one for each commit. 

There's nothing significant about the separate commits, they were just work in progress. Is there a simpler option than splitting them up, like a squashed commit or a gist?

pymol.rb
sjackman: It's not an issue on Linux because it `depends_on "python" => "with-tcl-tk"` rather than `depends_on :python`.

pymol.rb
telamonian: Ah, I see. So it will work with other env vars, not just `PATH`. I've changed it.

telamonian:
Following @ilovezfs request, I've changed the patch's source from a comparison of tags to a single gist raw. Assuming that your concern was something along the lines of Homebrew/brew#2837, does this address the problem?

I don't know if this thread is actually in the running for `most comments ever on a brew pull request`, but it would be nice to get this PR done ASAP. Are there any remaining issues?

telamonian:
Also, there was another successful test of the new formula on mac over on #5505:

> I managed to install the new PyMOL 1.8.6 formula from @telamonian and it works!

Thanks Romain (@romainstuder)

pymol.rb
ilovezfs: I mean `system "python", *Language::Python.setup_install_args(libexec)`

pymol.rb
```diff
@@ -13,17 +13,51 @@ class Pymol < Formula
...
+    depends_on "python" => "with-tcl-tk"
+  end
+
+  needs :cxx11
+
+  # Patch that makes the OS X native windowing system (Aqua) and PyMol play nicely together.
+  # Fixes https://sourceforge.net/p/pymol/bugs/187/ (05.09.17) and
+  # https://github.com/Homebrew/homebrew-science/issues/5505 (04.27.17), in which bad GUI calls were causing segfaults.
+  patch do
+    url "https://gist.githubusercontent.com/telamonian/494150cf2992d324272de9a0ded78e37/raw/9abce6d6f97b3943856b63f83ad5b7352c339926/osx_aqua_ext_gui_fix.diff"
```
ilovezfs: Please open a PR to https://github.com/Homebrew/formula-patches, which is where this should be hosted.

pymol.rb
```diff
@@ -13,17 +13,51 @@ class Pymol < Formula
   end
 
   depends_on "glew"
-  depends_on "python" => "with-tcl-tk"
-  depends_on "homebrew/dupes/tcl-tk" => ["with-threads", "with-x11"]
+  depends_on "msgpack"
   depends_on :x11
```
ilovezfs: Isn't the :x11 requirement Linux only now?

pymol.rb
```diff
@@ -13,17 +13,51 @@ class Pymol < Formula
...
+    depends_on :python
+  else
+    depends_on "freetype"
+    depends_on "gpatch" # see homebrew/homebrew-science#5102
+    depends_on "tcl-tk"
+    depends_on "libxml2"
+    depends_on "python" => "with-tcl-tk"
+  end
+
+  needs :cxx11
```
ilovezfs: This should be paired with `ENV.cxx11` in the install method in almost all cases.

ilovezfs:
>I don't know if this thread is actually in the running for most comments ever on a brew pull request

Not even close.

iMichka:
I'm closing this pull request.

@telamonian I am really sorry we could not fix this. I wanted also to say that I really appreciated all the hard work you put in this to make this run, even if at the end we could not find a solution that was acceptable for homebrew.

`pymol` has been removed from homebrew-science with #6420, so that there are no wrong expectations about it being it fixed.

Upstream needs to cut a release that builds and works correctly. The current state is unsatisfactory for our users, so we can not continue with that formula for the moment. We do not have the resources to keep working on this.

As homebrew-science is being phased out, the formula will need to be submitted to homebrew-core later down the road, which would be really great if that could happen one day.

PR #6072, opened 2017-08-05
cegma: Build a bottle for Linuxbrew

sjackman:


sjackman:
@tmozgach Is `cegma` in ORCA? It wasn't bottled previously. Could you please make a list of all the formulae in ORCA that are not bottled for Linuxbrew?

sjackman:
Can you also please look into build bottles for all the formulae listed at https://github.com/Linuxbrew/homebrew-science/pulls
See also https://github.com/Linuxbrew/homebrew-science/issues/3

tmozgach:
@sjackman yes, it is! I will.

Issue #6075, opened 2017-08-07
Revisions for jpeg 9 needed


ilovezfs:
- [x] astrometry-net (broken anyway #6079)
- [x] opencv
- [x] dlib
- [x] enblend-enfuse
- [x] hdf4
- [x] insighttoolkit
- [x] lammps
- [x] mathgl
- [x] nexusformat
- [x] opencv3
- [x] openimageio
- [x] paraview
- [x] pcl
- [x] pillow
- [x] siril
- [ ] statismo (broken anyway #5916)
- [x] sumo
- [ ] therion (broken anyway #5917)
- [x] vigra
- [x] visp
- [x] vtk
- [x] wxmaxima
- [x] xfig

Not sure about these since they depend on jpeg but aren't linked to it:
- [x] blast (remove jpeg dependency: #6157)
- [x] openalpr (openalpr needs a revision bump for opencv, making a note there #6186)

iMichka:
Made a comment for the two last ones which were unclear. All done, closing here.

PR #6085, opened 2017-08-08
Jpeg revisions for nexusformat, openimageio, siril

fxcoudert:
See https://github.com/Homebrew/homebrew-science/issues/6075


fxcoudert:
Revision bumps for `opencv` and `hdf4` need to be committed before this is rebuilt.

PR #6092, opened 2017-08-08
opencv: revision for jpeg

fxcoudert:
See https://github.com/Homebrew/homebrew-science/issues/6075


fxcoudert:
This is 👍 for Mac.

sjackman:
TravisCI
```
CMake Error at cmake/OpenCVCompilerOptions.cmake:21 (else):
  A duplicate ELSE command was found inside an IF block.
  CMakeLists.txt:432 (include)
```
https://travis-ci.org/Homebrew/homebrew-science/builds/262392885#L3425

sjackman:
`opencv` is failing on CircleCI when trying to build a precompiled header using `-x c++-header` (which I've actually never seen before).

```
g++ … -x c++-header -o opencv_core_Release.gch precomp.hpp
```

My best guess is that GCC 4.8 does not support `-x c++-header`. I'll confirm whether that guess is true. The error message is

```
(.text+0x20): undefined reference to `main'
```

So `g++` it's trying to link an executable, rather than build a precompiled header.

sjackman:
Superenv on Linuxbrew is to blame. Superenv thinks that it's linking and adds `-Wl,--dynamic-linker=/home/linuxbrew/.linuxbrew/lib/ld.so` to the command line. That causes `-x c++-header` to fail.

sjackman:
Disabling precompiled headers would work around this issue, if there's an option for that.
Using `env :std if OS.linux?` would likely work around this issue, until the bug is fixed in `Linuxbrew/brew`.
I'll work on a fix for `Linxubrew/brew`.

sjackman:
I've opened PR https://github.com/Linuxbrew/brew/pull/453 to fix the bug in Linuxbrew/brew.

sjackman:
I've merged PR Linuxbrew/brew#453, which fixes this bug in Linuxbrew/brew.

sjackman:
I've triggered a rebuild of CircleCI.

PR #6111, opened 2017-08-10
gnuastro: revision for jpeg

iMichka:


iMichka:
That's for Linux only. It is the only formula which conditionally needs jpeg.

PR #6114, opened 2017-08-10
circle.yml: do not tap dupes

iMichka:
That repository is now defunct and empty.


iMichka:
Got annoyed by the warning while reading the log files :)

sjackman:
Thanks, Michka.

PR #6121, opened 2017-08-12
simpleitk: Build a bottle for Linuxbrew

iMichka:


iMichka:
I tried to build simpleitk locally using the `brew-test-bot` command, and it worked; but it built a bottle with `rebuild 1`. So I need to have an open pull request to actually be able to build it correctly.

iMichka:
Nah this is not what I wanted to do, this just changes the circle.yml file, has no influence on the local build.

PR #6138, opened 2017-08-15
scram 0.15.0

rakhimov:
Version bump


iMichka:
@rakhimov Thanks for your contribution.

I think next time we can update the url to something like: `https://github.com/rakhimov/scram/archive/0.15.0.tar.gz`, so we do not have to put the tag and revision. Homebrew will take care of detecting the version number by itself.

rakhimov:
@iMichka Thanks for pulling.

> I think next time we can update the url to something like: https://github.com/rakhimov/scram/archive/0.15.0.tar.gz, so we do not have to put the tag and revision. Homebrew will take care of detecting the version number by itself.

I used to use the tar archive; however, I switched to git repo because of the need to pull Qt GUI icons for Mac, which is provided as a git submodule. The tar archive doesn't include git submodules by default.
Probably, there's a way to juggle with tar archive, but the current git pull seems to be the simplest.

iMichka:
Ho ok, I understand. Seems the only solution would be to build your own tar file and upload them somewhere. Not asking to do that. So it's fine to leave it like this :)

PR #6144, opened 2017-08-17
Help wanted, critic2 stable (new formula)

mkhorton:
Analysis of quantum chemical interactions in molecules and solids.

### ~~~Help wanted:~~~

~~~Formula is almost complete, I can get it to install, but I have to manually run `make` a second time ... the first time `make` is run, it gives a `BuildError: Failed executing: make` with:~~~

` USE Precision_Model, ONLY: stnd
    1
Fatal Error: Can't open module file 'precision_model.mod' for reading at (1): No such file or directory
compilation terminated.`

~~~For some reason, accessing the shell in interactive mode and running `make` a second time then works fine and the formula installs correctly and seems fully functional. I'm a bit confused as to what this issue is caused by.~~~

### Have you:

    - Note that author doesn't provide a version number, but this is a tagged release on github and marked as stable.
      - ~~~Help wanted here! See above.~~~

### New formula: have you

    - Test files are very large and not included in the source distribution, so test simply runs the program with an empty script to make sure it runs

### DIFF
--- a/critic2.rb
+++ b/critic2.rb
@@ -1,0 +1,44 @@
+class Critic2 < Formula
+  desc "Analysis of quantum chemical interactions in molecules and solids."
+  homepage "https://github.com/aoterodelaroza/critic2"
+  url "https://github.com/aoterodelaroza/critic2/archive/stable.tar.gz"
+  version "stable"
+  sha256 "5c0803d47fbbb061b9f0b44e56eeebd7aa3298b275b9253fecdae7847a1ff903"
+
+  depends_on "gcc" => :build
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+
+  option "with-libxc", "Compile with libxc support to calculate exchange and correlation densities"  
+  depends_on "libxc" => :optional
+  
+  def install
+    
+    system "autoreconf", "-i"
+                          
+    if build.with? "libxc"
+      system "./configure", "--with-libxc-prefix=#{HOMEBREW_PREFIX}/lib",
+                            "--with-libxc-include=#{HOMEBREW_PREFIX}/include",
+                            "--disable-debug",
+                            "--disable-dependency-tracking",
+                            "--disable-silent-rules",
+                            "--prefix=#{prefix}"
+    end
+
+    if build.without? "libxc"
+      system "./configure", "--disable-debug",
+                            "--disable-dependency-tracking",
+                            "--disable-silent-rules",
+                            "--prefix=#{prefix}"
+    end
+
+    system "make"
+    system "make", "install"
+    
+  end
+
+  test do
+    touch("empty.cri")
+    system "#{bin}/critic2", "empty.cri"
+  end
+end
--- a/critic2.rb
+++ b/critic2.rb
@@ -5,17 +5,16 @@ class Critic2 < Formula
   version "stable"
   sha256 "5c0803d47fbbb061b9f0b44e56eeebd7aa3298b275b9253fecdae7847a1ff903"
 
+  option "with-libxc", "Compile with libxc support to calculate exchange and correlation densities"
+
   depends_on "gcc" => :build
   depends_on "autoconf" => :build
   depends_on "automake" => :build
-
-  option "with-libxc", "Compile with libxc support to calculate exchange and correlation densities"  
   depends_on "libxc" => :optional
-  
+
   def install
-    
     system "autoreconf", "-i"
-                          
+
     if build.with? "libxc"
       system "./configure", "--with-libxc-prefix=#{HOMEBREW_PREFIX}/lib",
                             "--with-libxc-include=#{HOMEBREW_PREFIX}/include",
@@ -34,7 +33,6 @@ class Critic2 < Formula
 
     system "make"
     system "make", "install"
-    
   end
 
   test do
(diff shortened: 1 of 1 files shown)


critic2.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    if build.without? "libxc"
+      system "./configure", "--disable-debug",
+                            "--disable-dependency-tracking",
+                            "--disable-silent-rules",
+                            "--prefix=#{prefix}"
+    end
+
+    system "make"
```
jonchang: > For some reason, accessing the shell in interactive mode and running make a second time then works fine and the formula installs correctly and seems fully functional. I'm a bit confused as to what this issue is caused by.

Try `ENV.deparallelize { system "make" }` as it's likely a race condition in the Makefile. Also consider notifying the upstream maintainers of this issue.

critic2.rb
```diff
@@ -0,0 +1,42 @@
+class Critic2 < Formula
+  desc "Analysis of quantum chemical interactions in molecules and solids."
+  homepage "https://github.com/aoterodelaroza/critic2"
+  url "https://github.com/aoterodelaroza/critic2/archive/stable.tar.gz"
+  version "stable"
```
jonchang: I don't think we can accept a formula without real version numbers:

`version (stable) is set to a string without a digit`

Could you please ask the upstream maintainers of critic2 to make a release with an actual version number?

critic2.rb
```diff
@@ -0,0 +1,42 @@
...
+  sha256 "5c0803d47fbbb061b9f0b44e56eeebd7aa3298b275b9253fecdae7847a1ff903"
```
jonchang: Include `# doi "..."` and `# tag "chemistry"` lines. See `lammps.rb` for an example.

critic2.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  option "with-libxc", "Compile with libxc support to calculate exchange and correlation densities"
+
+  depends_on "gcc" => :build
```
jonchang: Are you sure this `depends_on "gcc"` and not `depends_on :fortran`?

critic2.rb
```diff
@@ -0,0 +1,42 @@
...
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libxc" => :optional
+
+  def install
+    system "autoreconf", "-i"
+
+    if build.with? "libxc"
+      system "./configure", "--with-libxc-prefix=#{HOMEBREW_PREFIX}/lib",
+                            "--with-libxc-include=#{HOMEBREW_PREFIX}/include",
```
jonchang: Do something like `--with-libxc-include=#{Formula["libxc"].opt_lib}"`

critic2.rb
```diff
@@ -0,0 +1,42 @@
...
+                            "--disable-silent-rules",
+                            "--prefix=#{prefix}"
+    end
+
+    if build.without? "libxc"
+      system "./configure", "--disable-debug",
+                            "--disable-dependency-tracking",
+                            "--disable-silent-rules",
+                            "--prefix=#{prefix}"
+    end
```
jonchang: Please refactor this section to remove duplication, by setting `args = [...]` then `if build.with? "libxc" args << "..."`. See `wcslib.rb` for an example of what I'm talking about.

critic2.rb
mkhorton: I raised this issue with the maintainers [here](https://github.com/aoterodelaroza/critic2/issues/6). They don't use minor version numbers in their code, instead using git for versioning, however this is a tagged stable release, so can be relied on to exist in future, and has an associated publication.

critic2.rb
mkhorton: Added these -- is there a list of tags homebrew-science uses? couldn't find a list in the wiki

critic2.rb
mkhorton: It was my understanding that gfortran has been rolled into gcc in homebrew, I'd tried "gfortran" as a dependency but it didn't work -- didn't realise `:fortran` could be used, thanks!

critic2.rb
mkhorton: Done

critic2.rb
mkhorton: Done

critic2.rb
mkhorton: This was exactly it! Thank you :)

mkhorton:
Thanks @jonchang for the detailed review. I've pushed some changes, I think it should be okay now?

critic2.rb
mkhorton: Ah, I think the full gcc dependency is required after all...

mkhorton:
(Looks like the only test failure now is "GitHub repository not notable enough" -- don't know what to say to this except that I find it useful! and the code has been in development since at least 2009 to the present day, and its associated papers have ~ 200 citations, so I think it's worth including! I don't think GitHub metrics are necessarily useful in this case.)

mkhorton:
Hi @jonchang, I just wanted to follow up on this -- are there any additional changes I need to make for this to be accepted / have I missed anything? If I've understood the CI output correctly, the test failure is just notability, I believe it's fine otherwise.

iMichka:

See also #6365 for the discussion about what is going to happen with homebrew-science.

Issue #4560, opened 2016-11-02
revisions for icu4c 58.1 needed


ilovezfs:
- [x] libfolia (libfolia 1.8 #6160)
- [x] ucto (ucto 0.9.6 #6163)
- [x] vislcg3 (vislcg3 1.0.1-alpha1 #6165)

RE: https://github.com/Homebrew/homebrew-core/pull/6505

CC @dpo @sjackman

dpo:
cf. #3522, #4092

PR #6160, opened 2017-08-19
libfolia 1.8

ilovezfs:


iMichka:
Follow up from #4560 ? :D

ilovezfs:
yup

PR #6165, opened 2017-08-19
vislcg3 1.1.0

ilovezfs:
also serves as revision for icu4c


ilovezfs:
https://github.com/TinoDidriksen/cg3/issues/3

Issue #6150, opened 2017-08-18
pymol not using current version of Glew


404random:
Installed pymol fine but when when trying to open, comes with the following error:

Traceback (most recent call last):
  File "/usr/local/Cellar/pymol/1.8.4.0/libexec/lib/python2.7/site-packages/pymol/__init__.py", line 72, in <module>
    import pymol
  File "/usr/local/Cellar/pymol/1.8.4.0/libexec/lib/python2.7/site-packages/pymol/__init__.py", line 536, in <module>
    import pymol._cmd
ImportError: dlopen(/usr/local/Cellar/pymol/1.8.4.0/libexec/lib/python2.7/site-packages/pymol/_cmd.so, 2): Library not loaded: /usr/local/opt/glew/lib/libGLEW.2.0.0.dylib
  Referenced from: /usr/local/Cellar/pymol/1.8.4.0/libexec/lib/python2.7/site-packages/pymol/_cmd.so
  Reason: image not found

After checking the glew version, it seems that I have 2.1 while this is referencing 2.0.0.

fxcoudert:
This will be fixed by #6178

PR #6178, opened 2017-08-20
pymol 1.8.6.0

fxcoudert:
- `tcl-tk` was migrated from dupes to core. Threads are enabled by default. X11 is not supported anymore
- Newer pymol version requires `msgpack`


iMichka:
Don't :D 

See this 99 comment long pull request:
https://github.com/Homebrew/homebrew-science/pull/6066

fxcoudert:
I know the PR, but it's not moved for 10 days and it is a heavy patch, not sure to be accepted upstream.

In the meantime, pymol is broken: https://github.com/Homebrew/homebrew-science/issues/6150

ilovezfs:
@telamonian any thoughts here?

iMichka:
I would propose
- To merge this as-is. Runtime will still be broken, but that is a different topic that can be fixed in the other PR
- Or remove pymol for the time being

ilovezfs:
I'm confused. What's the benefit of upgrading the version if it's broken either way? Shouldn't we just do the version upgrade in the other PR?

PR #6187, opened 2017-08-22
insighttoolkit 4.12.1

iMichka:


iMichka:
Times out on Linux because we lost the VTK bottles during the migration.
I hope I will be able to have a VTK bottle tomorrow:
https://github.com/Linuxbrew/homebrew-core/pull/3967

iMichka:
v4.12.1 has just been released. I'll change the commit tomorrow, when we have the vtk bottle for Linux.

iMichka:
```
homebrew/science/insighttoolkit:
  * Dependency vtk should not use option with-python3
  * Dependency vtk should not use option without-python
  * Dependency vtk should not use option with-python
```

PR #6196, opened 2017-08-23
openimageio: depend on opencv@2

iMichka:


iMichka:
Didn't have Linux bottles before and still fails to build there, so we can ignore Linux issues here.

jonchang:
Closing in favor of https://github.com/Homebrew/homebrew-science/pull/6230

PR #6216, opened 2017-08-26
Migrate to CircleCI 2.0

sjackman:


sjackman:
```
==> brew test --verbose circlator
Testing homebrew/science/circlator
==> /home/linuxbrew/.linuxbrew/Cellar/circlator/1.5.1/bin/circlator test outdir
  File "/home/linuxbrew/.linuxbrew/Cellar/circlator/1.5.1/bin/circlator", line 3, in <module>
    import argparse
ModuleNotFoundError: No module named 'argparse'
Error: homebrew/science/circlator: failed
<0> expected but was
<1>.
==> FAILED
```
https://travis-ci.org/Homebrew/homebrew-science/builds/268756867#L3617
@tmozgach Is `circlator` in ORCA, and does `brew test circlator` succeed?

tmozgach:
@sjackman yes, there is, it fails.

tmozgach:
@sjackman first of all, I got the following error after you have merged `spades` recently:

```
circlator.external_progs.Error: Found spades but couldn't get version.
Cannot continue
```
When I tried to check `spades` I got:

```
/home/linuxbrew/.linuxbrew/bin/spades.py --version
/usr/bin/env: 'python': No such file or directory
```
After that I created soft link, 
```
ln -s python2 /home/linuxbrew/.linuxbrew/bin/python
```
And `brew test circlator` now works but with warnings:
```
linuxbrew@5849923a70be:~$ brew test circlator
Testing homebrew/science/circlator
==> /home/linuxbrew/.linuxbrew/Cellar/circlator/1.5.1/bin/circlator test outdir
WARNING: SPAdes version 3.11.0 is being used. It will work, but better results are usually obtained from Circlator using SPAdes version 3.7.1. Although 3.7.1 is not the latest version, we recommend it for Circlator.
```

Should I just `inreplace` `/usr/bin/env: 'python'` to /usr/bin/env: 'python2' in`spades`?

sjackman:
I think that `#!/usr/bin/env python` is the correct shebang line for a Python program, and Linuxbrew should somehow be fixed so that such a shebang line works as expected. I'm fine with your solution of `ln -s python2 /home/linuxbrew/.linuxbrew/bin/python`.

tmozgach:
@sjackman I would like to just clarify, should I add a softlink in the formulae `spades`?

sjackman:
No, just in ORCA. I'll look into the `spades` test failure. It may require a fix to `Linuxbrew/brew`.

ilovezfs:
We're trying to get people to move to using python2 in the shebang lines in anticipation of the switch to having unversioned "python" mean python3. So I would recommend against such symlinks for unversioned "python" to mean python2 as the python formula was specifically modified not to have them, except in the libexec directory.

tseemann:
I am very concerned about making `python`==3.x and `python2`==2.x when we have a decade of software expecting it to be `python`==2.x and `python3`==3.x .  Is this for real?

iMichka:
Python 2 will die in 2020, and the world is going to go up in flames. So yes, it is real :D

ilovezfs:
I suggest you all read https://www.python.org/dev/peps/pep-0394/ ;)

ilovezfs:
With special attention to

>in preparation for an eventual change in the default version of Python, Python 2 only scripts should either be updated to be source compatible with Python 3 or else to use python2 in the shebang line.

sjackman:
> We're trying to get people to move to using python2 in the shebang lines in anticipation of the switch to having unversioned "python" mean python3. So I would recommend against such symlinks for unversioned "python" to mean python2

@ilovezfs If SPAdes works with either Python 2 or Python 3, should the shebang line be `#!/usr/bin/env python`? If it works with only Python 2 or only with Python 3, I understand that it should be either `python2` or `python3`.

PR #6226, opened 2017-08-29
field3d: revision 2 for boost

jonchang:
Created with `brew bump-revision-pr`.


jonchang:
openimageio failure addressed in https://github.com/Homebrew/homebrew-science/pull/6230

PR #6232, opened 2017-08-29
nixio: revision 1 for boost

jonchang:
Created with `brew bump-revision-pr`.


iMichka:
```
homebrew/science/nixio:
  * Dependency boost should not use option c++11
```

I think we can ignore this for the moment. Though we are using now gcc 5 on Xenial for the Linux builds, so at one point we could make boost c++11 by default.

iMichka:
@jonchang Please rebase for boost 1.65.1

PR #6254, opened 2017-08-30
ome-files: revision for boost

iMichka:


iMichka:
```
CMake Error at lib/ome/files/CMakeLists.txt:164 (add_library):
  Target "ome-files" links to target "Boost::log_setup" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

Seems as if it doesn't find boost

PR #6255, opened 2017-08-30
ome-xml: revision for boost

iMichka:


iMichka:
```
CMake Error at ome-xml/src/main/cpp/ome/xml/CMakeLists.txt:161 (add_library):
  Target "ome-xml" links to target "Boost::log_setup" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

Doesn't find boost

PR #6260, opened 2017-08-30
dgtal: revision for boost

iMichka:


iMichka:
```
homebrew/science/dgtal:
  * Dependency cgal should not use option with-eigen
```
Ignoring that as I am just revision-bumping.

And on Linux it can't be built because it only builds with a very old gcc. The issue is still open upstream. So not bottle there.

PR #6273, opened 2017-09-01
statismo: remove formula

iMichka:
This formula has been broken since some time,
as it fails to build with newest hdf5 and boost versions.

Upstream has no time/interest to take care of this for the moment.


iMichka:
@ilovezfs @sjackman 

See https://github.com/statismo/statismo/issues/258

> @iMichka Unfortunately I did not find the time to look at it yet. It also does not seem that I have much more free time in the near future. Unless somebody steps in, I think it is best to remove it from homebrew. In this way I will not be blocking you guys.
> The reason for the lack of resources that go into statismo from the side of the original developers is, that we have currently no projects that make use of statismo.

ilovezfs:
:skull_and_crossbones: R.I.P.

PR #6276, opened 2017-09-02
gubbins: remove "with-double" build option from fasttree dependency

tseemann:
I removed the `--with-double` on `fasttree` as it is now the default on the brew recipe.


ilovezfs:
I think you can just drop the pillow dependency.

ilovezfs:
>Install Pillow or its predecessor PIL (Python Imaging Library)
>if you want to create bitmaps with Bio.Graphics.")

Is that functionality we care about?

ilovezfs:
Let's assume it's needed. Hence: https://github.com/Homebrew/homebrew-science/pull/6277.

ilovezfs:
should be :green_apple: now.

ilovezfs:
Shipped. Thanks @tseemann!

PR #5502, opened 2017-04-26
rstudio-server: Depends on libedit ncurses for Linuxbrew

sjackman:

PR #6292, opened 2017-09-08
rstudio-server: update to 1.0.153

randy3k:
### Have you:

close #5502


randy3k:
There is an error on linuxbrew about boost, but I have no time to fix the linuxbrew formula.
Btw, the linuxbrew formula had never been working, so it should not hurt much.

rstudio-server.rb
```diff
@@ -101,9 +102,25 @@ class RstudioServer < Formula
     end
   end
 
-  unless build.head?
-    # this piece of code has been removed from RStudio master as R API has changed.
+  stable do
+    # the first hunk has been removed from RStudio master as R API has changed.
+    # the second hunk avoid modifying the directory on macosx server mode.
+    # https://github.com/rstudio/rstudio/pull/1507
     patch :DATA
```
jonchang: Can you please use the official commit from rstudio/rstudio? i.e. https://github.com/rstudio/rstudio/commit/b14242baeb7e24a133930412994b3670c5feaa0b.diff?full_index=1

rstudio-server.rb
```diff
@@ -101,9 +102,25 @@ class RstudioServer < Formula
...
+
+    patch do
+      # allow non-framework R installation
+      # https://github.com/rstudio/rstudio/pull/1504
+      url "https://github.com/randy3k/rstudio/commit/2dc776b.patch?full_index=1"
```
jonchang: ditto

rstudio-server.rb
```diff
@@ -101,9 +102,25 @@ class RstudioServer < Formula
...
+      sha256 "b3a507a7aef229b29b7aa658b158150d505764f68eb48a9cddd89460d7cd185c"
+    end
+
+    patch do
+      # set the default LANG to en_US.UTF-8
+      # https://github.com/rstudio/rstudio/pull/1506
+      url "https://github.com/randy3k/rstudio/commit/e80fc8b.patch?full_index=1"
```
jonchang: ditto

rstudio-server.rb
randy3k: The patch after DATA is different from that PR since DATA is a patch to 1.0.153 but the PR is patching `master`.

rstudio-server.rb
randy3k: will do

rstudio-server.rb
randy3k: will do

randy3k:
please note that `brew install rstudio-server --HEAD` requires a new formula [boost@1.63](https://github.com/Homebrew/homebrew-core/pull/17833#issuecomment-328199819).

randy3k:
it seems that the `boost@1.63` [will not be accepted](https://github.com/Homebrew/homebrew-core/pull/17833#issuecomment-328207641). I guess `--HEAD` cannot be supported for now.

rstudio-server.rb
randy3k: I was wrong. It still applies.

randy3k:
Now, `--HEAD` depends on `boost` rather than `boost@1.63`

randy3k:
Ok, now it is working with linuxbrew but doesn't pass the linkage test.
```
Unwanted system libraries:
  libtinfo.so.5
```

Any clues?

ilovezfs:
@randy3k HEAD now works with 1.65.1?

randy3k:
Both HEAD and stable work with boost 1.65.1 now. I spent some time in the source code to fix the boost 1.65.1 issue.

iMichka:
I'm working on updating Linuxbrew. We are a few commits away from boost, but at the same time llvm5 got out and llvm@4 was created, so I need to rebuild bottles for these first. Build times are quite long for all these, so I hope it will not take too long.

ilovezfs:
>Both HEAD and stable work with boost 1.65.1 now. I spent some time in the source code to fix the boost 1.65.1 issue.

@randy3k Thanks for doing that :heart:

randy3k:
How are the things? What should I do to get it merged?

jonchang:
> Ok, now it is working with linuxbrew but doesn't pass the linkage test.
> ```
> Unwanted system libraries:
>   libtinfo.so.5
> ```

The Linux version depends on `ncurses` which provides `libtinfo.so` so I'm not sure why `rstudio-server` isn't picking up that dependency. I'm just going to merge this without Linux bottles for now (instead of including broken bottles). If you could investigate this in a follow up issue that would be great.

jonchang:
Also note that our standard commit subject should be `rstudio-server 1.0.153`

jonchang:
Thanks @randy3k! This has been merged. If you have the time to investigate the Linux issue feel free to open a follow up pull request and mention me.

sjackman:
> The Linux version depends on ncurses which provides libtinfo.so so I'm not sure why rstudio-server isn't picking up that dependency. I'm just going to merge this without Linux bottles for now (instead of including broken bottles). If you could investigate this in a follow up issue that would be great.

My best guess would be that the package somehow includes a precompiled Linux executable that's linked against `libtinfo.so.5`. Executables compiled by Linuxbrew would not link against this library, but instead `libncursesw.so.6`.

randy3k:
Then it must be the [libclang](https://github.com/randy3k/homebrew-science/blob/e81400aa3b8ec274bfd2cb734ed7bf6a87a7a788/rstudio-server.rb#L103).

sjackman:
One way to fix it is to use `patchelf` to modify the ELF file. See https://github.com/Homebrew/homebrew-science/blob/master/mir-prefer.rb#L40-L43

PR #6341, opened 2017-09-22
beast: migrate to homebrew-core

jonchang:
https://github.com/Homebrew/homebrew-core/pull/18440


iMichka:
@jonchang @ilovezfs there is also a beast2 formula. beast will need to be renamed to beast@1, and beast2 will need to be migrated to core (+renamed to beast).

jonchang:
Beast and beast2 are separate, independently developed and updated pieces of software that happen to do the same thing and have the same name.

iMichka:
Well; that is original :)

Why not `beast-mcmc` to avoid confusions?

jonchang:
Well, the software is called "BEAST", so that's what the formula name should be: http://beast.community/

jonchang:
Believe me, I wish that the BEAST2 people had called their software something else so us regular BEAST users wouldn't be constantly questioned by reviewers for not using the "latest version" of the software, but it was named for arcane political reasons and it's stuck, so... 🤷‍♂️

Issue #6365, opened 2017-09-29
Deprecation of homebrew-science


iMichka:
Since #6348 we do not accept new formulae anymore. We also already migrated the most important formulae to core, and we have a list of other formulae that are being migrated: #6331

This is due to the fact that homebrew-science will loose OS X buildbot support / homebrew-core maintainer support around end of December.

I would like to discuss what happens with the rest of formulae and with homebrew-science in general after that date. There will be a massive amount of formulae that will not have been migrated.

I would like to have this discussion here, and be as transparent as possible about our plans.

@Homebrew/science @MikeMcQuaid

tdsmith:
What kind of options are under consideration? Splitting off into a separate organization? Closing shop and directing contributors to complementary efforts like conda-forge?

This is difficult in part because a lot of important scientific software has poor code quality and is difficult to package; build systems are not prioritized by tenure committees. :) Thanks to everyone who's fought to keep these tools building.

iMichka:
I think that some formulae could be hosted in private taps, maybe sorted by topic. There are a lot of formulae that have less than 10/20 downloads in the last 90 days, so these are really "low traffic" formulae.

This would allow to shift the maintenance effort to some users that really want to focus on a bunch of formulae and keep them working, while we provide a backbone of high-quality core formulae and dependencies.

sjackman:
When looking at download statistics, keep in mind that number of downloads in some time period doesn't indicate how many people installed the tool once ages ago have been using it ever since. Useful formula that are venerable and are no longer updated may not see recent downloads, even though they're still actively being used. That's one advantage that Debian Popularity Contest (Popcon) has over Homebrew Analytics. It reports the number of packages that are *currently* installed. Looking at the flux of package (number of new installations in a time period) is heavily biased toward formulae that were updated during that window.

sjackman:
@MikeMcQuaid How do you feel about doing a one-time test of all formulae in Homebrew/science, and purging any formulae that fail `brew audit`, `brew install`, `brew test`, or `brew linkage`, and keeping the formulae that pass all four? Note that I'm excluding `brew audit --online` for GitHub notability. If it were notable and passed all four tests, it would have been moved to Homebrew/core.

MikeMcQuaid:
[...]

PR #6408, opened 2017-10-21
remove mbsystem from homebrew-science

zimonkaizoku:
This is an attempt to remove mbsystem from homebrew-science as mbsystem is now maintained in a seperate tap including some additional dependencies (dwcaress/mbsystem). The mbsystem formula in homebrew-science is not up to date. There are missing dependencies and the build command lacks some arguments.


iMichka:
@zimonkaizoku Thanks! Makes sense as we are deprecating homebrew-science (#6365).

PR #6372, opened 2017-10-04
geant4 10.3.2

equal-l2:
### Have you:

- [x] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [x] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [x] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [x] Built your formula locally prior to submission with `brew install <formula>`?

### New formula:

We will no longer accept new formula in homebrew-science, as this tap is being phased out.
Please consider submitting your formulae to https://github.com/Homebrew/homebrew-core or host it in your own tap (https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap.html).

### Updates to existing formula: have you

- [x] Removed the `revision` line, if any, when bumping the version number?~~
- [x] Added/bumped the `revision` number if the changes affect the precompiled bottles?~~
- [x] Not altered the `bottle` section?
- [x] Checked that the tests still pass?

PR #6410, opened 2017-10-22
geant4 10.3.3

### LINKED ISSUE #6291: geant4 error, please fix
homebrew installed geant4

when source /usr/loacl/bin/geant4.sh

error:
   -bash: cd: /usr/local/bin/../share/Geant4-10.3.1/data/PhotonEvaporation4.3.2: No such file or directory
-bash: cd: /usr/local/bin/../share/Geant4-10.3.1/data/RadioactiveDecay5.1.1: No such file or directory 

I found PhotonEvaporation4.3 and RadioactiveDecay5.1 instead

could someone fix it

davidchall:
### Have you:

This should also fix #6291


davidchall:
In the previous PR (#6372) the linux bottle was too big to upload because it includes the physics data sets. Since they are provided as tarballs in the first place, I don't see any advantage to including these in the bottle. However, they are required for the test to pass. For this reason, I suggest that we download them in the `post_install` section.

We could either:
 - keep them as resources (originally done in #4736 to avoid failing builds due to network problems)
- use `geant4-config --install-datasets` (which would avoid human error causing bugs like #6291)

iMichka:
Is there maybe another test we could use, that does not need the sets? Or could we write a file which mocks a minimal set?

Network problems are annoying, but maybe their server is more stable since then?

davidchall:
> Is there maybe another test we could use, that does not need the sets?

Geant4 requires these datasets to run, and so they are needed for a meaningful test too. So I think it makes sense to include them in the formula, just not the bottle.

> Or could we write a file which mocks a minimal set?

Unfortunately, this is basically impossible in Geant4.

> Network problems are annoying, but maybe their server is more stable since then?

I have no idea if their server stability has improved. However, after more thought, I think the advantage of using resources is that these files are cached and can be reused for later versions (usually only a subset of datasets change when the geant4 version changes).

davidchall:
@iMichka 
I think I found a solution that simplifies the dataset download, to hopefully prevent mistakes like #6291.

I'm happy for this to be merged, so let me know if anything is blocking this.

jonchang:
Thanks @davidchall!

davidchall:
Thanks @jonchang - I think you can close #6291 now too

PR #6418, opened 2017-10-24
ipopt: fix test

dpo:


dpo:
@BrewTestBot test this please

PR #6420, opened 2017-10-25
pymol: remove formula

iMichka:
pymol has been broken since months. Patching seems complicated as huge patches
need to be applied.

Upstream needs to cut a release that works on mac, before pymol can
be added back.

This make clear that people should not expect the pymol formula to be working.
As we are also deprecating homebrew-science, the pymol formula will need to be added to homebrew-core as a new formula.


iMichka:
cc @Homebrew/science

sjackman:
I agree. cc'ing people who have contributed to this formula:
@adamv @cowsandmilk @derekbrokeit @dpo @ianml @ilovezfs @jacknagel @vszakats @xu-cheng

jonchang:
:+1:

PR #6459, opened 2017-11-19
scram 0.16.0

rakhimov:
version bump


iMichka:
@rakhimov Thanks for your contribution.

PR #6469, opened 2017-11-28
neuron: Added support for Python 3 bindings

tclose:
* Added support for installing Python 3 bindings of "neuron" formula with --with-python3 flag (overrides the installation of Python 2 bindings)
* Updated the downloaded Neuron version to 7.5


tclose:
These tests are failing because the tests require scipy. How should I add scipy as a dependency of the tests (not the package itself)?

sjackman:
You can't unfortunately add a dependency just for test. You can either add the dependency on scipy, or modify the test not to require scipy.

sjackman:
We're trying to avoid options when possible. Is it reasonable do you think to drop support for Python 2 and make Python 3 the default?

tclose:
I probably wouldn't do that until Python drops support for v2 in a couple of years time if it were up to me. I can only speak for myself, but I have only just got around to porting the majority of my code to Python 3.

What are the main motivations to avoid options? As if it is to avoid complications when installing the formula as a dependency, then I don't think it should be that much of an issue for Neuron, as I can't see it being a dependency for many other formulas (with the one I am writing for my Pype9 package a notable exception).

tclose:
Just checked the failing test, which is only on High Sierra. The problem is in the Neuron package and the formula as it currently stands does not build on High Sierra either.

I will report the issue on the Neuron forum.

tclose:
Here is the forum post if you would like to track it https://www.neuron.yale.edu/phpBB/viewtopic.php?f=4&t=3793&p=16251#p16251

sjackman:
For future pull requests, the short commit message should be `neuron 7.5`. You can add details below.

sjackman:
Merged. Thanks for your contribution to Linuxbrew, Tom.
`neuron` has over fifty installations in the last ninety days, so it's up for migration to Homebrew/core. This repo is being deprecated before the new year.

```
================================================================================
1 | homebrew/science/neuron                                       | 45 |  53.57%
2 | homebrew/science/neuron --with-python3                        | 35 |  41.67%
3 | homebrew/science/neuron --with-mpi --with-python3             |  3 |   3.57%
4 | homebrew/science/neuron --with-mpi                            |  1 |   1.19%
================================================================================
Total                                                             | 84 |    100%
================================================================================
```

PR #6474, opened 2017-12-01
Upgraded Neuron package so it builds on High Sierra

tclose:
This PR upgrades the Neuron version to contain bug fix for High Sierra. To do this I have 
set the neuron.rb formula to download from GitHub instead of the neuron.yale.edu downloads page.

As the Neuron authors are old-school physicists and don't believe in tagging point releases (see discussion I had with them here https://www.neuron.yale.edu/phpBB/viewtopic.php?f=6&t=3776&start=15) I have had to point it at recent commit SHA instead of a tagged version. However, this is probably a good idea as quite often bug-fixes get pushed to the most recent version on the downloads page without an increment in the version number (tsk, tsk), thus breaking the SHA check in the Homebrew formula. So ironically pointing the download directly at a SHA is probably going to be more stable.

As an additional bonus, the --HEAD option of the formula should work again (I assume it didn't previously), since I needed to add the autotools toolchain to build the package from scratch.


sjackman:
CircleCI had some trouble building…
```
gcc-5 -DHAVE_CONFIG_H -I. -I../.. -I../../src/nrnpython -I../../src/nrnoc -I../../src/oc -I../../src/parallel -I../../src/nrnjava -I../../src/nrncvode -I../../src/ivos -I../../src/sundials  -DNRNUNIT=1   -g -O2 -c -o lex.o lex.c
parse1.y: In function ‘yyparse’:
parse1.y:45:31: error: ‘yyps’ undeclared (first use in this function)
 if (!(ierr++))yyerror(arg); --yyps; --yypv; YYERROR
                               ^
```
https://circleci.com/gh/Homebrew/homebrew-science/6752

tclose:
The test failures are now just the audits failing due to the change in the download SHA.

sjackman:
Merged. Thanks, Tom!

PR #6481, opened 2017-12-04
ome-files 0.5.0

rleigh-codelibre:
### Have you:

- [Y] Followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-science/blob/master/.github/CONTRIBUTING.md) document?
- [Y] Checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-science/pulls) for the same update/change?
- [Y] Run `brew audit --strict --online <formula>` (where `<formula>` is the name of the formula you're submitting) and corrected all errors?
- [Y] Built your formula locally prior to submission with `brew install <formula>`?

- [Y] Removed the `revision` line, if any, when bumping the version number?
- [Y] Added/bumped the `revision` number if the changes affect the precompiled bottles?
- [N] Not altered the `bottle` section?
- [Y] Checked that the tests still pass?

Regarding bottles, they were disabled due to Boost incompatibilities.  With the new releases here, we no longer ship a local copy of FindBoost.cmake so shouldn't be in the situation where this will break again (at least, not until cmake FindBoost gets updated), so I've removed the line prohibiting bottles.

For ome-files, sphinx is used to build documentation at build time.  Unfortunately, the number of sphinx dependencies has grown to 18 separate python modules, and so I've updated to use a virtualenv rather than manually downloading and installing them.  Is the approach taken here OK?  It would be even nicer if I could just use "pip install sphinx" without listing all the dependencies, but it looks like this is explicitly prohibited.

Also, what's the timeframe to migrate to homebrew-core?


sjackman:
Merged. Thanks for your contribution to Homebrew/science, Roger!

Note that this tap Homebrew/science is being retired at the end of the year. You may wish to create your own tap to house this and similar formulae.
See https://github.com/Homebrew/homebrew-science/issues/6365
and https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap.html

rleigh-codelibre:
@sjackman Would formulae such as these be eligible for homebrew-core at all?  It's quite new so usage isn't very high yet.

If we need to move to our own tap, is the needed Jenkins CI configuration available so we can duplicate the homebrew CI setup for making bottles etc.?

Thanks,
Roger

sjackman:
The criterion for migration to Homebrew/core is >50 stars on GitHub, or >50 downloads in the previous ninety days. OME does not yet qualify.

Yes, I believe the CI configuration is available, though I'm not sure where for the Jenkins configuration. You'd need to ask one of the Homebrew/core developers.

Here's my configuration to build bottles for both macOS and Linux on CircleCI:
https://github.com/Linuxbrew/homebrew-bio/blob/master/.circleci/config.yml

PR #6493, opened 2017-12-13
bamtools: migrate to homebrew/core

EricFromCanada:
Companion PR: Homebrew/homebrew-core#21674


sjackman:
Thanks, Eric :canada:!

PR #6496, opened 2017-12-15
masurca 3.2.3

wookoouk:


masurca.rb
```diff
@@ -7,17 +7,19 @@ def message
...
+  url "ftp://ftp.genome.umd.edu/pub/MaSuRCA/latest/MaSuRCA-3.2.3.tar.gz"
+  sha256 "533d8bf2225c2f7701f82093025d8a7ec8830b1c6632801e8d6a0216b84d86e6"
   # doi "10.1093/bioinformatics/btt476"
   # tag "bioinformatics"
 
   bottle do
     sha256 "abb8cd1ad5d0333b2dfd8897effd514e31391d52d858daf58144eac1dbc96900" => :x86_64_linux
   end
 
+  depends_on LinuxRequirement
```
sjackman: Please leave this line where it was with the other dependencies.
You can replace it with
```ruby
depends_on :linux
```
and delete lines 1 through 8.

masurca.rb
```diff
@@ -27,26 +29,13 @@ class Masurca < Formula
...
-  depends_on "jellyfish"
   depends_on "parallel"
-  unless OS.mac?
-    depends_on "bzip2"
-    depends_on "zlib"
-  end
+  depends_on "boost"
+  depends_on "bzip2"
+  depends_on "zlib"
+  depends_on "perl"
```
sjackman: Please replace
```ruby
depends_on "perl"
```
with
```ruby
depends_on :perl => "5.18"
```

sjackman: Please sort the dependencies alphabetically.

masurca.rb
```diff
@@ -7,17 +7,19 @@ def message
 end
 
 class Masurca < Formula
-  desc "MaSuRCA: Maryland Super-Read Celera Assembler"
-  homepage "http://www.genome.umd.edu/masurca.html"
-  url "https://raw.githubusercontent.com/helios/bio-docker/master/stringtie/1.2.1/MaSuRCA-3.1.3.tar.gz"
-  sha256 "0708bc6ec5a61e177533b290380d2c3de598e49f1da640df9cd88fb538913e48"
+  desc "Maryland Super-Read Celera Assembler"
+  homepage "http://masurca.blogspot.com.au/``"
```
sjackman: There are stray backtick characters.
Please replace `.com.au` with `.com`

sjackman:
CircleCI failed with…
```
g++-5: internal compiler error: Killed (program cc1plus)
```
That usually means that it exceeded the 4 GB memory limit of CircleCI.
You can use `ENV["MAKEFLAGS"] = "-j4" if ENV["CIRCLECI"]` at the top of `def install` to see if that works around the issue.
See for example
https://github.com/Homebrew/homebrew-science/blob/master/cufflinks.rb#L27

wookoouk:
Resubmitted with requested changes:
https://github.com/Homebrew/homebrew-science/pull/6497

PR #6499, opened 2017-12-19
Unbottles for Boost 1.66.0

ilovezfs:
CC @sjackman


sjackman:
Thanks for the heads up.

PR #6507, opened 2017-12-21
remove rstudio-server

randy3k:
As homebrew-science is deprecating and homebrew core doesn't allow binary dependencies, and I am so far the only maintainer of the formula, I have moved the rstudio-server formula to https://github.com/randy3k/homebrew-rstudio-server

This PR removes rstudio-server formula from homebrew-science.


sjackman:
Thanks for your maintenance of `rstudio-server`, Randy. I appreciate it.

Migrated `rstudio-server` to https://github.com/randy3k/homebrew-rstudio-server/blob/master/rstudio-server.rb by @randy3k

PR #6508, opened 2017-12-21
tbl2asn 25.6

sjackman:


sjackman:
```
==> brew linkage homebrew/science/tbl2asn
Homebrew libraries:
  /home/linuxbrew/.linuxbrew/lib/libc.so.6 (glibc)
  /home/linuxbrew/.linuxbrew/lib/libm.so.6 (glibc)
  /home/linuxbrew/.linuxbrew/lib/libz.so.1 (zlib)
Missing libraries:
  libidn.so.11
Undeclared dependencies with linkage:
  zlib
```
https://travis-ci.org/Homebrew/homebrew-science/builds/319866071#L2280

PR #6510, opened 2017-12-21
bcftools 1.6

sjackman:
Depend on gsl by default.
Do not recommend samtools.


sjackman:
```
install: cannot stat ‘bcftools’: No such file or directory
```
https://travis-ci.org/Homebrew/homebrew-science/builds/319874345#L2609
Suspect a separate `make` is needed before `make install`.

PR #6511, opened 2017-12-22
samtools: migrate to homebrew/core

sjackman:
See https://github.com/Homebrew/homebrew-core/pull/22015


ilovezfs:
handled in d0ef8b58e

PR #6581, opened 2017-12-28
dssp 3.0.0

ilovezfs:
patch for C++11 and Boost >= 1.65


ilovezfs:
CC @iMichka this is the one where I attempted to use boost@1.60

PR #6590, opened 2017-12-29
kmergenie 1.7044

ilovezfs:
needs OpenMP
depend on gcc for macOS


ilovezfs:
Getting a weird
```
Testing presence of ntCard....
KmerGenie was not compiled correctly
make: *** [check] Error 1
```

